PR #1598, opened 2012-09-17
refactor ofFile and ofDirectory Poco::File into ofFilePath

arturoc:
the Poco::File in ofFile and ofDirectory have been refactored into ofFilePath and those classes inherit now from ofFilePath. ofFilePath can be used to compare and join paths using the overloaded operators:

ofFilePath("images") == ofFilePath("../data/images/")

returns true

ofFlePath("images) + ofFilePath("img.jpg")

returns "images/img.jpg" no matter the trailing slash

ofToDataPath has been refactored to use ofFilePath instead of string comparison

still needs some more testing


tpltnt:
@arturoc is this still relevant?

arturoc:
i still want to do this but since we changed the whole filesystem utils rencently the code is not valid anymore so i'm going to close it

PR #1657, opened 2012-10-15
Add ofGetVersionMajor(), ofGetVersionMinor() to ofUtils

nasser:
They return `OF_VERSION` and `OF_VERSION_MINOR` respectively. I wrote them because [Zajal](https://github.com/nasser/zajal) uses openFrameworks as [a shared library](https://github.com/nasser/zajal/tree/amsterdam/lib/core/lib), and can only see functions, not `#define`s. 

Although you can use the `#define`s from normal openFrameworks C++ projects, this patch brings internal version information under the more standard syntax.

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -661,6 +661,10 @@ unsigned int ofGetVersionMinor() {
 	return OF_VERSION_MINOR;
 }
 
+unsigned int ofGetVersionPatch() {
+	return OF_VERSION_PATCH;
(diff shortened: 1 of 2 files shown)


dimitre:
It would be great to have a way of differentiate GIT Master from 0.8.4
Git Master returns ``` ofGetVersionInfo() ``` as 0.8.4 too

bakercp:
@dimitre -- a timely request see

https://github.com/openframeworks/openFrameworks/issues/3796
https://github.com/openframeworks/openFrameworks/pull/3797

CB

Issue #76, opened 2009-12-06
FreeImage.h -> move BOOL typedefs etc to #defines and do #undef at the end


ofTheo:
this is so cocoa apps and other apps that have BOOL defines don't need FreeImage hacked

PR #1889, opened 2013-02-26
Fix for FreeImage.h typedef problems for OSX/iOS Xcode

danoli3:
Fix for FreeImage error when compiling on OSX for iOS (Xcode)

Problem:
"error: typedef redefinition with different types ('signed char' vs
'int32_t' (aka 'int'))
typedef signed char BOOL; "

Solution:
Use #define and #undef instead of typedef for the type definitions.

Updated for the new version of FreeType and to support the new 64bit data structures with the define/undef.
Based on the changes @ofTheo made 3 years ago to the same file
https://github.com/openframeworks/openFrameworks/issues/76

### DIFF
--- a/libs/FreeImage/include/FreeImage.h
+++ b/libs/FreeImage/include/FreeImage.h
@@ -136,22 +136,22 @@ FI_STRUCT (FIMULTIBITMAP) { void *data; };
 #ifndef _MSC_VER
 // define portable types for 32-bit / 64-bit OS
 #include <inttypes.h>
-#define BOOL int32_t
-#define BYTE uint8_t
-#define WORD uint16_t
-#define DWORD uint32_t
-#define LONG int32_t
-#define FIINT64 int64_t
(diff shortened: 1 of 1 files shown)


borg:
This issue is back in 0.9 on OSX using main.mm
Simple include of FreeImage.h throws the BOOL issue

Below shoudl fix it but I couldn't find the FREE_IMAGE_OSX_IOS_COMPILE_FLAG equivalent of OF_TARGET_OSX || OF_TARGET_IOS


// define portable types for 32-bit / 64-bit OS
#include <inttypes.h>
#ifdef (FREE_IMAGE_OSX_IOS_COMPILE_FLAG?)
#ifndef BOOL
  #define BOOL int
#endif
#else
typedef int32_t BOOL;
#endif


....end of FreeImage.h
#ifdef (OF_TARGET_OSX || OF_TARGET_IOS)
#undef BOOL
#endif

PR #1909, opened 2013-03-03
Fix iOS antialiased screen grab (#822).

### LINKED ISSUE #822: iOS - ofTexture.loadScreenData issue when antiAliasing is on
If I use loadScreenData with anti-Aliasing turned on in the program's startup like this:

ofAppiPhoneWindow \* iOSWindow = new ofAppiPhoneWindow();
iOSWindow->enableAntiAliasing(4);

then the texture will load in nothing but solid black. If I bypass this anti-aliasing, it works fine. I also cannot properly populate a texture using loadScreenData in a landscape orientation. Changing to Portrait orientation in the update loop, and then back to landscape after the grab fixes it.

mcforman:
Post-draw GL read-backs now work from antialiased framebuffer, to fix iOS screen grab issues such as with ofxiPhoneScreenGrab() and ofTexture::loadScreenData() (fix #822).

Done by setting display-mapped 'resolve' buffer (defaultFramebuffer) as the current framebuffer after a draw, rather than the fsaa buffer. The fsaa buffer is set as current again before next draw.

This means that any direct GL drawing code executed between update()+draw() cycles won't work, as it'll end up in the wrong framebuffer. _However_, I understand nobody should be doing that - if I'm wrong, well, dump it all!

Tested via ofxiPhoneScreenGrab() on an iPad2 and iPod Touch.

### DIFF
--- a/addons/ofxiPhone/src/gl/ES1Renderer.m
+++ b/addons/ofxiPhone/src/gl/ES1Renderer.m
@@ -183,6 +183,9 @@
 - (void)readPixels:(int)width:(int)height:(void *)buffer
 {
     if(fsaaEnabled) {
+ 		glBindFramebufferOES(GL_READ_FRAMEBUFFER_APPLE, fsaaFrameBuffer);
+		glBindFramebufferOES(GL_DRAW_FRAMEBUFFER_APPLE, defaultFramebuffer);
+		glResolveMultisampleFramebufferAPPLE();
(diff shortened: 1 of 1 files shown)


tpltnt:
This is the oldest open pull request right now. @mcforman could you try to resolve the conflicts with the main branch? Could you then ping @julapy for another review? Thank you in advance :)

PR #2275, opened 2013-07-16
ofxAndroidMagneticField proposal [Do not Merge]

pelayomendez:
First implementation of the MagneticField Sensor in Android. Needs testing and feedback from Android tablets and phones.  

The compass uses accelerometer to improve sensor feedback trying to avoid 2.3 API and gyroscope / gravity hardware so maybe it´s not the most accurate way.

Magnetic field is implemented in a different JAVA class, not sure if it should share class with other Android sensors similar to iOS CoreLocation.

An example is included for testing.

Best,

P.


tpltnt:
What is the current state of affairs?

roymacdonald:
Hi. Nothing from my side. Haven't looked at this in a long time. Right now I can't but I'll keep it in mind. I'll let you know about any progress.
best

Issue #2515, opened 2013-08-27
ofxTCPManager connect timeout feature


trentbrooks:
Hi, I was wondering if it would be useful to have a timeout for the Connect method in ofxTCPManager as there is for sending, receiving and accepting TCP messages? Without a timeout, an application attempting to connect to an invalid ip address (eg. "126.0.0.1") will hang for about 90 seconds then crash.

Here's a post on the forum which describes the problem with a few back links-
http://forum.openframeworks.cc/index.php/topic,13104.0.html

And here is a fix (with 5 second timeout) -
https://github.com/trentbrooks/openFrameworks/commit/9ef106ceadee7953bfc02c37b7a62d79065b7ae5

Trent

bilderbuchi:
closed by #2519

PR #2519, opened 2013-08-28
Feature tcpmanager connect timeout

trentbrooks:
Hi, I haven't submitted a PR before- so let me know when I'm off track, and if the coding style is wrong. I opened an issue here https://github.com/openframeworks/openFrameworks/issues/2515 which describes the problem.

I have set the connect timeout to 5 seconds by default. Not sure what the preferred time should be (and whether to use a #define for this)? But having no timeout, eg. OF_TCP_DEFAULT_TIMEOUT is not recommended as this would not solve the issue (blocking for 90 seconds then crashing when connecting to invalid ip). 

I have only tested on OSX.

Thanks OF,
Trent

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPManager.cpp
+++ b/addons/ofxNetwork/src/ofxTCPManager.cpp
@@ -180,17 +180,13 @@ bool ofxTCPManager::Connect(char *pAddrStr, unsigned short usPort)
             socklen_t len = sizeof so_error;
             getsockopt(m_hSocket, SOL_SOCKET, SO_ERROR, &so_error, &len);
             if (so_error == 0) {
-                ofLog() << "OK! tcp";;
                 return true;
             } 
         }
     }
     
-    if(!ret) ofxNetworkCheckError();
-    
+    if(!ret) ofxNetworkCheckError();    
 	return ret;
-    //ofLog() << "FAILED! tcp";
-    //return false;
 }
 
 //--------------------------------------------------------------------------------
--- a/addons/ofxNetwork/src/ofxTCPManager.cpp
+++ b/addons/ofxNetwork/src/ofxTCPManager.cpp
@@ -142,10 +142,7 @@ bool ofxTCPManager::Accept(ofxTCPManager& sConnect)
   return ret;
 }
 
-/*
-    Trent: adding a timeout for connection.
-    - based on http://stackoverflow.com/questions/2597608/c-socket-connection-timeout
- */
+
 //--------------------------------------------------------------------------------
 bool ofxTCPManager::Connect(char *pAddrStr, unsigned short usPort)
 {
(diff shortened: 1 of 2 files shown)


admsyn:
I *think* this broke ofxTCPManager::Connect on windows. Building on VS 2012 gets me:
```
Error	8	error C2065: 'F_SETFL' : undeclared identifier	c:\users\adam\workspace\openframeworks\addons\ofxnetwork\src\ofxtcpmanager.cpp	168	1	netTest
Error	9	error C2065: 'O_NONBLOCK' : undeclared identifier	c:\users\adam\workspace\openframeworks\addons\ofxnetwork\src\ofxtcpmanager.cpp	168	1	netTest
Error	10	error C3861: 'fcntl': identifier not found	c:\users\adam\workspace\openframeworks\addons\ofxnetwork\src\ofxtcpmanager.cpp	168	1	netTest
Error	11	error C2664: 'getsockopt' : cannot convert parameter 4 from 'int *' to 'char *'	c:\users\adam\workspace\openframeworks\addons\ofxnetwork\src\ofxtcpmanager.cpp	182	1	netTest
```

I was just trying see if #3365 could be closed, so I don't really know the windows equivalent to these calls.

trentbrooks:
Thanks @admsyn. I don't have access to windows/vs, but a quick google tells me these unix/posix calls will need win api equivalents. I should have investigated further back then before submitting the PR.

References for possible windows calls:
http://msdn.microsoft.com/en-us/library/windows/desktop/ms740632(v=vs.85).aspx
http://msdn.microsoft.com/en-us/library/windows/desktop/ms740126%28v=vs.85%29.aspx

This might need it's own issue, but it's related... wondering if it's time to convert all the ofxNetwork stuff to Poco (http://pocoproject.org/slides/200-Network.pdf) for cross platform ease? I implemented some variations of the TCP and UDP stuff a few months back and haven't needed ofxNetwork since- https://github.com/trentbrooks/ofxPocoNetwork. Anyway, worth mentioning as a starting point for discussion.

liquidzym:
is this fixed? not too sure about this,on windows something like
#ifdef TARGET_WIN32
	unsigned long imode = 1;
	if (m_dwTimeoutConnect != NO_TIMEOUT) ioctlsocket(m_hSocket, FIONBIO, (unsigned long*)&imode);
    #else
    if (m_dwTimeoutConnect != NO_TIMEOUT) fcntl(m_hSocket, F_SETFL, O_NONBLOCK);
	#endif

also the shoud be?  getsockopt(m_hSocket, SOL_SOCKET, SO_ERROR, (char*)&so_error, &len);

procedural:
@liquidzym thanks, I used your code in this PR that fixes VS 2012 and VS 2013 compile errors which @admsyn and I have: https://github.com/openframeworks/openFrameworks/pull/3784

PR #2575, opened 2013-09-19
Fix install_dependencies.sh for Fedora

bangnoise:
Incorrect package name: pulse-devel (is pulseaudio-libs-devel)
Missing package: openssl-devel

### DIFF
--- a/scripts/linux/fedora/install_dependencies.sh
+++ b/scripts/linux/fedora/install_dependencies.sh
@@ -10,7 +10,7 @@ if [ $EUID != 0 ]; then
    exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


manpaz:
Hello,

On Fedora 20 and 21, executing `./compileOF.sh` returns:
```
checking pkg-config libraries:   cairo zlib gstreamer-app-0.10 gstreamer-0.10 gstreamer-video-0.10 gstreamer-base-0.10 libudev freetype2 fontconfig sndfile openal openssl libpulse-simple alsa gl glu glew libmpg123 
makefileCommon/config.shared.mk:260: *** couldn't find some pkg-config packages, did you run the latest install_dependencies.sh?.  Stop.
there has been a problem compiling Debug OF library
please report this problem in the forums
chown: cannot access '../lib/linux64/*': No such file or directory
```
But `pkg-config --print-errors` reveals this
```
Package x11 was not found in the pkg-config search path.
Perhaps you should add the directory containing `x11.pc'
```
Which is fixed installing `libX11-devel`, so, I'd like to request to add that package to `install_dependencies.sh` for Fedora.

Thank you!

PR #2666, opened 2013-11-05
ofThread set-OS-thread name support

SoylentGraham:
via constructor or setThreadName you can now set the OS thread's name in windows. This is reflected in the visual studio debugger (and crashdumps IIRC).

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPClient.cpp
+++ b/addons/ofxNetwork/src/ofxTCPClient.cpp
@@ -224,7 +224,13 @@ string ofxTCPClient::receive(){
 	}
 
 	// check for connection reset or disconnection
-	if((length==-1 && (ofxNetworkCheckError() == ECONNRESET) ) || length == 0){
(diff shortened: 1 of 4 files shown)


bakercp:
Closing this PR as ofThread now supports thread names

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofThread.cpp#L34

SoylentGraham:
Just FYI, that doesn't name the thread in the OS/debugger, which was what my original PR did. Just setting a variable was no use for me when debugging (I wanted to see the threadname in xcode/visual studio/external thread monitoring tools).
That thread-name string has been around since before my PR.

A little irrelevant to me now as I don't work in OF any more, but AFAIK still doesn't support proper OS thread naming.
https://github.com/SoylentGraham/SoyLib/blob/master/src/SoyThread.cpp#L407
This is my implementation for various platforms, in case anyone ever did want to add this functionality.

PR #2690, opened 2013-11-25
Fix iosvideo grabber getPixelsRef

roymacdonald:
I was doing some CV stuff on iOS and found that the videoGrabber had the getPixelsRef function unimplemented.
I'd rather have someone like @julapy or @obviousjim to check what I've done. 
On the other hand, where is that I should document this fixes (sorry, I'm being lazy).
Best!


DomAmato:
isn't getPixelsRef depreciated now? also has things been rewritten?

arturoc:
yes this should be fixed now since getPixels now returns ofPixels

Issue #2741, opened 2013-12-20
tr 1 memory not found


brightredchilli:
I pulled master fa13d88 and tried to run the project generator in apps/devApps/projectGeneratpr/ and I was getting an error that <tr1/memory>  could not be found.
![image](https://f.cloud.github.com/assets/751191/1794475/9be70f9a-69ca-11e3-8388-e9e5c3d9e106.png)

I am not a c++ expert but I gathered that I wasn't using the right C++ library. To fix this, I set the C++ dialect and standard libraries in my project to match that of the openFrameworks static library target, like so:
![image](https://f.cloud.github.com/assets/751191/1794479/e82be4e8-69ca-11e3-8a24-bf55c0ec50ae.png)

I had a hard time figuring this out, and I hope someone might benefit. I was also running Mavericks. Should this be something that we incorporate into the project, using xcconfig files?

Also, slightly tangential, I think there should be some docs on how to get the projectGenerator up and running. The projectGenerator tutorials out there right now do not resemble what the it looks like in the project right now.

Issue #2695, opened 2013-11-27
iOS deployment target 7.0 leads to error: 'tr1/memory' file not found


ofZach:
apps have a really low deployment target (3.1), when you switch the deployment info under general->deployment info->deployment target or you pick a different device (ipad, iphone, etc) and the deployment target changes automatically, when it's set to iOS 7.0 you get the following memory tr1 error: 

In file included from ../../../libs/openFrameworks/utils/ofLog.h:5:
../../../libs/openFrameworks/types/ofTypes.h:9:10: fatal error: 'tr1/memory' file not found

the quick fix is to set the deployment target to 6.0, but I couldn't find a compile setting to get deployment target 7 to work ok.

rstecca:
I'm having this issue with 0.8.4 ios on Yosemite

PR #2764, opened 2014-01-18
Fix <tr1/memory> error on osx, ios

### LINKED ISSUE #2686: tr1/memory error
Hi all!

i have the annoying "tr1/memory file not found" error in most of my old projects and in the most of the add ons examples..

I'm on osx 10.9 , OF 0.0.8 and xCode 5.0.2

there are some threads on the forum about this but with no answers...
Does anyone knows how to solve this, i'm kind of suck with this error

hope someone can help
best
j

admsyn:
ALRIGHT, I think I figured it out :)

TL;DR openFrameworks as-is needs the c++ library version to be libstdc++. Setting the SDK to 10.8 kinda "fixes" that in a roundabout way. This lets fix lets us compile with the 10.9 SDK.

When using the 10.9 SDK, libc++ is made the "compiler default" standard library. This is what was causing the missing header error, as libc++ doesn't ship with the tr1 stuff. However, it _also_ doesn't work with the C++98 dialect we're using right now (meaning, an #ifdef to select `<memory>` instead of `<tr1/memory>` doesn't fix the problem).

OpenFrameworks itself has been specifying libstdc++ for a while, but the project template has just been "compiler default", which just became a problem now since the OSX 10.9 SDK makes libc++ the "compiler default".

This PR sets the project template's library version to libstdc++, and also switches the debugger to LLDB to get rid of the nag popup ("Your project has been updated to use the LLDB debugger.."). 

All of the above seems to apply to iOS as well when setting the deployment target to 7.0+

Tested by running the project generator; I was able to create a new project and run it without fiddling with build settings. Project templates have the post-fix `plutil -convert xml1` applied to get them back to regular XML.

Fixes #2686 #2741 #2695

### DIFF
--- a/scripts/ios/template/emptyExample.xcodeproj/project.pbxproj
+++ b/scripts/ios/template/emptyExample.xcodeproj/project.pbxproj
@@ -654,6 +654,8 @@
 				<string>YES</string>
 				<key>ARCHS</key>
 				<string>$(ARCHS_STANDARD_32_BIT)</string>
+				<key>CLANG_CXX_LIBRARY</key>
+				<string>libstdc++</string>
 				<key>CODE_SIGN_IDENTITY</key>
 				<string></string>
 				<key>CODE_SIGN_IDENTITY[sdk=iphoneos*]</key>
@@ -706,6 +708,8 @@
 				<string>YES</string>
 				<key>ARCHS</key>
 				<string>$(ARCHS_STANDARD_32_BIT)</string>
+				<key>CLANG_CXX_LIBRARY</key>
+				<string>libstdc++</string>
 				<key>CODE_SIGN_IDENTITY</key>
 				<string></string>
 				<key>CODE_SIGN_IDENTITY[sdk=iphoneos*]</key>
@@ -912,7 +916,7 @@
 			<key>isa</key>
 			<string>PBXReferenceProxy</string>
 			<key>path</key>
-			<string>libofxiPhone_iphoneos_Debug.a</string>
+			<string>libofxiOS_iphoneos_Debug.a</string>
 			<key>remoteRef</key>
 			<string>E41D410113B3A0D800A75A5D</string>
 			<key>sourceTree</key>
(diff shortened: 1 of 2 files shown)


iColpitts:
I'm having this issue trying to run examples on the most recent iOS release of OpenFrameworks. Is there a quick fix? I can't figure out exactly what I should have everything set at and where.

ofTheo:
@iColpitts two things that can easily fix this issue. 
- make sure that the build SDK for the OF project and the Example project are the same. If they are different you can get this error. 
- make sure that in "C++ Standard Library" setting in the OF project and the Example project are both changed from Compiler Default to libstdc++

alistaircolling:
@ofTheo Thanks for the info, this has got rid of the tr1/memory issue. I've tried your solution but now I'm getting:

`Apple Mach-O Linker error
-pie can only be used when targeting ios 4.2 or late
clang: error: linker command failed with exit code 1 (use -v to see invocation)`

I'm targeting to iOS 8.4, running of_v0.8.4_ios_release Xcode 6.4 Mac OS 10.10.5
I can't seem to get any of the simulators to run :/

I have tried the latest stable branch from github and have the same problem :/

I have tried running example projects and also using the projectGenerator.

Do you have any more suggestions?

Thanks in advance!
Alistair

ofTheo:
Hmm - 0.8.4 probably won't work with the iOS 8 SDK as it required 64bit libs. 
Can you try the latest nightly build: http://192.237.185.151/versions/nightly/of_v20150915_ios_nightly.zip

alistaircolling:
@ofTheo It worked!! That's great Theo. Thanks so much, spent the last 2 hours trying different solutions (including currently pulling the iOS 7.1 SDK from my old macbook with a broken screen).
Really grateful for your quick response. Much appreciated.
Best,
Alistair

ofTheo:
no problem - really glad its works for you!! :)

if you notice any bugs with the iOS code ( as its close to being our 0.9.0 release ) feel free to open an issue on github.

nujudrashed:
i try to fix it but until now it is not working 
can you help

PR #2773, opened 2014-01-23
Bugfix / Feature: ofSSLManager.

bakercp:
Currently, the defaults for SSL Context management are practically non-existent in the core.

In fact, the only place SSL Contexts are uses is in `ofURLFileLoader.h`.  In this case, an insecure default Client Context is created that accepts any server certificate.  For security reasons, we need to be a bit more careful about this and either automatically check the server certs against a Certificate Authority Bundle pem or require the user to "manually" accept the certificate.

This issue and other other issues are solved by the new `ofSSLManager`.  While it would be tempting to manage all default SSL context creation via `Poco::Net::SSLManager`, it is important that we set up openFrameworks-friendly defaults (e.g. the default location for client / server certificates, private keys, certificate authority bundles, etc).  Poco's SSL management framework is very easy to use and configure if you are inheriting from `Poco::Application`, but we do not, and don't need to.

Please see the well doxygen-documented `ofSSLManager` class for an in-depth discussion of why we need the manager and how it works.

This PR includes fixes for `ofURLFileLoader` and creates a framework for SSL certificate event callbacks.

If desired, it might be useful to register default event callbacks callbacks with useful error messages inside of `ofBaseApp`.  But that is slightly more intrusive and may not be needed because not all applications will need to initialize the `ofSSLManager` sub-system.

For an example of the SSL event callbacks in action, you can take a look at the development version of ofxSMTP here:

https://github.com/bakercp/ofxSMTP/blob/develop/example-basic-gmail/src/ofApp.cpp#L81

Your feedback is appreciated!


bakercp:
I'm going to close this for now.  With the removal of Poco, another solution will need to be found.  It lives here for now https://github.com/bakercp/ofxSSLManager.

PR #2835, opened 2014-03-09
ofMesh: fixed bug of saving mesh as ply ASCII file

sadmb:
There's a bug in ofMesh::save() function and I couldn't load ply ASCII file properly. (indices out of bounds)
I think binary is ok but not sure.


tpltnt:
@sadmb can you 
1. describe the steps to reproduce your bug and
2. solve the current conflicts with our master?
This would be very helpful. Thank you!

PR #2917, opened 2014-04-04
Restructuring OF

ofZach:
this isn't done yet -- just want to start a PR to try to re-organize the OF folder a bit.  

* most importantly: created a **tests** folder, we should figure out how to organize tests, maybe having different folders for different types of tests? (performance, platform, etc) 
* moving PG to one folder so we can start merging and working on the next gen PG.
* moved devapps to the tests folder for now

I also want to push here for a different project structure that doesn't put all project files in one place, this will help greatly for cross platform development and PG related issues.


arturoc:
we moved the PG to a different repo so this won't merge anymore

PR #2918, opened 2014-04-04
ofBox first commit. untested.

wasawi:
Hello OF Core devs,
This is basically same as ofRectangle but using the 3rd dimension.
I would like to know if it is a good idea to have a class like this in the core or otherwise why should not be there.

It has not been tested but I could refine it case you guys think that it could go to the core.

all the best!
j


bakercp:
It sounds like this hasn't gained much traction. @arturoc @ofTheo I'm closing this.  Please reopen if needed.

Issue #2931, opened 2014-04-13
xcode projects bug on 0.8.1 osx realease


roymacdonald:
I was testing the 0.8.1 osx release on osx 10.6.8 and xcode 4.2 and everytime I opened an xcode project it showed up empty. 
The solution to this is to remove the file project.xcworkspace from inside the .xcodeproj 
Xcode will rebuild the xcworkspace and it will work correctly. I guess that this is because of my old OS and those xcode projects were generate on a newer os.
On the other hand, if I create a new project using the project generator the newly created xcode project contains the project.xcworkspace file even if it is not inside the template project in scripts/osx/template so I guess that PG might be taking it's tamplate from somewhere else.
Hope someone can fix this in the release zip.

best

PR #2935, opened 2014-04-15
Xcode proj fix

ofZach:
this is a fix attempt for https://github.com/openframeworks/openFrameworks/issues/2931
note that this is against master, there's some changes already to ofUtils, etc....


DomAmato:
Is this PR still needed? just doing some house keeping

PR #3081, opened 2014-07-10
added getter for type long 

fx-lange:
Added getter() for type long long (long is rarely bigger than int on current architecture).

This way you can grab values > 2.147.483.647 from XML files. For example OpenStreetMaps node IDs.


fx-lange:
@bilderbuchi added it for ofXml in #3754

DomAmato:
if this was merged in ofXml can we close this PR

PR #3097, opened 2014-07-22
Added support for mouse wheel scroll. 

akupila:
Implements scroll wheel functionality in GLFW. Seems a bit too straight forward, not sure why this wasn't already in master.

__Only tested on OS X 10.10__

This enables the following:
```c++
void ofApp::mouseScrolled(float x, float y)
{
	ofLog() << "Scroll " << x << "x" << y << endl;
}
```


nneonneo:
This would be great, but how do you get the scroll direction?

We should get this integrated soon - it's pretty annoying not having scroll events!

mattfelsen:
Related mouse event PR at #3714 that would require the same effort of modifying examples/templates with new callbacks. After resolving merge conflicts & updating examples/templates, would this PR be ready to merge?

mattfelsen:
Looks like support for mouse scroll was already added via a merge by @arturoc in  f5b31db22efffa63a0d6a5dccc8f0f97680e5a5d. Looks like this can actually get closed instead?

nneonneo:
Yeah, looks like this could get closed.

PR #3101, opened 2014-07-23
fixes ofSerial::writeBytes on unix which fixes firmata sendPWM and servo

arturoc:
fixes firmata's pwm and servo on unix by fixing ofSerial::writeBytes to call select on the port if no bytes were received and by looping till all bytes are sent

also simplifies and fixes the communication examples and adds some simple methods to use ofSerial with ofBuffer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,16 @@ OF 0.9.0
 
 CORE
 ----
+### communication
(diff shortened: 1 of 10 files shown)


DomAmato:
This PR is stale and needs to be revamped for the newest revision of OF but its still an issue a la this forum thread

http://forum.openframeworks.cc/t/arduino-of0-9-0-of0-8-4/21305/20

PR #3120, opened 2014-07-31
fix #3061 (white rect issue on glfw)

### LINKED ISSUE #3061: GLFW on Windows (codeblocks IDE) takes long to start up
Launching an app using a GLFW window seems to take 3-4 seconds to start the app, leaving a blank (white) screen instead.

Replacing the same window with a GLUT one solves the problem, app launches more or less instantly.

igiso:
made 	glfwWindowHint(GLFW_VISIBLE,GL_FALSE);
global

fixes issue: https://github.com/openframeworks/openFrameworks/issues/3061

tested on PC&mac, please Test on LINUX

Note: 

this will "break" ofxSecondWindow addon because we now have to call show window after creating a new window, glfwWindowHint are global settings , we can fix it by reverting it from here but I think is better to fix the addon.

I will later make a seperate PR there as well.


igiso:
I think this was fixed in master Branch by 
settings.visible
visible(true);

closing as fixed

PR #3143, opened 2014-08-11
Apothecary: FreeImage 3.16.0

bakercp:
Update apothecary build scripts and header / static libs for FreeImage 3.16 (the latest tagged version).

Static libs for architectures included:
- [x] OSX FAT (32 bit w/ libstdc++ and 64 bit w/ libc++)
- [x] iOS FAT (armv7, armv7s, arm64, i386 and x86_64 w/ libc++)
- [x] Android - armeabi (@arturoc ?) 
- [x] Android - armeabi-v7a (@arturoc ?)
- [x] Android - x86 (@arturoc ?)
- [x] VS
- [x] win_cb

The following platforms use pkg-config system libs
- [x] Linux (32 bit)
- [x] Linux (64 bit)
- [x] Linux Armv6l
- [x] Linux Armv7l

Notes: 

- Before committing the `FreeImage.h` I had to convert the line endings to LF.  Apparently it is shipped with CRLF.
- Most linux distros seem to currently ship with v 3.15.*  Theoretically this shouldn't be a problem, as linux uses pkg-config headers.
- On win_cb / vs, we simply download extract and copy the precompiled libs.  The same .lib and .dll work on both win_cb and vs, resulting in duplicated files.  Using precompiled libs is much simpler than compiling manually w/ mingw, as their toolchain setup is very specific.


ofTheo:
I think we can close this now right? now that #3446 has been merged?

admsyn:
Going to close this, as it looks like #3446 has covered it

PR #3162, opened 2014-08-22
Pointer/Mouse/Cursor/Window Entered Events (replaces #2813)

bakercp:
replaces #2813


mattfelsen:
Hi all. This looked like a relatively easy PR targeted for 0.9 so I figured I'd give it a shot. Based on @bakercp's original PR plus comments in the discussion, I did the following things:
- added mouseEntered/mouseExited to ofEvents & ofBaseApp
- register/unregister these events in ofMainLoop
- added the GLFW callback

The events don't receive any arguments – no mouse x/y since it's a technically window event, and no state (ie entered vs left) since each state is broken out to a different event. We could pass the mouse x/y for convenience though using the last known values.

I tested this on OS X and it seems fine, though admittedly only with a single window app. Based on looking through the code, I noticed a few things:
- ofCoreEvents::notifyMouseEvent() looks like it isn't called from anywhere. Is this a remnant from the pre-multiwindow days? If so, does the same apply to notifyKeyEvent (and possibly others) and should they be removed?
- The existing windowEntry code is still in place for GLUT. Should it be updated to the mouseEntered/mouseExited paradigm as well? If so we can probably get rid of the ofEntryEventArgs class as well since this is the only place it's used.
- Does this mean the project templates & examples need to get updated with the addition of mouseEntered/Exited?

I pushed this change to my fork. Not sure if the preferred approach is to submit a new PR and reference this one, or to submit a PR to bakercp/windowEntry so it that it would get pulled in here once he merges. Let me know and I can do either!

https://github.com/mattfelsen/openFrameworks/commit/59301b9829947a8edb13fbbb7a9c31e9b2be70fc

bakercp:
Hey @mattfelsen if you PR against this branch (bakercp/windowEntry), then it will also be easier to see what changes you made.

arturoc:
`ofCoreEvents::notifyMouseEvent()` is used in some windows is just an alternative way of notifying a mouse event, i think android uses it.

> The existing windowEntry code is still in place for GLUT. Should it be updated to the mouseEntered/mouseExited paradigm as well?:

yes i'd say so

> Does this mean the project templates & examples need to get updated with the addition of mouseEntered/Exited?

that'd be great but a ton of work, i guess it could be done with a script that modifies all examples

mattfelsen:
Closed in favor of https://github.com/openframeworks/openFrameworks/pull/3702

Issue #2696, opened 2013-11-27
feature request: allow ofxOscReceiver to send messages (from the same port it receives)


totalgee:
One thing that is sometimes limiting about the `ofxOscSender/Receiver` classes (when trying to implement an OSC client/server model), is that each side involved in a networking exchange needs to know the other side's address/port beforehand. It would be convenient to be able to open an `ofxOscReceiver`, serving a specific port, and have it reply to whatever clients make requests. At the same time, it'd be nice to be able to write a client such that it only needs to know the server's address/port, but the server doesn't need to know anything about the client (e.g. doesn't need to know beforehand what port to reply to). The client simply makes a request (says "hello") to the server, and the server gets the reply port from the endpoint of the incoming message. But when the client has separate `ofxOscSender` and `Receiver` (as must be the case now), then each of these will have a different port -- client messages received by the server show the port of the client's sender, and replies back to that same port wouldn't be seen by its receiver. The server needs to know the client's receiver port -- so why not make it the _same_ as the sender port (in cases when this is desired?).

I made a fix (for my own use) for this problem. If you're interested, you can see it here: https://github.com/totalgee/openFrameworks/commit/b56f913d1dcc78d449be47ff23201a38dc206a4e

In summary: I added a `sendMessageTo()` method on `ofxOscReceiver`. This means a receiver listens for incoming messages (as usual), but also send messages from the same socket (which means they will show up on the receiving end with the same port number). The `ofxOscReceiver` is also not tied to send to a single address/port; it can send to different endpoints on the fly. This functionality is supported by the underlying oscpack library (well, by sockets themselves), but it's not exposed in oF.

With my change, you can do (in `setup()`):

```
mReceiver.setup(57003);
```

and receive messages as usual, in your `update()` method (using `mReceiver.getNextMessage()`). But then, you can also do:

```
mReceiver.sendMessageTo(message, "127.0.0.1", 54321);
```

[...]

PR #3173, opened 2014-08-26
ofxOsc: provide access to an ofxOscSender's local port

totalgee:
This was done in response to openframeworks/openFrameworks#2696.  The idea was to give access to an ofxOscSender's local port.  That way, an ofxOscReceiver could be created on the same port, and so all OSC communication could pass through a single port, simplifying client/server coding (a server can simply reply to clients at whatever port it receives their messages from).

A new method was added: `ofxOscSender:: getLocalPort()`.  Internally, this uses a new method exposed in oscpack: `UdpSocket::LocalEndpointForConnectedRemoteEndpoint()`.  The changes to oscpack were proposed by Ross Bencina, after discussing the issue with him.  Note that some optimization was also done in `UdpSocket::LocalEndpointFor()`, to avoid disconnecting and reconnecting, in the case where the local endpoint is being requested for the (already-connected) target.

Typical usage (in an oF client using ofxOsc) would be something like this:

```
mSender.setup("localhost", serverPort);
int localPort = mSender.getLocalPort();
mRecv.setup(localPort);
```


mrzl:
I can confirm this compiles on Windows 8.1 and Ubuntu 14.04. No changes to oscpack necessary.

arturoc:
i merged latest oscpack a couple of weeks ago, @totalgee if you want to remove your changes to oscpack, i think the rest will merge nicely and we can include it for 0.9 since it's such a simple addition now

totalgee:
Not sure what you mean.  My fix has changes to oscpack that would still be required, because I think it still doesn't include the functionality to provide access to the local port for a connected remote endpoint...

I can merge this with the latest master, but (unless I'm missing something?) I'd still need to make those changes in oscpack.

arturoc:
from what @mrzl said in the post previous to mine it seems the changes to oscpack are not required anymore, no?

mrzl:
Sorry for the confusion, of course there are changes to oscpack required. I'm not sure how easy it will be to merge the changes with the new version of oscpack. @totalgee if you get the chance, go ahead, otherwise I will check it out in the upcoming days.

mrzl:
Wouldn't it be good to provide an example for this additional functionality, too?

totalgee:
I merged with latest master, and pushed the (rebased) version here.  (if you previously were testing with my ofxOscLocalPort branch, please take it again, as the commits will have changed in the rebasing)

arturoc:
oh, ok sorry i thought this was already merged in oscpack, we'll wait until post 0.9 and then merge it. can you send a PR with the changes for oscpack to my copy of it at https://github.com/arturoc/oscpack better?

i'll merge the changes there and then we can merge this without changes to oscpack. i'm trying to put any modifications to oscpack in that repository so we can keep track of the changes to the original

totalgee:
Okay, I've made a PR in your oscpack repo with the required changes (the code is from Ross, after a discussion with him last year on how to do this).

A bit off-topic: if someone could point me to *how to build the OF examples* (when you're working from git on master), it would be helpful.  I understand I have to build projectGenerator -- and I did that -- but haven't managed to build .xcodeproj files for any examples (e.g. examples/addons/oscSenderExample).  If someone could point me to a document online (or in the repo) that explains it, I haven't found a working solution yet.  Then maybe I can make a simple client/server example illustrating this feature. Thanks!

arturoc:
you can just use the project generator simple that comes with any OF download but better if you get a recent one, from the nightlies for example or compile it your self. 

then point the path to OF/examples/addons

and set the name of the project to oscSenderExample

press generate project and it'll generate the project for the example, you can do the same for any example, just point the path to the level before the application root and set the name to then name of the example

you can also use the command line project generator now which you can copy to /usr/bin if you are in osx or linux and then from the OF root run:

projectGenerator examples -o . -r

and it'll create all the examples project files recursively

totalgee:
Thanks, but when I do that (the last example), from the top of my git OF repository, I get:

```bash
$ projectGenerator examples -o . -r
```

> [notice ] PG_OF_PATH not set, -o parameter needs to be set (-h for options)
> [ error ] ofPath seems wrong... not a directory
>
>
> [ error ] path to openframeworks (.) seems wrong, please check

But I am indeed at the top of the OF repo...  I was able to build `libs/openFrameworksCompiled` without problems.

arturoc:
yes, i just found that the command line tool doesn't understand . as a directory so better to just pass the absolute path by now.

totalgee:
Thanks, now it works...!

totalgee:
@mrzl I added a couple of examples, an OSC client and server.  The client (or multiple clients) sends requests to the server, and gets responses back.  The server also can timeout inactive clients.  It's not a super-complicated example, but neither is it trivial...hopefully it's actually useful.

arturoc:
that looks great, thanks so much! i've milestoned this for the next release so we'll merge it right away after 0.9

totalgee:
Okay.  Once you've integrated my stuff in arturoc/oscpack#23 (and brought it into OF master), I'll update this PR to use that, and no longer reference the required changes here.

PR #3196, opened 2014-09-07
Kinect sensor coordinate to color image coordinate conversion

micuat:
`ofxKinect::sensorToColorCoordinate(const ofPoint & p, float distance)` converts a sensor (depth w/o registration or IR image) coordinate to the corresponding color image coordinate. The other way round is not possible since this conversion requires a depth value.

I was only able to test with osx10.9.4+Kinect1473, and maybe there are some issues with other Kinect models so it should be tested before merging.

### DIFF
--- a/addons/ofxKinect/src/ofxKinect.cpp
+++ b/addons/ofxKinect/src/ofxKinect.cpp
@@ -396,6 +396,71 @@ float ofxKinect::getZeroPlaneDistance() {
 	return kinectDevice->registration.zero_plane_info.reference_distance;
 }
 
+#define REG_X_VAL_SCALE 256 // "fixed-point" precision for double -> int32_t conversion
+
+//------------------------------------
+int ofxKinect::sensorToColorIndex(int i) {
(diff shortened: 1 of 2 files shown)


tpltnt:
Please resolve the conflicts with the current master @micuat . Is anything else in the way of merging this pull request?

micuat:
@tpltnt this PR is only addition so resolving the conflicts won't be difficult. But it's been a while I added this and I haven't tested this feature recently so I'll test at least with oF 0.9.1 and push commits - it will be on Monday at the earliest as I don't have Kinect now.

micuat:
ok, so it seems like the conflicts are due to the addition of `const` keywords to the ofxKinect functions. Honestly I'm not sure about my implementation to be merged into master (it's not intuitive and anyways it will be rarely used).

The point is that `freenect_device* kinectDevice` should be somehow accessible from outside the class in order to do some low-level tricks. I would add

    freenect_device * getFreenectDevice() const {return kinectDevice;}

and move the coordinate conversion functions to ofApp or somewhere else.

PR #3228, opened 2014-09-12
ofSystemChoiceDialog

genekogan:
new ofSystem dialog which simply presents user with "Ok" or "Cancel" dialog and returns a bool for their choice. only implemented in OSX, needs implementations for other systems.


s-ol:
+1 - this is still relevant.

s-ol:
@arturoc done, check #6140. This can probably be closed? It seems the original repo/branch has been gone for some time already.

PR #3239, opened 2014-09-16
apothecary formula for glfw version OF is currently using. 

ofTheo:
Formula for glfw version OF is currently using. 
Builds 64bit 32bit combo on OS X. 
64bit needs to be added for windows.

@arturoc  - not sure if I am pulling from the right GLFW tree - let me know if it should be another one instead of https://github.com/arturoc/glfw/tree/feature-keysUnicode

Thanks!

### DIFF
--- a/scripts/apothecary/formulas/glfw.sh
+++ b/scripts/apothecary/formulas/glfw.sh
@@ -8,18 +8,18 @@
 
 FORMULA_TYPES=( "osx" "linux" "linux64" "vs" "win_cb" "ios" "android" )
 
-# define the version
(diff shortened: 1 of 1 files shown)


danoli3:
@ofTheo is this still good to merge?

ofTheo:
this is most likely not up to date anymore as mentioned above.

arturoc:
yes there's been some changes to glfw, it's about to release a new version actually, i'm waiting to see if a bug in ubuntu gets fixed and then will update the version in my branch.

PR #3279, opened 2014-10-01
Higher precission and monotonic timers

arturoc:
monotonic timers in platforms where is available (all except raspberry pi) and higher precision timers, mostly on windows where the precision was ms, now should be 1/100th of a ns. also sleeping should be more accurate, should solve:

http://forum.openframeworks.cc/t/emptyexample-running-at-50fps-in-release-mode/13731/10

and similar problems, we should probably enable the higher precision timing under windows in ofAppRunner:

``` cpp
    #ifdef WIN32_HIGH_RES_TIMING
        timeBeginPeriod(1);     // ! experimental, sets high res time
                                // you need to call timeEndPeriod.
                                // if you quit the app other than "esc"
                                // (ie, close the console, kill the process, etc)
                                // at exit wont get called, and the time will
                                // remain high res, that could mess things
                                // up on your system.
                                // info here:http://www.geisswerks.com/ryan/FAQS/timing.html
    #endif
```

since it seems to make a difference in the precision of the sleeping times

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -134,7 +134,7 @@ enum ofTargetPlatform{
 	#define GL_BGR_EXT 0x80E0
 	#endif
 
-	// #define WIN32_HIGH_RES_TIMING
+	#define WIN32_HIGH_RES_TIMING
 
 	// note: this is experimental!
 	// uncomment to turn this on (only for windows machines)


armadillu:
I had to use https://github.com/obviousjim/ofxMSATimer to get decently accurate on windows for https://github.com/armadillu/ofxTimeMeasurements, this is very welcome!

PR #3384, opened 2014-11-01
removes statics and globals. adds multiwindow

arturoc:
This removes almost all statics, singletons and global calls (like ofPushMatrix, ofGetWidth...) from the core, at least form the graphics and core events pipeline. To do that, the following function based apis have been transformed into classes:
- ofGraphics3d
- ofBitmapFont
- ofEvents -> ofCoreEvents

while keeping the api totally backwards compatible.

Windows now contain their own renderer and core events instance and notify that instead of the global one. Each renderer contains it's own ofGraphics3d and ofBitmapFont, which avoids singletons allowing multiwindow or multicontext even on different threads if everything is done through a explicit renderer.

There's a new class ofMainLoop which is the only singleton in the rendering pipeline and is responsible of creating new windows and calling them in a loop if the window doesn't provide it's own loop as is the case of glut, android and iphone.

the old syntax to create windows is still working but there's a new way which is the preferred one:

``` cpp
int main(){
    ofWindowSettings settings;
    ofCreateWindow(settings);

    return ofRunApp(new ofApp);  // now returns a value that can be 
                                // specified through ofExit(exitStatus)
}
```

where one can pass width, height and window mode (windowed, fullscreen...) to the settings like:

``` cpp
int main(){
    ofWindowSettings settings;
    settings.width=1024;
    settings.height=768;
    settings.windowMode = OF_WINDOW;
    ofCreateWindow(settings);

    return ofRunApp(new ofApp);
}
```

or:

``` cpp
int main(){
    ofWindowSettings settings;
    ofCreateWindow(settings);

    return ofRunApp(new ofApp);
}
```

there's also specializations of the window settings like:

``` cpp
int main(){
    ofGLWindowSettings settings;
    settings.setGLVersion(4,4);
    ofCreateWindow(settings);

    return ofRunApp(new ofApp);
}
```

or 

``` cpp
int main(){
    settings.setGLVersion(4,4);
    settings.decorations = false;
    ofCreateWindow(settings);

    return ofRunApp(new ofApp);
}
```

egl, iOS and android should have their own window settings class although that's not implemented yet. the EGL and iOS windows indeed still need to be refactored to use the new api.

To use multiwindow the syntax is:

``` cpp
int main(){
    ofWindowSettings settings;
    shared_ptr<ofBasewWindow> window1 = ofCreateWindow(settings);
    shared_ptr<ofApp> app1(new ofApp);
    ofRunApp(window1,app1);

    shared_ptr<ofBasewWindow> window2 = ofCreateWindow(settings);
    shared_ptr<ofApp> app2(new guiApp);
    ofRunApp(window2,guiApp);

    return ofRunMainLoop();
}
```

and simpler using c++11:

``` cpp
int main(){
    ofWindowSettings settings;
    auto window1 = ofCreateWindow(settings);
    auto app1 = make_shared<ofApp>();
    ofRunApp(window1,app1);

    auto window2 = ofCreateWindow(settings);
    auto app2 = make_shared<guiApp>();
    ofRunApp(window2,guiApp);

    return ofRunMainLoop();
}
```

There's also a way of creating a window without associated app and just bind some of it's events to an existing app. There's a couple of examples showing both cases in the events folder.

Also every class that can be drawn or binded can be called the traditional way like:

``` cpp
tex.draw(x,y)
```

or through a renderer like:

``` cpp
shared_ptr<ofBaseGLRenderer> gl;
...
gl->bind(fbo);
gl->bind(shader);
gl->bind(tex);
gl->draw(mesh);
gl->unbind(tex);
gl->unbind(shader);
gl->unbind(fbo);

gl->drawCircle(x,y,r);
```

which allows to use different gl contexts or even windows from different threads for example. this api is still experimental but there's an example on how to use it in apps/devApps/explicitRendererExample

### DIFF
--- a/addons/ofxiOS/src/app/ofAppiOSWindow.h
+++ b/addons/ofxiOS/src/app/ofAppiOSWindow.h
@@ -27,29 +27,129 @@
  * OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
  * OF THE POSSIBILITY OF SUCH DAMAGE. 
  *
- * ***********************************************************************/ 
-
+ * ***********************************************************************/
 
 #pragma once
 
 #import "ofAppBaseWindow.h"
 #include "ofxiOSConstants.h"
+#include "ofWindowSettings.h"
+
+class ofiOSWindowSettings: public ofGLESWindowSettings{
+public:
+    ofiOSWindowSettings()
+    :enableRetina(false)
+    ,retinaScale(0)
+    ,enableDepth(false)
+    ,enableAntiAliasing(false)
+    ,numOfAntiAliasingSamples(0)
+    ,enableHardwareOrientation(false)
+    ,enableHardwareOrientationAnimation(false)
+    ,enableSetupScreen(true) {
+        windowMode = OF_FULLSCREEN;
+        glesVersion = 1;
+    }
+    
+    ofiOSWindowSettings(const ofWindowSettings & settings)
+    :ofGLESWindowSettings(settings)
+    ,enableRetina(false)
+    ,retinaScale(0)
+    ,enableDepth(false)
+    ,enableAntiAliasing(false)
+    ,numOfAntiAliasingSamples(0)
+    ,enableHardwareOrientation(false)
+    ,enableHardwareOrientationAnimation(false)
+    ,enableSetupScreen(true) {
+        const ofGLESWindowSettings * glesSettings = dynamic_cast<const ofGLESWindowSettings*>(&settings);
+        if(glesSettings){
+            glesVersion = glesSettings->glesVersion;
+        } else {
+            glesVersion = 1;
+        }
+        const ofiOSWindowSettings * iosSettings = dynamic_cast<const ofiOSWindowSettings*>(&settings);
+        if(iosSettings){
+            enableRetina = iosSettings->enableRetina;
+            retinaScale = iosSettings->retinaScale;
+            enableDepth = iosSettings->enableDepth;
+            enableAntiAliasing = iosSettings->enableAntiAliasing;
+            numOfAntiAliasingSamples = iosSettings->numOfAntiAliasingSamples;
+            enableHardwareOrientation = iosSettings->enableHardwareOrientation;
+            enableHardwareOrientationAnimation = iosSettings->enableHardwareOrientationAnimation;
+            enableSetupScreen = iosSettings->enableSetupScreen;
+        } else {
+            enableRetina = false;
+            retinaScale = 0;
+            enableDepth = false;
+            enableAntiAliasing = false;
+            numOfAntiAliasingSamples = 0;
+            enableHardwareOrientation = false;
+            enableHardwareOrientationAnimation = false;
+            enableSetupScreen = true;
+        }
+    }
+
+    ofiOSWindowSettings(const ofGLESWindowSettings & settings)
+    :ofGLESWindowSettings(settings)
+    ,enableRetina(false)
+    ,retinaScale(0)
+    ,enableDepth(false)
+    ,enableAntiAliasing(false)
+    ,numOfAntiAliasingSamples(0)
+    ,enableHardwareOrientation(false)
+    ,enableHardwareOrientationAnimation(false)
+    ,enableSetupScreen(true) {
+        const ofiOSWindowSettings * iosSettings = dynamic_cast<const ofiOSWindowSettings*>(&settings);
+        if(iosSettings){
+            enableRetina = iosSettings->enableRetina;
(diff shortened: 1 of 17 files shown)


HalfdanJ:
@bakercp Have you already done the merging of the conflicting classes? Otherwise i'll do it tomorrow.

bakercp:
@HalfdanJ I have not started yet -- if you want to start, go for it!  Thanks!

HalfdanJ:
#3516 should close this merging.

ofTheo:
@procedural one big thing to note with multi window is that ofSetVerticalSync(false) needs to be called when each window's context is active. I would be curious if you could try your above test but with ofSetVerticalSync(false); in each windows draw function. 

Curious if that helps. 
We have a project which is using an older multi window approach and doing that helps fps immensely.

arturoc:
doing it in setup should be fine, each window context is active in it's setup. the multiwindow example does that and it goes up to 60fps without problem.

ofTheo:
thats awesome! 
:)

igiso:
I just ported one of our biggest projects in this branch to test things out.
but I am surprised on how well it works actually 
both with the "single app" style and the other syntax works fantastic both if used in setup and dynamically at runtime.

no problems whatsoever.

The esc sequence bug is solved in this branch - I didn't notice anything wrong.
The vertical sync behaves more or less as described with the exception of higher resolution usual problems (1080p>) not sure why but it's been the same in all OF versions, so it's cool
..cpu cycles seem to be more well behaved as well :+1: 
 

here are some minor differences (not bugs) I noticed just for documentation reasons,

1•
ofxFontstash doesn't draw, if the font is initialized in the context of the main window and then passed as pointer to the other windows, or used a class that is initialized in the same manner. 
(but this is expected and not an issue with our implementation)
sharing context doesn't help.

2•
when calling setWindowSize with the all setup resize_cb was not called, in the multiwindow setup it is called. 

3•
on my old own implementation of multiw-glfw, .. I was passing the current context to the main at the end of the loop, therefor when pressing closing the main window the rest of the windows were closed with it as well because glfwWindowShouldClose became true.
in this, if you close the main, the entire app waits for the other windows to close, which is fine.

4• why not implement the window title? in settings, I added it in anyway, but why not have it there before launch?


All in all this works fantastic for me, 
by the way,
can we add a silencer for the deprecations? 
 
:D 


screen shot 2015-01-13 at 8 47 58 am


also for me a define ofVersion would be handy 


Thank's for this great PR.

igiso:
I recommend we add  a bool closeOtherWindows in settings

so that in our main window creation we can do

settings.closeOtherWindows = true;

and when the main is closed,

in our exit_cb of glfw

if (ableTocloseOtherWindows)
call the 
	ofGetMainLoop()->shouldClose(0);


:)
I love this even more now that I've figured this out..
works BEAUTIFULLY!

ofTheo:
@igiso - thank you so much for testing this PR on a large project. That is actually one of the most helpful things people can do to help us make sure regressions don't slip in with these big changes.  Huge props to @arturoc too for this PR. 

For the deprecations warnings - you can silence via the Xcode build settings.

arturoc:
@igiso thanks for testing is really useful to know that it's working well with more complex projects.

the title idea sounds really good to me, do you want to add it and send a PR?

with the close other windows do you mean that when pressing ESC it should only close the current window and not the whole application?

also some resources can't be shared across contexts or even with shared contexts so it's normal that some things can't be shared.

igiso:
..  when pressing the red x button,
if you have let's say 1 main window with 10 secondary, then press X on the MAIN .. then all other should close with it..
or if you have 2 main windows, when pressing X on either one to close both.. etc..

it is really easy to set-up.. and I think most people will find it usefull..

basically (from what I can tell reading our implementation) just calling 
ofGetMainLoop()->shouldClose(0);
in the exit_cb will do the trick..

the bool in Settings will give the option, if a window is closed to close all windows.,,
(if that bool is set to true)


other glfw implementations I've seen used a slightly different approach but, in our case, this will work just fine..

arturoc:
oh, i see, yeah that sounds good, i guess we could have something like: `settings.isMainWindow`. if that flag is true then closing that window will close the application.

igiso:
yep, sounds just fine,

I should also add, that what applies for fontStash applies for ofImage as well,
meaning that if allocated in one context it can be drawn into another by passing a pointer.

bilderbuchi:
> also for me a define ofVersion would be handy 
see https://github.com/openframeworks/openFrameworks/blob/443bea41d04499582d9817dc0a480ffc2d392619/libs/openFrameworks/utils/ofConstants.h#L5-L7 :-)

igiso:
I am such an idiot ..I  knew I saw this somewhere…… 
:+1:

PR #3394, opened 2014-11-04
osx: introducing APP_VERSION

vdonnefort:
This new Makefile variable allows to set the software version shown into OSX
about window.


bakercp:
I'm all for this -- though I tend to use a single config.make file for all platforms.  It would be nice if there was a way have identical config.make files for all platforms to facilitate easier cross platform.

That said, I'd suggest putting 

```
APP_VERSION ?= 0.01
```

inside 
```
config.osx.default.mk
```

and let the user define it in whatever way they like (on the command line or manually in their cross platform config.make file.

PR #3399, opened 2014-11-06
Fix #3398 ,if click android back key,it will crash

### LINKED ISSUE #3398: ofxAndroid OFAndroid keyDown and keyUp Crash Bug!
master branch,on android,if u click the back key,it will crash.

11-06 23:09:45.158: E/libEGL(9627): call to OpenGL ES API with no current context (logged once per thread)

the problem is that u can't make native call for onBackPressed,it should be called in navtive GL thread,not in Android thread.
I fix this bug.
I have tested this!

iniwap:
on android,if u click the back key,it will crash.

11-06 23:09:45.158: E/libEGL(9627): call to OpenGL ES API with no current context (logged once per thread)

the problem is that u can't make native call for onBackPressed,it should be called in navtive GL thread,not in Android thread.
I fix this bug.
I have tested this!


arturoc:
i'm going to close this, the main problem is that you need to give a return value in the back pressed call or else you can't use the back button inside the app or if you always return true you will never go out the app with the return button.

probably we should add some comments in the events and in the docs

PR #3402, opened 2014-11-07
Fixes Missing SDKROOT for iOS+OFLib

danoli3:
When compiling the ```iOS+OFLib``` the ```SDKROOT``` was not being defined so it was having all sorts of trouble with the core framework locations (which require this variable).

screen shot 2014-11-08 at 1 05 26 am
screen shot 2014-11-08 at 1 05 38 am

It was falling into the ```/System/Library/Frameworks``` which is incorrect, although it had some for OS X with the same name, always failed on iOS specific frameworks.

This fix just defines the SDKROOT and all goes back to normal iOS SDK location.


danoli3:
This still is an issue under the hood in oF lib. Good to merge @julapy ?

julapy:
thanks @danoli3! looks good.

PR #3409, opened 2014-11-12
Fix compile error on Android: GL_HALF_FLOAT is only defined in GLESv3

nneonneo:


danoli3:
@arturoc has approached this instead with

``` #if defined(TARGET_ANDROID) || defined(TARGET_OF_IOS) ```

Which was apart of the new window branch.
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofGLUtils.h#L143

Should we instead use this ```TARGET_OPENGLES``` for raspberry pi as well? @arturoc

**Comment by arturoc** [comment]
i guess so, i haven't checked yet, @jvcleave tried it some days ago

**Comment by jvcleave** [comment]
@arturoc I had just commented it out as I tried to do the updates to `ofAppEGLWindow`

I did find a reference to GL_HALF_FLOAT but something wasn't allowing the OF to pick it up (load order maybe?)

````
$ grep -Ri GL_HALF_FLOAT /usr/include/
/usr/include/GL/glew.h:#define GL_HALF_FLOAT_ARB 0x140B
/usr/include/GL/glew.h:#define GL_HALF_FLOAT 0x140B
/usr/include/GL/glew.h:#define GL_HALF_FLOAT_NV 0x140B
/usr/include/GL/glext.h:/* reuse GL_HALF_FLOAT */
/usr/include/GL/glext.h:#define GL_HALF_FLOAT_ARB                 0x140B
/usr/include/GL/glext.h:#define GL_HALF_FLOAT                     0x140B
/usr/include/GL/glext.h:#define GL_HALF_FLOAT_NV                  0x140B
```

arturoc:
yeah, glew is not used in openGL ES so even if it's there it won't be included. can somebody fix this to make it mergeable with master again?

danoli3:
Hey @arturoc I'll patch this in now.

danoli3:
Okay fixed with PR #3582

PR #3415, opened 2014-11-14
GLFW for Stereoscopic monitors

Vamoss:
It enables the GLWF for Stereoscopic monitor like this from NVidia 3D Vision:
http://www.nvidia.com/object/3d-vision-displays.html

Usage:
	ofAppGLFWWindow * window = new ofAppGLFWWindow();
	window->setStereo(true);
	ofSetupOpenGL(window,1024,768,OF_WINDOW);
	
If you set stereo true without the correct monitor the application will not run correctly.


arturoc:
replaced by same functionality in #6cb8b70

PR #3479, opened 2014-12-12
ofSoundBuffer and ofSoundObject

kylemcdonald:
A number of cross-platform changes introducing an ofSoundObject and ofSoundBuffer. By @admsyn @bakercp @mazbox

ofSoundBuffer allows users to load sounds from disk and manipulate them on a sample-by-sample basis.

Classes inheriting ofSoundObject allow you to create filter chains.

This code is incredibly well documented, and I recommend we merge it soon :) I'm so psyched to have this, I'm creating the PR on behalf of @admsyn ;)

### DIFF
--- a/libs/openFrameworks/sound/ofSoundBuffer.cpp
+++ b/libs/openFrameworks/sound/ofSoundBuffer.cpp
@@ -188,6 +188,7 @@ void ofSoundBuffer::copyTo(float * outBuffer, std::size_t nFrames, std::size_t o
 	// if channels count matches we can just memcpy
 	if(channels == outChannels){
 		memcpy(outBuffer, buffPtr, nFramesToCopy * channels * sizeof(float));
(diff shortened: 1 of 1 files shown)


admsyn:
@arturoc the most recent ofSoundFile is on [my feature-sound-objects branch](https://github.com/admsyn/openFrameworks/tree/feature-sound-objects/libs/openFrameworks/sound)

@ofZach RE: ofScopedLock good point. Maybe for the sake of the example it'd make sense to just use mutex.lock() and mutex.unlock() to be a bit more obvious as to what's happening.

bakercp:
I tend to agree w/ @ofZach about the the conceptual aspect of scoped locks.  But they are useful. 

I added documentation / description of how they work with ofMutex a few days ago here:

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/types/ofTypes.h#L129-L165

PR #3502, opened 2014-12-27
Makefile OSX Updates - No Xcode Required!

bakercp:
Adds onto work by @lennart in https://github.com/openframeworks/openFrameworks/pull/3491.

- Removes 10.6 support.
- Adds 10.10 support.
- Adds configurable `MAC_OS_MIN_VERSION` version, defaults to 10.7 (like apothecary). 
- Adds support for compiling on 10.7-10.10 WITHOUT Xcode :+1: , just command line tools needed.

ping @pizthewiz @lennart


pizthewiz:
Good to merge?

bakercp:
Should be.  I haven't tried it with the latest master (shouldn't be an issue).  I'll test it and merge it in the next day or so if I don't find any problems.

bakercp:
This was again confirmed to work by another user here: 

http://forum.openframeworks.cc/t/compiling-error-in-the-isysroot-and-mmacosx-version-min-flags/18139

So, with that I'll merge it and let the fun begin.

PR #3247, opened 2014-09-19
+ ofSoundDevice

admsyn:
This closes #2785

It adds a new struct `ofSoundDevice`, which represents a sound device with a numeric ID, name, number of i/o channels and vector of sample rates, modeled after the `RtAudio::DeviceInfo` struct.

This PR changes `ofSoundStreamListDevices()` into a function which returns a `vector<ofSoundDevice>` and now *optionally* logs to the console. It also adds a function `ofSoundStreamGetMatchingDevices(...)` which lets you find a device by partial name, and optionally with an i/o channel count. You can also set an ofSoundStream's device with an ofSoundDevice directly (instead of just via ID). Does this fit your use-case @kylemcdonald ?

On my OSX laptop I get the following:

**ofSoundStreamGetMatchingDevices("Soundflower")**
```
(std::vector<ofSoundDevice, std::allocator<ofSoundDevice> >) $0 = size=2 {
  [0] = {
    name = "ma++ ingalls for Cycling '74: Soundflower (2ch)"
    deviceID = 2
    inputChannels = 2
    outputChannels = 2
    isDefaultInput = false
    isDefaultOutput = false
    sampleRates = size=6 {
      [0] = 44100
      [1] = 48000
      [2] = 88200
      [3] = 96000
      [4] = 176400
      [5] = 192000
    }
  }
  [1] = {
    name = "ma++ ingalls for Cycling '74: Soundflower (64ch)"
    deviceID = 3
    inputChannels = 64
    outputChannels = 64
    isDefaultInput = false
    isDefaultOutput = false
    sampleRates = size=6 {
      [0] = 44100
      [1] = 48000
      [2] = 88200
      [3] = 96000
      [4] = 176400
      [5] = 192000
    }
  }
}
```

**ofSoundStreamGetMatchingDevices("Apple", 0, 2)**
```
(std::vector<ofSoundDevice, std::allocator<ofSoundDevice> >) $0 = size=1 {
  [0] = {
    name = "Apple Inc.: Built-in Output"
    deviceID = 1
    inputChannels = 0
    outputChannels = 2
    isDefaultInput = false
    isDefaultOutput = true
    sampleRates = size=4 {
      [0] = 44100
      [1] = 48000
      [2] = 88200
      [3] = 96000
    }
  }
}
```

Tested on OSX, iOS and Windows, but the change touches all platforms due to the function signature change. The Android and Emscripten changes are a shot in the dark, pinging @arturoc

PR #3509, opened 2014-12-31
Adding missing ofSoundStreamSetDeviceID() functition

procedural:
I have different device id for sound output on Linux, without this function you can't switch it if a stream was created with `ofSoundStreamSetup()`.

### DIFF
--- a/libs/openFrameworks/sound/ofSoundStream.cpp
+++ b/libs/openFrameworks/sound/ofSoundStream.cpp
@@ -46,6 +46,11 @@ void ofSoundStreamListDevices(){
 }
 
 //------------------------------------------------------------
+void ofSoundStreamSetDeviceID(int deviceID){
+	soundStreamOutput.setDeviceID(deviceID);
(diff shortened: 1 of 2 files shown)


admsyn:
@procedural There's some discussion about not adding new global functions into the OF sound API happening in #3247 . The implication being that we're going to start deprecating the C-style API and that new code should prefer using the ofSoundStream class. IMO a good way forward would be to:

- add this function to the API, but mark all such functions as officially deprecated
- add static ofSoundStream member equivalents to the deprecated C-style functions

.. but it'd be good to have @arturoc and / or @kylemcdonald weigh in on that, since it's part of a larger OF API discussion.

FWIW if you need this functionality now you should be able to get it by adding an ofSoundStream member to your ofApp like the audioOutputExample

procedural:
@admsyn ok! There's no need for this PR if it will be deprecated and you can do the same with ofSoundStream then.

Issue #3514, opened 2015-01-02
ofVboMesh doesn't draw meshes on Mesa driver [1]


procedural:
```cpp
#include <ofMain.h>
#include "ofAppGLFWWindow.h"

class ofApp : public ofBaseApp
{
  public:
    void draw();
};

void ofApp::draw()
{
  glClearColor(0.1, 0.1, 0.1, 1);
  glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);

  ofSetColor(ofColor::tomato);
  ofDrawCircle(ofGetWindowRect().getCenter(), 80.0f);
}

int main()
{
  ofGLFWWindowSettings settings;
  settings.setGLVersion(2, 1);
  ofCreateWindow(settings);

  return ofRunApp(new ofApp());
}
```
On OpenGL 2.1:

![1](https://cloud.githubusercontent.com/assets/1072168/5595037/50634416-9279-11e4-833e-a79a52b835d0.png)

With `settings.setGLVersion(3, 3);`:

![2](https://cloud.githubusercontent.com/assets/1072168/5595039/539a2b04-9279-11e4-8eba-02d1e67f1317.png)

It's not just `ofDrawCircle`, every draw command is not working.

This happened after multidraw branch merge, before that programmable renderer worked fine.

ping @arturoc I guess :)

arturoc:
that's really strange, i've tested most graphics examples with both renderers but some last minute change must have broken something, let me take a look

arturoc:
mmh, it's working without problem for me, what OS and graphics card are you using? also do you see any errors on the console?

procedural:
```
Starting /mnt/space/dev/ofApp/bin/ofApp...
[verbose] ofShader: unload(): detaching and deleting GL_FRAGMENT_SHADER shader from program 25
[verbose] ofShader: unload(): detaching and deleting GL_VERTEX_SHADER shader from program 25
[verbose] ofShader: unload(): detaching and deleting GL_FRAGMENT_SHADER shader from program 15
[verbose] ofShader: unload(): detaching and deleting GL_VERTEX_SHADER shader from program 15
[verbose] ofShader: unload(): detaching and deleting GL_FRAGMENT_SHADER shader from program 5
[verbose] ofShader: unload(): detaching and deleting GL_VERTEX_SHADER shader from program 5
[verbose] ofShader: unload(): detaching and deleting GL_FRAGMENT_SHADER shader from program 13
[...]

PR #3526, opened 2015-01-05
Fix for Direct State Access on Windows

### LINKED ISSUE #3523: OpenGL drawing broken on Windows
Neither programmable GL nor fixed function GL seem to work currently under windows 8.1.

This is what I'm seeing with the advanced3D example (notice the missing primitives):
capture

I'm using VS2012, with a NVIDIA GT 750M with the latest Game Ready driver. Older, compiled projects still run, so we can rule out a driver issue.

I suspect it has something to do with an `ifdef` gone astray during one of the latest refactors, or a broken shader mapping. 

Trying to hunt this down currently, this issue in case someone else encounters this.

Possibly related to #3514

tgfrerer:
fixes #3523

* uses `glCreateBuffers` instead of `glGenBuffers` if direct state access (OpenGL 4.5+) is available.

glGenBuffers does not actually create a buffer, it just returns the next available name, and only a subsequent
call to bind() will actualy initialize the buffer in memory.

This is why, for direct state access, we need to call glCreateBuffers(), so that the buffer is initialized when we pin data to it using setData()

Before this change, OpenGL 4.5 enabled cards would not draw any VBO based graphics.

This might also be related to #3514

See also: https://www.opengl.org/registry/specs/ARB/direct_state_access.txt

### DIFF
--- a/libs/openFrameworks/gl/ofBufferObject.cpp
+++ b/libs/openFrameworks/gl/ofBufferObject.cpp
@@ -1,12 +1,32 @@
 #include "ofBufferObject.h"
 #include "ofConstants.h"
 #include "ofAppRunner.h"
+#include "GL/glew.h"
 
 ofBufferObject::Data::Data()
 :id(0)
 ,size(0)
 ,lastTarget(GL_ARRAY_BUFFER){
-	glGenBuffers(1,&id);
+	
+	// tig: glGenBuffers does not actually create a buffer, it just 
+	//      returns the next available name, and only a subsequent 
+	//      call to bind() will actualy initialize the buffer in
+	//      memory. 
+	//
+	//      This is why, for direct state access, we need to call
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!!

procedural:
> This might also be related to #3514

I'll test it right now!

arturoc:
just added some ifdefs to check that those features were really present or else it won't compile on older implementations

arturoc:
tim, if you can check that it still works for you and it's using the 4.5 features?

tgfrerer:
@arturoc  just checked it - it's all good =)

admsyn:
We have a winner! Looks like [this merge commit](https://github.com/openframeworks/openFrameworks/commit/636d520bdde2e5f8abbc050b4cd09914e0c39d4c) was commit number 10,000 :)

bakercp:
+1 :beers:

PR #3527, opened 2015-01-06
Apothecary Formula - openssl environment fixes

i-n-g-o:
Hello.
I ran into problems when using apothecary to build openssl on OSX 10.9.5.

i don't have LANG set in my environment, so the script failed when accessing $LANG.
Surrounding it with "if test ${LANG+defined};" helps.

my LC_CTYPE is set to UTF-8. got this error:
sed: RE error: illegal byte sequence

as pointed out here
http://stackoverflow.com/questions/19242275/re-error-illegal-byte-sequence-on-mac-os-x#19770395
it can be fixed setting LC_CTPYE=C

### DIFF
--- a/scripts/apothecary/formulas/openssl.sh
+++ b/scripts/apothecary/formulas/openssl.sh
@@ -106,13 +106,35 @@ function build() {
 
 			echo "Using Compiler: $THECOMPILER"
 
-			# patch the Configure file to make sure the correct compiler is invoked.
+			# unset LANG if defined
+			if test ${LANG+defined};
+			then
+				OLD_LANG=$LANG
+				unset LANG
+			fi
 
-			OLD_LANG=$LANG
-			unset LANG
-				sed -ie "s!\"darwin-i386-cc\",\"cc:-arch i386 -g3!\"darwin-i386-cc\",\"${THECOMPILER}:-arch i386 -g3!" Configure
-				sed -ie "s!\"darwin64-x86_64-cc\",\"cc:-arch x86_64 -O3!\"darwin64-x86_64-cc\",\"$THECOMPILER:-arch x86_64 -O3!" Configure
-			export LANG=$OLD_LANG
+            # unset LC_CTYPE if defined
+            if test ${LC_CTYPE+defined};
+            then
+                OLD_LC_CTYPE=$LC_CTYPE
+                unset LC_CTYPE
+            fi
+
+            # patch the Configure file to make sure the correct compiler is invoked.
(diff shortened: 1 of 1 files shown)


bakercp:
Strange -- do you have command line tools installed?  Or are your language preferences set to something other than the default?  Or ... is your bash config unsetting LANG?  

Just trying to understand why this hasn't come up before.  Thanks!

i-n-g-o:
good questions. i don't think that something is unsetting LANG.

i had to rescue my system a while ago. before that im am sure i had the command line tools installed. they are still sitting in /Library/Developer/CommandLineTools. but when i do a $ xcode-select --install it wants to install the command line tools. so it might be that the rescue screwed things up a bit...

bakercp:
I'd run `xcode-select --install` and let it install the latest command line tools then try to run the unmodified apothecary script again ...

i-n-g-o:
jep. i would. but it's currently not available from the Update Server. I'll try again later.

bakercp:
You can also download it directly from the developer.apple.com site.

i-n-g-o:
Just installed it. The installer did not change anything about my LANG setting... so the original script still fails.

bakercp:
OK.  good to know!  Thanks.  Any other ideas why your setup might be different?

i-n-g-o:
hm. no idea what could cause this. (i checked my terminal and system settings)
seems others have this issue as well:
http://stackoverflow.com/questions/7165108/in-osx-lion-lang-is-not-set-to-utf8-how-fix

bakercp:
Do you have everything set up according to these SO answers?

https://apple.stackexchange.com/questions/21096/where-does-lang-variable-gets-set-in-mac-os-x

bakercp:
(Sorry, I just saw that you already checked your system and terminal settings) ... so you may have already seen the last link).

i-n-g-o:
jep, i have found this post as well... thanks for checking.
weird.

danoli3:
Make sure to restart after installing command line tools. That can be a tricky undocumented gotcha.
Try again now with the latest master again.

i-n-g-o:
the problem on my machine is, that the locales are not fully set up.
fixing this would be easy.

the point is, that the build script should not fail because of that but rather should be able to handle not fully setup locales. (by testing variables before using them)

arturoc:
this might be a problem that only happens in computers setup in something different to english, the code seems to be taking into account that the variable is normally set so it shouldn't hurt to merge it and will be useful if it's the case that some computers have this problem.

danoli3:
Yeah good point @arturoc 
Okay just tested this for when it is defined and all is still working as expected

danoli3:
Okay good to merge now @arturoc

arturoc:
thanks!

PR #3528, opened 2015-01-07
fixes some minor issues with the compute shader particles example

tgfrerer:
* makes sure to issue the right number of workgroups + add comment outlining reasoning
* use GL_STREAM_DRAW as usage hint - although this does not make a difference on my driver implementation, but is more correct, since the particle data is constantly read and written on the GPU
* adds key toggle for fullscreen


arturoc:
thanks!

PR #3533, opened 2015-01-09
fix calculation of totalVerts for VBOs in setVertexBuffer()

tgfrerer:
+ if the buffer does not give us a stride, we try to calculate based on the data size and the number of coordinates (we assume each coordinate has the size of a float == 4 bytes)
+ if the buffer gives us a stride we use the attribute stride to calculate the total number of vertices.

### DIFF
--- a/libs/openFrameworks/gl/ofVbo.cpp
+++ b/libs/openFrameworks/gl/ofVbo.cpp
@@ -680,11 +680,11 @@ void ofVbo::setVertexBuffer(ofBufferObject & buffer, int numCoords, int stride, 
 #endif
 	positionAttribute.setBuffer(buffer, numCoords, stride, offset);
 	bUsingVerts = true;
-	// TODO: Check: we have no perfect way of knowing the new number of total vertices,
-	// since the buffer does not tell us, so we try to calculate based on the data size
-	// and the number of coordinates.
-	totalVerts = buffer.size() / (numCoords * sizeof(float));
-
+	// Calculate the total number of vertices based on what we know:
+	// + if the buffer does not give us a stride, we try to calculate based on the data size
+	//   and the number of coordinates.
+	// + else we use the attribute stride to calculate.
+	totalVerts = buffer.size() / (stride == 0 ? (numCoords * sizeof(float)) : stride );
 }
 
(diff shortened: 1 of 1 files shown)


ofTheo:
Thanks for this PR. 
It seems like an easy one to merge except we strongly discourage ternary operators in OF. 
( https://github.com/openframeworks/openFrameworks/wiki/oF-code-style )

Sorry for the hassle!

tgfrerer:
Hmm. In principle I would agree, and I wouldn't use ternary ops in examples and client facing code. I don't believe that there is a real performance benefit in using them (although there could be since it makes it easier for the compiler to optimise), but I thought:

     totalVerts = buffer.size() / (stride == 0 ? (numCoords * sizeof(float)) : stride );

was more readable than:

	int tmpStride = stride;
	if (stride == 0) tmpStride = numCoords * sizeof(float);
	totalVerts = buffer.size() / tmpStride;

since this way the heart of the method is buried in two extra lines you have to keep in memory when reading the code. I can change it if you insist =)

bakercp:
I know it's documented, established and all, but in internal cases like this I'd +1 ternary ...

ofTheo:
One of the reasons we don't encourage them either internally or in examples is they are not as easy to parse and that makes them a little harder to review, a little less clear what is going on and increases the likelihood of bugs.  

To me something like below would be a lot easier to follow, fix and debug 
```
int tmpStride = stride;  
if( tmpStride == 0 ){  
    //if we can't get a stride from the buffer lets calculate based off number of coords
    tmpStride = numCoords * sizeof(float);
}
totalVerts = buffer.size() / tmpStride; 
```
Not sure if there should be a dived by zero check in the last line as well?

Anyway I just wanted to explain a bit why we discourage them in the code style guide. 
It might be that this case is a fairly reasonable use of them, but if we had them used extensively I could see them causing more problems than they are solving.

bakercp:
@ofTheo understood.  And yes, probably check that div by zero ...

ofTheo:
thanks @tgfrerer and sorry to nitpick!  =)

tgfrerer:
it all makes sense =)

PR #3535, opened 2015-01-09
Fix #3534

### LINKED ISSUE #3534: ofxOsc doesn't make use of boolean OSC 1.1 specifications
`/to/my/boolean/path T`
nor this one:
`/to/my/boolean/path F`

I will propose you some PullReq to make it possible.
Interested?

groolot:


ofTheo:
Thanks for this PR! 
This looks fairly good to me. 

The only thing that I would be nervous about is changing the values of getTypeName(). 
If people are using that in their code and checking its value then the new values ( which are more correct in terms of the spec ), might cause issues. 

@bakercp @bilderbuchi what do you guys think?

addons/ofxOsc/src/ofxOscMessage.cpp
```diff
@@ -142,6 +154,26 @@ float ofxOscMessage::getArgAsFloat( int index ) const
...
+    if ( getArgType(index) != OFXOSC_TYPE_DOUBLE )
+    {
+        if ( getArgType( index ) == OFXOSC_TYPE_INT32 )
+        {
+            ofLogWarning("ofxOscMessage") << "getArgAsDouble(): converting double to int32 for argument " << index;
+            return ((ofxOscArgInt32*)args[index])->get();
+        }
+        else
+        {
+            ofLogError("ofxOscMessage") << "getArgAsDouble(): argument " << index << " is not a number";
```
bilderbuchi: what about the other number types (FLOAT, INT64) - those are not accounted for, and would erroneously trip this error.

addons/ofxOsc/src/ofxOscMessage.cpp
```diff
@@ -142,6 +154,26 @@ float ofxOscMessage::getArgAsFloat( int index ) const
 }
 
 
+double ofxOscMessage::getArgAsDouble( int index ) const
+{
+    if ( getArgType(index) != OFXOSC_TYPE_DOUBLE )
+    {
+        if ( getArgType( index ) == OFXOSC_TYPE_INT32 )
+        {
+            ofLogWarning("ofxOscMessage") << "getArgAsDouble(): converting double to int32 for argument " << index;
```
bilderbuchi: I think there's a mixup here - this function should return a double. If the arg is an int, is should cast the int to double (not the other way round). Also, why warn about it?

addons/ofxOsc/src/ofxOscMessage.cpp
```diff
@@ -237,12 +278,16 @@ ofxOscMessage& ofxOscMessage::copy( const ofxOscMessage& other )
...
-		if ( argType == OFXOSC_TYPE_INT32 )
+		if ( argType == OFXOSC_TYPE_TRUE or argType == OFXOSC_TYPE_FALSE )
+			args.push_back( new ofxOscArgBool( other.getArgAsBool( i ) ) );
+		else if ( argType == OFXOSC_TYPE_INT32 )
 			args.push_back( new ofxOscArgInt32( other.getArgAsInt32( i ) ) );
 		else if ( argType == OFXOSC_TYPE_INT64 )
 			args.push_back( new ofxOscArgInt64( other.getArgAsInt64( i ) ) );
-		else if ( argType == OFXOSC_TYPE_FLOAT )
-			args.push_back( new ofxOscArgFloat( other.getArgAsFloat( i ) ) );
+        else if ( argType == OFXOSC_TYPE_FLOAT )
```
bilderbuchi: indentation messed up here

addons/ofxOsc/src/ofxOscMessage.cpp
```diff
@@ -237,12 +278,16 @@ ofxOscMessage& ofxOscMessage::copy( const ofxOscMessage& other )
 	for ( int i=0; i<(int)other.args.size(); ++i )
 	{
 		ofxOscArgType argType = other.getArgType( i );
-		if ( argType == OFXOSC_TYPE_INT32 )
```
bilderbuchi: not really the submitter's fault, but why on earth is this not a switch statement instead of if/else tree? we're testing for enums, after all...

addons/ofxOsc/src/ofxOscReceiver.cpp
```diff
@@ -39,8 +39,12 @@ ofxOscReceiver::ofxOscReceiver()
 	listen_socket = NULL;
 }
 
-void ofxOscReceiver::setup( int listen_port )
+void ofxOscReceiver::setup( int listen_port, bool allowReuse )
```
bilderbuchi: where do these unrelated changes come from? they have nothing to do with the PR subject, and that's something that @ofTheo has already merged in, no? I'd be hesitant to touch the reuse code again, tbh.

addons/ofxOsc/src/ofxOscSender.cpp
```diff
@@ -47,12 +47,16 @@ ofxOscSender::~ofxOscSender()
 		shutdown();
 }
 
-void ofxOscSender::setup( std::string hostname, int port )
+void ofxOscSender::setup( std::string hostname, int port, bool enableBroadcast )
```
bilderbuchi: also, the broadcast changes appear here - was there some mixup with the PR branch root?

addons/ofxOsc/src/ofxOscSender.cpp
```diff
@@ -76,7 +80,7 @@ void ofxOscSender::sendBundle( ofxOscBundle& bundle )
 	socket->Send( p.Data(), p.Size() );
 }
 
-void ofxOscSender::sendMessage( ofxOscMessage& message )
+void ofxOscSender::sendMessage( ofxOscMessage& message, bool wrapInBundle )
```
bilderbuchi: also, the bundle changes are here - why??

bilderbuchi:
one of your commits - c061f658f562e4deabbb8c3c5f5a733c74065509 - dragged in some unrelated changes. also, it is an apparent imperfect duplicate of 99103c8257926d1baebd1acafb267a588fdb7ba7 - did you rebase your branch or some other fishy git thing?

bilderbuchi:
personally, I would think that `getTypeName` is only a cosmetic function, readable strings for logging or whatever. If people did comparisons, they should use `getType` and compare against the enum values (which is even easier). therefore I think it would be fine to change this. @bakercp do you agree?

bilderbuchi:
I think we should not merge this until the unrelated changes I mentioned above are purged from the PR. After that, and the stuff I commented inline is fixed, it looks good to me.

groolot:
I realise that I made some Git mistakes, and finaly haven't resolved correctly the issue.
I propose you to not take care of that issue and I will resend you some modifications, based on the 0.8.4 branch, without bakercp's ofxOSC-update double's implements.

Do you agree ?

bilderbuchi:
please just base your PR branch off of the current master, that would be best. :-)

bilderbuchi:
the double changes, if corrected in the implementation, would be fine for me, too.

Issue #3051, opened 2014-06-26
API bug : ofFbo should have a clear()


elliotwoods:
following `clear()` the GPU memory should be freed and `isAllocated()` should then return `false`

kylemcdonald:
@mattfelsen is tackling this one right now. his approach is like this:

1. rename `ofFbo::destroy()` to `ofFbo::clear()`
2. create a new `destroy()` that calls `clear()`
3. deprecate `destroy()` and replace instances
4. add `clearAlpha()`, `clearColorBuffers()`, `clearDepthBuffers()` (making a separate issue for this right now) this would cover theo's use case. the naming comes from a discussion with @ofzach and @mattfelsen.

kylemcdonald:
closed by #3540 thanks again @mattfelsen

PR #3538, opened 2015-01-10
Add ofFbo::clear() for consistency with other classes

mattfelsen:
Just added a clear() method to ofFbo, which just calls destroy(), but adds consistency with other classes like ofImage and ofTexture. One thing to point out: destroy() is a private method, so I'm adding a public method which only calls a private method...seems kind of weird?

Addresses issue #3051


kylemcdonald:
ah i didn't explain this correctly, let me describe it again :)

mattfelsen:
Closing this in favor of pull #3540

PR #3540, opened 2015-01-10
Deprecate ofFbo::destroy() in favor of clear()

mattfelsen:
A different approach to issue #3051

### DIFF
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -287,7 +287,7 @@ ofFbo::ofFbo(const ofFbo & mom){
 
 ofFbo & ofFbo::operator=(const ofFbo & mom){
 	if(&mom==this) return *this;
-	destroy();
+	clear();
 	settings = mom.settings;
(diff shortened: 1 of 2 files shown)


kylemcdonald:
looks good! i'm going to wait for one more person to sign off before merging.

bilderbuchi:
looks like a straightforward-enough replacement.

kylemcdonald:
this got a few eyes on it from discussion in another PR, so i'm merging.

arturoc:
the only problem i see with this is that some legacy ofFbo implementations used to have a clear which cleared the background, even if we are not going that route having a clear method might be confusing for some people

mattfelsen:
How about an ofLogWarning() if you call begin() with an unallocated FBO?

kylemcdonald:
@arturoc good point. i think the best solution could be to add a brief line of doxygen documentation above `clear()` that explains exactly what it does (maybe explicitly mentions that you should use `ofClear()` if you want to clear the background) and then also add the error message (i think it's an error, not a warning) if you call begin() unallocated.

Issue #3539, opened 2015-01-10
ofFbo::clearAlpha(), clearDepthBuffers(), clearColorBuffers()


kylemcdonald:
this is an alternative proposal so solve a use case from @ofTheo in #3051 which was also present in many of the older ofxFbo/ofxFBO addons before ofFbo.

for an example of clearing alpha, see https://github.com/kylemcdonald/DohaInstallation/blob/master/addons/ofxFbo/src/ofxFbo.cpp#L142-L151

kylemcdonald:
closed, going in a different direction as per discussion in #3542

PR #3542, opened 2015-01-10
Add ofFbo::clearAlpha()

mattfelsen:
Adds clearAlpha() as mentioned in #3539 from old ofxFbo implementation

### DIFF
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -287,7 +287,7 @@ ofFbo::ofFbo(const ofFbo & mom){
 
 ofFbo & ofFbo::operator=(const ofFbo & mom){
 	if(&mom==this) return *this;
-	destroy();
+	clear();
 	settings = mom.settings;
 	bIsAllocated = mom.bIsAllocated;
 
@@ -319,7 +319,7 @@ ofFbo & ofFbo::operator=(const ofFbo & mom){
 }
 
 ofFbo::~ofFbo(){
-	destroy();
+	clear();
 }
 
 int	ofFbo::maxColorAttachments() {
@@ -338,7 +338,7 @@ int	ofFbo::maxSamples() {
 }
 
 
-void ofFbo::destroy() {
+void ofFbo::clear() {
 	if(fbo){
 		releaseFB(fbo);
 		fbo=0;
@@ -365,6 +365,10 @@ void ofFbo::destroy() {
 	for(int i=0; i<(int)colorBuffers.size(); i++) releaseRB(colorBuffers[i]);
 	colorBuffers.clear();
 	bIsAllocated = false;
+}
+
+void ofFbo::destroy() {
+	clear();
 }
 
 bool ofFbo::checkGLSupport() {
@@ -426,7 +430,7 @@ void ofFbo::allocate(int width, int height, int internalformat, int numSamples) 
 void ofFbo::allocate(Settings _settings) {
 	if(!checkGLSupport()) return;
 
-	destroy();
+	clear();
 
 	// check that passed values are correct
 	if(_settings.width <= 0 || _settings.height <= 0){
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -371,6 +371,17 @@ void ofFbo::destroy() {
 	clear();
 }
 
+void ofFbo::clearAlpha() {
+	begin();
+
+	glColorMask(0, 0, 0, 1);
+	glClearColor(0, 0, 0, 1);
+	glClear(GL_COLOR_BUFFER_BIT);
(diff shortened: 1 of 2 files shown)


kylemcdonald:
again, looks good to me. it mixes some other work from today into here, but presumably the other PR gets merged and it's not a problem.

i suspect there might be some oversimplification here wrt one of two things:

1. newer versions of opengl not support these functions (seems to always be the case.. not sure here)
2. something special we need to do to handle the multiple color buffers

if someone can check that this isn't completely wrong, it would be appreciated :)

tgfrerer:
+1 on the `destroy` deprecation, and clearing that up.

Not sure about `clearAlpha()`.

Why not clear the alpha within `ofFbo::begin()` / `ofFbo::end()` in your draw()? We can already do this using `ofClearAlpha`. It would look like:

	ofFbo myFbo;
	
	ofApp::draw(){
		myFbo.begin();
		ofClear(0); /// clearing my fbo's alpha and current color buffer
		ofSetColor(ofColor::white);
		ofDrawRectangle();
		ofClearAlpha();   /// now my fbo's alpha channel is cleared
		... /// some more drawing happens here
		myFbo.end();
	
		ofClearAlpha(); // clear screen alpha
		ofClear(0); // clear screen alpha + color
		myFbo.draw(); // draw fbo to screen
	}

-- 
I'm worried that if we allow an FBO to clear its own alpha we're working a bit against DRY and our own current model of how we abstract OpenGL in openFrameworks. Also, 

1. it changes the bind / unbind state of the FBO as a side-effect (so we'd have to introduce some complexity to only bind/unbind if necessary) otherwise - 
2. it binds / unbinds the FBO unnecessarily - and binding / unbinding FBOs is one of the single most costly GL operations.

If we implemented `ofFbo::clearAlpha`, why shouldn't we also have `ofFbo::clearColor` (which might name-clash with the above proposed, better use of the function) or even an `ofFbo::setColor()`?

I understand an `FBO` / `ofFbo` as a **drawing context** and not as a *drawing object* - and although we program in an object oriented language, OpenGL and it's abstractions for the most part isn't really object oriented.

That said, it seems that OpenGL and GLNEXT are moving into a more truly object oriented system of abstractions, but until then, i think our code might be easier to maintain if we try to keep consistent with the current OpenGL model =)

kylemcdonald:
I didn't realize there was already ofClearAlpha()! This also makes me feel
that we should stick with FBO as a context not an object.

@ofZach @ofTheo is ofClear() and ofClearAlpha() enough, or is there a good
argument for making it a method of an FBO?

Maybe the better thing to add is just ofClearDepth().

ofZach:
I think it's fine to stick closer to opengl -- I really do miss the clear function from earlier ofxFbos but it's ok...  I generally feel like p5's pgraphics style is much easier to read and understand -- but I think that's not really possible in OF: 

https://processing.org/reference/PGraphics.html

I do think producing some more FBO examples would be helpful -- especially wrt alpha which generally trips people up (understanding how clear, alpha, drawing with alpha in an FBO, etc all interact in unexpected ways).

arturoc:
I agree with tim, except that i think ofFBO is really object oriented :) the reason to not have clearAlpha, setColor... is just a matter of responsibility: an FBO is a rendering surface, a window is a rendering surface too. We don't have window.clear() or window.setColor and the same way it doesn't make sense to have fbo.clear() or any other method related to rendering.

The equivalent to a PGraphics in OF would be an ofRenderer not an ofFBO which is just a way of using in an OO manner the corresponding GL object.

with the changes i did for multiwindow i also refactored the renderers quite a bit making it possible to use them explicitly as it's shown in apps/devApps/explicitRendererExample. it's still an unstable api since it's not as complete as the c one in ofGraphics... (mainly misses the overloads to be able to do draw(something,x,y) or draw(something,pos). 

then a renderer has a rendering surface which is usually a window but can be set to an FBO

another reason is that adding methods like clear* or anything else related to rendering will make the fbo dependent on the global renderer which i've minimized to only the draw methods in the multiwindow refactoring to make it possible to use OF from different threads... as long as you use a explicit renderer instead of the C api and the draw methods.

something that i think we should add though is a push/pop style in the fbo begin/end, i always have problems thinking the state of the renderer is the same as it was before beginning an fbo and get confused for a while until i realize that i left the color at black for example and that's why the fbo seems to not be drawing.

kylemcdonald:
ok, thank you @arturoc for the very complete explanation, and @ofzach for signing off :)

@mattfelsen thanks again for your work on this. i'm going to close this PR since we've decided to go in a different direction, but it's essential to have real code to look at and discuss when making these kinds of decisions i think.

mattfelsen:
@kylemcdonald No worries!

I didn't want to lose arturo's suggestion about push/popping styles (which I think is a great idea) so I created issue #3547

And my mistake on the previous commit creeping in – my stash wouldn't apply cleanly on upstream/master so I started the feature branch of my local master which already had the destroy -> clear change in there. Whoops :P

PR #3551, opened 2015-01-12
fix TARGET_WIN32 compile error in ofAppGlutWindow

tgfrerer:
Replaces ofAppPtr by instance->events() as glfw window does it.
+ this only affects Windows builds, as it only appears within the TARGET_WIN32 #ifdef

@arturoc, I hope this is the right way to make sure the right window instance receives the drag event - I copied GLFW window behaviour.


arturoc:
yep, thanks!

PR #3544, opened 2015-01-11
Update ofMath.h

manish-juneja:
added overloaded ofDistSquared and ofDist function to calculate distance between 2 points in 3D space.

PR #3562, opened 2015-01-15
Remove errant + chars in ofMath

admsyn:
Looks like #3544 pulled in some + chars in ofMath, breaking compilation (I guess a copy+paste from a git diff?).

`<joke>`
There's too much math in ofMath. This makes there be less math.
`</joke>`

ping @ofTheo / @manish-juneja


ofTheo:
thanks! @admsyn  :)   math -- !

Issue #1758, opened 2012-12-17
Xcode MacOSX - iOS use "rysnc" rather than "cp" when copying "bin/data" files


danoli3:
Rsync offers faster speed overall in build times when copying files over to the device or to the target.
It uses compressed transfers, checksum comparisons (so not re-copying the same unchanged files over and over again) and is even used by Xcode when transferring files linked in projects structure

Proposed change to Run Script:
```
rsync -avz --exclude='.DS_Store' "${SRCROOT}/bin/data/" "${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
```
Manual page for Rsync:
http://developer.apple.com/library/mac/#documentation/Darwin/Reference/ManPages/man1/rsync.1.html

The current run script is:
```
cp -rf bin/data/ "$TARGET_BUILD_DIR/$PRODUCT_NAME.app"
```
Manual page for cp:
http://developer.apple.com/library/mac/#documentation/Darwin/Reference/ManPages/man1/cp.1.html

This system still has the same problem as the "cp" command by not removing files in the target location that have been removed from the source directory. (Which can only currently be resolved by cleaning the project, or manually deleting the files in the app structure).

Rysnc does allow for you though to delete everything within the destination folder / app other than certain files and folders, so this could be adapted to not delete Xcode objects / folders however I think the clean method is okay for now.

PR #3563, opened 2015-01-16
Fixes: #1758 iOS use "rysnc" rather than "cp" when copying "bin/data" files

danoli3:
See issue: #1758

This follows the same method that Xcode uses to copy files to App
packages.
- Will not copy files if already on destination (this will increase
performance).
- If file has been updated and already exists on the target app, you
will need to clean build (same workflow as Xcode native adding
resources).
- Tested working with iOS projects to simulator and device.

Currently CP copys the files to the .app folder everytime, updating the modification datetime and therefore causing the copy times to devices every single build. So say you build a project to your iPad... every single time you have to transfer all the files in your data folder to the target device, even if nothing has changed.

The new method will copy a new file once, like Xcode, to the .app and then not update it until a clean.
This means that the files are cached and build / test times are almost instant after the first deployment.

### DIFF
--- a/scripts/ios/template/emptyExample.xcodeproj/project.pbxproj
+++ b/scripts/ios/template/emptyExample.xcodeproj/project.pbxproj
@@ -523,7 +523,7 @@
 			<key>shellPath</key>
 			<string>/bin/sh</string>
 			<key>shellScript</key>
-			<string>cp -rf bin/data/ "$TARGET_BUILD_DIR/$PRODUCT_NAME.app"
+			<string>rsync -avz --exclude='.DS_Store' "${SRCROOT}/bin/data/" "${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
 
 
 


julapy:
@danoli3 has given me a demo copying over a 700MB file to a iOS device during build,
and rsync makes a huge difference. the `cp` method copies over the files every time, where as rsync copies only the file that have changed... in this case its the difference between an instant build and a 10sec wait.

@ofTheo @admsyn you guys want to take a quick gander?
+1 from me.

bilderbuchi:
in general, I have only made good experiences (backing up stuff) with rsync. :-)

admsyn:
Same here, looks like a pretty clear win :+1:

kylemcdonald:
i just noticed my .app starting to get really big (i have a few GB of data in my data/ folder), and i can see that this script is also running on OSX on desktop. is that supposed to happen?

danoli3:
Not at all. Not sure what's happening on your end man

ofZach:
@kylemcdonald what script do you mean wrt desktop ?

this is what we have in the template: 

```
rsync -aved ../../../libs/fmodex/lib/osx/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/"; install_name_tool -change ./libfmodex.dylib @executable_path/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME";
mkdir -p "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Resources/"
rsync -aved "$ICON_FILE" "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Resources/"
rsync -aved ../../../libs/glut/lib/osx/GLUT.framework "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Frameworks/"
```

it's just handling icon file and frameworks but doesn't do anything with your data folder.

kylemcdonald:
sorry, by desktop i meant just meant OSX :)

when i download a fresh copy of OF 0.9.2 and create a project with the project generator, then check "run script" under "build phases" i see this:

```
mkdir -p "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Resources/"
# Copy default icon file into App/Resources
rsync -aved "$ICON_FILE" "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Resources/"
# Copy bin/data into App/Resources
rsync -avz --exclude='.DS_Store' "${SRCROOT}/bin/data/" "${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/data/"
# Copy libfmod and change install directory for fmod to run
rsync -aved ../../../libs/fmodex/lib/osx/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Frameworks/";
install_name_tool -change @executable_path/libfmodex.dylib @executable_path/../Frameworks/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME";
# Copy GLUT framework (must remove for AppStore submissions)
rsync -aved ../../../libs/glut/lib/osx/GLUT.framework "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Frameworks/"
```

you can see the line that says `rsync -avz --exclude='.DS_Store' ...` copies the data folder into resources, if i understand correctly.

afaict it's coming from this line in the osx template https://github.com/openframeworks/openFrameworks/blob/035af4555da07ec7cd332cec4ee75af9e1bffdda/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj#L654

ofZach:
ping @danoli3 -- it's seems to have been introduced by your commit 

https://github.com/openframeworks/openFrameworks/commit/89fbd2df7733e8c97a1ddc37e1fe5fbd1f0976a8

I think this commit has something to do with code signing for OSX desktop applications?   I don't think this rsync of data folder is something we should do by default typically and I prefer to maybe include this as a special template and not the default OSX template if it's useful to have this around (similar to how we have a no fmod project template) 

ping @ofTheo -- this seems really important to look at....

danoli3:
Gotcha, yeah I was confused by the desktop ref haha.

The reasoning for this change was to support and make it up to the current standard of OS X App packaging that Xcode does by default for included files.

See https://github.com/openframeworks/openFrameworks/issues/4619

This command allows for sand boxing and packaging of A given OS X app onto the AppStore or Developer distribution. (Approved way).

If you don't want to package the data into the app for an implementation, comment out that rsync command(obviously if you just want a private app that is modifying and writing to the bin/data folder it may not be necessary to package the assets).

Might be a good idea to split the rsync command into a separate script command in Xcode to make it easier to comment out.

ofZach:
I personally think this rsync command should rather be commented out by default :)    -- ie, 99% of the cases don't need this command and it's kind of doing something weird and not necessarily useful (or transparent).  

also, we have templates, and these changes that are specific to app store submission feel maybe more appropriate for a template that's specific to app store apps than a general OSX template...

ofZach:
PS:  I think it's great that we have this, I just think the templates help make it more clear when a project is doing something specific.  See the "nofmod" one for example, which is what we use for the project generator itself -- it removes fmod so we can make a command line app without the baggage, etc.

danoli3:
It is a deviation from the standard OSX template yeah in some respects, however it also brings commonality between iOS and OSX templates. The same thing happens on iOS.

I'll check out the no fmod when I'm back in Syd

ofTheo:
hey just catching up on this. 
I agree we shouldn't be doing the rsync by default as that could end up being a really quick way to fill up your hard drive. 

@danoli3 could we do the rsync command as part of the Archive build process?
Or simply have a third target Deploy 

ie: myApp Debug
     myApp Release
     myApp Deploy  

Deploy would be the only one to copy the files to the .app bundle. 
We could also use a Macro to change the data path to the copied data folder location.

danoli3:
Yes we could add a deploy target (distribution may be better word)

@julapy and I spoke about making a distribution target for iOS a while ago as well, so might be a good approach for The Xcode projects.

I have some code to make this happen, I believe you just define a new command like DISTRIBUTION=1 in the commands where DEBUG=1 and then use that in the bash script to determine what is used.

arturoc:
agreed this shouldn't be default, the problem kyle was having is that his application was probably generating some files in data which get copied in next compile. That's a pretty common thing to do in an OF application and you don't want those files to end up in the bundle.

bilderbuchi:
you could filter the contents of /data so that it only sync over stuff that is not gitignored (as generated files would be).

ofZach:
nice suggestion @bilderbuchi but I really think this behavior should be opt-in.   The package contents aren't readily accessible (you have to right click a file and say explore package contents) and this generally breaks with how we've been doing things for years.   The nice thing about the template system from 0.9.0 is that this sort of behaviour is perfect for a  template since it represents a kind of edge case.  The templates came kind of late in the PG dev cycle but it will be great to expose them more in the gui going forward (they are available w the command line tool).

danoli3:
Okay I have a working solution that fixes it for all parties :D! See PR.

Issue #861, opened 2012-01-02
ofxSynth needs an overhaul


viniciolindo:
ofxsynth doesn't wont to compile because this line:
source->audioOut( working, numFrames, numChannels, tickCount ) in ofxSoundUnit.cpp is wrong.
the canddates are:

virtual void audioOut( float * output, int bufferSize, int nChannels, int deviceID, long unsigned long tickCount  )
virtual void audioOut( float * output, int bufferSize, int nChannels)

PR #3565, opened 2015-01-16
Removes ofxSynth from core addons

### LINKED ISSUE #3564: Remove ofxSynth
This was discussed around the time of the YCAM dev meet, and tucked into the mega-sound-PR that has since been broken down into smaller parts.

Should we remove ofxSynth from the core addons? If I remember correctly, the rationale is that at this point it's just a heavily outdated version of [ofxMaxim](https://github.com/micknoise/Maximilian) and we could be pointing people towards that instead.

Related: #861 

@openframeworks/owners @openframeworks/sound

admsyn:
Closes #3564 and #861

After this, running `ack ofxSynth` on the OF root only gets a hit in the changelog.


kylemcdonald:
:+1: thanks for handling the other platforms, build scripts, readme, etc!

Issue #3558, opened 2015-01-15
ofPolyline getCentroid2D() can return NAN / infinity coordinates


armadillu:
On perfectly decent ofPolylines (meaning with valid non-NAN points), you do get sometimes a centroid with x / y that's NAN. I suspect its related to the calculated area being zero. I can't fully test now, but I was getting NANs which where propagating to other parts of the code and ultimatelly leading to crashes while working with lots of user-generated ofPolylines.

I switched to .getBoundingBox().getCenter() which was good enough for the project, and I got no more NAN's.

ofZach:
good test example (straight line = 0 area): 

    ofPolyline line;
    
    line.addVertex(0,0);
    line.addVertex(20,20);
    line.addVertex(40,40);
    
    cout << line.getCentroid2D() << endl;

armadillu:
Ah of course, I didn't think of this case. I guess it's not a really a bug then...

ofZach:
I've always felt the centroid is considering the polyline as a shape, but I prefer to just compute the centroid of the points as a point set (ie, add up all the pts and find the average).  I'm not sure of what the difference is, but I still feel like a polyline without an area should have a midpt.

ofZach:
ah I take this back thinking about it a bit more (I think this makes sense if the line is uniformily sampled, but not so much if it's not uniform), but maybe, we should have some error handling when the area < epsilon, etc.

bilderbuchi:
mathematically, it's pretty clearly defined:

> In mathematics and physics, the centroid or geometric center of a two-dimensional region is the arithmetic mean ("average") position of all the points in the shape. 

Here, "points" are the infinite points (e.g. of mass) making up the shape, not the vertices of a polygon.

However, there's a [formula](http://en.wikipedia.org/wiki/Centroid#Centroid_of_polygon) given for the centroid of a polygon, are we using that one? I haven't yet done the math manually to see if this holds up with degenerate points, but I'd assume so from a quick glance.

bilderbuchi:
ok, the formula [we use](https://github.com/openframeworks/openFrameworks/blob/39e374059aa18f62c5590789ca0055302490c2c5/libs/openFrameworks/graphics/ofPolyline.cpp#L962-L980) is the same as the one from WP, with a workaround for the fact that our polygons are not closed, in that the first and last point are not identical - is that still correct?
It could also be that the requirement of non-selfintersecting polygons is violated in some polygons where this fails?
[...]

PR #3567, opened 2015-01-18
fixed bug with ofPolyline::getCentroid2D() returning NaN and infinity

kylemcdonald:


kylemcdonald:
going to merge this since two weeks have passed without objection, and it seems to be the agreement of everyone in #3558

next step is possibly to replace this and other zero comparisons with ofIsZero()

PR #3569, opened 2015-01-18
Serial and arduino doxygen documentation from ofSite

HalfdanJ:
Documentation of communication folder

PR #3570, opened 2015-01-18
Serial enumerateDevices deprecation

HalfdanJ:
Function was already marked as deprecated in comment. Based on the #3569 branch, so merge that first


kylemcdonald:
good find, and this looks correct.

bakercp:
+1 for this.

bakercp:
The #3569 was already merged, so this should be fine to merge now.

PR #3571, opened 2015-01-18
ofSoundDevice doxygen comments, fix warning

admsyn:
This PR:

- Makes ofSoundDevice a class (instead of a struct), makes it conform tighter to OF style
- Fixes a warning RE: above for ofSoundDevice (forward declared as a class, but actually a struct)
- Adds doxygen comments to ofSoundDevice, fixes outdated ones in ofSoundStream
- Fixes the inconsistent whitespace in ofSoundStream.h


kylemcdonald:
looking good!

Issue #2548, opened 2013-09-09
Compile times on Visual Studio 2012 slow


elliotwoods:
Currently our compile times on Windows are pretty slow.

I'm wondering if either @leocolomb might have some thoughts on this, or if @obviousjim could bend an ear at Microsoft to give some suggestions.

Each individual file is slow to compile, and VS doesn't compile in parallel as much as XCode (it compiles seperate projects in parallel within a solution, but not the files within the project). And I presume the duration is simply because there are so many headers associated with each file (so essentially each source file is enormous by the time it makes it the compiler).

I presume the official line is something like 'use precompiled headers!', and i'm curious what is involved in implementing that.

Maybe LLVM/GCC is doing some magic to automatically speed up the process (e.g. auto-precompiling headers), and perhaps we need to just squeeze MSVC in the right way to get it to act in the same way. Or VS2013 might bring along magic fixes.

I haven't had any positive experiences with precompiled headers yet, and am not pressing for them to be a core strategy.

LeoColomb:
Fixed by #3572 ? (and move to #255)

bilderbuchi:
I don't know (not a Win user), what's your opinion?

LeoColomb:
I don't know what is the targeted time...
With multiprocessor tag, the compile time is incredibly improved, but still around 1 minute.

Nevertheless, @elliotwoods's original remark is now fixed.

> Each individual file is slow to compile, and VS doesn't compile in parallel as much as XCode

admsyn:
IMO this can be closed, unless there's any other low-hanging fruit like the compiler flag? Especially given that vs2012 is presumably on the way out.

bilderbuchi:
alright then :-)

elliotwoods:
revisiting this. i'm looking into precompiled headers again
multi-processor compilation definitely helps! but i think good use of precompiled headers still might save the 8 tonnes of carbon dioxide being emitted each year from people compiling cpp's in oF/VS 
will make a new issue once i understand what's what / do some tests

elliotwoods:
Got this working in Visual Studio also

For a 2 file (main.cpp, ofApp.cpp) project with 1 line of code in draw
VS2015 / Xeon 6-core / x64 / Debg build:

Using PCH:

`
1>      283 ms  ClCompile                                  1 calls
1>     3408 ms  Link                                       1 calls
========== Build: 1 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========
`

[...]

PR #2571, opened 2013-09-17
Feature Visual Studio 2013 - WIP

LeoColomb:
Patch to begin the support of Visual Studio 2013, and Windows 8.1.

* Support full C++11 - #2335   
* Improve compile time   

---
- [x] Projects version `120`
- [x] Improve solutions with VS2013 features
- [x] Scripts
- [x] x64 compatible solutions (#2758)
- [ ] **Recompile libs**
   * Win32
   * x64
- [ ] Check compilation (errors/bugs)
- [ ] Changelog!

---
Resolve #2570

PR #3572, opened 2015-01-18
Turn on multiprocessor compilation for visual studio 2012

admsyn:
As mentioned by @LeoColomb in #2548 , this turns on multiprocessor compilation (`/MP`) for visual studio 2012. It also turns off "minimal rebuild" (`/Gm`), because that's an incompatible option.

Build times (for all of OF + empty example) :
- Current master 2:37
- With /MP 1:11

(My Windows HDD is elderly)


LeoColomb:
:smile: :+1: 

Already done in "current" VS2013 transition ( #2571 )

Issue #3430, opened 2014-11-20
project generator ADDON_CFLAGS only sets C++ Flags in Xcode


danomatika:
ADDON_CFLAGS in addon_config.mk actually sets the *Other C++ Flags* in Xcode. This is fine for C++ libs but does't work when compiling C libs.

For instance, ofxPd uses libpd which is a C library and requires a couple of defines to be set to build correctly. Those flags have to be set in *Other C Flags* to work in Xcode but they are only set in *Other C++ Flags*, so generated projects do not work.

The real fix would require two variables: ADDON_CFLAGS and ADDON_CPPFLAGS. Naturally, all add ons would need to be updated to use the new ADDON_CPPFLAGS variable, so this may not be the easiest solution.

danomatika:
Any interest in this? Still a showstopper for ofxPd.

The easiest solution would be to introduce a new variable for the CFlags like ADDON_C_CFLAGS or ADDON_LOW_CFLAGS or some variant.

bakercp:
Definitely interested, just not enough bandwidth at the moment ...

danomatika:
Ok, have this working in the PG: https://github.com/ofZach/projectGeneratorSimple/pull/62

danomatika:
FIx for this is in: https://github.com/openframeworks/openFrameworks/pull/3573

danomatika:
This is basically done with my core PR and the PG PR.

bilderbuchi:
thanks Dan, closing, then.

PR #3573, opened 2015-01-19
added ADDON_CPPFLAGS Makefile var so ADDON_CFLAGS can be used by PG for pure C

danomatika:
This variable is added mainly for Xcode projects generated from the PG (see https://github.com/openframeworks/openFrameworks/issues/3430) and is added to the Makefiles for completeness. The Makefile simply appends it to PROJECT_ADDONS_CFLAGS. Tested on OS X Makefile build.

### DIFF
--- a/libs/openFrameworksCompiled/project/makefileCommon/config.addons.mk
+++ b/libs/openFrameworksCompiled/project/makefileCommon/config.addons.mk
@@ -67,6 +67,7 @@ define parse_addon
 	$(eval ADDON_DATA= ) \
 	$(eval ADDON_INCLUDES= ) \
 	$(eval ADDON_CFLAGS= ) \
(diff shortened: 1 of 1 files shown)


danomatika:
Ok, the makefile var has been updated to ADDONS_CPPFLAGS and tested on iOS & OS X projects with the PG update. Both are working and fix a number of issues, mainly allowing for Pure C cflags in Xcode projects.

This will not break compatibility with existing addison since the ADDONS_CFLAGS & ADDONS_CPPFLAGS are appended to the same PROJECT_ADDONS_CFLAGS variable in the build system

ofTheo:
looks good to me!

danomatika:
Decision made. :D hah

PR #3575, opened 2015-01-19
Add missing include path in ofxOsc's addon_config

admsyn:
ofxOsc's addon_config didn't specify an include path of `libs/oscpack/src`, which is necessary for some of oscpack's internal includes. Presumably this didn't affect the PG since it adds basically all paths, but it should probably be specified anyway.

### DIFF
--- a/addons/ofxOsc/addon_config.mk
+++ b/addons/ofxOsc/addon_config.mk
@@ -29,7 +29,7 @@ common:
 	# include search paths, this will be usually parsed from the file system
 	# but if the addon or addon libraries need special search paths they can be
 	# specified here separated by spaces or one per line using +=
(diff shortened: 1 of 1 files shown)


arturoc:
the way the makefiles and the PG work is that by default they include every path in libs and src recursively. if you want to do something different to that then you can specify it in the addons_config.mk that way most addons just work without any further config.

bilderbuchi:
so, close this unmerged, @arturoc ? I think I agree, since I don't remember compilation problems since we switched to that setup.

admsyn:
Makes sense, I misinterpreted the setting as general include paths instead of specifically external / "extra" paths.

PR #3268, opened 2014-09-29
Apothecary - Tess2

danoli3:
- [ ] OSX FAT (32 bit w/ libstdc++ and 64 bit w/ libc++)
- [x] iOS FAT (armv7, arm64, i386 and arm64 and x86_64 all libc++) [Build Status](https://travis-ci.org/danoli3/apothecary-den)
- [ ] VS 2012 >> ??
- [ ] win_cb / mingw32 >> ??

PR #3577, opened 2015-01-20
Apothecary - Tess2 Formula [iOS]

danoli3:
Former PR's #3268 and older #3150

Formula's for:
- [x] iOS (armv7, arm64, i386, x86_64 all with libc++) [Build Status](https://travis-ci.org/danoli3/apothecary-den/branches)


tgfrerer:
hey @danoli3, I was looking into getting an os x forumla for tess2, but it seems that the recipe downloads from a google-code repo which has been deprecated and seems to be linked to the currently github-hosted version: https://github.com/memononen/libtess2 

Since that has been updated, the patch stage fails, which means that all recipes appear broken. 

Also, libtess seem to have moved from `cmake` to `premake`, which throws in lua as a dependency (awrr)... What's the best way forward?

danoli3:
Oh wow you are right! Nice find. Okay sweet this may explain some issues I saw with tess2 in iOS  branch I was testing.

Someone has a pull request to incorporate cmake again.. https://github.com/memononen/libtess2/pull/3

scripts/apothecary/formulas/tess2/tess2.sh
```diff
@@ -15,6 +15,12 @@ VER=1.0
...
 
+CSTANDARD=c11 # c89 | c99 | c11 | gnu11
+CPPSTANDARD=c++11 # c89 | c99 | c11 | gnu11
+COMPILER_CTYPE=clang # clang, gcc
+COMPILER_CPPTYPE=clang++ # clang, gcc
+STDLIB=libc++
+
 # download the source code and unpack it into LIB_NAME
 function download() {
 	curl -LO https://libtess2.googlecode.com/files/libtess2-$VER.zip
```
danoli3: This one snuck in... damn nice find @tgfrerer

arturoc:
oh it's actually great that mononen is maintaining this library i thought we would have need to change it again at some point since it was unmaintained for some time.

we should recompile it for other platforms too. something to be careful is that i patched the original to remove some crashes mostly from assertions but it would just sufficed to add -DNDEBUG while compiling to remove them.

i guess the current premake toolchain takes care of that but worth looking into it

danoli3:
Yeah I've updated your patch file that had those fixes to work with the latest libtess2 and all is working well. (just using that original cmake list)
Good point with the ```-DNDEBUG```, will keep that in mind to kill those asserts

PR #3578, opened 2015-01-21
Move OSX system frameworks to CoreOF.xcconfig

### LINKED ISSUE #3290: ofAVFMovieRenderer linker errors
Hi there,
whenever the ofAVFMovieRenderer is involved in a project, i get linker errors:
> Undefined symbols for architecture i386:
  ".objc_class_name_AVPlayer", referenced from:
      pointer-to-literal-objc-class-name in libopenFrameworks.a(ofAVFMovieRenderer.o)
  ".objc_class_name_AVPlayerItem", referenced from:
      pointer-to-literal-objc-class-name in libopenFrameworks.a(ofAVFMovieRenderer.o)
  ".objc_class_name_AVPlayerItemVideoOutput", referenced from:
      pointer-to-literal-objc-class-name in libopenFrameworks.a(ofAVFMovieRenderer.o)
  ".objc_class_name_AVURLAsset", referenced from:
      pointer-to-literal-objc-class-name in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_AVMediaTypeVideo", referenced from:
      ___28-[AVFMovieRenderer loadURL:]_block_invoke_2 in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_AVPlayerItemDidPlayToEndTimeNotification", referenced from:
      ___28-[AVFMovieRenderer loadURL:]_block_invoke_2 in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_CACurrentMediaTime", referenced from:
      -[AVFMovieRenderer update] in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_CMTimeGetSeconds", referenced from:
      -[AVFMovieRenderer duration] in libopenFrameworks.a(ofAVFMovieRenderer.o)
      -[AVFMovieRenderer currentTime] in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_CMTimeMakeWithSeconds", referenced from:
      -[AVFMovieRenderer setCurrentTime:] in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_CMTimeMaximum", referenced from:
      -[AVFMovieRenderer setCurrentTime:] in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_CMTimeMinimum", referenced from:
      -[AVFMovieRenderer setCurrentTime:] in libopenFrameworks.a(ofAVFMovieRenderer.o)
  "_kCMTimeZero", referenced from:
      ___28-[AVFMovieRenderer loadURL:]_block_invoke_2 in libopenFrameworks.a(ofAVFMovieRenderer.o)
      -[AVFMovieRenderer stop] in libopenFrameworks.a(ofAVFMovieRenderer.o)
      -[AVFMovieRenderer setCurrentTime:] in libopenFrameworks.a(ofAVFMovieRenderer.o)
ld: symbol(s) not found for architecture i386

Seems there are no 32 bit symbols. This is done on OSX 10.9.5 with Xcode 6 installed, with OF current master.
Calling make from cmdline.
Is this an OF issue or am I doing something wrong?

admsyn:
This closes #3290, and sprung out of the conversation on how to handle shifting OSX system framework dependencies.

The problem at the moment is that an app's linked frameworks are basically set in stone at the time the project is created, leading to [tricky linker errors](http://forum.openframeworks.cc/t/fix-for--cvdisplaylinkrelease-etc-errors/) when the dependencies shift after the fact (note the view count on that forum post..). 

This PR extracts the framework dependency list from the project template as well as OF itself, moving it to CoreOF.xcconfig. Since each app's settings are already being pulled from this config file, this will allow us some level of control over "old" projects after the point at which they were created.

A side benefit of this is that it makes the projects look much leaner in Xcode, which IMO is pretty nice. I'm trying to resist the urge to bikeshed any further about the project template in this PR :)

screen shot 2015-01-20 at 10 54 25 pm

For this to work, a slight modification to Project.xcconfig is needed, changing:
```
OTHER_LDFLAGS = $(OF_CORE_LIBS)
```
to:
```
OTHER_LDFLAGS = $(OF_CORE_LIBS) $(OF_CORE_FRAMEWORKS)
```
...meaning any current projects won't benefit from this change, but shouldn't be affected since they've already got the current required system framework list baked into them.

This could be applied to iOS as well, if this is a direction we'd like to move in.

ping @openframeworks/macos @openframeworks/ios


bakercp:
+10.

pizthewiz:
Yep, I'm a big fan of this too.

ofZach:
+1 -- this could be a **huge help** for allowing older projects to keep compiling (and it cleans up the project nicely).  Are there any change in the build phases for the template that need to be looked at?  I know there are some framework linking and build step stuff there.

ofTheo:
only issue I see here is with GLUT. 
we aren't using the system GLUT framework we're using the custom one in OF/libs/ 

if we're happy to completely remove GLUT from master then we could just drop that framework from the list. ( I think we talked about this already in another PR / Issue ). 

otherwise we would need to manually have the GLUT framework in the sidebar as it used to be. 

Huge thanks for this PR - love having less stuff in the sidebar!

admsyn:
Good point @ofTheo . I've been fiddling with Xcode things for a bit and I don't think there's a sane way around having the GLUT framework included in the project directly.

What if we remove GLUT from the project template as per this PR, but leave it in libs/glut/lib/osx for the time being? Old projects will continue on as if nothing had happened, and you can still add the GLUT framework with "Add Files to Project.." if you need the fallback for new projects. It's not much more complicated than adding an addon, really.

ofTheo:
Great, sounds good to me - feel free to merge!

PR #3579, opened 2015-01-21
ofxGPS - cross platform GPS addon for iOS and Android

tallavi:
The addon itself is at: https://github.com/tallavi/ofxGPS

It's a GPS + compass addon that exposes the same interface for iOS & Android. I use it for the past month on both OSs and it works great for me.

This PR is related to the android part:

1) removed references of the old gps addon from ofxAndroid (there were name clashes)
2) Updated OFAndroidGPS.java.
3) Updated OFAndroid.java. This file is too big so I put all GPS related stuff in OFAndroidGPS.java. Now OFAndroid doesn't need to know there's OFAndroidGPS.java. If my scheme will be accepted, this method can be used to clean up OFAndroid.java from allot of lines that should be in the addon files.

Since this addon has a java side, I see no choice other than suggesting to include this addon as part of the core. I think it can be super useful to any OF-er that needs location/direction services.

Tal


bilderbuchi:
pinging @openframeworks/android

arturoc:
looks great, there's something weird though, you are deleting all the cpp files but not adding any way to access the new API from c++?

tallavi:
@arturoc, the new API is the ofxGPS addon in the link I provided.

Would it be better to not have it in a different repo, but copy it here as part of this PR?

It will also require a change in the .gitignore.

kylemcdonald:
@tallavi given that the old functionality was an addon (but only inside ofxAndroid) then the best thing to do would be to add ofxGPS as a "core addon" (i.e., add it to your openFrameworks branch that you are submitting as a PR).

the only complication i can see is that we usually deprecate features before removing or refactoring them, and i'm not sure how it would work here.

but in the meantime, add your ofxGPS as a "core addon" like ofxKinect or ofxAndroid and we can all test it then merge it!

tallavi:
@kylemcdonald: ok, done as you requested, although I'm not completely sold that it needs to be a core addon anymore. ofxGPS is an exercise in modularity. 

ofxAndroid has to be a core addon, since there's no OF in android without it, but now that the entire GPS functionality (well, at least the cpp side. The java side is a more complicated issue as already discussed above) was separated away, I don't see a good reason to always have ofxGPS.

My other concern is automatically adding this addon in iOS project. I did that manually. Dragged the common and iOS folders to xcode. I think the project builder may have a problem with the sub folders, but I'm not sure what to do about this, since I haven't seen a precendence of an addon with specific files from iOS and for Android.

The android side is not problem, btw, since the android compiler just ignores mm files.

arturoc:
i agree with kyle, since this has been part of OF before it makes sense to make it a core addon.

a couple of things that i would change is that we don't return shared_ptr directly anywhere else in OF also the way it's done you can call ofxGPS::create but also create an instance. i would probably just publish the event through a function and have a system static instance in the cpp file. we use that same pattern in other places.

also we should keep the old api working but deprecated for some time

arturoc:
also, thanks so much for looking into this, we really need to make this apis more uniform for android and iOS and have a common base for any future platform

tallavi:
First of all - my pleasure! As my app is cross platform, it's a priority to me too, and I'm glad to give something back to OF when I can.

Regarding your comment, yeah, you're right, that's quite an old design of mine, from a period of time I thought the ofEvents system will let me pass control from the JNI thread to the UI thread.

So I'm changing it to be with a static start/stop methods (no constructor/destructor) and reporting back in a few days.

tallavi:
Side note regarding static vs. instance in OF: More than once I find myself in a situation that I don't really want an instance, but am forced to have one because I want to register to ofEvents. Isn't there a way to have events call a static method? What's the best practice in such cases?

arturoc:
no there's no way to do that currently, the problem is that ofEvents are based on POCO which is a very OO library (too much in my opinion :)  so it doesn't even consider the possibility of adding a raw function as a listener. we could modify the events in poco and send them a PR. i've added some features before and they accepted them without problem

tallavi:
Done. 

Check out the updated, fully static interface in ofxGPS.h

Tested on iOS and Android.

bakercp:
@tallavi I'm quite interested in this -- any chance of bringing the PR up to date so I can test it on iOS / Android in the next week or so?

tallavi:
Hi @bakercp, I'd love to help!

I no longer maintain a up to date PR for ofxGPS only since I started a bigger endeavor of improving life cycle handling on Android. 

Because ofxGPS was not accepted and merged when I started the life cycle thing, and both features touch OFAndroid.java, they now come together, I hope it’s ok.. I think OF will benefit from the life cycle changes too and I would LOVE to get them both merged. I use them both constantly in my iOS/Android app and I actively maintain and improve them.

Coincidently, I rebased my branch just yesterday and here it is: https://github.com/tallavi/openFrameworks/tree/feature_android_life_cycle_ofxGPS

but:

1) it has the xCode crashing issue on iOS, It’s not specifically related to this branch, but it currently prevents me from using this branch.

2) Something has changed in surface created on Android and now it crashes when returning to the app after minimized. I'm on it. I’m sure it’s a small fix (don’t know if in my code or somewhere in OF)

3) As I said, I think the life cycle handling is a step in the right direction, as it allows more complicated scenarios for OF enabled android apps, but.. it requires a small backwards compatibility break in the “bootstrapping files” - main.cpp & OFActivity.java. No changes are necessary in the ofApp files. I was in contact with @arturoc regarding this, and he believes it's a prefect time to put in some breaking changes, since everything else is broken too :-) 

Notice that the life cycle part is still work in progress as we reach more and more complicated scenarios in our multi-activity app, but it doesn’t affect the simple OF examples scenario either way - the simple scenario of a single OF activity is supported without a problem, like before.

So, my suggestion is to rebase this branch on latest master (hoping the xCode thing is resolved by the latest commits), fix the android issue, and create a new PR instead of this one. I can make it in the time table you need. Do you think we can get this thing merged?

(if not, I can probably do a ofxGPS only PR, but I prefer not to work twice :-()

Tal

tallavi:
This PR is cancelled in favor of: https://github.com/openframeworks/openFrameworks/pull/4260 that contains all of this and more.

PR #3581, opened 2015-01-21
Android get result from listbox

i-n-g-o:
hi.

changes to AlertListBox for changing background of list-item.
add function to retrieve the result of ofxAndroidAlertListBox with ofxAndroidGetLastAlertListSelection()


bilderbuchi:
ping @openframeworks/android

arturoc:
when do you call this method? if i understand correctly while the list is showing the app still runs in the background right? wouldn't it be better to make a callback the same way the alert boxes work?

i-n-g-o:
The changes are influenced by ofxAndroidAlertTextBox() and ofxAndroidGetTextBoxResult() and work the same way: first call ofxAndroidAlertListBox() then ofxAndroidGetLastAlertListSelection().

imho the selection should be handled via a callback.
as should an ok-click in an AlertTextBox call back into the app with the entered string.
no need for storing the last textBoxResult or listBoxResult...

so maybe introducing a callback for ok-click in the AlertTextBox and a listBox would be better?
There would be another callback needed for multi-selection listBoxes.

okPressed(string text); // text-box callback
okPressed(int selection); // list-box callback
okPressed(vector<bool> selections); // multi-selection list-box callback

i-n-g-o:
closing this PR and proposing a new one with callbacks as mentioned above

PR #3582, opened 2015-01-23
OpenGLES - GL_HALF_FLOAT is only defined in GLESv3 - Merge Fix for #3409

danoli3:
Reference Pull request:  Fix compile error on Android: GL_HALF_FLOAT is only defined in GLESv3 #3409


danoli3:
Good to merge now @arturoc ?

arturoc:
thanks!!

PR #3585, opened 2015-01-23
Apothecary - FreeImage Formula Only - Fix

danoli3:
- Fixes not copying the include (if only building iOS)


danoli3:
@arturoc Just merging this bitesize fix now. Preq for iOS building to fix header not updating issues.

PR #3587, opened 2015-01-24
Remove ofSoundObject for now, closes #3586

### LINKED ISSUE #3586: bug in ofSoundObject
this in line 130:

```cpp
std::shared_ptr<ofBaseSoundOutput> ofSoundMixer::getChannelSource(int channelNumber) {
	if (channelNumber < channels.size()) {
        return std::shared_ptr<ofBaseSoundOutput>(channels[channelNumber]);
	}else{
		return std::shared_ptr<ofBaseSoundOutput>();
	}
}
```

is wrong, it's creating a shared_ptr from an already existing pointer which will destroy the original once it goes out of scope, the correct thing to do seems to store shared_ptr in the vector.

ping @bakercp since it seems he added this code

admsyn:
This PR removes ofSoundObject.cpp/.h, the references to them from IDE files, and the corresponding soundObjectExample. As discussed in #3586, the API isn't ready for general use yet and will need some more love before being included in an OF release.

ping @arturoc , @openframeworks/sound


kylemcdonald:
going to merge this as @arturoc already requested it in #3586

PR #3588, opened 2015-01-24
Fix invalid itr increment in ofMainLoop on windows/vs

admsyn:
When closing an app window on windows, there's a debug assertion raised:
```
Expression: map/set iterator not incrementable
```
..on this for loop condition in ofMainLoop:
```
void ofMainLoop::loopOnce(){
    >   for(auto i = windowsApps.begin();i!=windowsApps.end();i++){
        ...
```
This PR modifies the increment condition and the erase call to `windowsApps.erase(i++);`, which stops it from firing. I'm not 100% up to speed on the new multiwindow changes, but I tested the changes with @arturoc 's explicitRendererExample on windows + osx and everything seems to work correctly after this fix.

(This is Windows 7 and visual studio 2012 express)

### DIFF
--- a/libs/openFrameworks/app/ofMainLoop.cpp
+++ b/libs/openFrameworks/app/ofMainLoop.cpp
@@ -112,14 +112,15 @@ int ofMainLoop::loop(){
 }
 
 void ofMainLoop::loopOnce(){
-	for(auto i = windowsApps.begin();i!=windowsApps.end();i++){
+	for(auto i = windowsApps.begin();i!=windowsApps.end();){
 		if(i->first->getWindowShouldClose()){
 			i->first->close();
-			windowsApps.erase(i);
+			windowsApps.erase(i++);
 		}else{
 			currentWindow = i->first;
 			i->first->update();
 			i->first->draw();
+			++i;
 		}
 	}
 	if(pollEvents){
--- a/libs/openFrameworks/app/ofMainLoop.cpp
+++ b/libs/openFrameworks/app/ofMainLoop.cpp
@@ -112,10 +112,13 @@ int ofMainLoop::loop(){
 }
 
 void ofMainLoop::loopOnce(){
-	for(auto i = windowsApps.begin();i!=windowsApps.end();){
+	auto i = windowsApps.begin();
+	while(i!=windowsApps.end()) {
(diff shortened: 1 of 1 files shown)


bilderbuchi:
man, a for loop without an incrementor looks very strange. is this a real fix or just masking the fact that you erase an element out of an iterator during execution (isn't that a Bad Thing (tm) to do?). on which line does the debug assert fire?

admsyn:
I agree it looks a little odd :) [This SO answer is a decent explanation of what's going on](http://stackoverflow.com/a/4600592/704596).

> ...erase an element out of an iterator during execution (isn't that a Bad Thing (tm) to do?)

Only in the case that it invalidates all iterators (with a std::map it's fine). [From cppreference.com on std::map::erase](en.cppreference.com/w/cpp/container/map/erase):

> References and iterators to the erased elements are invalidated. Other references and iterators are not affected. 

The issue as I understand it is that erasing in a for loop like this:
```
for(...; i++) {
   ...
   map.erase(i);
   ...
}
```

..means you end up erasing the object the iterator points to (invalidating the iterator), then trying to increment it in the next trip around the for loop. Calling `map.erase(i++)` exploits the fact that the postfix increment returns the original value of 'i' instead of the incremented one. So the logic becomes "move the iterator, erase the old thing" instead of "erase the thing, move the iterator".

Since this is so awkward, in C++11 erase returns an iterator you can use like:
```
i = map.erase(i);
```
..and not really have to deal with this :) I assume at this point we can't be assuming C++11 features in all standard libraries OF will be compiled with, otherwise this fix could be a tad cleaner.

bilderbuchi:
I see. it would probably be more beautiful/idiomatic to, like in the SO answer you linked to, use a while loop instead of the for-without-increment, right?

arturoc:
wow, yeah didnt' understood what was different. if we could use something clearer? this is kind of too clever, if someone has to modify that code it'll be really easy to introduce a bug again.

admsyn:
Sure, I'll re-work it into a more explicit while loop. I agree with the "too clever" assessment, though I've seen the pattern many times elsewhere and sort of tuned out the fact that it's a fairly opaque construct.

arturoc:
already fixed by #3664 thanks!

Issue #3242, opened 2014-09-17
Unify OS X and iOS AV Foundation Player


ofTheo:
With the merging of #3200 we now have an AVFoundation player/grabber for OS X.
We also have a separate one for iOS. 

I think the goal of AVFoundation was to have a shared API for video on both iOS and OS X, so it would be worth looking at consolidation. 

It could result in more consistent behavior and less bugs to fix. 

ping @julapy @admsyn @prisonerjohn @openframeworks/video @openframeworks/ios

julapy:
hey, just wanted to give this a little push since OF activity is ramping up for 0.9.0
have @prisonerjohn @openframeworks/video folks had a chance to look at this yet?

for now it might be an idea to replace the current iOS video player with ofxAVFoundationVideoPlayer (https://github.com/julapy/ofxAVFoundationVideoPlayer). ive been concentrating my dev efforts on the new ofxAVFoundationVideoPlayer and leaving the OF iOS player stagnant... so to take advantage of the new features and bug fixes, its probably worth doing the swap now for 0.9.0. keep in mind that ofxAVFoundationVideoPlayer stems from the OF iOS player so its essentially the same, but now also works on OSX.

@ofTheo @admsyn - you guys ok with this?

ofTheo:
@julapy I am super happy to do this. 
We're actually talking about doing a bunch of video stuff now anyway with #3543 so I think its the perfect time. 

Please do a PR for it! It will be great to have a single video player for both. 
Not sure if #3543 should be done first @pizthewiz @bakercp ? thoughts?

prisonerjohn:
Sorry I haven't had a chance to dig through yet, I only took a quick glance. I agree that a unified video player for iOS and OS X is def the way to go! I could test it in my recent and current projects if it gets pulled in the main repo.

admsyn:
:+1:

julapy:
spent a couple days merging the OSX and iOS video player code (PR is here - #3590)

its working well across the two platforms in pretty much the same way.

one difference worth noting is that the current OSX video player in master uses the AVPlayerItemVideoOutput class which is only available in iOS6+ and currently ofxiOS supports iOS5 and up. so in this version, AVAssetReader is used for reading video sample data instead AVPlayerItemVideoOutput. this is ok i think, as they both do the same thing and when we bump up the minimum iOS version to iOS6+, then we can swap out AVAssetReader for AVPlayerItemVideoOutput.

[...]

Issue #2283, opened 2013-07-17
deprecate QuickTime classes, make AV Foundation default


danomatika:
These don't compile with newer OSX sdks without Carbon right?

pizthewiz:
I am going to close this out now that video playback has transitioned to AV Foundation for the 0.9.0 release, while recognizing we still need a grabber implementation #3382 which will come in a later release.

DyInCN:
We have a old project used the QuickTime.framework. In order to playback the video with special sound effects，we used the QTAudioContextInsertRegistryInfo API,and set the processDataCallback of the 
QTAudioContextInsertRegistryInfo object.The processDataCallback is said like this:

<typedef CALLBACK_API( OSStatus , AudioContextInsertProcessDataCallback )(void *inUserData, AudioUnitRenderActionFlags *ioRenderFlags, const AudioTimeStamp *inTimeStamp, UInt32 inNumberFrames, AudioBufferList *inInputData, AudioBufferList *outOutputData);>

When the program excutes this callback, we can get the inInputData and outOutputData. And then, 
there is a third C++ API,which need the parameters are from the inInputData and outOutputData.
This API can run audio effects for the playing data.

Now we need to abandon the QuickTime.framework, and replace it by AVFoundation.framework.
The third C++ API was made into a static library, and we can not modify the source code about it.
So there is a question is that how can I get the inInputData and outOutputData by AVFoundation?
If someone can help us, it is so much thanks。

PR #3590, opened 2015-01-25
OSX and iOS unified AVFoundation video player.

julapy:


ofTheo:
wow - awesome to see this @julapy 
any regressions or possible issues to look out for?

julapy:
@ofTheo, no regressions as far as im aware.
things should work all the same on OSX and iOS.
be good to give it a thorough test though...

kylemcdonald:
just for reference, this would help with #2283 and would close #3242. further description of this PR is in https://github.com/openframeworks/openFrameworks/issues/3242#issuecomment-71353036

i suggest where either:

1. trust @julapy and merge immediately to give as much time to test before 0.9.0 is released :) (my preference)
2. wait for 0.9.0, and include this in 0.9.1 as is currently planned (which could be around 6-8 months...)

julapy:
thats what i was thinking as well... since we are going forward with the unified video player, then its better to switch to it now so we can test it and improve it if any issues arise. it will also ensure that any future developments to the video player work on both osx and ios.

kylemcdonald:
@julapy great, then i'd say give this 2-3 more days, then merge if there are no objections from @openframeworks/video and we will make sure to squash any bugs before the next release :)

bilderbuchi:
tbh, I'd rather we not add more features/PRs/issues and testing debt to 0.9.0, and instead slate this for immediate merge after 0.9.0 release - that way, the testing time is maximised before this hits the "public", and we don't have to allocate yet more time to testing this.
either way, I just wanted to give my 2 cents from a release management p.o.v., and will defer to @openframeworks/video for the final decision.

kylemcdonald:
@bilderbuchi i generally agree, but given that the video player is already in flux and will need testing regardless i don't think this adds any testing debt -- just redirects it. and, ideally, it reduces debt by consolidating two implementations.

kylemcdonald:
thanks @ofTheo!

PR #3593, opened 2015-01-27
Allow access to pixel color through index (like setColor)

nykwil:
It is quite useful to be able to get access to an image through index.
It's possible to get pixel by index so there's precedent.

### DIFF
--- a/libs/openFrameworks/graphics/ofImage.cpp
+++ b/libs/openFrameworks/graphics/ofImage.cpp
@@ -939,6 +939,12 @@ ofColor_<PixelType> ofImage_<PixelType>::getColor(int x, int y) const {
 
 //------------------------------------
 template<typename PixelType>
+ofColor_<PixelType> ofImage_<PixelType>::getColor(int index) const {
+	return pixels.getColor(index);
+}
+
+//------------------------------------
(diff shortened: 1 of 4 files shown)


kylemcdonald:
this looks great to me! it also looks like it's implemented correctly in terms of how it rewrites getColor(x,y).

this is how the pixels array works in processing, and i imagine someone coming from processing would want to play with pixels this way https://processing.org/reference/pixels.html

if no one has any problems with this in the next week, someone will merge this :)

thanks @nykwil!

arturoc:
accessing through the index is usually quite inefficient, from 0.9 on you can do:

```cpp
for(auto line: pixels.getLines()){
    for(auto pixel: line.getPixels()){
        pixel.getColor()
    }
}
```

or even:
rough the index is usually quite inefficient, from 0.9 on you can do:

```cpp
for(auto line: pixels.getPixels()){
    pixel.getColor()
}
```

which does the same as the typical for loop with x,y or an index but way more efficiently and with a more concise syntax. it also avoids any assumptions about the number of channels in the pixels... but given that we already have accessors based on the index i'm going to merge it

thanks!

PR #3596, opened 2015-01-29
fix issue where ofFbo would be leaking its viewport

tgfrerer:
because ofFbo did not restore its viewport to openGL in ofFbo::end() but only push/pop'ed the matrix stack, OpenGL and openFrameworks would get out of state sync once an fbo with different size than the screen would be unbound, leaving the viewport for the main screen at the Fbo's size.

This fixes this by using the renderer's own push/popView methods, which forward the state changes to the GPU / OpenGL upon unbind().


arturoc:
Thanks Tim.

Issue #3260, opened 2014-09-25
Feature: make ofGetCurrentRenderer work across multiple threads. 


ofTheo:
If you have an app with multiple GL contexts each in its own thread - all the current OF calls fight each other as the current renderer can change back and forth between the threads. 

I saw this recently when trying to get the threaded Kinect V2 GPU decoding stable in @hanasaan gpu branch https://github.com/hanasaan/ofxKinectV2/tree/of_gpu_dev

Basically the multiple GL contexts are totally fine as each are in their own thread. Pure GL calls work fine. But calls to OF renderer related functions can cause really strange issues ( viewports inverting on the main thread, strange graphics glitches etc ). This is because a call in the second thread mixes its OF renderer commands in with the first thread. 

Thinking about it I thought that it would be impossible to keep our current API and support multiple threaded contexts. 

ie: calls like ofPushMatrix(); ofPopMatrix(); etc. 

However I think it could be possible that ofGetCurrentRenderer() would be able to register renderers based on the thread they are running in. So a call to ofGetCurrentRenderer()->draw(video, 0, 0, 320, 240);  Could internally check the thread id and if there was a renderer registered for that thread it would use that renderer. 

This would allow the renderer's to operate completely separately while still supporting the global functions that we have in our API. If I understand it right we could make this fix without changing or breaking any of our current API. 

Curious what people's thoughts are on this.

Issue #1863, opened 2013-02-07
ofFbo needs better support for MRT


tobiasebsen:
ofFbo does support MRT, but it is not very flexible i regards to targets of various formats, nor is it possible to set the format and attachment of the render buffer.

I suggest:
1. createAndAttachTexture() and createAndAttachRenderbuffer() are made public.
2. adding a function called attachTexture(const ofTexture& tex) that allows attaching various types of textures.
3. making all private variables and functions "protected", so that it becomes easier to inherit from ofFbo.

Issue #1742, opened 2012-12-11
Need stack based handling of FBO bindings?


timscaffidi:
I ran into the inverted bitmap string issue (#1474) again when I was drawing to an FBO which was itself drawn inside another FBO. The "nested" FBO was actually an internal FBO used by an addon's draw method, neatly hidden away from view.

example:

``` c++
void testApp::draw(){
    fbo1.begin();
    addon.draw(); // binds draws then releases an internal FBO

    ofDrawBitmapString(...);
    fbo1.end();

    fbo1.draw(); // inverted text
}
```

The outer FBO remains bound after the internal one releases, so openGl must be using a stack like mechanism. I can tell because the FBO contains the text, it is just inverted. The OF gl renderer only has support for one FBO, so once the internal one gets released, it thinks no FBO is bound, and won't re-invert the bitmap strings.

I fixed the issue by calling `fbo1.end(); fbo1.begin();` prior to drawing the bitmap string, but a proper fix would be using a stack to keep track of bound FBOs. This will also affect anything else using the `currentFbo` variable of `ofGLRenderer`.

PR #3601, opened 2015-01-31
Fix MSAA FBO and fix MSAA MRT FBO

### LINKED ISSUE #3598: ofFbo with multisampling enabled, does not clear.
the code below should make the fbo all white.
instead its just showing gpu garbage.

it works correctly when disabling multisampling => `fbo.allocate(w, h, GL_RGBA, 0);`
so its an issue with multisampling on the fbo i believe.

````
void ofApp::setup(){
    fbo.allocate(512, 512, GL_RGBA, 4);
}
````
````
void ofApp::draw(){
    fbo.begin();
    ofClear(255, 255);
    fbo.end();
    fbo.draw(0, 0);
}
````

### PR DESCRIPTION
**Pull request by tgfrerer**
This PR is a fix for #3598, and it also adds the ability to use mutlisampling (MSAA) with fbos that have multiple render targets (MRT) -- this used to be an issue in the past and was also left over with a big TODO in the inline comments of `ofFbo` =P

I got a bit carried away fixing the initial issue, but figured it would be nice to be able to use more than one MSAA buffer with `ofFbo`.

@julapy, would you mind checking out if this fixes #3598 - also, maybe, if everything still works as expected on iOS?

I've split this PR up into more or less self-contained commit units, each with some more information in the commit message. Also, I've added some inline comments where it seemed to make sense.

I'm too tired now to test on OS X and Linux, and with the fixed function renderer (and about to head out to get a very late Friday :beer:), but Windows/VS2012 programmable works a breeze. If interested parties could give it a spin, i can see to fix anything that pops up on Monday when I'm back at work.

### Test Project
Here is a one-file project as a gist, which I used to test this against:
https://gist.github.com/tgfrerer/94a1e0443a8a62866019

### Test Render
I'm rendering the same scene into 3 FBOs with each 3 render targets,
left has 4xMSAA, middle 0, right 16xMSAA. Top are render targets 0, middle render targets 1, bottom render targets 2. For best results, look at the image with 1:1 magnification =)

### CODE REVIEW

**Comment by kylemcdonald** [comment]
tested on osx 10.9.5 but with gl 4.0 (highest i have available) and they look the same to me.

screen shot 2015-02-01 at 22 13 54

**Comment by julapy** [comment]
hey @tgfrerer,
works great for me on osx.
but on ios, nothing renders to the fbo.
here is my test code => https://gist.github.com/julapy/abef5208d492fa1fd273

when i roll back to master, the fbo renders correctly.

**Comment by tgfrerer** [comment]
Thanks for testing, Kyle,  Lukasz! 

@julapy, after spending some time (and £79 on an iOS dev membership :/ ) I seem to have found the issue: in iOS with MSAA enabled in the windowing system, instead of going straight to the screen render buffer, a *default framebuffer* is bound, which is not known to ofFbo since it bypasses the openFrameworks pipeline at generation. Therefore, ofFbo can't track the bind state of GL_FRAMEBUFFER_BINDING correctly, without using these costly `glGetIntegerv()` methods. 

As a fix, i've added an `#ifdef` to `ofFbo::bind()` so that, for GLES, it looks up the current GL binding state as we had it before.

`- (BOOL)createFramebuffer:(CAEAGLLayer *)layer ` 

Is the function, where in both renderers, GLES1, and GLES2 the default framebuffer is set up. It could make sense to, from this method, let ofFbo know about the current default buffer, maybe by initialising the framebuffer through ofFbo, but i wonder if that's possible right now.

I've pushed a fix to the PR, tested on iOS/iPhone 5c, with GLES1 and GLES2.

**Comment by julapy** [comment]
hey Tim, sorry that you got sucked into the ios membership scam :)
but great work figuring out the issue! i'll spend some time later today testing this out across some ios projects.

**Comment by tgfrerer** [comment]
It is a scam, isn't it? At least i know now in part where these record quarterly profits came from.

Great! i've just tested it on linux, ubuntu 14.10, with the latest binary nvidia driver / dwm (unity is broken for GLFW), and it works perfectly. I say if @julapy's tests come back ok, we should be ready to merge.

**Comment by julapy** [comment]
yep, looks good on iOS too!

**Comment by tgfrerer** [comment]
thanks, brilliant! i'd say we're ready to merge =) ping @arturoc

**Comment by arturoc** [comment]
looks correct to me except that with the multiwindow branch i also removed most global state ans static variables to be able to use OF GL objects from different threads. this introduces a static stack for the framebuffers which reintroduces some kind of global state. is it necessary to fix this problem? if not i would just remove it by now and if it is try to move it to the renderers instead

**Comment by tgfrerer** [comment]
Makes perfect sense to add it to the renderers! Even better, iOS would know about it's current renderer/fbo binding state and wouldn't have to deal with `glGets`...  
Let me look into it after lunch; will append PR.

**Comment by tgfrerer** [comment]
The latest changes implement fbo binding state as a member variable of their respective GL renderers, like @arturoc suggested. 

+ since fbo.bind changes the state of the renderer, such parts are executed within the context of the renderer, which adds the ability to extend these methods in the future with more renderer-specific code. It also adds the possibility to interface with windowing systems like iOS that might run the screen as a FBO.

Tested on windows, programmable + fixed function GL, with the above test project, ofFboTrails example, and another project, which gave me this result (best viewed on a big screen): 
https://cloud.githubusercontent.com/assets/423509/6044151/82a32732-ac89-11e4-93a4-df757a9fe0aa.png)

Will test on iOS and OS X now...

File: libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp [hunk 1]
```diff
@@ -1315,12 +1321,46 @@ void ofGLProgrammableRenderer::bind(const ofFbo & fbo, bool setupPerspective){
 }
 
 //----------------------------------------------------------
-void ofGLProgrammableRenderer::unbind(const ofFbo & fbo){
-	glBindFramebuffer(GL_FRAMEBUFFER, 0);
+void ofGLProgrammableRenderer::end(const ofFbo & fbo){
+	fbo.unbind();
```
arturoc: this should now be `unbind(fbo)` to avoid global calls when calling through the renderer

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
```diff
@@ -1315,12 +1321,46 @@ void ofGLProgrammableRenderer::bind(const ofFbo & fbo, bool setupPerspective){
 }
```
arturoc: how does this start the fbo at all? fbo.begin() calls this but none bind the fbo, right?

libs/openFrameworks/gl/ofGLRenderer.cpp
```diff
@@ -458,7 +464,7 @@ void ofGLRenderer::bind(const ofFbo & fbo, bool setupPerspective){
 }
 
 //----------------------------------------------------------
-void ofGLRenderer::unbind(const ofFbo & fbo){
+void ofGLRenderer::end(const ofFbo & fbo){
 	fbo.unbind();
```
arturoc: same here `unbind(fbo)`

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
tgfrerer: hmm, i need the indirection to mark the buffers dirty which have been rendered to, if using MRT.

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
tgfrerer: `bind` does not start the fbo, `begin` does =).

Keeping bind and begin separate might help us in the future, if we need to bind the fbo without setting up matrices, like you would in some more advanced cases.

tgfrerer: and the fbo is bound in the renderer upon bind(fbo), as the renderer needs to keep track of the fbo binding state in its local scope.

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
arturoc: mmh ok, let me take a look later. in general calling bind/unbind begin/end or draw in GL objects is now the only operations that are thread unsafe so to work with threads you should be able to do:

```
renderer.bind(fbo)
```

and that should be safe but if that internally calls fbo.bind and fbo calls ofGetGLRenderer then it'll not be safe any more. 

why do we need the frameBuffer stack? does that solve some problem? i've been also trying to avoid keeping this kind of state even inside the renderer since it's really prone to errors. it would be weird to have nested fbo binds no?

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
arturoc: oh sorry yeah, i didn't realize that is calling fbo.bind at the end since the diff view was hiding it :S

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
tgfrerer: yes, nested fbos seem weird to me too. We have an issue open about them #1742. Would be nice to do without, but i felt i needed them to reliably keep track of fbo state.

tgfrerer:
related #1863

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
tgfrerer: Hey @arturoc, can you explain a little more what the thread-issue is? 

I've got a bit of time to fix things if neccessary, but it's a bit of a pain keeping this mergeable. And since MRT is broken since (or shortly before) multiwindow it would be nice to have it working again for 0.9.

I see that multithreading was discussed in #3260, but i'm not sure how current the discussion there is.

arturoc: hey, yeah sorry. the idea is that if you wanted to use OF with several gl contexts from different threads, now you can do it as long as you don't use:
```
someglobject.draw(..)
```

or

```
someglobject.bind()
```

which are not thread safe because they call the ofGetCurrentRenderer which is a singleton and will fail if called from multiple threads simultaneously. instead now you can do:

```
renderer.draw(someglobject)
```

and

```
renderer.bind(someglobject)
```

so all the draw and bind methods in a renderer should not access any singleton even indirectly. if a renderer calls object.bind, object.unbind or object.draw from any of it's methods then it breaks the thread safety.

there's an example in apps/devApps/explicitRendererExample that doesn't use threads but uses this alternative syntax to bind and draw gl objects through a renderer instead of directly through the object.

tgfrerer: So if i understand correctly, you are talking about thread-safety and a race-condition around accessing `ofGetCurrentRenderer()` on the CPU (openFrameworks) side, and not on the GPU/OpenGL side?

I think i could find a way to address that.


---


Just to make sure: we are not talking about the implications of multiple contexts, each on its own thread, sharing resources, hopefully. Because this would be where things would get really, really complex, and i'm not sure oF can handle it 100% currently without introducing sync objects, mutexes and the like =) 

Also, it promises to be super slow (http://blog.gvnott.com/some-usefull-facts-about-multipul-opengl-contexts/) because of possible per-command context swaps.

In such a scenario it appears FBOs would not be shareable, because they are Container objects, like VAOs (https://www.opengl.org/wiki/OpenGL_Object#Container_objects). But how even to document this - or flag this in code? Or make an ofFbo not shareable?

It's really hard to find good documentation on this, i trawled the specs and the OpenGL registry with very meagre results. The most promising thing i could find was this: http://www.seas.upenn.edu/~pcozzi/OpenGLInsights/OpenGLInsights-AsynchronousBufferTransfers.pdf and it didn't appear too enthusiastic.

Sorry if i seem a bit confused, but i've been rewriting, and banging my head against this for 4 hours now, without finding a solid solution, and i'm a bit scared about the maintenance implications of `ofGLFWWindowSettings.shareContextWith = mainWindow;` to be honest =)

arturoc: yeah, only talking about thread safety in the cpu, particularly about avoiding singletons (right now only the current window which contains a current renderer). not even about the renderer itself being thread safe, each thread should create it's own renderer

everything in the GL side is just responsibility of the user.

shareContextWith allows 2 windows usually in the same thread to share resources like textures or buffers but it's something that GLFW itself allows and again responsibility of the user to use correctly. 

you could even create 2 windows in different threads that share no GL objects at all but they can work without crashing by not accessing any singletons

sorry if it wasn't clear before

arturoc: > In such a scenario it appears FBOs would not be shareable, because they are Container objects, like VAOs (https://www.opengl.org/wiki/OpenGL_Object#Container_objects).

also this is why now every window contains it's own renderer, some resources can't be shared so the programmable renderer wouldn't work if there was only one renderer for all the windows as it was till now, the shaders won't work across different contexts even if they are shared.

but this is not something we should care about for all the objects, if someone wants to use openGL in such an advanced scenario i think it's safe to suppose that they'll now which objects can be shared or not

tgfrerer: Thanks so much for clearing this up- it makes all sense now =)

I'll see to send a fix tomorrow.

Btw would it be a good idea to somewhere formalise or externalise the design guidelines for the API and the renderer that were developed over time?

 I think it could be a helpful reminder for us and anyone wanting to send PRs...

I could start whacking something up into an issue or the GitHub wiki with the label "design strategy" which we could link to from the Contribute readme.

Some general principles around:
- which states the renderer tracks and which not
- where the renderer's responsibility ends 
- how GL state will be restored in core if in doubt
- gl best practices for the renderers

tgfrerer:
The latest updates do not round-trip to the fetch the current renderer upon bind().
+ also removed the framebuffer binding stack, because it seemed like overcomplicating things
+ added a bunch of documentation entries to the relevant methods, to incorporate some of the throughts outlined in the above line notes, and to make future maintenance a tad simpler
+ Additional information in the commit message of the latest commit.

* tested on windows, fixed function (GL 2.0) and programmable (GL 4.3).

arturoc:
yay! :)

tgfrerer:
[:beers:]

libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
arturoc: by the way, yes a document specifying some design guidelines would be
great to have!

Issue #3602, opened 2015-01-31
ofEllipse has extra line when saved to PDF


AmnonOwed:
I tried to draw an ofEllipse to PDF, however I noticed that an extra line is added in the PDF export. This is especially noticeable when it is drawn as stroke-only. The extra line is not visible in the application, but only in the exported PDF.

This issue has been reproduced in OF versions 0.8.4. and git development build on Windows 8 64-bit using Code:Blocks. The expected behavior is a regular circle, also in the PDF export. The current behavior is a circle plus an extra straight line running through it. This image shows what an ofEllipse currently looks like when it is saved to PDF:

![ellipsepdf](https://cloud.githubusercontent.com/assets/4075846/5988924/0e68e594-a975-11e4-9e5d-bfc729be3ca7.jpg)

A small code example reproducing the issue:

```c++
ofBeginSaveScreenAsPDF("Export-" + ofGetTimestampString() + ".pdf", false);
ofBackground(225);
ofNoFill();
ofSetColor(0);
ofEllipse(100, 100, 150, 150);
ofEndSaveScreenAsPDF();
```

AmnonOwed:
@youandhubris Great find! But I think somehow you sent the pull request to your forked master instead of to openframeworks:master.

youandhubris:
@AmnonOwed Yes, my bad! I'm new to everything related to programming. And Github, in this case. I think, now, I did it right. (Doing it at 6 a.m. wasn't a great idea, also. : )  )

PR #3605, opened 2015-02-01
Fix for #3602: ellipse not exported correctly

youandhubris:
Solves issue [#3602](https://github.com/openframeworks/openFrameworks/issues/3602). It wasn't an extra line, it was circling around.


arturoc:
Thanks!

PR #3613, opened 2015-02-03
fix apothecary libtess for iOS

tgfrerer:
* pulls libtess source from git, using a revision hash to make sure the correct version is downloaded, fixes an issue outlined in #3577
* keeps cmake as build environment (installed via homebrew)

### DIFF
--- a/scripts/apothecary/formulas/tess2/tess2.sh
+++ b/scripts/apothecary/formulas/tess2/tess2.sh
@@ -21,12 +21,14 @@ COMPILER_CTYPE=clang # clang, gcc
 COMPILER_CPPTYPE=clang++ # clang, gcc
 STDLIB=libc++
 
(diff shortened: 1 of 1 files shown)


danoli3:
The link for the git repo is already in the script here: https://github.com/tgfrerer/openFrameworks/blob/fix-apo-libtess2-ios/scripts/apothecary/formulas/tess2/tess2.sh#L15

It just was using the old google code link by accident.

Might be better to use the link at line 15 in your download script part, for continuity with other scripts (which would have solved this issue to begin with D:)

danoli3:
Okay worked it out @tgfrerer thanks for pointing all this out. I think I'll merge your branch into mine and I'll create another PR with all the other changes.

I've updated the tess2 script to memos latest as well as recreated and fixed the patch.
The cmakefile we have is still solid for the new version. 
The patches fix all the issues I was having with tess2 in 0.9.0 test branch!

PR #3615, opened 2015-02-04
Apothecary Formula - Tess2 fixes

danoli3:
Relevant Former PR: https://github.com/openframeworks/openFrameworks/pull/3613

Build Status: [Build Status](https://travis-ci.org/danoli3/apothecary-den/builds/49404443)

### DIFF
--- a/scripts/apothecary/formulas/tess2/tess2.patch
+++ b/scripts/apothecary/formulas/tess2/tess2.patch
@@ -1,7 +1,65 @@
-diff -ru tess2/Source/geom.c tess2/Source/geom.c
---- tess2/Source/geom.c
-+++ tess2/Source/geom.c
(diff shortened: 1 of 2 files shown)


danoli3:
Verified Build passing.

Merging now so others can build up the tess2 script.

tgfrerer:
looking good! 

one thing: if we use the REV sha `24e4bdd4158909e9720422208ab0a0aca788e700` instead of `master`, we can be sure that we're locked to the correct revision for the patches to apply cleanly.

danoli3:
Yeah good idea. I'll patch that directly.

tgfrerer:
i've just discovered that the osx version would not link against the proper stdlibs. have updated the recipe and included the REV, PR in a sec...

Issue #3599, opened 2015-01-30
ofAppiOSWindow::Settings not working in current master


minusplusminus:
Hello, I've read in the issues list that there are currently some changes in ofAppiOSWindow::Settings.  

      //  here are the most commonly used iOS window settings.
         //------------------------------------------------------
         ofAppiOSWindow::Settings settings;
         settings.enableRetina = true; // enables retina resolution if the device supports it.
         settings.enableDepth = false; // enables depth buffer for 3d drawing.
         settings.enableAntiAliasing = true; // enables anti-aliasing which smooths out graphics on the screen.
         settings.numOfAntiAliasingSamples = 2; // number of samples used for anti-aliasing.
         settings.enableHardwareOrientation = true; // enables native view orientation.
         settings.enableHardwareOrientationAnimation = false; // enables native orientation changes to be animated.
         settings.rendererType = OFXIOS_RENDERER_ES1; // type of renderer to use, ES1, ES2, etc.
    
         ofAppiOSWindow window(settings);
    ofSetupOpenGL(&window, 0, 0, OF_FULLSCREEN);
    
    window.startAppWithDelegate("BleepAppDelegate");

When I change  it to ofiOSWindowSettings, I get an crash on:

     //------------------------------------------
     void ofCoreEvents::notifySetup(){
     	ofNotifyEvent( setup, voidEventArgs );
     }

danoli3:
Yeah seems to be an issue in the custom delegate only.

Something deeper with the new windowing code and how it works with custom
delegates.

I did notice this when updating the iOS example main functions.

@julapy any ideas?

danoli3:
More information from @tallavi 
> My application is initializing from within a viewController inside a storyboard. I copied the idea from the > iosStoryBoard example.

>Now it seems that the init process is changed. I tracked the unneeded destruction to these lines in >ofxIOSEAGLView.mm:

>if(app != ofGetAppPtr()) {              // check if already running.
>    ofRunApp(shared_ptr<ofBaseApp>(app));    // this fallback only case occurs when app not created in >main().
>}
>The code goes into the if since the app was not created but not ran, and after exiting this line, I get the >destructor because is seems that no one on the inside took that shared_ptr and saved it! I followed the >code inside and I verified that the app is being put in the map that maps windows into apps. I don't >understand why it's getting released! 
>http://forum.openframeworks.cc/t/help-crashing-because-my-app-is-being-destoryed-by->ofxioseaglview/18410/2

[...]

PR #3617, opened 2015-02-04
ios mainloop fixes

julapy:
this is a fix to get ofxiOS working correctly with the new multi-window feature.

@arturoc,
found one issue where after `windowLoop()` was called, `windowsApps` was being cleared.
this resulted in `ofGetAppPtr()` returning NULL when while iOS (and Android) apps are running.

to fix this ive created an `exit()` function in `ofMainLoop` which is called after the while loop is finished and the app is ready to exit. the `exit()` function is also called by ofxiOS when it needs to remove apps and remove the event listener.

let me know if you see any issue with these modifications.
be good to get this in soon as its causing crashes with some of the ios examples.

### DIFF
--- a/addons/ofxiOS/src/core/ofxiOSEAGLView.mm
+++ b/addons/ofxiOS/src/core/ofxiOSEAGLView.mm
@@ -17,6 +17,7 @@ static ofxiOSEAGLView * _instanceRef = nil;
 
 @interface ofxiOSEAGLView() {
     BOOL bInit;
+	shared_ptr<ofBaseApp> appSharedPtr;
 }
 - (void)updateDimensions;
 @end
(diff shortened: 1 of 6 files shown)


danoli3:
Fixes the issue reported in #3599

:+1: 

- Tested iOS Storyboards examples
- Tested iOS Native View
- Tested normal iOS examples

More solid than before XD

arturoc:
looks good, the only problem i see is that for windows with their own loop, (glut and android) the exit sequence is not called any more, right? the exit call should be moved after the else as it was before?

the checks for this are that apps don't crash at the end, in both glfw and glut, that exit is called for apps in both windows

if that works android should be fine too  but i'll take a look when i have a moment once it's merged

nneonneo:
Yeah, currently Android is crashing because `MainLoop::loop()` frees the window, after which `setup()` crashes the app.

julapy:
@arturoc 
on ofxiOS when `windowLoop()` is called, the run loop is created using a asynchronous timer class which fires on every frame to call update and draw. since its async, after `windowLoop()` is called, it will then continue on through the function and clear `windowsApps`... it doesn't circle around in a while loop like osx does.

arturoc:
oh ok, then it should be good. thanks!

PR #3618, opened 2015-02-04
fix apothecary fat osx lib for tess2

tgfrerer:
+ before, only libc++ based library slices were generated for tess2 on os x 
+ this uses a more explicit build for os x, making sure we have the correct std libraries linked to the i386 and x84_64 slices.
+ also compiles libtess2 with o3 (fastest) for desktop os x
+ cleans up osx build script slightly


danoli3:
Awesome! I'll give it a whirl on travis

danoli3:
[Build Status](https://travis-ci.org/danoli3/apothecary-den)

:+1:

PR #3622, opened 2015-02-05
List view text box callback fix

i-n-g-o:
TextBox:
- using callback okPressed passing the text as an argument

ListView:
- using callbacks for cancelPressed, listItem pressed and okPressed
- okPressed uses int or a boolean array as an argument depending if it is single- or multiselection
- able to control multiselection when opening listView
- selection-color for list-items


bilderbuchi:
pinging @openframeworks/android

Issue #3614, opened 2015-02-04
Dependencies are missing their License Agreements with their Source/Headers in /libs/


danoli3:
Just noticed in the https://github.com/openframeworks/openFrameworks/tree/master/libs folder for all of the repositories we are missing the license agreements from all or most of the libraries.

BSD (Poco for example) https://github.com/pocoproject/poco/blob/develop/LICENSE
> ... The copyright notices in the Software and this entire statement, including
>the above license grant, this restriction and the following disclaimer,
>**must be included in all copies of the Software**, in whole or in part, and
>all derivative works of the Software, unless such copies or derivative
>works are solely in the form of machine-executable object code generated by
>a source language processor. ...

Do we have to include the LICENSE.txt/.md from these 3rd party dependencies with their headers?

I know we are acknowledging them in the https://github.com/openframeworks/openFrameworks/blob/master/LICENSE.md with links to the original source, however are we still breaking some sort of agreement?

@arturoc

kylemcdonald:
i don't think we need to modify any of the header files, but it would be great to get the license files and put them adjacent to the libraries.

danoli3:
Yeah I didn't mean modify the source headers whoops, I meant add the license files with the header locations / libraries as you suggested :D!

bilderbuchi:
strange that they aren't in there yet, aren't we mainly just dropping the libs in? in any case, :+1: to adding the license files.

kylemcdonald:
i know that apothecary has a working directory for checking out source, but it would be AMAZING if after cloning OF and running apothecary it was also responsible for copying the headers and licenses of these libraries into the libs/ folder (and we left everything but the openFrameworks folder empty).

bilderbuchi:
that would probably also ensure that headers and compiled libs always stay in sync.

danoli3:
Yeah I think Apothecary should have another step to the API - like ```copy()``` except only for the license files.

So we can do 
```./apothecary -t ios license tess2```

For example to just copy the license files.
And also make this a default step in the update process.

@danomatika

danomatika:
No, that's too fine grained a step to create a new function for. Simply add the cp line to the copy() function in the formulas. Super easy.

I would have added this when I started Apothecary, but my original goal was to replicate the existing libs layout at the time which didn't include LICENSE files, as fas as I can recall. Definitely should add that now.

danomatika:
For Poco this would be simply:

    cp -v LICENSE $1/include/Poco

[...]

PR #3623, opened 2015-02-06
Apothecary Formula Only - License Copy Process

danoli3:
Fixes: **Dependencies are missing their License Agreements with their Source/Headers in /libs/**
Issue Reference: https://github.com/openframeworks/openFrameworks/issues/3614

**Apothecary script formula only.**
Tested and working via apothecary-den for iOS / OS X scripts (most crossover)
Double checked all the others that I couldn't build directly.

**Fixes minor issues in:**
- scripts/apothecary/formulas/kiss.sh (broken if statement)
- scripts/apothecary/formulas/videoInput.sh (missing branch @ofTheo had merged to master)

### DIFF
--- a/addons/ofxAssimpModelLoader/scripts/formulas/assimp.sh
+++ b/addons/ofxAssimpModelLoader/scripts/formulas/assimp.sh
@@ -320,6 +320,9 @@ function copy() {
 	else
 		cp -Rv lib/libassimp.a $1/lib/$TYPE/assimp.a
 	fi
+
+    # Copy License File
+    cp -v LICENSE $1/LICENSE
 }
 
 # executed inside the lib src dir
--- a/scripts/apothecary/formulas/videoInput.sh
+++ b/scripts/apothecary/formulas/videoInput.sh
@@ -9,7 +9,7 @@
 FORMULA_TYPES=( "vs" "win_cb" )
 
 # define the version
-VER=update2013
+VER=master
 
 # tools for git use
 GIT_URL=https://github.com/ofTheo/videoInput.git
@@ -17,7 +17,7 @@ GIT_TAG=$VER
 
 # download the source code and unpack it into LIB_NAME
 function download() {
-	git clone https://github.com/ofTheo/videoInput.git -b update2013
+	git clone https://github.com/ofTheo/videoInput.git
 }
 
 # prepare the build environment, executed inside the lib src dir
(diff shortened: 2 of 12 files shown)


danoli3:
@danomatika @kylemcdonald @arturoc 

Thumbs up for merge? :D!

danoli3:
> #### So new outlook after testing all of the scripts and the structure 
> I'm probably still for a pure ```license``` folder in each lib directory. If one and or many files.
>
>#### Reasons for the license folder: 
>- Developers at a glance know where the license information is
>- Encapsulates ambiguous source license filenames. "readme" "randomlibname" "COPYING" "COPYING.txt" "COPYING.md" "LICENSE" "Lisence.txt"  "LICENSE.txt" "gplv2" "BSD.txt" etc...
>- Easier to update and keep track of
>- Multiple license files in one location for each lib (if applicable)
>- Doesn't mess up structure when navigating into libraries headers or lib folder.
>
>#### Reasons against: 
>- Extra folder

https://github.com/openframeworks/openFrameworks/issues/3614#issuecomment-73185896

scripts/apothecary/formulas/videoInput.sh
```diff
@@ -9,7 +9,7 @@
 FORMULA_TYPES=( "vs" "win_cb" )
 
 # define the version
-VER=update2013
```
danomatika: Was this a change you intended to include in the license commit?

addons/ofxAssimpModelLoader/scripts/formulas/assimp.sh
```diff
@@ -320,6 +320,9 @@ function copy() {
 	else
 		cp -Rv lib/libassimp.a $1/lib/$TYPE/assimp.a
 	fi
+
+    # Copy License File
+    cp -v LICENSE $1/LICENSE
```
danomatika: There's no need to name the file when copying it here (and in the other formulas). Just do:

    cp -v LICENSE $1/

scripts/apothecary/formulas/videoInput.sh
danoli3: As mentioned, yes

ofTheo:
This looks good to me to merge. Either with or without the fix @danomatika mentioned.

kylemcdonald:
merging this in it's current state as future changes are not difficult to make if someone wants to put the work into making them.

danomatika:
Here you go @kylemcdonald #3631

PR #3624, opened 2015-02-06
apothecary: fix cairo + dependencies

tgfrerer:
+ cairo binary was in fact not properly fat: this is now fixed (and verified).
+ use fat clang flags to build cairo + dependencies (https://gist.github.com/tgfrerer/8e2d973ed0cfdd514de6)
+ simplifies build scripts
+ makes sure that binaries are linked against the correct stdlib for their build type.
+ removes multicore build (-j)


tgfrerer:
manual merge to "freshen up" PR after licence inclusion jumbled up things. tested correct build.

ofTheo:
can we merge this now as its only the formulas?

PR #3625, opened 2015-02-06
apothecary: add osx-fat-fmodex script

tgfrerer:
+ this is a download + copy-only script, since fmodex appears not to be available in source.
+ downloads fat binary release of fmodex for os x (64/32bit), 
+ linking verified against libstdc++ and libc++.

The current oF:master dylib supplied for fmodex on os x is slim. 
+ downloads headers and fat binary dylib for fmodex from fmod.org. 
+ Version downloaded (42221) is the version closest to the current headers in master, since the version referenced in the current headers (42220) is not available as a download for os x.

### DIFF
--- a/scripts/apothecary/formulas/fmodex.sh
+++ b/scripts/apothecary/formulas/fmodex.sh
@@ -10,7 +10,7 @@
 FORMULA_TYPES=( "osx" "vs" "win_cb" )
 
 # define the version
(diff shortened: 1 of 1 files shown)


bilderbuchi:
@tgfrerer curious that it's not available as source - what license is this under, and can we even use it?

Issue #3620, opened 2015-02-05
Support for the Raspberry Pi 2 B


pizthewiz:
As reported in the forum thread [Raspberry Pi 2 install failing](http://forum.openframeworks.cc/t/raspberry-pi-2-install-failing), it sounds as though the existing Raspberry Pi support is detected via the cpu type which has gone from ARM11/`ARMv6` in the A/+ and B/+ to Cortex-A7/`ARMv7` with the 2 B.

kylemcdonald:
i'm adding this to 0.9.0 under the assumption that it's not a big change, but it's important. if it turns out to be very complicated, or testing compatibility is delaying the release further, we can push it back.

bilderbuchi:
pinging @openframeworks/linuxarm who should know how to best fix this.

bakercp:
I'm waiting on one as well, so I should be able to give a little better input.  Here's what I'd do:

1.  Copy this file into the linux armv6l folder 

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/linuxarmv6l/config.linuxarmv6l.default.mk

and call it `config.linuxarmv7l.rpi2.mk`.

2. Then I'd change My remove the `pkg-config` checks so to allow me to figure out what (if any) additional libraries need to be added to the install-dependencies script.

3. Then I'd make obvious changes like changing `armv6l` to `armv7l` [here](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/linuxarmv6l/config.linuxarmv6l.default.mk#L100),

4.  Then I'd try to run it directly on the quad core RPI2 with a command like this:

```
$ export MAKEFLAGS="-j4 -s PLATFORM_VARIANT=rpi2"
$ make
```

rc1:
I can confirm this works when the architecture is set to: `armv7-a`.

PTS93:
Indeed I can confirm that, when using armv7-a I can successfully compile.

bakercp:
Great!  Can one of you submit a PR with the rpi2 configuration variant?  I'll edit my instructions above as well.  So -- in addition to compiling, does it seem to be running OK?

arturoc:
probably we don't need a specific variant an instead just detect the platform architecture? that way is easier to maintain if we add libraries... since it seems everything else is compatible

bakercp:
[...]

PR #3626, opened 2015-02-06
Added fixed make file for RPi2

PTS93:
Added fixed make file as described in this issue
https://github.com/openframeworks/openFrameworks/issues/3620


bakercp:
Confirmed working by many here https://github.com/openframeworks/openFrameworks/issues/3620.

PR #3631, opened 2015-02-08
seem small apothecary updates

danomatika:
Some small apothecary updates:

* Don't need explict names when copying license files, related to #3623 
* Don't assume we have the android sdk installed, only set sdk vars when targeting android and exit with error if we can't find the path file. Also, this was opening a file with a relative path *before* the script actually changes dir so it would break if you we're not running apothecary from it's own dir. Apothecary is designed to be called from any location so it's easier to script.
* Updated readme with info on custom echo wrappers
* Small formula whitespace fixes

I also noted the following which I have questions about:

1. There are two kiss formulas right now, one should be deleted: `kiss/kiss.sh`& `kiss.sh`

2. There is lots of [commented stuff in openssh.sh](https://github.com/danomatika/openFrameworks/blob/master/scripts/apothecary/formulas/openssl.sh#L243), can we delete it?

3. freetype.sh has [commented copy commands for older versions](https://github.com/danomatika/openFrameworks/blob/master/scripts/apothecary/formulas/freetype.sh#L343), if we're not using version < 2.5, can we remove it?

4.  Better feedback/debugging? Whoever updated freeimage & poco added lots of redundant echos like "built successful" etc. A lot of this info is actually printed by apothecary already especially when you run it with the `-v` verbose flag (which is **highly** recommended for debugging). Also, there are a number of custom echo commands (echoWarning, echoError, echoInfo, echoVerbose) which provide better feedback through coloring. `echoVerbose` is especially useful for debug prints. Anyway, I updated the [readme to note these](https://github.com/danomatika/openFrameworks/tree/master/scripts/apothecary#custom-apothecary-echo-wrappers) and I suggest not using `echo` at all, if possible.

5. There are a fair number of bash style clashes ... which is not a suprise I guess with numerous authors at this point

### DIFF
--- a/scripts/apothecary/README.md
+++ b/scripts/apothecary/README.md
@@ -469,7 +469,7 @@ Bash provides an easy way to debug scripts by `set -x`. Put this command in your
 
 **PLEASE PLEASE PLEASE comment your formulas!** Writing build scripts sucks! Don't make it even more hell for people that have to try and maintain what you've done. Note *any* assumptions you're making **especially** regarding things that may change based on newer releases of a library.
 
-Last, if you write a formula but can't finish/test it on all platforms, etc make a note of this using the `echoWarning` print command. This is a function provided by apothecary that prints text in <span style="color: #FF0">yellow (caution, right?)</span>. Start your info string with with "TODO: ":
+If you write a formula but can't finish/test it on all platforms, etc make a note of this using the `echoWarning` print command. This is a function provided by apothecary that prints text in <span style="color: #FF0">yellow (caution, right?)</span>. Start your info string with with "TODO: ":
 
 	if [ "$TYPE" == "win_cb" ] then ;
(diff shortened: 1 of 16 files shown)


kylemcdonald:
great. would be good to get a quick check from @openframeworks/apothecary 

does it make sense to enable verbose by default? people might be adding echoes because they don't realize the behavior is there.

danomatika:
>does it make sense to enable verbose by default?

No. It's verbose for a reason :D It's for debugging.

arturoc:
seems good to me,  mostly the android stuff should be merged right away. haven't been following the licensing part so not sure about that. i'll merge it unless someone has any complaints :)

danomatika:
The licensing stuff is merely a small detail. The bigger points for me are to remove one of the kiss formulas and cleaning up the commented formulas.

arturoc:
perfect, thanks!

PR #3137, opened 2014-08-11
Apothecary: OpenCv 2.4.9

ofTheo:
Static libs, headers and apothecary formulas for architectures included:
- [x] OSX FAT (32 bit w/ libstdc++ and 64 bit w/ libc++)
- [x] iOS FAT (armv7, armv7s, arm64, i386, x86_64 all with libc++)
- [x] Linux (32 bit)
- [x] Linux (64 bit)
- [x] Linux Armv6l
- [x] Linux Armv7l
- [x] Android - armeabi (@arturoc ?) 
- [x] Android - armeabi-v7a (@arturoc ?)
- [x] Android - x86 (@arturoc ?)
- [ ] VS (?)
- [ ] win_cb (?)

----

Creates a single FAT opencv.a lib with O3 optimization and 64bit libc++ and 32bit libstdc++

don't merge yet. 
still to do. 

- [ ] don't use build root - use build folder. 
- [x] find a better way to keep _buildroot folder clean ( right now lipo will suck up any lib in _buildroot/lib/)
- [x] remove libtiff lib 

ping @pizthewiz 

changes
- removed OpenCL as it was causing errors 
- removed some other libs not needed ( CUDA etc ) 
- lipo for loop wasn't working - fixed. 
- error suppression for lipo incase lib is already FAT 
- libtool into a single opencv.a 

formula based off of @pizthewiz gist - addresses #3039

PR #3635, opened 2015-02-09
apothecary formula for opencv

ofTheo:
This is a just the apothecary formula for OpenCv. 
Windows isn't currently working all other platforms should be good. 

This is pulled from #3137 and should be good to merge without breaking the current master by updating the libs. 

@danoli3 @bakercp - if this looks good I'll merge it in. 

@elliotwoods once this is in master it should be pretty easy to try it out on Windows.


arturoc:
android and emscripten are also missing, it's probably better to leave the apothecary PRs open until all platforms are fixed so it's easier to keep track of what's missing

PR #3637, opened 2015-02-10
allowing other addons directories

quinkennedy:
Adding the ability in the Project Genenerator to specify alternate
Addons directories to load addons from.

Outstanding Items:
* This includes the examples and .md files etc in the addon's root, which I don't think are included when using the current project generator
* Only tested on OSX 10.9
* Not tested with an addon that includes a framework

suggestions for test cases welcome!

### DIFF
--- a/apps/devApps/projectGenerator/src/addons/ofAddon.cpp
+++ b/apps/devApps/projectGenerator/src/addons/ofAddon.cpp
@@ -344,11 +344,9 @@ void ofAddon::fromFS(string path, string platform){
 
     clear();
     this->platform = platform;
-	name = ofFilePath::getFileName(path);
-	addonPath = ofFilePath::join(getAddonsRoot(),name);
-
+    name = ofFilePath::getFileName(path);
     string filePath = path + "/src";
-    string ofRootPath = ofFilePath::addTrailingSlash(getOFRoot()); //we need to add a trailing slash for the erase to work properly
+    addonPath = path;//ofFilePath::join(getAddonsRoot(),name);
 
     ofLogVerbose() << "in fromFS, trying src " << filePath;
 
@@ -356,8 +354,8 @@ void ofAddon::fromFS(string path, string platform){
     getFilesRecursively(filePath, srcFiles);
 
     for(int i=0;i<(int)srcFiles.size();i++){
-    	srcFiles[i].erase (srcFiles[i].begin(), srcFiles[i].begin()+ofRootPath.length());
-		//ofLogVerbose() << " srcFiles " << srcFiles[i];
+        srcFiles[i] = getRelPath(getOFRoot(), ofFilePath::getEnclosingDirectory(srcFiles[i])) + ofFilePath::getFileName(srcFiles[i]);
+		ofLogVerbose() << " srcFiles " << srcFiles[i];
     	int init = 0;
 #ifdef TARGET_WIN32
     	int end = srcFiles[i].rfind("\\");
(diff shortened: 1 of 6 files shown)


kylemcdonald:
this is a great feature to have, but i'm worried that contributions to this app will be short lived.

i think new work on the PG is happening at https://github.com/openframeworks/projectGenerator and @ofzach and @nongio are leading that work right now. maybe they can provide some feedback?

ofZach:
can you explain a use case for this change?

quinkennedy:
@ofZach Yea, I can try to explain. 

tl;dr: directory structure flexibility

ntl;tdr:
When working on projects, I tend to do at least one of two things:

* include addons in the repository to make it easier for people working on the project to make sure they are using the same version of the same addon
* have multiple openFrameworks apps in a repository

The biggest impetus for this change was the first bullet point. In the past I have generated a project using the project generator and then do the following manually to migrate to in-repo addons:

* Edit the Header Search Paths
* Edit the file locations in the project explorer
* Hope

-------
There is also a hope here to break the requirement of having projects always 3 levels deep inside the oF base directory. This comes from three different directions:
* being offended by Cinder folk claiming that oF is "crippled by its directory-structure reliance"
* wanting to include oF as a submodule in my repository for versioning, rather than a note in the README saying "download this particular version of oF"
* wanting to organize oF projects in my repository without being restricted to X number of levels depending on where I check it out.

quinkennedy:
@kylemcdonald I would be happy to look at migrating my changes to the current PG repo. The reason that I did it on this version was because I have had problems using the apps/projectGenerator/projectGeneratorSimple in the past and have found apps/devApp/projectGenerator to be more reliable. That is probably coming from my being more familiar with the latter (as opposed to any inherent usability/reliability difference) though.

arturoc:
i agree it's great to have this. it makes it so much easier to work with addons when collaborating with other people through some versioning system. we should add it to the makefile system too

arturoc:
btw most of this classes are shared between the PG simple and this so it should be relatively easy to merge this changes. i think there's been some recent changes from  @danomatika to add support for cpp flags but apart from that most files should be identical in both projects

quinkennedy:
Is there an official sun-setting schedule for the PG in this repo? What I'm wondering is if I should be attempting to continue developing this feature in both repositories.

ofZach:
I've moved PG code here, although it's not really a complete repo yet and just has a command line project against the PG code: 

https://github.com/openframeworks/projectGenerator

give me a few days, and I'll get everything in this repo, then we can take issues and changes / PRs there

quinkennedy:
ah, I got confused between openframeworks/projectGenerator and ofZach/projectGeneratorSimple. So I'll wait for @ofZach to give me the go-ahead to migrate these changes to openframeworks/projectGenerator.

So now my question is, should these changes also be migrated to ofZach/projectGeneratorSimple?

ofZach:
i think the should be in neither!!   my plan is to move both PGs (devApps / simple) to the new repo, since they have pretty much the same codebase.   Then, we can make adjustments to the underlying library and all the PGs can have that benefit and no more worry about sync, multiple places to make PRs, etc.  My overall plan is to make just one command line tool that will have different front ends -- there's currently an example of that command line tool in that new repo, but it's slightly wonky to use it on OSX given how app bundles work, etc.

quinkennedy:
Awesome @ofZach that is what I was hoping for. Lemme know when you are comfortable with me digging into it.

danomatika:
This sounds good @ofZach.

mattfelsen:
Hey @quinkennedy, this sounds great – exciting to check it out soon. At the last NYC OF meetup, we talked about how we structure our OF projects at Local Projects using submodules so that we can version both OF itself and addons, and keep the source for all of those things outside the project repo. It's written up here, hit me up if you have more questions! http://local-projects.github.io

The kicker is that since we keep project addons outside OF, the PG doesn't know about them. So your changes would be very welcome for when you're creating a new app & know that you want to add a bunch of addons upfront. For existing projects, the Xcode OFPlugin does the trick for setting relative addon paths correctly.

@ofZach finally got around to checking out the command line PG since bugging you about it a few months go :stuck_out_tongue_closed_eyes: Looks great, and excited to play with this as well :+1:

arturoc:
closing this since it was fixed by adding support for "local addons" which allows to add addons from anywhere

thanks!

Issue #3621, opened 2015-02-05
bringing our glfw branch up to date with glfw:master without breaking keyInput


igiso:
are these the only changes we've made to the glfw core?:
https://github.com/arturoc/glfw/compare/feature-keysUnicode

if so,

is it ok If I fork the glfw master, and add these 8 changes to my branch and then make a pull request to arturoc's glfw  so we can be up-to date?

I can also add libglfw3.a to master

tgfrerer:
or we could point apothecary to @arturoc's fork, that way, we'd be up to date for 0.9.

arturoc:
hey, yeah i've been meaning to do it for a while, those are the only changes so merging with latest master shouldn't be too problematic. @igiso if you can send me a PR to that branch that'd be super useful.

and yes we should point apothecary to that branch

igiso:
I think you might want to do some offline testing before merging to your online branch


I've implemented your code and it works fine, but while I was editing I noticed that they have added
some  "const unichar" inside the cocoa_window.m (keyDown)func
(that weren't there before)
not sure if they have toggled this in a different manner, (that might help us go back to master)

Also:
current glfw master seems to behave a bit differently with our current branch of openFrameworks anyway
(regardless our changes) //


the window appears like.. all over the place,..  (on start up )

to reproduce it 
here is my compiled library for mac 32bit and the two headers:
https://www.sendspace.com/file/rwi9cf 

just put them in and
call:
    ofSetWindowPosition(100, 300);
in setup
 clean and build...

maybe it is related to the other issue as well..

I can make a pull request anyway.. but I think you should take a look first.

igiso:
ok. I've solved the windowing "problem"
… 

it is because position is initialised as zero and is called in  ofAppGLFWWindow:setup
and then if you call set windowPos in ofAppsetup it goes all over the place..

igiso:
we can make propper gui with the new glfw floating undecorated windows..
![screen shot 2015-02-07 at 7 46 17 pm](https://cloud.githubusercontent.com/assets/2058400/6093063/0969d818-af02-11e4-8f23-cec03f0da544.png)

pizthewiz:
I think this needs the 0.9.0 milestone as it is required by #3627.

igiso:
here it is..
https://github.com/igiso/glfw/compare/glfw:master...master

arturoc:
[...]

PR #3646, opened 2015-02-12
Apo glfw arturoc patch

tgfrerer:
Use @arturoc patch of glfw to support unicode #3621


bilderbuchi:
@tgfrerer would you put a `TODO` or something (an issue?) in that portion of code to at least mark this for a later update, when we can swith to glfw again?
Also, the comment you added is misleading: `# temporary fix until https://github.com/openframeworks/openFrameworks/issues/3621 is resolved.`, when #3621 is, currently, marked as resolved. ;-)

PR #3650, opened 2015-02-14
Apothecary License Update With License Folder

danoli3:
- Adds Copy License Process for OpenCV
- Adds folders for all types for encapsulation and ease of
identification due to ambiguous license file names in same libraries
cases.
- Validated Working via https://github.com/danoli3/apothecary-den

Related Issue: https://github.com/openframeworks/openFrameworks/issues/3614


danoli3:
Be good to get this merged before the think about building the libs for OS X. @kylemcdonald @tgfrerer

bilderbuchi:
man, all this often-duplicated code makes me really file like the license copy part should be in a separate function ;-)

danoli3:
Yeah it probably could be a separate apothecary function which works like
download / update / copy and be then accessible from the command line for
if the Dev just wanted to copy licenses (say they already compiled before
this PR :P )

The function could do the folder rm and mkdir and then implement a call to
the child script like copy does.

Then custom Filenames could be copied at that level (as currently).

The function would have to first call download then the license.
This could reshuffle the current code PR Around, but would still require
the core source license to destination copy commands.

Still think this is solid for now. Will need to get dan involved for the
core apothecary reshuffle.

bilderbuchi:
I didn't mean as a user-accessible function, but as a bash-internal one, following the DRY (don't repeat yourself) principle.

danoli3:
The only repeated part is the remove folder if exists and make folder... not really necessary at the higher level I don't think. There are many repeated parts like: 
https://github.com/openframeworks/openFrameworks/blob/master/scripts/apothecary/formulas/freetype.sh#L322 for example following the same kind of idea.

If we were to make it at a higher level, it would have to be around here I believe:
https://github.com/openframeworks/openFrameworks/blob/master/scripts/apothecary/apothecary#L674

I still think this is solid enough as is and should merge to consolidate and being the scripts in line again.

danoli3:
I'd like to merge this soon. @bilderbuchi 's idea I think we can put aside for now and think about later.

@bakercp keen?

bakercp:
+1 sorry for the delay.

PR #3655, opened 2015-02-14
fixing compileOF.sh (for ARMv7)

tpltnt:
Hi *,

this patch removes an overly specific path to the "chown" command during the compilation of oF. At least on ARMv7 the $LIBSPATH (of compileOF.sh) conflicts/is out of sync with the $PLATFORM_LIB_SUBPATH (of config.shared.mk). The former is used for the "chown" command, the latter for the actual compilation output. The "libs/openFrameworkCompiled/lib" directory should be owned by the user anyway.

Cheers,
tpltnt


bilderbuchi:
ping @arturoc

arturoc:
thanks!

PR #3661, opened 2015-02-17
fix ofEasyCam segfault when en/disableMouseInput within ofApp::setup()

tgfrerer:
* since `ofEasyCam.events` will only get set upon first use (`begin`) of the camera, enabling or disabling `mouseInput` for a camera would push a nullpointer onto the `ofEvents` stack for the camera in question, with dreadful results.

This fix addresses the issue by checking `events` against null, and applying the `mouseInput` only when possible. it also makes sure to honour any settings for the camera which have been passed during `ofApp::setup`.

Also adds some inline comments to explain what's happening to facilitate code maintenance.

### DIFF
--- a/libs/openFrameworks/3d/ofEasyCam.cpp
+++ b/libs/openFrameworks/3d/ofEasyCam.cpp
@@ -160,35 +160,56 @@ char ofEasyCam::getTranslationKey(){
 
 //----------------------------------------
 void ofEasyCam::enableMouseInput(){
-	if(!bMouseInputEnabled){
-		bMouseInputEnabled = true;
-		bEventsSet = true;
+	if(!bMouseInputEnabled && events){
 		ofAddListener(events->update, this, &ofEasyCam::update);
 		ofAddListener(events->mouseDragged , this, &ofEasyCam::mouseDragged);
 		ofAddListener(events->mousePressed, this, &ofEasyCam::mousePressed);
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

admsyn:
I think this means #1836 can be closed

tgfrerer:
Yes, indeed. 

Thanks for spotting this - strange thing is that it seemed to have been fixed for a little while, but a couple of months ago, probably a refactor in the events system must have surfaced the segfault again. Anyway, it should be fixed now =)

kylemcdonald:
@tgfrerer this breaks ofEasyCam for me :( minimal example below:

```c++
#include "ofMain.h"

public:
    ofEasyCam cam;
    void draw() {
        cam.begin();
        ofDrawBox(100);
        cam.end();
    }
};
int main() {
    ofRunApp(new ofApp());
}
```

behavior of 5b538b940f4b81e7b02eb46505d0b9a69545e863 is that the screen is white and clicking anywhere doesn't seem to change that. behavior before with parent commit 662830f5e6a7d8caa3779707edfa2b6a3bdbdf34 was a box was drawn in the center of the screen and you could drag to rotate.

tgfrerer:
Could it be that the camera starts out without mouse input enabled?

Do you mind testing if it works if you call `camera.enableMouseInput` in `setup()` - knowing this would make it easier to fix.

PR #3664, opened 2015-02-19
fix mainloop segfault upon window close

tgfrerer:
fixes an issue where closing the main window via mouse could crash an app.

+ if a window or app was closed through system input (via a mouse clicking on tiny x'es) or Command-Q, the ofMainLoop was bound to overshoot and cause a postmortem segfault.

This is because the iterator used to loop through all open windows was invalidated after a closed window was erased. This PR proposes to set the window iterator to the return value of `erase()` should the `windowsApps` `map` get modified. 

`erase()` is guaranteed to return the next position after the deleted element, and `::end` if there is no more element, so the loop can function as expected.

### DIFF
--- a/libs/openFrameworks/app/ofMainLoop.cpp
+++ b/libs/openFrameworks/app/ofMainLoop.cpp
@@ -113,15 +113,16 @@ int ofMainLoop::loop(){
 }
 
 void ofMainLoop::loopOnce(){
-	for(auto i = windowsApps.begin();i!=windowsApps.end();i++){
+	for(auto i = windowsApps.begin(); !windowsApps.empty() && i != windowsApps.end() ;){
 		if(i->first->getWindowShouldClose()){
 			i->first->close();
-			windowsApps.erase(i);
+			i = windowsApps.erase(i); ///< i now points at the window after the one which was just erased
 		}else{
 			currentWindow = i->first;
 			currentWindow->makeCurrent();
 			currentWindow->update();
 			currentWindow->draw();
+			i++; ///< continue to next window
 		}
 	}
 	if(pollEvents){


arturoc:
i think @admsyn had a fix for this issue but cannot find it right now, i'm merging this cause that one seemed a little bit less understandable. @admsyn if you can close yours in case it's still open?

tgfrerer:
Oh, sorry @admsyn! - I did a check on the open PRs list and #3588 must have slipped through my net...

admsyn:
Fair enough :) The i = erase(i) is more elegant, I was assuming we couldn't use those c++11 goodies internally yet, so good to know we can!

tgfrerer:
;) since i is an iterator of the right type `i= erase()` is plain old, Scott Meyer cpp98 =)

admsyn:
FWIW this is giving me a compile error on Xcode 6 (OSX 10.10) (`windowsApps.erase(...)` is returning void). 

screen shot 2015-02-19 at 12 45 50 pm

Apologies if I'm missing something RE OSX C++11 developments, though, haven't had the time to keep up with OF the last few weeks

tgfrerer:
Adam, you are 100% right! I am sorry & stand corrected. 

I rashly assumed that `windowsApps` was a `std::vector` and had double-checked against the reference of `std::vector` - where `erase()` returned an iterator already in cpp98. 

It is, in fact (and as you pointed out in #3588) a `std::map`, where the iterator as a return value is only returned with cpp11.

I believe the idea is to keep the core cpp98 for now, which allows you to use cpp98 and cpp11 in ofApps.

There is however a simple fix for it, and I'll send it in a moment, sticking to strict cpp98 (and getting rid of the `auto` for that matter), so that OS X is un-broken again.

PR #3665, opened 2015-02-20
fixes the segfault on exit fix, sticking strictly to cpp98 conventions.

tgfrerer:
* makes #3664 OS X cpp98 compatible - and un-breaks OS X builds.
* uses an explicit declaration of the iterator type in favour of c++11 'auto' (it aint pretty but it's backwards compatible and essentially the same)
* uses the idiom lined out on http://stackoverflow.com/questions/8234779/how-to-remove-from-a-map-while-iterating-it

Tested on 
- [x] Windows VS2012
- [x] OS X


arturoc:
i think this is actually @admsyn first solution :) 

thanks!

PR #3668, opened 2015-02-20
refactor renderers to remove some unused methods

### LINKED ISSUE #3667: ofMesh::draw() slow vs ofGetGLRenderer()->draw(ofMesh) in fixed pipeline
```
ofMesh mesh;

mesh.draw();
//vs
ofGetGLRenderer()->draw(mesh, false, true, false);
```
When using the fixed pipeline renderer, those two behave very differently when drawing lots of meshes. The ofMesh::draw() method is waaay slower, around ~10x slower when drawing 1000 ofMeshes. After some digging, came to the conclusion that this happens  because each ofMesh::draw() does a glPushAttrib() and a glPopAttrib() before drawing the mesh.

This was noticed on Mac OS X 10.9.

Arturo thinks this might not be necessary. See here: http://forum.openframeworks.cc/t/ofmesh-draw-slow-vs-ofgetglrenderer-draw-ofmesh/18593.

arturoc:
- fixes #3667
- removes renderer.draw(mesh) without polygon mode which was behaving
different in fixed and programmable
- fixes programable::drawInstanced(mesh) which was setting the wrong
polygon mode afterwards

### DIFF
--- a/libs/openFrameworks/app/ofAppNoWindow.h
+++ b/libs/openFrameworks/app/ofAppNoWindow.h
@@ -60,7 +60,6 @@ private:
 	void draw(const ofPath & shape) const{}
 	void draw(const of3dPrimitive&, ofPolyRenderMode) const{}
 	void draw(const ofNode&) const{}
-	void draw(const ofMesh & vertexData, bool useColors, bool useTextures, bool useNormals) const{}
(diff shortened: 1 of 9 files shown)


tgfrerer:
+1

PR #3670, opened 2015-02-21
move global texture references to private

i-n-g-o:
moving global TextureRef and TextureCacheRef to protected instance variables
each movie-player should own their own Texture and TextureCache


i-n-g-o:
joined this PR with this one: https://github.com/openframeworks/openFrameworks/pull/3736

PR #3672, opened 2015-02-21
make sure a pending timeObserver gets cleaned up use finalize...

i-n-g-o:
In case bWillBeUpdatedExternally is set to false a timeObserver would get attached to AVPlayer.
Even if AVPlayer never starts playing, the timeObserver (and all scheduled blocks) holds a reference to self (ofAVFoundationVideoPlayer)
Therefore releasing the videoPlayer in ofAVFoundationPlayer would actually not deallocate the object and leak memory

using a finalizer gives the chance to remove the timeObserver before releasing the object...

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -131,7 +131,7 @@ void ofAVFoundationPlayer::close() {
         videoTexture.clear();
 		
         videoPlayer.delegate = nil;
-		[videoPlayer release];
+		[videoPlayer finalize];
         
         if(bTextureCacheSupported == true) {
             killTextureCache();
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -131,7 +131,7 @@ void ofAVFoundationPlayer::close() {
         videoTexture.clear();
 		
         videoPlayer.delegate = nil;
-		[videoPlayer finalize];
+		[videoPlayer cleanupAndAutorelease];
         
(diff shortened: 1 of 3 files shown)


pizthewiz:
I think we should avoid the `finalize` signature, that is an existing and deprecated method on [`NSObject`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSObject_Class/index.html#//apple_ref/occ/instm/NSObject/finalize) that is/was used for garbage collection. Could the cleanup not be performed in `-[ofAVFoundationPlayer dealloc]`?

i-n-g-o:
hm.
i wanted to have just one method to be called on ofAVFoundationVideoPlayer to cleanup and release it.
i don't think it should be released in `[ofAVFoundationVideoPlayer cleanup]`.
so either we add a `[ofAVFoundationVideoPlayer release]` after calling `[ofAVFoundationVideoPlayer cleanup]` in `[ofAVFoundationPlayer dealloc]` or find another name.

pizthewiz:
Sorry I am not following, why not perform the clean up tasks in `dealloc` or are you saying there is some sort of retain cycle that prevents the player object from being deallocated?

i-n-g-o:
exactly. we cannot (should not) call cleanup in `dealloc` but have to cleanup before we release ofAVFoundationVideoPlayer. the object would keep beeing retained by timeObserver if `bWillBeUpdatedExternally == false`. but timeObserver only gets released in dealloc.

this PR is fixing is a potential memory-leak.
A similar issue in ofxAVVideoPlayer was fixed with such a construction.

i-n-g-o:
joined this PR with this one: https://github.com/openframeworks/openFrameworks/pull/3736

PR #3674, opened 2015-02-23
initially enable mouseInput for ofEasyCam

tgfrerer:
This restores the default behaviour of ofEasyCam, which used to have mouse input enabled by default, and addresses an issue reported by @kylemcdonald on #3661.


kylemcdonald:
fixed it, thanks :)

PR #3682, opened 2015-02-26
Fix #3677 ofBuffer::allocate()

### LINKED ISSUE #3677: ofBuffer.allocate() behaviour is inconsistent with other class functions
ofBuffer buffer;
    buffer.allocate(10);
    long length = buffer.size();
    // length == 9

ofBuffer (for better or worse) adds one byte to its internal vector size everywhere (?) except in `allocate()`. This means if you use `allocate()` you end up with a buffer which appears to be shorter than requested.

Fixing this will presumably break a lot of user and ofxAddon code that works around it by adding one to the value passed to `allocate()`.

As tracking the extra byte seems to be beyond the codebase, a more robust solution might be to ditch the extra byte entirely and to have functions which treat the buffer as a string check for non-NULL-terminated strings (guaranteeing NULL-termination appears to be the reasoning behind the extra byte).

OF 0.8.4 and git HEAD, all platforms.

bangnoise:
ofBuffer::allocate() didn't zero-pad the buffer, now it does. This matches behavior elsewhere so ofBuffer::size() correctly returns the allocated size. Fixes #3677.

### DIFF
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -111,7 +111,7 @@ void ofBuffer::clear(){
 void ofBuffer::allocate(long _size){
 	clear();
 	//we always add a 0 at the end to avoid problems with strings
-	buffer.resize(_size + 1);
+	buffer.resize(_size + 1, 0);
 }
 
 //--------------------------------------------------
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -110,7 +110,8 @@ void ofBuffer::clear(){
 //--------------------------------------------------
 void ofBuffer::allocate(long _size){
 	clear();
-	buffer.resize(_size);
+	//we always add a 0 at the end to avoid problems with strings
+	buffer.resize(_size + 1);
```
arturoc: actually can you change it to:

```
buffer.resize(_size + 1,0)
```

so the buffer is set to 0, else the final char can be anything and won't work as null terminator for strings

bangnoise:
oops, done

arturoc:
thanks!

PR #3627, opened 2015-02-06
oF 64 bit, c++11 support (only merge once os x apothecary build binaries are in master)

tgfrerer:
I guess an image says it better than 1.000 words =):

screenshot 2015-02-06 16 18 30

This sets up the OS X openFrameworksLib project files and the openFrameworks project template so that you can select via drop-down whether you want to compile your openFrameworks app as 64 bit c++11, or, legacy 32 bit cpp98. Addresses #2016, #1864

---
# Use

1. Use the project generator to create or update an openFrameworks project. 
2. Open your project. You'll see this in the XCode top bar: 
screenshot 2015-02-06 16 15 20

3. Select the build target (in my case `emptyExample Debug`) to either compile the `Debug` or the `Release` configuration. 
screenshot 2015-02-06 16 37 59

4. Then select the target architecture (32 or 64 bit).
screenshot 2015-02-06 16 18 30

You can click the architecture dropdown to switch between architectures. Don't worry about selecting the target architecture for openFrameworks itself, that one is automatically detected from your current build target. 

Also, the correct stdlib is automatically used to link everything together. In 32 bit mode, you can use cpp98 like you always did. =)
screenshot 2015-02-06 16 42 39
screenshot 2015-02-06 16 19 33

But, once you switch over to 64 bit mode, you get full c++11 support:
screenshot 2015-02-06 16 16 44

---

## How this works

In XCode, you can set per architecture build flags. These build flags make sure that the 64 bit and the 32 bit version are compiled with the correct toolchains. The Project is set to `Universal` (which gives you the architecture selection dropdown), but the projects themselves are set up to only build the active architecture. This gives you either a 32 bit binary or a 64 bit binary.

I put the necessary project build settings into the CoreOF.xcconfig file. This file is included in both the App project compile settings and the openFrameworksLib project compile settings. 

For the app settings, it is included at `Target` level, meaning the settings in CoreOF.xcconfig it will take precedent over the settings selected in the `project build settings` panel. (If you need to, you can still override the settings using the `target build settings` panel).

To update "old" project files to 64 bit, using the projectGenerator with "update project" in devApps should do.

---

# Pre-Requisites

This works once apothecary has been used to compile out all necessary fat libs: 

	./apothecary update glfw
	./apothecary update FreeImage
	./apothecary update cairo
	./apothecary update openssl
	./apothecary update poco
	./apothecary update tess2
	./apothecary update fmodex

---

# Preview

[...]

PR #3684, opened 2015-02-27
apothecary: remove harfbuzz dependency from freetype

tgfrerer:
As @carthach pointed out in #3627, freetype might fail to work in some cases since an optional dependency to harfbuzz (a utility to parse extended OpenType features) has not been met. 

Since we haven't included & compiled harfbuzz in the past (and freetype for homebrew is fine without it) it seems a sensible choice to compile freetype without it.

If we wanted to include Harfbuzz, we'd need to add a dependency for harfbuzz 0.9.19 or newer: <http://sourceforge.net/projects/freetype/files/freetype2/2.5.3/> 

But this would create a catch22 while compiling, because harfbuzz, in turn, depends on Freetype.

We also have had a flag to exclude harfbuzz from android builds for a while. I wonder if we should exclude it from iOS builds, as well... 

ping @bakercp

### DIFF
--- a/scripts/apothecary/formulas/freetype.sh
+++ b/scripts/apothecary/formulas/freetype.sh
@@ -58,7 +58,7 @@ function build() {
 
 		local TOOLCHAIN=$XCODE_DEV_ROOT/Toolchains/XcodeDefault.xctoolchain 
 
-		./configure --prefix=$BUILD_TO_DIR --without-bzip2 --enable-static=yes --enable-shared=no \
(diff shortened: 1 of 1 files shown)


arturoc:
hey, yeah we've talked about including harbuzz but will leave it for next release so if you can exclude it from iOS too?

tgfrerer:
done =)

arturoc:
great!

bakercp:
Yep, harfbuzz is an important part of our ofFont future, but we're skipping it for now.

PR #3692, opened 2015-03-05
Fix missing return in ofSoundPlayer::loadSound, update examples

admsyn:
This PR:

- Returns a bool from inside loadSound
- Updates the sound player examples to use the non-deprecated "load" syntax


arturoc:
thanks!

PR #3693, opened 2015-03-06
bHide now means 'hidden', not 'shown'

camb416:
Small point I found confusing in the GUI examples. bHide meant "shown" or "not hidden". Still defaults on, but uses !bHide to draw.


ofZach:
this seems like a straightforward bug - thanks for the fix...

PR #3699, opened 2015-03-08
ofVecXf based ofNoise functions.

andreasmuller:


arturoc:
thanks!

libs/openFrameworks/math/ofMath.h
```diff
@@ -386,24 +386,42 @@ float ofNoise(float x);
...
 float ofNoise(float x, float y, float z);
 
+/// \brief Calculates a three dimensional Perlin noise value between 0.0...1.0.
+float ofNoise( ofVec3f p );
+
 /// \brief Calculates a four dimensional Perlin noise value between 0.0...1.0.
 float ofNoise(float x, float y, float z, float w);
 
+/// \brief Calculates a four dimensional Perlin noise value between 0.0...1.0.
+float ofNoise( ofVec2f p );
```
bakercp: Also this is incorrect.  Should be ofVec4f.

bakercp: Not exactly sure how it compiled.

arturoc: it compiles because it's just a declaration of  a function, you can declare something as many times as you want, but yeah the function for 4d is not accessible as it is right now

PR #3701, opened 2015-03-08
Const and argument corrections in ofMath.h/.cpp.

bakercp:
A few fixes for https://github.com/openframeworks/openFrameworks/pull/3699 related to const correctness + additional const corrections and a fix of an incorrect `ofNoise(const ofVec2f&)` signature.


arturoc:
thanks chris

bakercp:
No prob :)

PR #3702, opened 2015-03-09
Add mouseEntered/Exited events

mattfelsen:
Replaces https://github.com/openframeworks/openFrameworks/pull/3162

Hi all. This looked like a relatively easy PR targeted for 0.9 so I figured I'd give it a shot. Based on @bakercp's original PR (linked above) plus comments in the discussion, I did the following things:
- added mouseEntered/mouseExited to ofEvents & ofBaseApp
- register/unregister these events in ofMainLoop
- added the GLFW callback

The events don't receive any arguments – no mouse x/y since it's a technically window event, and no state (ie entered vs left) since each state is broken out to a different event. We could pass the mouse x/y for convenience though using the last known values.

I tested this on OS X and it seems fine, though admittedly only with a single window app. Based on looking through the code, I noticed a few things:
- ofCoreEvents::notifyMouseEvent() looks like it isn't called from anywhere. Is this a remnant from the pre-multiwindow days? If so, does the same apply to notifyKeyEvent (and possibly others) and should they be removed?
- The existing windowEntry code is still in place for GLUT. Should it be updated to the mouseEntered/mouseExited paradigm as well? If so we can probably get rid of the ofEntryEventArgs class as well since this is the only place it's used.
- Does this mean the project templates & examples need to get updated with the addition of mouseEntered/Exited?

### DIFF
--- a/libs/openFrameworks/events/ofEvents.h
+++ b/libs/openFrameworks/events/ofEvents.h
@@ -213,6 +213,8 @@ class ofCoreEvents {
 	ofEvent<ofMouseEventArgs> 	mousePressed;
 	ofEvent<ofMouseEventArgs> 	mouseReleased;
 	ofEvent<ofMouseEventArgs> 	mouseScrolled;
+	ofEvent<void>				mouseEntered;
+	ofEvent<void>				mouseExited;
 
 	ofEvent<ofTouchEventArgs>	touchDown;
 	ofEvent<ofTouchEventArgs>	touchUp;
@@ -253,6 +255,8 @@ class ofCoreEvents {
 	void notifyMouseDragged(int x, int y, int button);
 	void notifyMouseMoved(int x, int y);
 	void notifyMouseScrolled(float x, float y);
+	void notifyMouseEntered();
+	void notifyMouseExited();
 	void notifyMouseEvent(const ofMouseEventArgs & mouseEvent);
 
 	void notifyExit();
@@ -286,6 +290,8 @@ void ofRegisterMouseEvents(ListenerClass * listener, int prio=OF_EVENT_ORDER_AFT
 	ofAddListener(ofEvents().mousePressed,listener,&ListenerClass::mousePressed,prio);
 	ofAddListener(ofEvents().mouseReleased,listener,&ListenerClass::mouseReleased,prio);
 	ofAddListener(ofEvents().mouseScrolled,listener,&ListenerClass::mouseScrolled,prio);
+	ofAddListener(ofEvents().mouseEntered,listener,&ListenerClass::mouseEntered,prio);
+	ofAddListener(ofEvents().mouseExited,listener,&ListenerClass::mouseExited,prio);
 }
 
 template<class ListenerClass>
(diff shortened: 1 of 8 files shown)


libs/openFrameworks/events/ofEvents.h
```diff
@@ -213,6 +213,8 @@ class ofCoreEvents {
 	ofEvent<ofMouseEventArgs> 	mousePressed;
 	ofEvent<ofMouseEventArgs> 	mouseReleased;
 	ofEvent<ofMouseEventArgs> 	mouseScrolled;
+	ofEvent<void>				mouseEntered;
+	ofEvent<void>				mouseExited;
```
arturoc: isn't there any way to know where the mouse entered? when it exists we can for sure use the last position and when it enters we can probably use it too if there's a mouse position triggered before mouse entered?  it would be useful to report that too even if GLFW doesn't do it by default

mattfelsen: GLFW only tells you whether the mouse entered or left, but we can pass along ofEvents::currentMouseX/Y. I'll add it in my next pass at this PR

mattfelsen:
I milestoned this to 0.9.0 based on...
- this PR replaces an existing one milestoned for 0.9.0
- because it changes the ofRegisterMouseEvents API, it seems good to do that on a major release

mattfelsen:
Based on @arturoc's comments [here](https://github.com/openframeworks/openFrameworks/pull/3162#issuecomment-77811160) and [here](https://github.com/openframeworks/openFrameworks/pull/3702#discussion-diff-26018751) I have...
- added x/y args to the events
- converted GLUT to use the same new events

The GLUT code compiles but I'm not sure how to test it. Also, I still don't see anywhere that ofCoreEvents::notifyMouseEvent() is called, but I added code there as well just to cover all the bases. And finally, I also removed the now-unused windowEntry/ofEntryEventArgs code.

admsyn:
:+1: tested on OSX.

One thing I wasn't expecting is that the x/y args in the callback are from the frame *before* the mouse entry event (due presumably to the way the entry callback happens before the mouse motion callback). So if I come in fast from the left side of the window, I get an x of -24, for example. Might be totally fine, but it just wasn't what I was expecting off the bat :)

mattfelsen:
Yea, the callback only provides the state of whether the mouse is over the window or not – it doesn't actually provide the mouse x/y.

http://www.glfw.org/docs/latest/input.html#cursor_enter

For convenience, I added the last known x/y but as you mentioned, they're a frame behind. I see a couple options here...
- Keep this implementation but specify in the docs that the coords are behind
- Delay notification of this event until the next motion callback, and then pass an averaged x/y of the frames before & after entry. This seems hacky/weird.
- Remove x/y from the args and force the user to call ofGetMouseX/Y if they want it, which in theory should make it more obvious that you're getting the last known mouse coords

admsyn:
IMO keeping this implementation and specifying it in the docs is the way to go. Is the callback order consistent, though, or perhaps implementation-specific? If we document that it's "last frame" and then that's not true on e.g. linux then that's worse than no docs :)

Even better would be if callback order could be specified by OF to make this a non-issue, but a quick look around GLFW leaves me thinking that's not the case.

mattfelsen:
Good points, @admsyn. I'll do some tests to make sure the order doesn't vary across frames, and also want to make sure this works in multi window apps (in both the app-per-window and single-app-multiple-window models). Is there a way to test this on GLUT under OS X? I don't know much about the underlying windowing systems and how to force OF to use one or the other.

I'll submit updates to the CHANGELOG and docs in the next pass at this PR as well.

admsyn:
GLUT gives me the same behaviour as GLFW on OSX

admsyn:
If the implementation can stand to get to get a bit more complex, we could buffer the mouse entered/exited event as a bool, then trigger it during the next mouse motion callback. With a line/rect intersection between the previous and current mouse positions and the window that'd give the exact point where the event "really" occurred, regardless of how frequently the motion callback is triggered.

May not be worth cluttering up the implementation just for that, though.

mattfelsen:
Ah yes, the line/rect intersection would be better than just the midpoint. I'll leave it up to others to decide if it's worth it though :smile: 

For multiwindow, I'm getting the events on both windows if there is an app per window (from events/multiWindowExample). If there is a single app for both windows (from events/multiWindowOneAppExample), I'm only seeing events for the main window. However, this is the same behavior for the other events (mouseMoved, keyPressed, etc.) so I'm guessing this is the expected behavior.

PR #3710, opened 2015-03-12
Linux64 freeimage

jedahan:
So, here are three commits that help make `./compileOF.sh` work on my x86_64 Linux 3.19 system.

I am not sure if the first patch should just be a moving of the **FreeImage.h** file up a directory, or if what I did was correct.

I am not sure if we want to upgrade to 3.16, though it seems like not much is changed, just a few things added.

I am not sure if we want apothecary to handle FreeImage, but I would prefer that it did.

Thankfully, this pull request can start a discussion on if the direction I'm heading is right or wrong, as I may be adding linux and linux64 support to apothecary as needed to get things running peachy keen.


arturoc:
mmh, this is not very correct, we are not using the freeimage that is includes at all we instead use the one in the system since it's already installed in every linux install with a desktop. also installing a different version that the one in the system would be problematic for other libraries

apart from that what you are doing would use the freeimage file in the include directory but the rest of includes and the library from the system which again would be problematic.

i'm going to close this, can you open an issue specifying which error are you getting?

jedahan:
FreeImage is not packaged on my distro, even though I have a desktop. That was the main reason for this pull request.

What is the downside to having apothecary handling FreeImage, thereby linking against a known version on any distribution?

I edited the Makefile to link against the apothecary built FreeImage in bba95d651c3caaae599f81e4ef8e15e6c8e16713 . What if that Makefile linked against system FreeImage and only falls back to apothecary FreeImage if that fails?

arturoc:
what distro are you using? as far as i know free type is the only oss library too read TTF so its really strange that any distro won't have it

jedahan:
FreeType is installed and happy on my system, it's FreeImage that is has no package for [Exherbo Linux](http://exherbo.org). This kinda stuff comes up very infrequently as the assumption is that everyone is on ubuntu.

I [submitted a FreeImage package](https://galileo.mailstation.de/gerrit/766) for exherbo, which should make things work as expected.

I was just wondering what the goals of apothecary are at a high level - to only build things that are not incuded in debian/ubuntu? To be able to build any library dependency of oF no matter what system they are on? To be able to know we have a fixed version of oF library dependencies across platforms? 

I don't mean to distract from good work going on, so if thinking about 'do we bundle dependencies or not' is draining too much time from more important work, feel free to say so and I'll either maintain my own fork or run stuff in a vm if needed.

arturoc:
oh, sorry i always get confused with freeimage and freetype,  it was super weird that freetype wasnt' already there :).

we support debian, ubuntu, archlinux, and fedora which should also give support for any derivative of those, which i'd say covers 99% of existing desktop distributions or at least users of OF on linux. 

in any case every distribution has some kind of packaging system so replicating it in apothecary seems kind of pointless for linux: in all the distros we officially support, there's only 2 libraries that aren't already there poco and tess2. we haven't included any package yet for linux apothecary and i don't think will do. i'd prefer to submit official packages or create a ppa or the equivalent for other distros. 

also if you want to contribute a install_dependencies script for exherbo that'd be great too

arturoc:
btw, the main problem with maintaining our own versions is that for some libraries like gstreamer it would be a lot of space and would for sure conflict with the libraries in the system. also gstreamer for example or cairo depend on freetype and freeimage so if we tried to leave gstreamer and cairo out of apothecary but have apothecary formulas for freetype and freeimage it would be super problematic since programs will try to link against 2 different versions of each library.

in general in linux it's not a good idea to pack libraries and it's better to rely on the systems package manager. at some point we would even like to have openframeworks native packages for the most used distros.

jedahan:
Thanks for taking the time for explaining the ideal future, i'll work towards that vision of packaging oF natively, and if I come up with a good install_dependencies script, share that as well.

Issue #3694, opened 2015-03-06
ofXml::setValue as int


camb416:
A tiny override to accept an int when using setValue...

https://github.com/openframeworks/openFrameworks/compare/master...camb416:feature-xmlsetint

bilderbuchi:
ping @joshuajnoble

joshuajnoble:
Looks fine to me

bilderbuchi:
reopened by #3800

kylemcdonald:
@camb416 could you submit this again but with a more general solution, like:

```c++
template <class T>
bool setValue(const string& path, const T& value) {
    return setValue(path, ofToString(value));
}
```

i think this is what @arturoc is looking for.

sorry for the delay in merging your work! thanks for the contribution :)

PR #3711, opened 2015-03-13
ofXml::setValue as int

camb416:
Adding support for ofxml::setValue as int.
feature #3694


bilderbuchi:
already preliminarly reviewed by @joshuajnoble

arturoc:
kyle this was wrong, we are using getTypeValue and setValue<Type>(...) to avoid anbiguities. even the string and int types can be ambiguous.

kylemcdonald:
thanks, i didn't look too closely i just noticed that it was a small change done correctly, had been checked by a few people, and left open for months without any negative comments. looking at the ofXml API i see what you're talking about.

PR #3712, opened 2015-03-14
squash non-unicode ✗

jedahan:
how did that x sneak in?!


admsyn:
Going to assume this doesn't require thorough core review :)

Issue #3301, opened 2014-10-12
Review current examples for deprecated API usage / older non-current OF code usage. 


ofTheo:
ie: 
- use of char * arrays for filenames. 
- sprintfs for joining strings. 
- printf instead of ofLog() 
- legacy / deprecated OF code. 

A suggestion was made to also nominate examples that might not be useful in their current state to be flagged for removal. 
Feel free to suggest them in the comments. 

I also want to use this review to check for regressions / bugs in the 0.9.0 release - so this issue will get closed close to the RC release phase.

laserpilot:
current locations of files with sprintf or printf:

Sprintf:
a23-0-5-175:examples laser$ grep -lr "sprintf" *
android/androidSoundPlayerExample/src/ofApp.cpp
events/eventsExample/src/ofApp.cpp
graphics/blendingExample/src/ofApp.cpp
graphics/imageSaverExample/src/ofApp.cpp
ios/audioOutputExample/src/ofApp.mm
ios/eventsExample/src/ofApp.mm
ios/opencvExample/src/ofApp.mm
ios/soundPlayerExample/src/ofApp.mm

Printf
a23-0-5-175:examples laser$ grep -lr "printf" *
3d/cameraParentingExample/src/ofApp.cpp
android/androidSoundPlayerExample/src/ofApp.cpp
events/eventsExample/src/ofApp.cpp
gl/geometryShaderExample/src/ofApp.cpp
gl/glInfoExample/src/glewOutput.cpp
gl/glInfoExample/src/ofApp.cpp
graphics/blendingExample/src/ofApp.cpp
graphics/imageSaverExample/src/ofApp.cpp
graphics/imageSequenceExample/src/ofApp.cpp
ios/audioOutputExample/src/ofApp.mm
ios/eventsExample/src/ofApp.mm
ios/iPhoneGuiExample/src/gui/MyGuiView.mm
ios/MapKitExample/src/ofApp.mm
ios/opencvExample/src/ofApp.mm
ios/soundPlayerExample/src/ofApp.mm

arturoc:
closed by #3713

PR #3713, opened 2015-03-15
Review of current examples for deprecated API usage

julianadenauer:
replaced sprintf, printf
and for the ios examples replaced the deprecated loadImage, loadFont and loadSound with load
issue #3301


kylemcdonald:
wow, just seeing this now -- excellent work @julianadenauer

PR #3716, opened 2015-03-16
Remove outdated PocoODBC lib & references in CoreOF.xcconfig

mattfelsen:
Apothecary doesn't build this part of POCO, but it was still being linked against
in Xcode. This commit removes the outdated i386 lib (since no FAT replacement will
be built) and removes it from OF_CORE_LIBS in CoreOF.xcconfig


mattfelsen:
I should add that I couldn't find any references to ODBC in the core, or in any addon names on ofxaddons.com, so it seems safe to remove.

tgfrerer:
+1

danoli3:
Thanks @mattfelsen. 

This was fixed though with this commit: https://github.com/openframeworks/openFrameworks/commit/6eccef2891bb459144262c5b61903fdb5503b041
With the creation of the new libc++ 32 fat libs for OSX.

PR #3723, opened 2015-03-20
ofNode parenting issue, setParent maintaining global transform had the same...

### LINKED ISSUE #3722: ofNode setParent maintaining global position doesn't work as expected
Both :
```
ofNode::setParent( p, true );
ofNode::setParent( p, false );
```
do the same. When `bMaintainGlobalTransform == true` we should compute the future transform the child will have and then set it. Right [now](https://github.com/openframeworks/openFrameworks/blob/31d24eac22f817534c2a9c17d6fe3b6d01957a70/libs/openFrameworks/3d/ofNode.cpp#L20) the transform is reevaluated from the new parent, hence modifying the global transform. 

Here's what I think it should be:

```
    ofMatrix4x4 postParentGlobalTransform = parent.getGlobalTransformMatrix().getInverse() * getGlobalTransformMatrix();
    this->parent = &parent;
    setTransformMatrix(postParentGlobalTransform);
} else {
    this->parent = &parent;
}
```

[PR](https://github.com/openframeworks/openFrameworks/pull/3723) sent!

chuckleplant:
... effect as not maintaining it.

Fixes #3722

### DIFF
--- a/libs/openFrameworks/3d/ofNode.cpp
+++ b/libs/openFrameworks/3d/ofNode.cpp
@@ -15,9 +15,9 @@ ofNode::ofNode()
 //----------------------------------------
 void ofNode::setParent(ofNode& parent, bool bMaintainGlobalTransform) {
     if(bMaintainGlobalTransform) {
-        ofMatrix4x4 globalTransform(getGlobalTransformMatrix());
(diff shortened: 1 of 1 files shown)


chuckleplant:
Matrix multiplication is swapped! :sweat_smile:

PR #3730, opened 2015-03-23
Fix scroll up issue on ofxIntSlider

### LINKED ISSUE #3728: Scroll up on ofxIntSlider doesn't work (ofxGui)
When scrolling up on a `ofxIntSlider` the value doesn't increase whereas it decreases when scrolling down. However it works well with an `ofxFloatSlider`.

I think the problem comes from the method `bool ofxSlider<Type>::mouseScrolled(ofMouseEventArgs & args)` (line 106) in `ofxSlider.cpp`:

	template<typename Type>
	bool ofxSlider<Type>::mouseScrolled(ofMouseEventArgs & args){
		if(mouseInside){
			if(args.y>0 || args.y<0){
				double range = value.getMax() - value.getMin();
				range /= b.width*4;
				Type newValue = value + ofMap(args.y,-1,1,-range, range);
				newValue = ofClamp(newValue,value.getMin(),value.getMax());
				value = newValue;
			}
			return true;
		}else{
			return false;
		}
	}

If `range < b.width*4` then range will be `< 1` after dividing  and then `ofMap(args.y,-1,1,-range, range)` will be `< 1` too. So if `Type` is an `int` it will be truncated.

elaye:
Fixes #3728

### DIFF
--- a/addons/ofxGui/src/ofxSlider.cpp
+++ b/addons/ofxGui/src/ofxSlider.cpp
@@ -118,6 +118,23 @@ bool ofxSlider<Type>::mouseScrolled(ofMouseEventArgs & args){
 	}
 }
 
+template<>
+bool ofxSlider<int>::mouseScrolled(ofMouseEventArgs & args){
+	if(mouseInside){
+		if(args.y>0 || args.y<0){
+			double range = value.getMax() - value.getMin();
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!!

PR #3734, opened 2015-03-26
fix ofShader::load() return value

tgfrerer:
Currently, `ofShader::load()` loads, compiles, and links GLSL shader source.

`ofShader::load()` returns a bool, which is set solely based on the current shader object's internal `bLoaded` variable, which does not say anything about whether the shader has been loaded, compiled and linked successfully.

It will also attempt to compile and link shader stages, even if an earlier shader stage compile or load failed.

---

This PR suggests the following changes:

* If an error is encountered in any of the stages of ofShader::load (`load`, `compile`, and `link`), return `false` early at the point of error, print an error messge, and do not continue processing this shader. This makes it easier to debug shader errors, since subsequent errors only fill up the log, and are most likely just results of the first error reported.

* ofShader::linkProgram only returns true if *both* load and `linkStatus`  were true, that is, when the shader was successfully linked.

With this, the return value of `ofShader::load()` is a reliable indicator of whether the shader was loaded, compiled, and linked properly.

### DIFF
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -132,13 +132,23 @@ bool ofShader::load(string shaderName) {
 
 //--------------------------------------------------------------
 bool ofShader::load(string vertName, string fragName, string geomName) {
-	if(vertName.empty() == false) setupShaderFromFile(GL_VERTEX_SHADER, vertName);
-	if(fragName.empty() == false) setupShaderFromFile(GL_FRAGMENT_SHADER, fragName);
+	if(vertName.empty() == false) {
+		bool success = setupShaderFromFile(GL_VERTEX_SHADER, vertName);
+		if (success == false) return false;
+	}
+	if(fragName.empty() == false) {
+		bool success = setupShaderFromFile(GL_FRAGMENT_SHADER, fragName);
+		if (success == false) return false;
+	}
 #ifndef TARGET_OPENGLES
-	if(geomName.empty() == false) setupShaderFromFile(GL_GEOMETRY_SHADER_EXT, geomName);
+	if(geomName.empty() == false) {
+		bool success = setupShaderFromFile(GL_GEOMETRY_SHADER_EXT, geomName);
+		if (success == false) return false;
+	}
 #endif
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/gl/ofShader.cpp
```diff
@@ -132,13 +132,23 @@ bool ofShader::load(string shaderName) {
 
 //--------------------------------------------------------------
 bool ofShader::load(string vertName, string fragName, string geomName) {
-	if(vertName.empty() == false) setupShaderFromFile(GL_VERTEX_SHADER, vertName);
-	if(fragName.empty() == false) setupShaderFromFile(GL_FRAGMENT_SHADER, fragName);
+	if(vertName.empty() == false) {
```
arturoc: won't this be better like?

```cpp
if(!vertName.empty() && !setupShaderFromFile(GL_VERTEX_SHADER, vertName){
    return false;
}
```

so many comparisons with false look confusing but otherwise seems good

libs/openFrameworks/gl/ofShader.cpp
```diff
@@ -447,13 +458,16 @@ bool ofShader::linkProgram() {
 
 		glLinkProgram(program);
 
-		checkProgramLinkStatus(program);
+		linkStatus = checkProgramLinkStatus(program);
```
arturoc: this can break old code since there's checks somewhere else for bLoaded, for example, line 526:

```
void ofShader::setUniformTexture(const string & name, int textureTarget, GLint textureID, int textureLocation) const{
	if(bLoaded) {
```

and you can do:
```
shader.load(...);
...
if(shader.isLoaded()){
....
}
```

which will be broken. we could keep bLoaded with the same meaning it has now (everything worked) and add individual flags like bShadersCompiled, bLinked, bBinded that you can query to check if every step individually was correct

langolierz:
any news on this pr ? I would [also like it](https://forum.openframeworks.cc/t/identifying-when-ofshader-hasnt-linked/30626) if isLoaded was a reliable indicator of whether the shader was loaded. anything I can do / to help ? @tgfrerer @arturoc

Issue #3729, opened 2015-03-23
AVFoundationPlayer: Problems if not run in the main thread


urshofer:
I just encountered huge memory leaks and problems if AVFoundationPlayer is not run in the main thread. To avoid this I've added an autorelease pool to `init()`in ofAVFoundationVideoPlayer.m:

    if (![NSThread isMainThread]) {
      pool = [[NSAutoreleasePool alloc] init];
    }

and in `dealloc()`:

    if ([NSThread isMainThread]) {
      [super dealloc];
    }
    else {
      [pool drain];
    }

As far as i can see the memory usage is now quite steady. I'm really not into AVFoundation but autorelease w/o pool is only flushed in the main thread, isn't it?
Added a gist here: https://gist.github.com/urshofer/a8e36889dcf55c558bbc

Best,
Urs

pizthewiz:
Only the main `NSThread

pizthewiz:
The thread creator is responsible for putting a pool in place on secondary threads. Generally, the pool should be hoisted into the thread execution wrapping all operations, not intermixed in a particular object use:

```Objective-C
- (void)viewDidLoad {
    self.thread = [[NSSthread alloc] initWithTarget:self selector:@selector(runner:) object:nil];
    [self.thread start];
}

- (void)runner:(id)object {
    @autoreleasepool {
        // do stuff here 👍
     }
}
```

pizthewiz:
It should be noted though that not all Cocoa/Cocoa Touch classes support execution on secondary threads, some require active runloops and what not.

urshofer:
Yes - execution on secondary threads is only meant for background tasks. Not only cocoa stuff probably does not run, also openGL textures.

i-n-g-o:
does this PR https://github.com/openframeworks/openFrameworks/pull/3736 solve this issue? (did not come around to test it myself...)

urshofer:
closed issue since it is resolved with PR https://github.com/openframeworks/openFrameworks/pull/3736

Issue #3715, opened 2015-03-16
AVFoundationPlayer: Repetitive calls to movie.load() and movie.close() is leaking memory in OSX 10.9


urshofer:
I'm encountering a strange memory effect when calling `load()`or `loadAsync()` and `close()` repeatedly:

On my MBP with 10.8.5 an application with the commands mentioned above runs w/o leaking memory for days. On a Mac Mini with 10.9.5 the application is leaking memory. It does not depend with which SDK (i've tested 10.9 and 10.10) the Application has been compiled. According to Xcode Instruments, the leak happens in MediaToolbox and AudioToolbox is constantly allocating more and more memory.

The problem is probably apple related, but I actually don't know really. I'm just posting this because it could be crucial for long running video installations.

![bildschirmfoto 2015-03-16 um 10 11 20](https://cloud.githubusercontent.com/assets/116976/6663567/34175f94-cbc8-11e4-993c-5767d0a45c2d.png)

urshofer:
Update: Just updated the system to 10.10 - looks like the leaking is gone. Probably 10.9 includes a buggy AVFoundation implementation?

![bildschirmfoto 2015-03-16 um 14 26 35](https://cloud.githubusercontent.com/assets/116976/6666898/b0b9bbc6-cbe8-11e4-9a19-f144329c4198.png)

bilderbuchi:
ping @openframeworks/macos

admsyn:
Yeah that's a reasonable conclusion, going to close this.

bakercp:
So, is there any suggested way around this issue?  We've been having loads of trouble with the master branch AVFoundation player when running our app [https://github.com/kibio/kibio] (compiled on 10.10) on 10.8 / 10.9 @brannondorsey ...

admsyn:
This issue specifically or loads of trouble with the AVFoundation player in general? Judging from the output in Instruments, the leaks are just 4KB every reload which doesn't seem like it'd actually cause issues in practice.

Is there an issue on your repo with more info about the trouble you're running into? Couldn't find one with a quick scan.

urshofer:
The leaks were not big, but after a night the memory consumption raised from 40mb at start to 520mb. I'm working with network triggered fragments of really short videos, from 1 up to 5 seconds. So there are plenty of load and unload calls. Not an ideal situation for a permanent installation.

To sum up the troubles I've encountered so far: 
[...]

Issue #3704, opened 2015-03-10
Make loadAsync() in AVFoundationPlayer run without close() before.


urshofer:
First: The loadAsync function is really a great improvement for background loading of clips. 

But repeated loading of movies without a movie.close() before does not work. As a workaround, you can of course always call .close() on the video player object before .loadAsync() again. But close() is quite slow, on my MBP about 50ms.

Like this, the loadAsync() functions looses a lot of it's intended functionality.

julapy:
ok thats interesting.
so does calling `close()` usually take that long regardless if you have loaded the video sync or async?
or does it take longer to close when loading async only?

any thoughts on how this could be improved?

also pinging @HalfdanJ who added the async loading code.

urshofer:
I'll provide some tests tomorrow. I'm quite new to the AVFoundation Framework, coming more from the gStreamer-Side so please excuse my not so accurate descriptions...

Back in the days when the AVFoundationPlayer was still an addon, it was possible to call load without close before. Although load was slower than loadAsync now, the whole process was a little bit faster without the necessity of close(). By then, I did some tests with removing the init function from the load function if the player was already loaded, which was the fastest solution but crashed after a while.


Am 11.03.2015 um 03:08 schrieb lukasz karluk <notifications@github.com>:

urshofer:
okay, i did some tests. the function which is responsible for loading is here:


    void ofxGaplessVideoPlayer::_appendMovie(string _name)
    {
        int actionTimeout = ofGetElapsedTimeMillis();
        video.close();
        actionTimeout = ofGetElapsedTimeMillis() - actionTimeout;

        int loadTime = ofGetElapsedTimeMillis();
        video.loadAsync(_name);
        loadTime = ofGetElapsedTimeMillis() - loadTime;

        ofLogError() << "Loading: " << loadTime << "ms Closing: " << actionTimeout << "ms";
    }

Which results in this output:

__2015-03-12 14:16:40.969 AC PlayerDebug[26841:890f] video loaded at 640 x 358
[ error ] Loading: 1ms Closing: 124ms__

Closing is between 50 and 130ms.

Tests proved that changing to __.load()__ instead of __.loadAsync()__ has no influence on the closing time.

arturoc:
[...]

Issue #2530, opened 2013-09-02
ofVideoPlayer makes the app not responding after a while if (especially large) video file are reloaded.


l-s:
app stalls/blocks and osx multicolor wheel starts rotating after a while if movie files are (re)loaded. 
The problem happens with frequent reloading (every 2 or 3 update calls or after played 2 or 3 frames checking getCurrentFrame) and sooner with large video file (2 GB).  This happens to me after few seconds or several minutes depending on video size and reloading frequency. Note that this makes the app blocking faster, but i experienced this problem even with loading a clip on a movie loop video with a sensor input (i.e. randomly).
A small video file (50 MB) blocks the app if loaded after few update (reloadUpdate=2 in the example below), just after the previous one ( if myMovie.isLoaded(){load another one} ) or after the movie is loaded and played a bit (a couple of frames, reloadAfterFrame=2)
A big video file loading blocks the app sooner even with not so fast reloading frequency (i.e. reloadUpdate=20 or reloadAfterFrame=20).
Also when the app freezes the memory allocated (in the activity monitor report) doubles (it's not releasing the previous movie? anyway even if i close the movie before loading a new one things don't change).
I start experiencing this problem with version 0072, now i'm using v008, v0071 works fine to me. 
osx 10.8.1 on mac mini 2,3 Ghz Core i5, solid state hd.
Quiktime Photojpg codec.

``` c++
void testApp::setup(){
    ofSetVerticalSync(true);
    currMovie=0;
    myMovie.loadMovie(ofToString(currMovie)+".mov");
    myMovie.play();
    changeMovie=0;

    reloadUpdate=20;// =2 this blocks sooner the app
    reloadAfterFrame=20;// =2 this blocks sooner the app
}

//--------------------------------------------------------------
void testApp::update(){
    changeMovie++;
    if (changeMovie>reloadUpdate)
   // if (myMovie.isLoaded() && myMovie.getCurrentFrame()> reloadAfterFrame) 
    // checked also if movie was loaded and played some frames, same result
    {
        cout <<"reload "<<ofToString(currMovie)<<".mov"<<endl;
        changeMovie=0;
        currMovie++;
        currMovie=currMovie%3;
        myMovie.loadMovie(ofToString(currMovie)+".mov");
        myMovie.play();
    }
    myMovie.update();

}
```

i-n-g-o:
might be fixed by this PR: https://github.com/openframeworks/openFrameworks/pull/3736
could someone test?

[...]

PR #3736, opened 2015-03-26
oFAVFoundationPlayer gc and memory fixes

i-n-g-o:
adding garbage collection for AVFoundationVideoPlayer disposal making movie loading faster.
also fixes some memory leaks reported here:
https://github.com/openframeworks/openFrameworks/issues/3704
https://github.com/openframeworks/openFrameworks/issues/3729
https://github.com/openframeworks/openFrameworks/issues/3715

might fix this as well: https://github.com/openframeworks/openFrameworks/issues/2530 (did not test)

also includes these pullrequests: 
https://github.com/openframeworks/openFrameworks/pull/3672: using a cleanup+release method for internal cleanup. refactored `finalize` to `cleanupAndAutorelease`
https://github.com/openframeworks/openFrameworks/pull/3670: making texture objects private

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -115,14 +115,15 @@ class ofAVFoundationGC : public ofThread {
 	
 public:
 	
-	static ofAVFoundationGC* instance();
+	~ofAVFoundationGC();
+	
+	static ofAVFoundationGC& instance();
 	void addToGarbageQueue(ofAVFoundationVideoPlayer*);
 	
 	
 private:
 	
-	ofAVFoundationGC(){}; //use instance()!
-	static ofAVFoundationGC*	singleton;
+	ofAVFoundationGC(); //use instance()!
 	
 	vector<ofAVFoundationVideoPlayer*> videosPendingDeletion;
 	
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -85,12 +85,12 @@ public:
     
 protected:
 	
-	bool loadPlayer(string name, bool bAsync);
+    bool loadPlayer(string name, bool bAsync);
 
 #ifdef __OBJC__
-	ofAVFoundationVideoPlayer * videoPlayer;
+    ofAVFoundationVideoPlayer * videoPlayer;
 #else
-	void * videoPlayer;
+    void * videoPlayer;
 #endif
     
     bool bFrameNew;
@@ -100,8 +100,8 @@ protected:
     bool bTextureCacheSupported;
 	
     ofPixels pixels;
-	ofPixelFormat pixelFormat;
-	ofTexture videoTexture;
+    ofPixelFormat pixelFormat;
+    ofTexture videoTexture;
 };
 
 
@@ -115,19 +115,19 @@ class ofAVFoundationGC : public ofThread {
 	
 public:
 	
-	~ofAVFoundationGC();
-	
-	static ofAVFoundationGC& instance();
-	void addToGarbageQueue(ofAVFoundationVideoPlayer*);
+    ~ofAVFoundationGC();
+
+    static ofAVFoundationGC& instance();
+    void addToGarbageQueue(ofAVFoundationVideoPlayer*);
 	
 	
 private:
 	
-	ofAVFoundationGC(); //use instance()!
-	
-	vector<ofAVFoundationVideoPlayer*> videosPendingDeletion;
-	dispatch_semaphore_t sema;
-	void threadedFunction();
+    ofAVFoundationGC(); //use instance()!
+
+    vector<ofAVFoundationVideoPlayer*> videosPendingDeletion;
+    dispatch_semaphore_t sema;
+    void threadedFunction();
 	
 };
 #endif
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -9,6 +9,7 @@
 #include "ofBaseTypes.h"
 #include "ofPixels.h"
 #include "ofTexture.h"
+#include "ofThread.h"
 
 #ifdef __OBJC__
 #import "ofAVFoundationVideoPlayer.h"
@@ -103,3 +104,29 @@ protected:
 	ofTexture videoTexture;
 };
 
+
+
+//--------------------------------------------------------------
+// garbage collector for videoplayers
+// destroying an AVAssetReader takes a while,
+// so do it on a thread
+#ifdef __OBJC__
+class ofAVFoundationGC : public ofThread {
+	
+public:
+	
+	static ofAVFoundationGC* instance();
+	void addToGarbageQueue(ofAVFoundationVideoPlayer*);
+	
+	
+private:
+	
+	ofAVFoundationGC(){}; //use instance()!
+	static ofAVFoundationGC*	singleton;
+	
+	vector<ofAVFoundationVideoPlayer*> videosPendingDeletion;
+	
+	void threadedFunction();
+	
+};
+#endif
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -47,6 +47,7 @@ public:
     ofPixels & getPixels();
     ofTexture * getTexturePtr();
     void initTextureCache();
+    void killTexture();
     void killTextureCache();
 	
     float getWidth() const;
@@ -90,6 +91,7 @@ public:
 protected:
 	
     bool loadPlayer(string name, bool bAsync);
+	void disposePlayer();
 
 #ifdef __OBJC__
     ofAVFoundationVideoPlayer * videoPlayer;
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -118,30 +118,3 @@ protected:
 #endif
 };
 
-
-
-//--------------------------------------------------------------
-// garbage collector for videoplayers
-// destroying an AVAssetReader takes a while,
-// so do it on a thread
-#ifdef __OBJC__
-class ofAVFoundationGC : public ofThread {
-	
-public:
-	
-    ~ofAVFoundationGC();
-
-    static ofAVFoundationGC& instance();
-    void addToGarbageQueue(ofAVFoundationVideoPlayer*);
-	
-	
-private:
-	
-    ofAVFoundationGC(); //use instance()!
-
-    vector<ofAVFoundationVideoPlayer*> videosPendingDeletion;
-    dispatch_semaphore_t sema;
-    void threadedFunction();
-	
-};
-#endif
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -126,7 +126,7 @@ private:
 	ofAVFoundationGC(); //use instance()!
 	
 	vector<ofAVFoundationVideoPlayer*> videosPendingDeletion;
-	
+	dispatch_semaphore_t sema;
 	void threadedFunction();
 	
 };
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -44,12 +44,8 @@ bool ofAVFoundationPlayer::load(string name) {
 //--------------------------------------------------------------
 bool ofAVFoundationPlayer::loadPlayer(string name, bool bAsync) {
 	
-    if(videoPlayer != NULL) {
-        // throw this player away, add it to GC
-        ofAVFoundationGC::instance().addToGarbageQueue(videoPlayer);
-        videoPlayer = NULL;
-    }
-
+	// dispose videoplayer, texture and texture-caches
+	close();
 
     // create a new player
     videoPlayer = [[ofAVFoundationVideoPlayer alloc] init];
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -56,7 +56,7 @@ bool ofAVFoundationPlayer::loadPlayer(string name, bool bAsync) {
 	
     if(videoPlayer != NULL) {
 		// throw this player away, add it to GC
-		ofAVFoundationGC::instance()->addToGarbageQueue(videoPlayer);
+		ofAVFoundationGC::instance().addToGarbageQueue(videoPlayer);
 		videoPlayer = NULL;
 	}
 	
@@ -139,7 +139,8 @@ void ofAVFoundationPlayer::close() {
         videoTexture.clear();
 
 		// dispose videoplayer
-		ofAVFoundationGC::instance()->addToGarbageQueue(videoPlayer);
+		ofAVFoundationGC& inst = ofAVFoundationGC::instance();
+		inst.addToGarbageQueue(videoPlayer);
 
 		videoPlayer = NULL;
 		
@@ -747,18 +748,10 @@ ofTexture * ofAVFoundationPlayer::getTexture() {
 
 //--------------------------------------------------------------
 //--------------------------------------------------------------
-// GC singleton
-ofAVFoundationGC* ofAVFoundationGC::singleton = NULL;
-
-ofAVFoundationGC* ofAVFoundationGC::instance(){
-	
-	if (!singleton){   // Only allow one instance of class to be generated.
-		singleton = new ofAVFoundationGC();
-		singleton->startThread();
-	}
+ofAVFoundationGC& ofAVFoundationGC::instance(){
+	static ofAVFoundationGC singleton;
 	return singleton;
 }
-
 
 void ofAVFoundationGC::addToGarbageQueue(ofAVFoundationVideoPlayer * p){
 	lock();
@@ -767,6 +760,15 @@ void ofAVFoundationGC::addToGarbageQueue(ofAVFoundationVideoPlayer * p){
 	unlock();
 }
 
+ofAVFoundationGC::~ofAVFoundationGC(){
+	// end thread and clean up
+	waitForThread(true);
+}
+
+// private constructor
+ofAVFoundationGC::ofAVFoundationGC(){
+	startThread();
+}
 
 void ofAVFoundationGC::threadedFunction(){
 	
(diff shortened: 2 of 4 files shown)


admsyn:
Looks sensible, gave it a test and it seems to perform well. It's tricky to reason about the implications of how it's handling memory without deeper knowledge of the AVFoundation player, though. What would be good to test?

Pinging @julapy / @openframeworks/video?

mattfelsen:
Hi all. I've been tracking the discussions in the mentioned issues and am super happy to see the work that's gone into the AVF player recently! I'm working on a project where I'm unloading some combination of multiple 1080, 4K, and 6K videos at the same time, so I'll try to do some additional testing and report back.

urshofer:
I'll test it also extensively later this day and report back. Thanks, @i-n-g-o !

urshofer:
With the recent fixes my test application crashes:

bildschirmfoto 2015-03-27 um 09 56 04

This was not the case with this version: https://github.com/i-n-g-o/openFrameworks/commit/4473858a088881b8cf0025d7b8f32d5ca393def2

i-n-g-o:
found a comment i wanted to re-enable before pushing, but forgot. sry.
can you check if it works again?

urshofer:
still crashing. slightly different debug output, but quite the same position. 

bildschirmfoto 2015-03-27 um 12 12 18

about the memory leak: In my case yesterday's fork is absolutely not leaking and now running constantly since 20 hours.

i-n-g-o:
can you show me how you start the player/load a file?
or share a test-application?

urshofer:
hm

it's not that easy - I'm using the player in a network triggered playback system called www.automatic-cinema.com, mainly in a plugin called [ofxGaplessVideo](https://github.com/urshofer/ofxGaplessVideo). This plugin is adding load and trigger commands on a stack and triggers two players alternately.

I admit that my system is treating the playback system very harsh. It switches states often and is loading in very short intervals. Due to the network nature, it can also be that load calls can have some irregularities.

I think right now the player crashes after two consecutive `async()` loads before the first load has completed:

    video.loadAsync(_name);
    video.loadAsync(_name);

These two calls clearly don't make sense, but I think it can happen quite often in oF applications by mixing the sync'ed nature of the update()-draw() cycle of a oF app and the new async'd loading functions.

i-n-g-o:
hm. cannot reproduce that behaviour here, but got a different one by calling loadAsync 4-times in a row.

do you `video.loadAsync(_name);` on the main-thread?

urshofer:
> hm. cannot reproduce that behaviour here, but got a different one by calling loadAsync 4-times in a row.
> 
probably it's a timing issue, depending on the loaded clips? In my case, it's also not necessarily after two calls
but twice the two calls, which makes also four in the end.

> do you video.loadAsync(_name); on the main-thread?
> 

Yes - the video player is running on the main thread, only the command queue, which has actually nothing to do with the player, is filled from another thread.

Do you encounter memory leaks with the older `_ofAVFoundationPlayer_add_GC_for_videoplayer` fork? I don't, and it is running very stable...

i-n-g-o:
very strange this.
@urshofer: since i cannot reproduce the crashes, would you mind trying out some things?
(we also could do that in a one-to-one session on IRC...)

no, i don't experience memoryleaks with `_ofAVFoundationPlayer_add_GC_for_videoplayer` per se. this PR would also deal with two other PRs.
moving `_videoTextureRef` and `_videoTextureCache` from global to private seems quite critical to me and should not make problems.
releasing videoplayer with `cleanupAndAutorelease` would solve a potential leak, which might also be addressed in a different way.

urshofer:
I need to run but I will do some more tests. It's also hard for me to reproduce the bug. It a nasty runtime thing and the crash can happen just anytime, after 10 minutes or (as in my last try) after two hours.

It happens when calling `update()` in some certain moments and it is definitely a threading problem.

Would it be an option just to merge the branches just without the `cleanupAndAutorelease` part? Just guessing, but probably the latter part is responsible for the crashes?




Am 27.03.2015 um 15:14 schrieb Ingo Randolf <notifications@github.com>:

i-n-g-o:
the `cleanupAndAutorelease` part does not affect what happens in `update`. as long as open() or close() and update() are called from the same thread there is no way cleanup() would mess up update().
but who knows.

more tests would be good. also to verify that `_ofAVFoundationPlayer_add_GC_for_videoplayer` (maybe with local TextureRef-objects) is really stable.

a threading issue definitely. maybe the problem comes from somewhere else...?

urshofer:
I'm also sure that it has nothing to do with `update()` itself. I forked your code and created a version as you proposed for another test run: `_ofAVFoundationPlayer_add_GC_for_videoplayer` with local Texture-Refs:

https://github.com/urshofer/openFrameworks/tree/_uh_ofAVFoundationPlayer_threadfix

Of course, problems could come from elsewhere in my application. But it's a fact that it runs stable with other video frameworks like GStreamer on both Linux and OSX (I used it as a fallback solution on OSX if stability remains an issue). Would be nice to have only one framework on all platform ;)

i-n-g-o:
cool. can you already say how this version performs?

i was not accusing your application. what i did mean was, that the problem also could come from somewhere else from within the player maybe caused by the changes.

i'm very curious what the cause is. if it is really the cleanupAndAutorelease function, then we need to get rid of it and address the problem it does in some other way.

i-n-g-o:
@urshofer: do you still have this crashlog around you attached lately? (the second one) just figured it was thread 2 which crashed... its not visible in the screenshot.
would like to have a look if possible.

urshofer:
> cool. can you already say how this version performs?
> 
hi ingo. the version with private textures and gc just crashed. best experiences so far with the gc only version.
> i was not accusing your application. what i did mean was, that the problem also could come from somewhere else from within the player maybe caused by the changes.
> 
> 

no problem, I didn't feel accused :) I just know that I make many mistakes.

> i'm very curious what the cause is. if it is really the cleanupAndAutorelease function, then we need to get rid of it and address the problem it does in some other way.
> 
> 

I'll give some feedback tomorrow...

urshofer:
my bad I was screenshotting the wrong section. will provide better info as soon as I encounter new crashes…

best,
Urs

Am 28.03.2015 um 18:55 schrieb Ingo Randolf <notifications@github.com>:

urshofer:
ok - here a few updates:

I got three version of my application currently running. Whereas the version with the GC only and the one with GC and local textures are running stable for 12 hours now, the one with cleanupAndAutorelease is definitely crashing more often:

    Thread 14 Crashed:: Dispatch queue: com.apple.avfoundation.avasset.completionsQueue
    0   libobjc.A.dylib               	0x9634d0b4 objc_msgSend + 36
    1   com.apple.MediaToolbox        	0x97a2f0d5 0x977ff000 + 2293973
    2   com.apple.MediaToolbox        	0x97a32bad 0x977ff000 + 2309037
    3   com.apple.avfoundation        	0x980870f2 -[AVAssetReaderTrackOutput     _enableTrackExtractionReturningError:] + 1243
    4   com.apple.avfoundation        	0x980159b9 -[AVAssetReaderOutput     _prepareForReadingReturningError:] + 72
    5   com.apple.avfoundation        	0x980156c0 -[AVAssetReader startReading] + 342
    6   cc.openFrameworks.ofapp       	0x005442cd -[ofAVFoundationVideoPlayer     createAssetReaderWithTimeRange:] + 1804
    7   cc.openFrameworks.ofapp       	0x0054382f __47-[ofAVFoundationVideoPlayer loadWithURL:async:]_block_invoke_2 + 593
    8   libdispatch.dylib             	0x9b09ce12 _dispatch_call_block_and_release + 15
    9   libdispatch.dylib             	0x9b098130 _dispatch_client_callout + 50
    10  libdispatch.dylib             	0x9b09bd05 _dispatch_queue_drain + 1017
    11  libdispatch.dylib             	0x9b09d99d _dispatch_queue_invoke + 186
    12  libdispatch.dylib             	0x9b09af89 _dispatch_root_queue_drain + 395
    13  libdispatch.dylib             	0x9b0aa63d _dispatch_worker_thread3 + 97
    14  libsystem_pthread.dylib       	0x9175b1da _pthread_wqthread + 724
    15  libsystem_pthread.dylib       	0x91758e2e start_wqthread + 30

slightly different, but quite similar in another test run:

    Thread 5 Crashed:: Dispatch queue: com.apple.avfoundation.avasset.completionsQueue
    0   com.apple.CoreFoundation      	0x9551bf48 CFArrayAppendValue + 40
    1   com.apple.MediaToolbox        	0x97a2f0d5 0x977ff000 + 2293973
    2   com.apple.MediaToolbox        	0x97a32bad 0x977ff000 + 2309037
    3   com.apple.avfoundation        	0x980870f2 -[AVAssetReaderTrackOutput _enableTrackExtractionReturningError:] + 1243
    4   com.apple.avfoundation        	0x980159b9 -[AVAssetReaderOutput _prepareForReadingReturningError:] + 72
    5   com.apple.avfoundation        	0x980156c0 -[AVAssetReader startReading] + 342
    6   cc.openFrameworks.ofapp       	0x004b72cd -[ofAVFoundationVideoPlayer createAssetReaderWithTimeRange:] + 1804
    7   cc.openFrameworks.ofapp       	0x004b682f __47-[ofAVFoundationVideoPlayer loadWithURL:async:]_block_invoke_2 + 593
    8   libdispatch.dylib             	0x9b09ce12 _dispatch_call_block_and_release + 15
    9   libdispatch.dylib             	0x9b098130 _dispatch_client_callout + 50
    10  libdispatch.dylib             	0x9b09bd05 _dispatch_queue_drain + 1017
    11  libdispatch.dylib             	0x9b09d99d _dispatch_queue_invoke + 186
    12  libdispatch.dylib             	0x9b09af89 _dispatch_root_queue_drain + 395
    13  libdispatch.dylib             	0x9b0aa63d _dispatch_worker_thread3 + 97
    14  libsystem_pthread.dylib       	0x9175b1da _pthread_wqthread + 724
    15  libsystem_pthread.dylib       	0x91758e2e start_wqthread + 30

I can force a crash by loading clips into already playing instances, and I guess that the problem is somhwere situated in the `cleanupAndAutoreleas` code.

i-n-g-o:
hm. very strange. cleanup should not interfere with `createAssetReaderWithTimeRange`, in fact `cleanupAndAutorelease` is called on a different instance.
`cleanup` does the same as what `dealloc` did before... i added a [_player release] so that it is exactly the same as before.

worth to try would be to call `cleanup` from outside the object and then do a normal `autorelease` afterwards. the simplest would be to just remove `[self autorelease]` in `cleanupAndAutorelease` and call `[toDel autorelease];` in the GC. maybe also getting rid of the override of `dealloc` in the VideoPlayer.
i could push that...

i changed how the videoPlayer gets disposed in `loadPlayer()` and just call `close()`. this also disposes the current textureRef and destroys the textureCache. this might not be necessary, but worth a try.

urshofer:
I'll give it a try later. Probably it would also be an elegant option to `dispatch_async` the cleanup process instead of using the `ofThread` class. This would render the GC class obsolete.

The players without `cleanupAndAutorelease` are still running though.

urshofer:
still crashing also after the latest commits with the same exception. the old branch is still running.
    
    Thread 13 Crashed:: Dispatch queue: com.apple.avfoundation.avasset.completionsQueue
    0   com.apple.CoreFoundation      	0x9551bf48 CFArrayAppendValue + 40
    1   com.apple.MediaToolbox        	0x97a2f0d5 0x977ff000 + 2293973
    2   com.apple.MediaToolbox        	0x97a32bad 0x977ff000 + 2309037
    3   com.apple.avfoundation        	0x980870f2 -[AVAssetReaderTrackOutput _enableTrackExtractionReturningError:] + 1243
    4   com.apple.avfoundation        	0x980159b9 -[AVAssetReaderOutput _prepareForReadingReturningError:] + 72
    5   com.apple.avfoundation        	0x980156c0 -[AVAssetReader startReading] + 342
    6   cc.openFrameworks.ofapp       	0x00532f1d -[ofAVFoundationVideoPlayer createAssetReaderWithTimeRange:] + 1804
    7   cc.openFrameworks.ofapp       	0x00532484 __47-[ofAVFoundationVideoPlayer loadWithURL:async:]_block_invoke_2 + 593
    8   libdispatch.dylib             	0x9b09ce12 _dispatch_call_block_and_release + 15
    9   libdispatch.dylib             	0x9b098130 _dispatch_client_callout + 50
    10  libdispatch.dylib             	0x9b09bd05 _dispatch_queue_drain + 1017
    11  libdispatch.dylib             	0x9b09d99d _dispatch_queue_invoke + 186
    12  libdispatch.dylib             	0x9b09af89 _dispatch_root_queue_drain + 395
    13  libdispatch.dylib             	0x9b0aa63d _dispatch_worker_thread3 + 97
    14  libsystem_pthread.dylib       	0x9175b1da _pthread_wqthread + 724
    15  libsystem_pthread.dylib       	0x91758e2e start_wqthread + 30

i-n-g-o:
thanks for testing!
i changed it such that `cleanup` and `autorelease` is called from outside the class.
also true, to do that on a concurrent queue. i got rid of the GC-class and dispose the players with async_dispatch...
the changes work fine in a simple test-app. curious how it behaves in your test-scenario.

urshofer:
hm - apparently it does not help. with your latest commits it's still crashing. same output:

    Thread 32 Crashed:: Dispatch queue: com.apple.avfoundation.avasset.completionsQueue
    0   com.apple.CoreFoundation      	0x9551bf48 CFArrayAppendValue + 40
    1   com.apple.MediaToolbox        	0x97a2f0d5 0x977ff000 + 2293973
    2   com.apple.MediaToolbox        	0x97a32bad 0x977ff000 + 2309037
    3   com.apple.avfoundation        	0x980870f2 -[AVAssetReaderTrackOutput _enableTrackExtractionReturningError:] + 1243
    4   com.apple.avfoundation        	0x980159b9 -[AVAssetReaderOutput _prepareForReadingReturningError:] + 72
    5   com.apple.avfoundation        	0x980156c0 -[AVAssetReader startReading] + 342
    6   cc.openFrameworks.ofapp       	0x004b4cad -[ofAVFoundationVideoPlayer createAssetReaderWithTimeRange:] + 1804
    7   cc.openFrameworks.ofapp       	0x004b4208 __47-[ofAVFoundationVideoPlayer loadWithURL:async:]_block_invoke_2 + 593
    8   libdispatch.dylib             	0x9b09ce12 _dispatch_call_block_and_release + 15
    9   libdispatch.dylib             	0x9b09bd05 _dispatch_queue_drain + 1017
    10  libdispatch.dylib             	0x9b09d99d _dispatch_queue_invoke + 186
    11  libdispatch.dylib             	0x9b09af89 _dispatch_root_queue_drain + 395
    12  libdispatch.dylib             	0x9b0aa63d _dispatch_worker_thread3 + 97
    13  libsystem_pthread.dylib       	0x9175b1da _pthread_wqthread + 724
    14  libsystem_pthread.dylib       	0x91758e2e start_wqthread + 30

(whole log here: https://gist.github.com/urshofer/0f48ff9a0a57006263e8#file-gistfile1-txt)

i-n-g-o:
ok. last try before i dump this branch...
i moved the cleanup-code back into `dealloc`. now it is the same as it originally was.
if this does not help, i will continue working on the "add_GC_for_videoplayer" branch. and add from the current working code changes one-by-one.

urshofer:
ok - got it running over here. the first 10 minutes successful. but it needs some time for possible runtime errors to show up. anyway, looks promising since I always had the feeling that the `close()` and `dealloc()` somehow overreleased the player.

i-n-g-o:
hm. how would the player be over-released? maybe i don't see something...
about the longer loading time: since a couple of commits when loading a file it does also destroy the texture-cache. could this be responsible for the longer loading? are you sure it is the `dispatch_async` call?

urshofer:
update: still running – I guess I would call it quite stable like this.
I was just wondering that in the old version, `cleanup` was called and then once again by `dealloc`. As I am probably not so firm in the code as you are I was just guessing that this duplicity could probably cause troubles.

I forgot the texture cache, probably that's the difference in loading time. You're right.

i-n-g-o:
ok. cool.
some changes:
on loadPlayer() the textureCache is not getting destroyed, therefore loading is faster again
for a test i skipped the `dealloc` and only cleanup in `cleanup`. curious if this shows the same symptoms as the safety-cleanup call in `dealloc` before.

i-n-g-o:
@urshofer 
i wonder: do you call setPosition(), firstFrame() or stop() on your ofAVFoundationPlayer from a different thread? maybe shortly after you load a file? or can it happen that one of those functions get called while a file is loading?

urshofer:
No - all movie control (load/pause/...) is run from the main thread. I check in the update function if there are commands on a queue. If so, the command is sent to the player.  The queue however is filled from another thread and locked/unlocked.

(I used to have a player run on a different thread in a non-textured mode. But this is quite slow since the pixels have to be cloned to a texture)

It can totally happen after many loads that a setPosition() / firstFrame() call happens on an unloaded clip. But I assume this should not be a problem since all these functions have a check like

    if(videoPlayer == NULL) {
        return;
    }

btw: You're recent fixes crash with about the same debug output.

Am 30.03.2015 um 22:58 schrieb Ingo Randolf <notifications@github.com>:

i-n-g-o:
ok. i give up on that for now.
moved everything back into `dealloc`, how it was in the working version.

urshofer:
couldn't we pull this into master ?

Pinging @admsyn @julapy / @openframeworks/video?

i-n-g-o:
i think with the last commits it's fine to merge.

in case the videoPlayer gets updates internally, we can avoid the retainCycle so that `dealloc` gets actually called. since we call `autorelease` from a different thread i am not quite sure, what side-effects this might have. maybe this is better to be rewritten, so that we use our own queue instead of the main_queue for `timeObserver`. then we could synchronize that queue after removing the timeObserver.
but by default videoPlayer is set to be updated externally anyway.

urshofer:
running tests now with the latest commit. confirming that load time is back to almost zero. very good.

btw: this PR is also resolving https://github.com/openframeworks/openFrameworks/issues/3729, which can be useful if the player is run in non-textured mode for background processing of video images.

urshofer:
Test application is running stable now for 30 hours. Seems ready to merge into master in my opinion…

i-n-g-o:
cool. thanks for all the testing!

bilderbuchi:
pinging @openframeworks/video @openframeworks/macos for a merge decision.

urshofer:
After 50 hours of constant loading / unloading, triggering, rewinding I stop the test. Looks stable.

ofTheo:
Looks good to me! thanks for the PR. 
Merging.

arturoc:
@i-n-g-o can you close the related issues? i'm not sure any more which ones are fixed :)

urshofer:
closed the ones related I've opened:

https://github.com/openframeworks/openFrameworks/issues/3729
https://github.com/openframeworks/openFrameworks/issues/3719
https://github.com/openframeworks/openFrameworks/issues/3715
https://github.com/openframeworks/openFrameworks/issues/3704


Am 03.04.2015 um 19:58 schrieb arturo <notifications@github.com>:

bilderbuchi:
thanks for the extensive load testing, @urshofer !

mattfelsen:
Hi all. Sorry for the much delayed response here...I'm running on an OF fork from Nov of last year so it was difficult to test these changes with my app. I noticed lots of video performance issues with large files (not related to work in this PR) which I'm opening a separate issue about.

Issue #3743, opened 2015-04-02
windows shouldn't move on creation


arturoc:
http://forum.openframeworks.cc/t/github-master-glfw-window-moves-to-top-left-when-created/19100

we should make the position private in the settings class and add a method so when the position is set it marks a flag that allows to know if we need to move the window when it's created

PR #3744, opened 2015-04-02
Bug 3743: Make ofWindowSettings.position protected in order to track if ...

trojanfoe:
...it's been set.

See https://github.com/openframeworks/openFrameworks/issues/3743


arturoc:
thanks!

PR #3745, opened 2015-04-04
ofXml: better parse error messages & file existence check

danomatika:
* catch SAXException for better parse error info
* check if file exists before trying to load and print error if not

Now I get:

    [ error ]  ofXml: parse error: Tag mismatch in line 10 column 3

and:

    [ error ] ofxXml: couldn't load, "glslColorScheme.xml" not found

instead of:

    [ error ] ofXml: loadFromBuffer(): DOM ERROR


admsyn:
:+1:

arturoc:
thanks!

PR #3754, opened 2015-04-09
added getter for type long (ofXml & ofUtils)

fx-lange:
Added getter methods for type long in ofXml (as in #3081 for ofxXmlSettings). 

I used long long instead of long for maximal value range.

Tested single methods + serialize and deserialize in combination with ofParameter (Ubuntu only).

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -475,9 +475,9 @@ double ofToDouble(const string& doubleString) {
 }
 
 //----------------------------------------
-long long ofToLong(const string& longString) {
-	long long x = 0;
-	istringstream cur(longString);
+int64_t ofToInt64(const string& intString) {
+	int64_t x = 0;
+	istringstream cur(intString);
 	cur >> x;
 	return x;
 }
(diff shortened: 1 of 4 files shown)


arturoc:
looks good, i think we should begin to use int64_t and uint64_t for this types. it's much shorted and unrderstandable, getLong that returns `long long` is kind of confusing and it's also really easy to do:

`long v = xml.getLong()`

where you could loose precision and it's a really difficult error to figure out

fx-lange:
yes I agree, thought a lot about it. it is confusing and depending on system/compiler/... long and int have the same range or long and long long . I think I should at least rename the functions to getLongLong and ofToLongLong but that's quite long :) 

haven't really seen int64 in oF land so far. so you mean functions like ofGetElapsedTimeMillis() should also return int64 some day?

arturoc:
yes we haven't used it much till now, only in ofGetElapsed* but ofSoundBuffer has quite a few of them now. for the return types we can change it without problem since it won't break anything but for naming methods i guess we should decide on it before adding them.

fx-lange:
hey @arturoc, changed everything related to int64_t. what do you think?

bilderbuchi:
I think this is a misconfig in @arturoc Travis setup for the linux side (it's a bit of a hack currently) - Arturo, that Details link points to the generic build list, not to a specific build, maybe something with the PR build config?

arturoc:
yes i was doing some changes in the config before, surely the PR arrived while i was moving things around and failed. i'm going to merge it. thanks!

Issue #1005, opened 2012-02-29
feature ofRandom(ofVec3f) 


elliotwoods:
Propose the following functions:

``` c++
ofVec2f ofRandom(const ofVec2f&);
ofVec3f ofRandom(const ofVec3f&);
ofVec2f ofRandom(const ofVec2f& range1, const ofVec2f& range2);
ofVec3f ofRandom(const ofVec3f& range1, const ofVec3f& range2);
```

also lots of other candidates for this, e.g. `ofClamp`, `ofMap`, etc
some clever templating possible?

artificiel:
this can be closed as there are now templated methods to apply std::random_distributions to glm::vec — see https://github.com/openframeworks/openFrameworks/issues/3198#issuecomment-1745243359 for more info.

Issue #224, opened 2010-08-26
gaussian noise


arturoc:
can this be useful?
http://www.openframeworks.cc/forum/viewtopic.php?f=25&t=4500

DomAmato:
How should we name this feature? ofGaussian with an overloaded function? or is this meant to replace the ofRandom functions?

kylemcdonald:
i'm think the api would just be `ofRandomGaussian(float mean = 0., float std = 1.)` and it would not replace `ofRandom`.

DomAmato:
Since there was discussion of moving to the light weight pcg random, they have a c++ implementation. its pretty simple, like 3 files

https://github.com/imneme/pcg-cpp

DomAmato:
float ofRandomGaussian(float mean, float stddev) {
	std::default_random_engine generator;
	std::normal_distribution<float> distribution(mean, stddev);
	return distribution(generator);
    }

normal distributions were added in c++11 and I believe all of our platforms are now c++11 compliant?

kylemcdonald:
yes, i think that's correct. though i'd look at how the seed for ofRandom works, reinitializing it on every call might be a problem?

regardless, if you could write a short example to demonstrate behavior and submit a PR that'd be great.

DomAmato:
I dont think that the seeding works because it deals with a different subset of code. std::normal_distribution requires the <random> header to be included and an engine to be created. We could create a static initialization. I think we would need multiple functions though because in order to set the parameters of the distribution (if someone decides to change it like how it is posted above) you have to create a new parameter and pass it in.

like so:

    static std::default_random_engine generator;
    static std::normal_distribution<float> distribution(0, 1);

    float ofRandomGaussian(float mean, float stddev) {
	    std::normal_distribution<float>::param_type _param(mean, stddev);
	    distribution.param(_param);
	    return distribution(generator);
    }

    float ofRandomGaussian() {
        //this assumes the parameters dont change just get another number from the distribution
        return distribution(generator);
     }

DomAmato:
also c++11 added in a lot more random features:


bernoulli_distribution
binomial_distribution
cauchy_distribution
chi_squared_distribution
discrete_distribution
exponential_distribution
extreme_value_distribution
fisher_f_distribution
gamma_distribution
geometric_distribution
lognormal_distribution
negative_binomial_distribution
normal_distribution
piecewise_constant_distribution
piecewise_linear_distribution
poisson_distribution
student_t_distribution
[...]

PR #3757, opened 2015-04-12
ofMath: variable clean-up

piotrts:
Removes unnecessary variables and variable assignment in ofRandom(), ofRandomf() and ofRandomuf().


bakercp:
Seems reasonable to me.

bilderbuchi:
+1 looks reasonable
@piotrts some other small issues that may be interesting to you: #224, amybe #1005

arturoc:
thanks!

PR #3761, opened 2015-04-15
Fix ofxAndroidSoundPlayer getVolume

jknight-nc:
getVolume was using an incorrect jni method signature. It was using "(V)F", when it should actually be "()F". getVolume was triggering a crash on Android devices.


jknight-nc:
The java method is getVolume located at:

https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidSoundPlayer.java

Clearly, it has no parameters.

arturoc:
thanks! i'm was almost sure that V was for void not sure how that has been like that for so long

PR #3764, opened 2015-04-18
Update link (minor fix)

kant:


docs/osx.md
```diff
@@ -9,7 +9,7 @@ Installing Xcode
 ----------------
 openFrameworks is primarily developed with Xcode 4+, but may work with other versions. 
 
-To install Xcode, you need to register as an Apple developer and download the Xcode disk image from: https://connect.apple.com/
+To install Xcode, you need to register as an Apple developer and download the Xcode disk image from: https://developer.apple.com/xcode/downloads/
```
bakercp: I realize this is just a quick edit -- but I think we should mention that you can also get Xcode from the app store for more recent OS versions.

bakercp:
+1

PR #3773, opened 2015-04-21
Bugfix make look at

### LINKED ISSUE #3772: ofMatrix4x4::makeLookAtMatrix seems to do the wrong thing
I'm pretty sure makeLookAtMatrix does the wrong thing, both setting the first three columns using incorrect basis vectors, and accidentally setting the perspective transform value to 0. Compare to ofMatrix4x4::makeLookAtViewMatrix. I've got a commit ready which fixes this.

mhintz:
Fixes #3772

### DIFF
--- a/libs/openFrameworks/math/ofMatrix4x4.cpp
+++ b/libs/openFrameworks/math/ofMatrix4x4.cpp
@@ -934,7 +934,7 @@ void ofMatrix4x4::makeLookAtMatrix(const ofVec3f& eye,const ofVec3f& center,cons
 	_mat[0].set(xaxis.x, xaxis.y, xaxis.z, 0);
 	_mat[1].set(yaxis.x, yaxis.y, yaxis.z, 0);
 	_mat[2].set(zaxis.x, zaxis.y, zaxis.z, 0);
-	_mat[3] = eye;
+	_mat[3].set(eye.x, eye.y, eye.z, 1);
 }
 
 void ofMatrix4x4::getLookAt(ofVec3f& eye,ofVec3f& center,ofVec3f& up,float lookDistance) const
--- a/libs/openFrameworks/math/ofMatrix4x4.cpp
+++ b/libs/openFrameworks/math/ofMatrix4x4.cpp
@@ -931,10 +931,10 @@ void ofMatrix4x4::makeLookAtMatrix(const ofVec3f& eye,const ofVec3f& center,cons
 	ofVec3f xaxis = up.getCrossed(zaxis).getNormalized();
 	ofVec3f yaxis = zaxis.getCrossed(xaxis);
 
-	_mat[0].set(xaxis.x, xaxis.y, xaxis.z, 0);
-	_mat[1].set(yaxis.x, yaxis.y, yaxis.z, 0);
-	_mat[2].set(zaxis.x, zaxis.y, zaxis.z, 0);
-	_mat[3] = eye;
+    _mat[0].set(xaxis.x, yaxis.x, zaxis.x, 0);
+    _mat[1].set(xaxis.y, yaxis.y, zaxis.y, 0);
+    _mat[2].set(xaxis.z, yaxis.z, zaxis.z, 0);
+    _mat[3].set(eye.x, eye.y, eye.z, 1);
 }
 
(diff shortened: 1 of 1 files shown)


mhintz:
P.S. I was able to test this and whereas before, I was seeing nothing after applying the result of makeLookAtMatrix to an object, now I'm seeing what appears to be the correct position and orientation. (on OSX 10.10)

libs/openFrameworks/math/ofMatrix4x4.cpp
```diff
@@ -931,10 +931,10 @@ void ofMatrix4x4::makeLookAtMatrix(const ofVec3f& eye,const ofVec3f& center,cons
 	ofVec3f xaxis = up.getCrossed(zaxis).getNormalized();
 	ofVec3f yaxis = zaxis.getCrossed(xaxis);
 
-	_mat[0].set(xaxis.x, xaxis.y, xaxis.z, 0);
-	_mat[1].set(yaxis.x, yaxis.y, yaxis.z, 0);
-	_mat[2].set(zaxis.x, zaxis.y, zaxis.z, 0);
-	_mat[3] = eye;
```
tgfrerer: That's a great catch! An `ofVec3f` implicitly cast to `ofVec4f` will have `.w` set to 0, which will lead to invalid homogenous coordinates.

tgfrerer:
Looks good to me, although in the .h file the comment on top of the declaration of `makeLookAtViewMatrix` says that 
`makeLookAtViewMatrix` is the inverse of `makeLookAtMatrix`. 

It seems that `makeLookAtMatrix` is not used anywhere in the core, and I wonder what it was originally intended to be used for. 

`ofNode` implements its own lookAt calculations in ofNode.cpp#L241 ff., which are similar to the original implementation:

		ofMatrix4x4 m;
		m._mat[0].set(xaxis.x, xaxis.y, xaxis.z, 0);
		m._mat[1].set(yaxis.x, yaxis.y, yaxis.z, 0);
		m._mat[2].set(zaxis.x, zaxis.y, zaxis.z, 0);

Which, to me, seems correct. What seems odd in the current implementation of `makeLookAtMatrix` in `ofMatrix.cpp` is that it introduces a translation (by applying eye to _mat[3]), which does not immediately seem to make sense, since a lookAt Matrix should be a straightforward rotation matrix. 

@mhintz does it work for you if you get rid of the `_mat[3] = eye;` line?

libs/openFrameworks/math/ofMatrix4x4.cpp
arturoc: ofNode, which is used in lots of places, internally sets the lookAt like;

```
		ofMatrix4x4 m;
		m._mat[0].set(xaxis.x, xaxis.y, xaxis.z, 0);
		m._mat[1].set(yaxis.x, yaxis.y, yaxis.z, 0);
		m._mat[2].set(zaxis.x, zaxis.y, zaxis.z, 0);
```

so except for that 4th row i'd say is correct. the difference between lookAtMatrix and lookAtViewMatrix is that lookAtMatrix is supposed to be used for objects while lookAtViewMatrix is used to set up a view matrix for a camera which is the inverse

arturoc:
didn't saw your last comment tim :)   

the eye translation is because it's also supposed to move the object to the eye position and lookAt center. 

we should move to glm asap anyway.

tgfrerer:
=) yes, +1k for glm!

@mhintz does it work if only the last line is fixed in the way you suggested?

mhintz:
Fixing just the last line shows the object on screen, but I'm not sure it works in the intended manner. I made what I think is a basic test case, which shows an object using the existing matrix in purple and one using the proposed style in green. (https://gist.github.com/mhintz/5f229781bfa5dbe350b6). I'm pretty sure that the existing way positions the object with the wrong position and orientation. I'm not entirely sure what the difference is between this case and ofNode, but will take a look in order to better understand the whole system.

mhintz:
Ok, looked at this again this week and realized that there was a mistake in the test case I was using :(. You're right, the only thing that ought to be changed is that last line. I've reverted the commit I originally made so it's just the one-liner.

arturoc:
thanks!

PR #3777, opened 2015-04-23
use backingscale to re-position window on os x

tgfrerer:
* take HIDPI / LODPI into account when repositioning fullscreen window on OS X

before this, HIDPI fullscreen windows would not cover the full screen.


bakercp:
How does this factor interact with the existing `pixelScreenCoordScale` that is calculated as `        pixelScreenCoordScale = framebufferW / windowW;`?

tgfrerer:
It's essentially the same, but manually calculated. i've changed the PR to use the existing `pixelScreenCoordScale`. Cheers!

bakercp:
+100 :)

tgfrerer:
Swell. I'd say this is ready to merge, what do you think @arturoc ?

PR #3781, opened 2015-04-25
small updates: guiExample code consistency & OF_TTF string as static const

danomatika:
**guiExample code consistency**: some small updates to resolve multiple coding styles and the use of multiple ifs instead of if else statements.

**OF_TTF string as static const**: this fixes a SWIG error when generating OF bindings (aka for ofxLua). Simply swapping "const static" with "static const" solves it. C++ doesn't care either way, so this change [shouldn't effect anything](http://stackoverflow.com/a/12006039/2146055). Besides, "static const" is the recommended order in C anyway.


admsyn:
:+1: might as well convert the deprecated `loadSound` to `load` as well.

arturoc:
thanks!

PR #3782, opened 2015-04-25
Fix a typo

procedural:
A typo which causes a compile-time error on VS 2013.


admsyn:
:+1: going to just pull this now, since it is quite clearly a typo

PR #3783, opened 2015-04-25
Fix compile time errors for Windows and C++11

procedural:


procedural:
[Here you can](https://github.com/openframeworks/openFrameworks/commit/73f7e17) see `char *data = new char[ MAX_BUFFER_SIZE ];` line was accidentally replaced with `data = new char[ MAX_BUFFER_SIZE ];`

procedural:
As for `reinterpret_cast`, Clang on Linux with C++11 enabled forced me to do it explicitly by causing compile-time error.

procedural:
You can see cast even in example of MSDN docs: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740476%28v=vs.85%29.aspx

bilderbuchi:
See, this is why I really don't like patching upstream libraries manually - the second part is [already correct](https://code.google.com/p/oscpack/source/browse/trunk/ip/win32/UdpSocket.cpp#433) in oscpack itself. The first change I can't relate to upstream oscpack code, is it also something we modified ourselves?

procedural:
@bilderbuchi when you view the file on github it has "Blame" button on top right corner which shows who introduced each line... Here's that commit for first change: https://github.com/procedural/openFrameworks/commit/b64cb41

procedural:
It seems like the same changes are merged now: https://github.com/openframeworks/openFrameworks/pull/3793

Closed.

PR #3784, opened 2015-04-25
Changing ofxNetwork to fix compile-time errors on Windows and VS 2013

procedural:


procedural:
I tested it already, it compiles and works for MinGW and VS 2013, the code just replaces unavailable on Windows `fcntl` with `ioctlsocket` and adds a cast for `getsockopt` which is also necessary to avoid compile error on Visual Studio (see [related MSDN page](https://msdn.microsoft.com/en-us/library/windows/desktop/ms738544%28v=vs.85%29.aspx))

arturoc:
thanks!

PR #3787, opened 2015-04-26
Add guard to avoid segfault on uninitialized Video Grabber

Elettronik:
If the ofVideoGrabber is not initialized on application start the grabber member is null, resulting in a segmentation fault, if  we try to check for the initialization of object; The performance penality of this guard are negligible.


admsyn:
:+1:

arturoc:
thanks!

PR #3789, opened 2015-04-28
Fixes ofxOpenALSoundPlayer const issues

danoli3:
Fixes ofxOpenALSoundPlayer const issues associated with the new const changes for 0.9.

Picked up by Xcode under libc++. It has stricter compiler checking so this was an error.

Issue: Calling a non-const function (Update) and assigning a variable under a const function.

openal

### DIFF
--- a/addons/ofxiOS/src/sound/ofxOpenALSoundPlayer.cpp
+++ b/addons/ofxiOS/src/sound/ofxOpenALSoundPlayer.cpp
@@ -223,15 +223,22 @@ void ofxOpenALSoundPlayer::setPaused(bool bP) {
 		return;
 	
 	if(iAmAnMp3)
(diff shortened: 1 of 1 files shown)


danoli3:
Need this one in main. Impossible to compile for clang for iOS.

Someone please give it a once over (remembering you can't modify a member or call another non-const function in a const)

bilderbuchi:
pinging @bakercp @openframeworks/ios

julapy:
hey, 

ive had a look through the proposed fix and its hard to tell if this will have any knock on effects.
ofxOpenALSoundPlayer is a really old class and a bit all over the place... so its a bit risky moving things around too much IMO.

ive PR'd another fix #3815
which removes the `const` for the `isPlaying` method,
so the method can internally remain the same.

figured this will be a safer patch for now.

what does everyone else think of this approach?

julapy:
fixed in #3815

Issue #3609, opened 2015-02-02
error C2065: 'data' : undeclared identifier in ofxOSC, UdpSocket.cpp


pierrep:
Visual studio 2012, github OF... @ofTheo ?

pierrep:
Also  error C2664: 'setsockopt' : cannot convert parameter 4 from 'unsigned long *' to 'const char *'

bilderbuchi:
sounds maybe like a wrong include in the VS project? mind that there's a posix and a win folder [here](https://github.com/openframeworks/openFrameworks/tree/master/addons/ofxOsc/libs/oscpack/src/ip), but your project should only use one.

pierrep:
Yes I'm aware of that...but no it's not that. The windows codebase is different and seems to be missing an undeclared data variable...

yty:
@pierrep @bilderbuchi 

I submitted before... 

https://github.com/yty/openFrameworks/commit/d726292e8a13b8a58d2f66f03f22f1973ab6dc7a

pierrep:
@yty Yes, that fixes it, thanks!

arturoc:
why are there errors in the libraries files? are we still using a patched oscpack? i lost track wit @ofTheo last modifications

bilderbuchi:
unfortunately, @yty 's commit contains strange stuff, like the Win32 ifdef in a posix file (which should not be included in windows), so if there's a fix in there we should tease it out. 
@yty did you ever submit this fix upstream with the oscpack developer? that's the proper place where such a fix should land, and I seem to remember that I asked if you could do that before?

PR #3793, opened 2015-04-29
Fixes compiler errors for win32 version of ofxOSC.

tgfrerer:
+ Does not affect posix systems (linux, OSX), since these compile against the `/posix` version of the affected `.cpp` file.

This adds a missing declaration and a (char *) cast to make the windows compiler happy. Otherwise apps using ofxOSC will not compile on Windows.


bilderbuchi:
mind reporting this one [upstream](https://code.google.com/p/oscpack/issues/list), Tim?

bilderbuchi:
connected issue #3609?

kylemcdonald:
@yty can you confirm that this also fixes the bug you were seeing? if it does, we can merge it.

ofTheo:
this looks good to merge. 
Both fixes are clear bugs.

PR #3794, opened 2015-04-30
ofXml cleanup

bakercp:
Lots of cleanup for ofXml.

No functional changes.

### DIFF
--- a/libs/openFrameworks/utils/ofXml.cpp
+++ b/libs/openFrameworks/utils/ofXml.cpp
@@ -1,8 +1,7 @@
-
 #include "ofXml.h"
 
 ofXml::~ofXml() {
-	releaseAll();
+    releaseAll();
 }
 
 ofXml::ofXml(const std::string& path) {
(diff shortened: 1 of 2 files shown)


arturoc:
great, thanks, i think we should wait until post 0.9 to merge this. we should do this with every class in the core so it would be weird to have one class using one style and the rest another.

we should change at least all the headers to use std::, we can discuss also if we want to use std::string... in the cpp's too or instead use `using namespace std;` there

bakercp:
Agreed. I was basically trying to understand ofXml better and was cleaning up as I went.

joshuajnoble:
Like it!

arturoc:
@bakercp if you want to clean this i'll merge it, we've been merging stuff like this already so no need to wait to do all classes at once

bakercp:
@arturoc ok.

bakercp:
Closing this because it's crazy.  Using #4159.

PR #3795, opened 2015-04-30
remove ofThreadLogger

arturoc:
instead catch the exceptions and log them locally on each thread.
simplifies the architecture and seems a better idea than having
some global state taking care of this.

i've just realized that ofThreadLogger and Poco's ErrorHandler is doing just this. i also wanted to make the thread detach itself if it failed which i would need to have done in ofThreadLogger.

i think this makes much more sense and simplifies things a bit but i might be missing something about what Poco's ErrorHandler is doing. ping @bakercp

### DIFF
--- a/libs/openFrameworks/app/ofAppRunner.cpp
+++ b/libs/openFrameworks/app/ofAppRunner.cpp
@@ -69,7 +69,6 @@ void ofInit(){
 	static bool initialized = false;
 	if(initialized) return;
 	initialized = true;
-	Poco::ErrorHandler::set(new ofThreadErrorLogger);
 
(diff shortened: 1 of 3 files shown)


bakercp:
The main advantage of using Poco's error handler is that it will catch _all_ uncaught threaded exceptions from Poco threads, not just those caused by ofThread.

bakercp:
Just to take it a little further, I do see the need for the auto-detatch in ofThread -- perhaps we could keep the current poco thread handling functionality and add the custom ofThread handling capability here w/ the detatch?

arturoc:
don't think anyone is going to use Poco threads directly even less now that we have support for c++11 ( we should actually move to c++11 thread asap) but i'm ok leaving it by now

Issue #3796, opened 2015-05-01
more useful semantic versioning on master between official releases


bakercp:
As we're getting nearer to the 0.9.0 release, I'm trying to keep my addons up-to-date with new deprecations, etc.  Currently the version on master still says 0.8.4

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofConstants.h#L5-L7

Would it be acceptable to adopt a versioning scheme where the develop branch is incremented as soon as a release is made so that addons can potentially support both the master branch and releases simultaneously with if-defs, etc?

My goal is basically to more gracefully transition during deprecations and avoid lots of warnings in my addons.

One solution might be to use patch-version extensions, for example, by adding a patch version extension of `develop` we could indicate that this is still a development version.  So, for instance, right now the version might be:

`0.9.0-develop`

Anyway, this would require one or two minor changes, but would all still constitute legal sem-ver 2.0.0 syntax, etc.

Thoughts?

arturoc:
there's a similar PR already from @armadillu can't find it right now. anyway i think just updating the version number right after we do the release should be enough. other projects do that and it works without problem

armadillu:
Hey @arturoc  I made no PR, I just asked in the forum as I was just wondering how others handled that. I don't care so much about warnings/deprecations, but I do care about missing/renamed methods that don't let the compile. In my case, its ofSetPixelStorei vs ofSetPixelStoreiAlignment:

https://github.com/armadillu/ofxProgressiveTextureLoad/commit/72f565d104b0f0eaea7eba08f3bae8b6957af481

We use branches inside the addon, but it gets out of control very quickly when several people are using the addon in different projects. I really can't see a good solution...

http://forum.openframeworks.cc/t/of-version-safe-addons/18886

bakercp:
Just added a possible solution that incorporates semver version tags.

bilderbuchi:
+1, this has been an itch for me for quite a while. I also know of other projects which increment/change the version immediately after release.

bakercp:
@bilderbuchi do you have any thoughts on what the "tag" should be between releases?  Basically I want to be able to check and see if that tag is an empty string, I can know that it is an official release.  Perhaps it should be something more descriptive than "master" -- like "prerelease"?  And eventually "rc1, rc2, etc". until the release at which point it would just be a blank string?

bilderbuchi:
[...]

PR #3797, opened 2015-05-01
Add version tag support and increment version.

bakercp:
Reference https://github.com/openframeworks/openFrameworks/issues/3796

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -739,7 +739,7 @@ string ofGetVersionInfo(){
 	stringstream sstr;
 	sstr << OF_VERSION_MAJOR << "." << OF_VERSION_MINOR << "." << OF_VERSION_PATCH;
 
-	if (sizeof(OF_VERSION_PRE_RELEASE) > 1)
+	if (!std::string(OF_VERSION_PRE_RELEASE).empty())
 	{
 		sstr << "-" << OF_VERSION_PRE_RELEASE;
 	}
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -739,9 +739,9 @@ string ofGetVersionInfo(){
 	stringstream sstr;
 	sstr << OF_VERSION_MAJOR << "." << OF_VERSION_MINOR << "." << OF_VERSION_PATCH;
 
-	if (sizeof(OF_VERSION_TAG) > 1)
+	if (sizeof(OF_VERSION_PRE_RELEASE) > 1)
 	{
-		sstr << "-" << OF_VERSION_TAG;
+		sstr << "-" << OF_VERSION_PRE_RELEASE;
 	}
 
 	sstr << std::endl;
@@ -760,8 +760,8 @@ unsigned int ofGetVersionPatch() {
 	return OF_VERSION_PATCH;
 }
 
-std::string ofGetVersionTag() {
-	return OF_VERSION_TAG;
+std::string ofGetVersionPreRelease() {
+	return OF_VERSION_PRE_RELEASE;
 }
 
 
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -737,7 +737,14 @@ void ofLaunchBrowser(const string& _url, bool uriEncodeQuery){
 //--------------------------------------------------
 string ofGetVersionInfo(){
 	stringstream sstr;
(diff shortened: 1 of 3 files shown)


arturoc:
i would call it master or nightly and then stable once we do a release

bakercp:
Well, ideally (in my mind at least) for releases it will just be an empty string `""`.  Since we have moved away from the develop branch model I suppose it should be called "master" ... or in some future build-system it should be updated with build tools :)

I'll update the PR w/ master for now.

bilderbuchi:
+1 on this, and thank you for making it fully semver-compatible.
Also, +1 on having this as an empty string for release versions, maybe this should be added as a comment, and as info in the release checklist on the wiki.
finally, maybe it's a good idea to not call this "TAG" but "PRERELEASE_INDICATOR" (or somethign shorter) to make the meaning clearer for people not intimately familiar with the semver specification (i.e. that hyphenated postfixes indicate prerelease versions).

bakercp:
@bilderbuchi I see you've answered my question over here.  Regarding what to call it -- I can't seem to find a clear "spec" answer ... i've seen patch-extension and seen it referred to as a tag ... (older versions of the spec seem to call it tag).

Anyway, I'd prefer to follow the semver2.0.0 naming convention if we can figure out what that is ...

bilderbuchi:
I think the spec only details the contents of the version string, and not the "name" of the parts making it up, but if you follow the system of the spec items 6 to 9, `PRERELEASE` (or more precisely `PRE-RELEASE`, but that seems strange for a preprocessor variable) seems to be most consistent with the semver specification.

bakercp:
@bilderbuchi I think you nailed the name -- it should indeed be a version of pre-release. I'd suggest `OF_VERSION_PRE_RELEASE`.  I did take a look to see if we had any other examples of collapsing hyphens/dashes in variable names (constants) in oF and I can't seem to find any.  It comes down to an aesthetic thing for me since we can't really lean on consistency and .. for whatever reason, even though it's long, I find PRE_RELEASE easier to read.

bilderbuchi:
ok, let's do that, then. Incidentally, it also makes the meaning clear without any additional comment or documentation.

bakercp:
As soon as this is merged, I can update the wiki checklist https://github.com/openframeworks/openFrameworks/wiki/Release-preparation-checklist

libs/openFrameworks/utils/ofUtils.h
```diff
@@ -820,6 +820,20 @@ unsigned int ofGetVersionMinor();
 /// \returns The patch version number.
 unsigned int ofGetVersionPatch();
 
+/// \brief Get the pre-release version of openFrameworks.
+///
+/// openFrameworks uses the semantic versioning system.
+///
+/// For pre-release versions of openFrameworks, including development versions,
+/// this string will describe the pre-release state. Examples might include
+/// "master", "rc1", "rc2", etc.  For all official releases, this string will be
```
bilderbuchi: maybe replace "official" with "full" or "proper"/"main" - prereleases are also official in that we (should) decide which commit is the prerelease state, and set a git tag, etc.

bakercp: Ah good point.  "For all stable releases, ..."

libs/openFrameworks/utils/ofUtils.cpp
```diff
@@ -737,7 +737,14 @@ void ofLaunchBrowser(const string& _url, bool uriEncodeQuery){
 //--------------------------------------------------
 string ofGetVersionInfo(){
 	stringstream sstr;
-	sstr << OF_VERSION_MAJOR << "." << OF_VERSION_MINOR << "." << OF_VERSION_PATCH << endl;
+	sstr << OF_VERSION_MAJOR << "." << OF_VERSION_MINOR << "." << OF_VERSION_PATCH;
+
+	if (sizeof(OF_VERSION_PRE_RELEASE) > 1)
```
arturoc: using sizeof here is kind of weird no? something like `string(OF_VERSION_PRE_RELEASE).length()` makes more sense i think

bakercp:
@arturoc  -- fixed that sizeof thing.  Initially I was having problems with string literals in preprocessor defines and found that solution on stackoverflow.  Clearly there was a different problem, so I fixed it using a string.  Much prettier now.  Good to merge?

arturoc:
thanks!

bakercp:
+1 :)

bakercp:
@dimitre - master now reports a more useful version string.  CB

armadillu:
Thx guys! :)

dimitre:
pure bliss, thank you guys! :)

PR #3798, opened 2015-05-02
Documentation matrix4x4

mhintz:
Hi all,

Partly as a learning exercise, partly as an effort to add some more thorough documentation, I went through ofMatrix4x4 and added some comments explaining the uses of each function. I'm sure most of it is obvious to people who know what they're looking at when inspecting a matrix class, but documentation is for those of us who don't always know! I also added some text pointing out OF's use of row-vector left-to-right style to the top.

Split up into multiple commits so each section can be examined independently.

### DIFF
--- a/libs/openFrameworks/math/ofMatrix4x4.h
+++ b/libs/openFrameworks/math/ofMatrix4x4.h
@@ -33,6 +33,20 @@
 /// kidding, bazzlemunging is not a thing) stuff in OF is handled internally by
 /// objects like ofNode or ofCamera.
 /// 
+/// OF uses row-vector style by default, meaning that when transforming a vector
+/// by multiplying with a matrix, you should put the vector on the left side and
+/// the matrix (or matrices) to its right. When multiplying by multiple matrices,
+/// the order of application of the transforms is left-to-right. This means that
+/// the standard order of manipulation operations is vector * scale * rotate * translate.
+/// 
+/// Note that in GLSL, this convention is reversed, and column-vector style is used.
+/// OF uploads the matrices correctly, but you should reverse the order of your manipulations
+/// to, e.g. translate * rotate * scale * vector.
+/// 
+/// OF still lets you do matrix-vector multiplication with the vector
+/// on the right if that's your preferred style. You can perform matrix transformations
+/// that use the OpenGL style by using functions like glTranslate, glRotate, and glScale.
+/// 
 class ofMatrix4x4 {
 public:
 //	float _mat[4][4];
--- a/libs/openFrameworks/math/ofMatrix4x4.h
+++ b/libs/openFrameworks/math/ofMatrix4x4.h
@@ -417,21 +417,27 @@ public:
 	/// ofMatrix4x4 T = ofMatrix4x4::newTranlationMatrix(b);
 	/// ofVec3f c = a*R*T;
 	/// ~~~~~
-	///
-	inline ofVec3f postMult( const ofVec3f& v ) const;
 	/// where `*` is calling preMult()
+	/// 
 
+	/// \brief Matrix * vector multiplication. This operation
+	/// implicitly treat vectors as column-matrices
+	inline ofVec3f postMult( const ofVec3f& v ) const;
 	inline ofVec4f postMult( const ofVec4f& v ) const;
 
-	/// \brief Basic Matrixf multiplication
+	/// \brief post-multiply by another matrix. This matrix becomes `this * other`
 	void postMult( const ofMatrix4x4& );
 
+	// \brief Vector * matrix multiplication. This
+	// operation implicitly treats vectors as row-matrices
 	inline ofVec3f preMult( const ofVec3f& v ) const;
 	inline ofVec4f preMult( const ofVec4f& v ) const;
 
+	// \brief pre-multiply by another matrix. This matrix becomes `other * this`
 	void preMult( const ofMatrix4x4& );
 
-
+	// \brief Equivalent to calling postMult(other) but you can do
+	// someMatrix *= someMatrix without breaking const-correctness
 	inline void operator *= ( const ofMatrix4x4& other ) {
 		if ( this == &other ) {
 			ofMatrix4x4 temp(other);
@@ -439,6 +445,7 @@ public:
 		} else postMult( other );
 	}
 
+	// \brief creates a new matrix from the product of two matrices
 	inline ofMatrix4x4 operator * ( const ofMatrix4x4 &m ) const {
 		ofMatrix4x4 r;
 		r.makeFromMultiplicationOf(*this, m);
@@ -446,12 +453,12 @@ public:
 	}
 	
 
-	/// \brief Calls postMult()
+	/// \brief Matrix * vector multiplication. Calls postMult()
 	inline ofVec3f operator* (const ofVec3f& v) const {
 		return postMult(v);
 	}
 	
-	/// \brief Calls postMult()
+	/// \brief Matrix * vector multiplication. Calls postMult()
 	inline ofVec4f operator* (const ofVec4f& v) const {
 		return postMult(v);
 	}
--- a/libs/openFrameworks/math/ofMatrix4x4.h
+++ b/libs/openFrameworks/math/ofMatrix4x4.h
@@ -640,23 +640,23 @@ public:
 	void glScale( const ofVec3f& v );
 	/// \}
 
-	//---------------------------------------------
-	/// get methods: return matrix components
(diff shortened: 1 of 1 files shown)


bakercp:
Cool! A couple things.  Doxygen comments need 3 slashes. ///

Also, oF is the "official" abbreviation, rather than OF.

Thanks for the help!

mhintz:
Whoops, didn't know about the doxygen triple slash thing! I changed that

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -30,15 +30,30 @@
...
+/// on the right if that's your preferred style. You can perform matrix transformations
+/// that use the OpenGL style by using functions like glTranslate, glRotate, and glScale.
+/// 
 class ofMatrix4x4 {
 public:
 //	float _mat[4][4];
 	
 	/// \cond INTERNAL
 	// Should this be moved to private?
+	/// The values of the matrix, stored in row-major style
```
bakercp: Please use sentences with periods or other punctuation (https://github.com/openframeworks/openFrameworks/wiki/openFrameworks-doxygen-documentation-style-guidelines).  So in this case, just add a period.

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
 		makeIdentityMatrix();
 	}
+
+	/// \brief You can pass another ofMatrix4x4 to create a copy
 	ofMatrix4x4( const ofMatrix4x4& mat) {
 		set(mat.getPtr());
 	}
+
+	/// \brief You can pass a pointer to floats, and the first 16 contents will be extracted into this matrix.
+	/// Note: the validity of these values is not checked!
```
bakercp: Notes can be notated with:

```
/// \note The validity of these values is not checked!
```

Or better yet, use a `\warning`.

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
 	ofMatrix4x4( float const * const ptr ) {
 		set(ptr);
 	}
+
+	/// \brief Rotation matrices can be constructed from a quaternion
```
bakercp: + period, etc.

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
 	void makeRotationMatrix( float angle1, const ofVec3f& axis1,
 	                 float angle2, const ofVec3f& axis2,
 	                 float angle3, const ofVec3f& axis3);
 
 
-	// init related to another matrix
+	/// init related to another matrix
+	/// \brief Matrix becomes the inverse of the provided matrix.
 	bool makeInvertOf( const ofMatrix4x4& rhs);
+	/// \brief Matrix becomes an orthonormalized version of the provided matrix.
```
bakercp: In general, `\brief` should be no more than a single line.  It basically ends up as the one sentence description in the method list in the documentation.  The remaining paragraph should be separated by a space and can be as long as needed.

See https://github.com/openframeworks/openFrameworks/wiki/openFrameworks-doxygen-documentation-style-guidelines#document-the-class

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
+	/// The basis vectors (the 3x3 chunk embedded in the upper left of the matrix)
```
bakercp: For some of these more complicated mathematical concepts, it would be totally acceptable and helpful to give a link to the relevant wikipedia article.  See other examples in [ofMath](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/math/ofMath.h#L216).

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
-	//---------------------------------------------
-	// init as opengl related matrix for perspective settings
-	// see opengl docs of the related funciton for further details
-
-	// glOrtho
+	///---------------------------------------------
+	/// init as opengl related matrix for perspective settings
+	/// see opengl docs of the related funciton for further details
+
+	/// glOrtho
```
bakercp: Not sure what this line is for ... ?  Can it be deleted?

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
+	/// \brief Matrix becomes an orthographic projection matrix with a
+	/// box-shaped viewing volume described by the six parameters.
+	/// left, right, bottom, and top essentially specify coordinates in
+	/// the zNear clipping plane.
 	void makeOrthoMatrix(double left,   double right,
 	               double bottom, double top,
 	               double zNear,  double zFar);
 
-	// glOrtho2D
+	/// glOrtho2D
```
bakercp: Same here.

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -48,91 +63,145 @@ class ofMatrix4x4 {
...
+	/// \brief Matrix becomes a 2D orthographic projection matrix with a 
+	/// box-shaped viewing volume described by the four parameters
+	/// plus, implicitly, a zNear of -1 and a zFar of 1
```
bakercp: finish with period.

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -146,29 +215,27 @@ class ofMatrix4x4 {
 	                                   float angle3, const ofVec3f& axis3);
 	inline static ofMatrix4x4 newRotationMatrix( const ofQuaternion& quat);
 
-	
+	/// create new matrices related to glFunctions. See the description of the make* methods for more info.
```
bakercp: Capitalize first word.

libs/openFrameworks/math/ofMatrix4x4.h
```diff
@@ -376,25 +453,25 @@ class ofMatrix4x4 {
...
 	}
 
 
 	//---------------------------------------------
-	// specialized postMult methods, usually you want to use this
-	// for transforming ofVec not preMult
-	// equivalent to postMult(newTranslationMatrix(v)); */
+	/// specialized postMult methods, usually you want to use this
+	/// for transforming ofVec not preMult
+	/// equivalent to postMult(newTranslationMatrix(v)); */
```
bakercp: What is the */ doing on this line?  Can we get rid of it?

bakercp:
@mhintz Thanks again for all of the work on this.  I think people have been waiting on this class because it's challenging!

So my comments above were mostly just style things -- punctuation, capitalization, etc.

If you add those in I'll merge this and we can keep on moving!

Thanks again!

kylemcdonald:
@bakercp "OF" is the capitalization on the abbreviation and "openFrameworks" is the full name, if the [about page](http://openframeworks.cc/about/) is to be trusted.

bakercp:
I was going off this recent discussion https://github.com/openframeworks/ofBook/issues/134

mhintz:
@bakercp Thanks for linking to the doxygen style guide! There were some things there that I must have missed the first time over.

I also noticed that this should probably be a PR against the `feature-documentation` branch. Is that true?

libs/openFrameworks/math/ofMatrix4x4.h
mhintz: I'm looking around for a good link and can't find anything ... unfortunately the wikipedia article for "Normal Matrix" is not very helpful or practical to this case. I'm pretty sure the material I read to (sort of) understand the use for this was all in "A 3D Math Primer for Graphics and Game Development". How do you feel about citing paper sources?

bakercp:
Regarding the linking to sources -- rather than paper sources -- is there a part of ofBook that you could link to?

bilderbuchi:
The master branch as target is correct. The only permanent branches in the repo are master and stable, everything else is temporary (and I think in need of a cleanup, actually)

mhintz:
@bakercp Just pinging because I pushed some edits which fix the formatting issues you mention and add several \name groups which I hope will be helpful.

ofTheo:
going to merge this as it looks really helpful and thorough.

PR #3801, opened 2015-05-03
fix addons.make not getting parsed properly

ofZach:
I just pulled master and something seems broken with how the addons.make is getting parsed on OSX. 

@arturoc can you take a look at this small change (I wonder if any changes to ofBuffer line parsing might have contributed).  Essentially some OSX projects were generated without addons, because the iterating through ofBuffer was not totally working.  

happy to dig in further, I just did a minor re-write and this seems to help.

### DIFF
--- a/apps/devApps/projectGenerator/src/projects/baseProject.cpp
+++ b/apps/devApps/projectGenerator/src/projects/baseProject.cpp
@@ -173,16 +173,24 @@ void baseProject::addAddon(ofAddon & addon){
 
 void baseProject::parseAddons(){
 	ofFile addonsMake(ofFilePath::join(projectDir,"addons.make"));
-	ofBuffer addonsMakeMem;
-	addonsMake >> addonsMakeMem;
+    ofBuffer addonsMakeMem;
+    addonsMake >> addonsMakeMem;
+    
+    string line = addonsMakeMem.getFirstLine();
+    
 	while(!addonsMakeMem.isLastLine()){
-	    string line = addonsMakeMem.getNextLine();
-	    if(line[0] == '#') continue;
-		ofAddon addon;
-		cout << projectDir << endl;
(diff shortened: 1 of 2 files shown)


ofZach:
this also fixes an issue with LD_FLAGS -- when we add a library to a project, we wind up writing LD flags into the project setting.  This add OF_CORE_FRAMEWORKS to that (since our inserting of LD_FLAGS overwrites the configs), so that we are not getting linking errors...  

this is a very important fix to get OSX projects working on master....

arturoc:
there might be some small difference in how ofBuffer parsing by lines works now that has broken this. i'll take a look tomorrow, mostly to check that ofBuffer keeps working the same as before.

ofZach:
thanks - It felt like it was somehow missing the first line on getNextLine(), I rewrote this to be more explicit (and careful) but it would be good if the old code worked also.   What I had noticed were several examples that needed ofxGui not having them added which is why I digged into this. 

for sure the other change is a very needed change for OSX, some changes to the config get broken when we add libs, this helps make that not happen.

arturoc:
my last commit should fix the old way of using it but if you can check?

also now you can do:

```
ofFile addonsMake(ofFilePath::join(projectDir,"addons.make"));
ofBuffer addonsMakeMem;
addonsMake >> addonsMakeMem;
for(auto line: addonsMake.getLines()){
     if (!line.empty() && line[0] != '#') {        // line is not a comment
            ofAddon addon;
            addon.pathToOF = getOFRelPath(projectDir);
            addon.fromFS(ofFilePath::join(ofFilePath::join(getOFRoot(), "addons"), line),target);
            addAddon(addon);
      }
}
```

arturoc:
this was fixed by fixing ofBuffer to parse lines exactly as it did before

PR #3815, opened 2015-05-12
ios ofxOpenALSoundPlayer compiler bug fix.

julapy:


bilderbuchi:
Why remove the `const` from a getter method, this looks weird?

julapy:
have added some reasoning in #3789
not the best solution but it is the safest for a quick patch.

danoli3:
Sounds good for now, we can open an issue to fix the const ideas later.

We just can't compile without this fix, lets get this on master now.

danoli3:
I'll merge this into the iOS Changes 0.9 PR, as those compiler settings seems to be having the biggest issue with it and just continue from there.

julapy:
that sounds like a good plan.
i'll merge this and open up an issue to change the `isPlaying` back to `const` and update that function internally to work as a const function.

danoli3:
Nice!

PR #3826, opened 2015-05-14
OS X 32bit and 64bit libs from Apothecary formulas

ofTheo:


julapy:
since the header filed have changed for OSX, do these libs need to be recompiled for iOS as well?

ofTheo:
@julapy I think not. most of the .h files were either files that iOS didn't include ( like OpenCV/gpu ) or just line ending fixes. so I think iOS should be all good.

danoli3:
@julapy Yeah should be fine :+1:

PR #3835, opened 2015-05-15
Added Poco Libs for Linux64.

bakercp:


arturoc:
i had added them yesterday :) 

there's a problem where libraries compiled in something newer won't work properly with ubuntu 14.04 so probably we'll have to include a call from install_dependencies to apothecary update

arturoc:
btw, what have you used to compile these ones?

bakercp:
Ahh, strange.  When I pulled master there were still old libraries in there and some that were no longer included (MySQL ODBC, etc).  Also, when I tried to compile on linux with master, they were out of date ...  anyway ... strange.

I compiled these and the 32 bit on Ubuntu 14.04.  I'm compiling arm6/arm7 on rpi1 and rpi2.

arturoc:
oh ok, these work on 15.04 so it might be ok to leave them. 

mysql and odbc are not compiled any more in the apothecary formula

bakercp:
Also -- sorry for repeating your commit -- @ofTheo asked me submit linux libs, so I was going off that.

Yeah, mysql and odbc are no longer in apothecary, which is why i deleted the old static libs in this merge.

PR #3837, opened 2015-05-15
Small fixes

bakercp:
Warnings showing up during linux compile.


bakercp:
@arturoc thoughts here?

arturoc:
i think i added this because at some point that substraction can return a negative number and that will probably crash the app trying to access a wrong position in memory. perhaps we can just cast the other side to int64_t?

bakercp:
Since `nFrames` and `fromFrame` are both now unsigned, how about:

```
if ((fromFrame + nFrames) >= getNumFrames())
{
      nFramesToCopy = getNumFrames() - fromFrame;
}
```

bakercp:
ping @arturoc -- solution above work?  I'm trying to avoid casts ...

bakercp:
@admsyn does this look ok to you?

bakercp:
If so, I'll update the PR and merge.

bakercp:
@arturoc I implemented the above fix.  Let me know if it looks OK and I'll merge (or feel free to do it yourself).

arturoc:
yeah that looks correct, thanks!

bakercp:
+1.  Thanks @arturoc

PR #3838, opened 2015-05-15
ofMesh of index type fixes

bakercp:
These fixes make the use of index variables in ofMesh more consistent and also get rid of various type comparison warnings on linux.


arturoc:
thanks!

bakercp:
There is more coming ...

PR #3839, opened 2015-05-15
ofMesh of index type fixes

bakercp:
@arturoc here are the rest to remove the warnings.

I'm not exactly sure why we have ofIndexType (it resolves to a tess index which resolves to an unsigned int).  Would be better to use use `std::size_t` for indices throughout (don't have to check < 0, etc) and guaranteed to be correct bit-width for the platform.  Anyway ...

Check it!

### DIFF
--- a/libs/openFrameworks/3d/of3dPrimitives.cpp
+++ b/libs/openFrameworks/3d/of3dPrimitives.cpp
@@ -145,7 +145,7 @@ void of3dPrimitive::mapTexCoords( float u1, float v1, float u2, float v2 ) {
     //setTexCoords( u1, v1, u2, v2 );
     ofVec4f prevTcoord = getTexCoords();
     
-    for(int j = 0; j < getMesh().getNumTexCoords(); j++ ) {
+	for(std::size_t j = 0; j < getMesh().getNumTexCoords(); j++ ) {
         ofVec2f tcoord = getMesh().getTexCoord(j);
         tcoord.x = ofMap(tcoord.x, prevTcoord.x, prevTcoord.z, u1, u2);
         tcoord.y = ofMap(tcoord.y, prevTcoord.y, prevTcoord.w, v1, v2);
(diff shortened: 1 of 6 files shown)


arturoc:
we use ofIndexType cause GLES only supports unsigned short while GL supports unsigned int so we need a type that is the correct size for each platform

arturoc:
thanks!

bakercp:
Got it.  Thanks!

bilderbuchi:
> we use ofIndexType cause GLES only supports unsigned short while GL supports unsigned int so we need a type that is the correct size for each platform

but wouldn't std::size_t correctly reflect this for each platform?

arturoc:
no, siize_t might be different in different platforms according to the cpu or os architecture but this is a GPU limitation

bilderbuchi:
okay, but `TESSindex`, which is what ofindexType boils down to, is also [only switched according to platform, not GPU](https://github.com/openframeworks/openFrameworks/blob/ef44006cf219a1aed09e9891e3997a724386fd36/libs/tess2/include/tesselator.h#L123-L127)?

arturoc:
yeah sorry is not a limitation of the GPU itself but of the version of openGL used, openGL vs openGL ES

until now we know which version of openGL we are using from detecting the platform. we've been adding the different types of TESSindex as we added new platforms, it's not in the original code.

there is the posibility of runnning openGL ES on plain linux and some arm linux support plain openGL for example but until now we haven't had support for those

bilderbuchi:
so, just for clarity, wouldn't it  be better to do this selection right where ofindextype is defined, and not hidden somewhere in a manually patched Tess header?

bakercp:
+1 for more clarity around how / where / why ofIndexType is defined.

arturoc:
tess also needs the same type to work well with openGL ES so we rely on that. that way we also avoid that they could get out of sync. perhaps we can add some comments but the way it's done by typedefing one from another is correct in my opinion

PR #3841, opened 2015-05-17
Fix ubuntu install deps

jamet:
PR to fix install dependencies.sh problem described in this thread : http://forum.openframeworks.cc/t/install-dependencies-issues-with-current-github-master-and-ubunut-14-04/19620

tested only on two laptops!


arturoc:
thanks!

PR #3846, opened 2015-05-18
remove public vars from ofImage and ofVideo*

arturoc:


kylemcdonald:
awesome. looks good to me, glad to have `getImageType()` too.

PR #3856, opened 2015-05-20
Fix #3537

### LINKED ISSUE #3537: ofxOsc doesn't make use of OSC 1.1 types specifications
`/to/my/boolean/path T`
nor this one:
`/to/my/boolean/path F`

groolot:


ofTheo:
This looks pretty good to me. 
Thanks for working on all the 1.1 types @groolot 

ping @bilderbuchi @openframeworks/core @arturoc for a double check.

bilderbuchi:
Unfortunately, I've been out of ofxOsc too long to do a meaningful review of this.

bilderbuchi:
one thing: has this been tested? with our OSC examples, too?

arturoc:
the only thing i see that might break old code is how the type names have changed from int to i, int64 to h....

i guess most people would be using the enum so it should not be so problematic.

groolot:
@arturoc Indeed, I arbitrary change the string value returned according to what is most relevant to return while waiting for a representation of the type.
I based on the OSC 1.1. specifications to set correctly the values.
It may be preferable to return directly the OFX_OSC_TYPE from the enum, but couldn't be used with this kind of enum: should use a ``std::map``

groolot:
@bilderbuchi : I tested this code with the examples shipped with 0.8.4 = works well
because of no test suite, everything is done manually : not perfect.

BTW, I use this code since 3 months in a project (see https://github.com/PMO-Ecluse/mallarmeGenerateur3D) and it work great without any issue.
It is also connected with the ``ofxOscRouter`` addons by @bakercp and works like a charm :)

ofTheo:
great - thanks for the look over. 
I think it would be fine to get this merged then - @groolot you'll need to pull in master to get the PR to merge though.

bilderbuchi:
Closing this as @groolot has opened another PR for this.

PR #3857, opened 2015-05-20
Fix #3537 , PR within master

### LINKED ISSUE #3537: ofxOsc doesn't make use of OSC 1.1 types specifications

groolot:
as @ofTheo advise me, this is the pull-request based on the master branch.
This bugfixes #3537 which is about OSC 1.1 specifications.

### DIFF
--- a/addons/ofxOsc/src/ofxOscSender.cpp
+++ b/addons/ofxOsc/src/ofxOscSender.cpp
@@ -147,8 +147,10 @@ void ofxOscSender::appendParameter( ofxOscMessage & msg, const ofAbstractParamet
 		msg.addIntArg(parameter.cast<int>());
 	}else if(parameter.type()==typeid(ofParameter<float>).name()){
 		msg.addFloatArg(parameter.cast<float>());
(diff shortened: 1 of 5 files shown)


addons/ofxOsc/src/ofxOscSender.cpp
```diff
@@ -177,11 +179,25 @@ void ofxOscSender::appendMessage( ofxOscMessage& message, osc::OutboundPacketStr
 		if ( message.getArgType(i) == OFXOSC_TYPE_INT32 )
 			p << message.getArgAsInt32( i );
 		else if ( message.getArgType(i) == OFXOSC_TYPE_INT64 )
-			p << (osc::int64)message.getArgAsInt64( i );
```
groolot: I'm not sure about that.
Maybe an error to not cast :/

ofTheo:
great - I'm going to merge this if there are no objections!

PR #3858, opened 2015-05-20
fix ofVbo copy constructor

galsasson:
This patch fixes ofVbo copy and assignment constructors when not using the programmable pipeline.

Without it vaoID will be garbage and 'clear' tries to release it.


arturoc:
thanks!

Issue #3859, opened 2015-05-21
ofxOSC addon updated today - error "Use of overloaded operator '<<' is ambiguous"


dimitre:
just dragging the addon files to my project returns two errors in ofxOscSender.cpp file
In the following lines:
```c++
		else if ( message.getArgType(i) == OFXOSC_TYPE_INT64 )
			p << message.getArgAsInt64( i );

		else if ( message.getArgType( i ) == OFXOSC_TYPE_TIMETAG )
			p << message.getArgAsTimetag( i );
```

ofTheo:
what OS are you on @dimitre ? 

pinging @groolot

dimitre:
OS X 10.10, Xcode 6.3.2 @ofTheo

ofTheo:
The #3859 PR fixes it for me. 
@groolot could this present any issues?

armadillu:
fixes it for me too on Xcode 6.3.1. 10.10 sdk

groolot:
I've not tied it, but it seems to be pretty better. This is coherent, but, finally, this should not be good for people using the shipped ospack dev library (like in Debian as me).
But there was not any issue with the ospack on Debian 64bits machine.
I think that #3861 is fine.

ofTheo:
great - thanks all!

groolot:
An alternative is to come back to the @arturoc's solution from https://github.com/openframeworks/openFrameworks/blame/58682920929899135497313d10715ce53fd7cd84/addons/ofxOsc/src/ofxOscSender.cpp#L180
and cast the returniing value of the message like this:
```
else if ( message.getArgType(i) == OFXOSC_TYPE_INT64 )
            p << (osc::int64)message.getArgAsInt64( i );
```

I think the @arturoc solution is better than modifiing the oscpack source code, due to the coherence broken with shipped liboscpack package on Debian (for example).

bilderbuchi:
@groolot @ofTheo I have to say I'd also prefer to fix this on our side instead of patching oscpack itself, to keep us as close as possible to vanilla oscpack.

arturoc:
@groolot can you send a new PR with that commit and reverting the changes to oscpack? 

this seems to be a bug in oscpack, i've submitted an issue to the original repo: https://code.google.com/p/oscpack/issues/detail?id=21

ofTheo:
@groolot I also had to do the cast for the timetag

see the change in the last file of this commit:
[...]

PR #3861, opened 2015-05-21
osc int64 bugfix for #3859

ofTheo:


groolot:
this probably why @arturoc staticaly casted to ``osc::int64`` the returned value of the message.

PR #3862, opened 2015-05-21
Fix issue 3859

groolot:


groolot:
Should revert ff6c6a4dd932f74f92cb4fadf91255e2aa5cf237 then.

Sorry for the mistake.

PR #3863, opened 2015-05-22
Feature Request: Exposing Cairo context and surface internal variables for addons

procedural:
I'm making an addon for Cairo which allows to override window's transparency and draw vector shapes more easily,  to make this happen and use `ofDraw...` commands I need to assign different surfaces and cairo contexts to internal variables which are currently not exposed.

To make addon work, either I need to completely duplicate functionality of `ofCairoRenderer` or with only two functions I propose here draw shapes directly with openFrameworks!

These two functions don't break anything, they're not even exposed to anyone except for people who can create a custom pointer to a renderer, i.e. advanced users.

Here's what will be possible :)

### DIFF
--- a/libs/openFrameworks/graphics/ofCairoRenderer.cpp
+++ b/libs/openFrameworks/graphics/ofCairoRenderer.cpp
@@ -1375,8 +1375,16 @@ cairo_t * ofCairoRenderer::getCairoContext(){
 	return cr;
 }
 
+void ofCairoRenderer::setCairoContext(cairo_t * _cr){
+  cr = _cr;
+}
+
 cairo_surface_t * ofCairoRenderer::getCairoSurface(){
 	return surface;
+}
+
+void ofCairoRenderer::setCairoSurface(cairo_surface_t * _surface){
+  surface = _surface;
 }
 
 ofPixels & ofCairoRenderer::getImageSurfacePixels(){
--- a/libs/openFrameworks/graphics/ofCairoRenderer.h
+++ b/libs/openFrameworks/graphics/ofCairoRenderer.h
@@ -157,7 +157,9 @@ public:
 
 	// cairo specifics
 	cairo_t * getCairoContext();
+	void setCairoContext(cairo_t * cr);
 	cairo_surface_t * getCairoSurface();
+	void setCairoSurface(cairo_surface_t * surface);
 	ofPixels & getImageSurfacePixels();
 	ofBuffer & getContentBuffer();
 


bakercp:
+1 from me.

arturoc:
mmh, it's kind of strange to set the internals like that, what options do you need that are not there yet? we could have different setup methods that allow other settings but passing the surface like that seems like breaking encapsulation a lot.

procedural:
> what options do you need that are not there yet?

I need to switch between multiple Cairo contexts at runtime :) I've uploaded the addon here:

https://github.com/ofnode/ofxCairoX11

[This line](https://github.com/ofnode/ofxCairoX11/blob/3299b4f/src/ofxCairoX11.cpp#L83), for example, let me switch to the context that overrides window's shape, when I finish shaping it I can switch back to [the context for drawing](https://github.com/ofnode/ofxCairoX11/blob/3299b4f/src/ofxCairoX11.cpp#L91) and draw shapes as usual. Here's the example of such code:

```cpp
  cairo.begin();

    cairo.shapeBegin(); // <-- Switching to the context for drawing window shape

      if (cairo.resized) {
        cairo.shapeModeAdd();
        ofDrawRectangle(0, 0, ofGetWidth(), ofGetHeight());
      }

      cairo.shapeModeSub();
      if (MousePressed && ofGetMouseX() != 0 && ofGetMouseY() != 0)
        ofDrawCircle(ofGetMouseX(), ofGetMouseY(), ofRandom(BrushRadius));

    cairo.shapeEnd(); // <-- Switching back to the context for normal color drawing

    ofBackground(200);

    ofSetColor(ofColor::lightYellow);
    ofDrawCircle(ofGetMouseX(), ofGetMouseY(), BrushRadius * 2.0f);

  cairo.end();
```

arturoc:
i think we could have something like `setupForXWindows` or `setupWith(cairo_context, cairo_surface)` then you can have 2 ofCairoRenderer in your application and switch between them. the main difference is that this way is a setup call which sets the initial parameters to work with an external context and not rely on any initial settings. also that the context won't change during the whole life of the renderer which makes it easier to maintain.

arturoc:
btw, in that example why would you want to use X and not just an image surface which is faster? it would also be multiplatform instead of linux only

procedural:
> i think we could have something like setupForXWindows or setupWith(cairo_context, cairo_surface) then you can have 2 ofCairoRenderer in your application and switch between them

This would be great too! This PR is just a proposal. If there's a better way to do this -- I'm all in for this!

> btw, in that example why would you want to use X and not just an image surface which is faster? it would also be multiplatform instead of linux only

Because I'm dumb! I mean I only vaguely know some X11 commands to make them do what I want, if there's other ways to do the same and in multiplatform way I'm open for alternatives.

But I don't know multiplatform way to make window's shape different, in that addon I use `XShapeCombineMask` which uses X11 Shape extension and as far as I know there's no other way to do that for Windows and OS X in Cairo?

procedural:
> why would you want to use X and not just an image surface which is faster?

Apparently drawing an image surface is actually much slower, because you have to call `swapRgb` to swap R and B channels on each frame, which is crazy expensive:

```bash
CPU: AMD64 family15h, speed 2.1e+06 MHz (estimated)
Counted CPU_CLK_UNHALTED events (CPU Clocks not Halted) with a unit mask of 0x00 (No unit mask) count 50000
samples  %        linenr info                 symbol name
155087   26.5821  algorithmfwd.h:171          void std::swap<unsigned char>(unsigned char&, unsigned char&)
97903    16.7807  move.h:102                  std::remove_reference<unsigned char&>::type&& std::move<unsigned char&>(unsigned char&)
82759    14.1850  ofPixels.h:735              ofPixels_<unsigned char>::Pixel::operator[](int)
70461    12.0771  ofPixels.h:416              ofPixels_<unsigned char>::swapRgb()
67328    11.5401  ofPixels.h:681              ofPixels_<unsigned char>::Pixel::operator++(int)
52828     9.0548  ofPixels.h:674              ofPixels_<unsigned char>::Pixel::operator++()
45623     7.8198  ofPixels.h:723              ofPixels_<unsigned char>::Pixel::operator!=(ofPixels_<unsigned char>::Pixel const&) const
861       0.1476  ofPixels.h:230              ofPixels_<unsigned char>::set(unsigned char)
826       0.1416  glew.c:230                  _glewStrCLen
508       0.0871  ofVec4f.h:46                ofVec4f::operator[](int) const
398       0.0682  ofMatrix4x4.cpp:439         ofMatrix4x4::makeFromMultiplicationOf(ofMatrix4x4 const&, ofMatrix4x4 const&)
173       0.0297  ofVec4f.h:42                ofVec4f::operator[](int)
122       0.0209  ofVec4f.h:38                ofVec4f::getPtr() const
```

Maybe there's other way to draw image with Cairo without swapping, but I didn't found one so far.

procedural:
Here's the solution: https://github.com/openframeworks/openFrameworks/pull/3867

bilderbuchi:
closed by  70ab62c I think.

procedural:
Well, no, this PR is about exposing context and surface variables really... Oh well... :)

bilderbuchi:
I thou

procedural:
@tgfrerer this PR will be closed eventually anyway, so you could keep it close... :)

I really wish there was some kind of `setupWith(cairo_context, cairo_surface)` command... :children_crossing:

procedural:
Will close this for now with this wall of hope:

@arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc @arturoc

PR #3867, opened 2015-05-25
Adds GL_BGRA format

procedural:
Makes possible to draw Cairo or other BGRA swizzled images without calling super expensive `swapRgb` function.

### DIFF
--- a/libs/openFrameworks/gl/ofGLUtils.cpp
+++ b/libs/openFrameworks/gl/ofGLUtils.cpp
@@ -177,6 +177,8 @@ int ofGetGLFormatFromInternal(int glInternalFormat){
 #endif
 			case GL_ALPHA:
 				return GL_ALPHA;
+			case GL_BGRA:
+				return GL_BGRA;
 
 			default:
 				ofLogError("ofGLUtils") << "ofGetGLFormatFromInternal(): unknown internal format " << glInternalFormat << ", returning GL_RGBA";
@@ -253,6 +255,8 @@ int ofGetGlTypeFromInternal(int glInternalFormat){
 
 		case GL_STENCIL_INDEX:
 			return GL_UNSIGNED_BYTE;
+		case GL_BGRA:
+			return GL_BGRA;
 
 		default:
 			ofLogError("ofGLUtils") << "ofGetGlTypeFromInternal(): unknown internal format " << glInternalFormat << ", returning GL_UNSIGNED_BYTE";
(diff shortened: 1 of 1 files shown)


tgfrerer:
+1 =)

libs/openFrameworks/gl/ofGLUtils.cpp
```diff
@@ -253,6 +255,8 @@ int ofGetGlTypeFromInternal(int glInternalFormat){
 
 		case GL_STENCIL_INDEX:
 			return GL_UNSIGNED_BYTE;
+		case GL_BGRA:
```
tgfrerer: are you sure that should be `GL_BGRA`?  i believe it should say GL_UNSIGNED_BYTE, in which case you could add it  above next to GL_RGB

libs/openFrameworks/gl/ofGLUtils.cpp
```diff
@@ -253,6 +255,8 @@ int ofGetGlTypeFromInternal(int glInternalFormat){
...
+			return GL_BGRA;
```
tgfrerer: Are you sure that should say `GL_BGRA`? i believe it should say `GL_UNSIGNED_BYTE`, in which case you could add it above next to `GL_RGB` etc.

procedural: You're right. Fixed!

arturoc:
The internal format of a texture can't be GL_BGRA only GL_RGBA: https://www.opengl.org/sdk/docs/man/html/glTexImage2D.xhtml. 

Then you upload using loadData(data,GL_BGRA)  with the new ofPixels based on pixel format instead of number of channels this should be happening automatically already just by allocating the pixels as OF_PIXELS_BGRA

arturoc:
this: https://github.com/openframeworks/openFrameworks/commit/70ab62ca9cce49b0867ae6b4db7d89702c879979 should fix it, now you can do `tex.loadData(cairo.getImageSurfacePixels())` without specifying the format at all since the pixels already have the correct format. also when saving it'll avoid swapping the pixels twice as i think was doing before.

procedural:
>> Then you upload using loadData(data,GL_BGRA) with the new ofPixels based on pixel format instead of number of channels this should be happening automatically already just by allocating the pixels as OF_PIXELS_BGRA

Ah, you're right! The trick was to allocate ofTexture *as well*:

```cpp
  texture.allocate(w, h, GL_RGBA);
  buffer.allocate(w, h, OF_PIXELS_BGRA);
```
Don't know why, but explicitly allocating it as `GL_RGBA` fixes the render.

arturoc:
yes the autoallocate in loadData is kind of handy for very basic things but as soon as you want to do something slightly different is not very reliable. it passes the format as internal which in lots of cases is not correct. i added a TODO some days ago to fix it by adding a function that guesses the internal from the format

PR #3873, opened 2015-05-27
Fixes #3872.

### LINKED ISSUE #3872: ofPixels: pixelsSize variable not set after rotate / resize, resulting in null ptr.
New cached `pixelsSize` variable is not set to new count after resize / rotate operations.

It shouldn't be required for rotate operations (the total number of pixels remains the same), but it should be reset after resize operations.

e.g.

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofPixels.cpp#L1138-L1141

If not set, any iterations (e.g. swapRGB during image save fails with null ptr).

bakercp:
See #3872


bakercp:
ping @arturoc

bakercp:
@arturoc could you do a quick review and see if I missed any?  It doesn't apply to operations that use `swap` and `==` as you got the variables covered in the `swap` definition.

arturoc:
looks correct to me, thanks!

PR #3876, opened 2015-05-29
Vs2015 scripts

DomAmato:
updates for the upcoming vs2015 with 32 and 64 bit options


scripts/apothecary/formulas/FreeImage/FreeImage.sh
```diff
@@ -10,20 +10,20 @@
 FORMULA_TYPES=( "osx" "vs" "win_cb" "ios" "android" )
 
 # define the version
-VER=3160 # 3.16.0
+VER=3170 # 3.16.0
 
 # tools for git use
 GIT_URL=https://github.com/danoli3/FreeImage
-GIT_TAG=3.16.0
+GIT_TAG=3.17.0
```
danoli3: I guess this means we'll make the switch to 3.17.0 @bakercp @arturoc @ofTheo 

I can do the changes for OS X / iOS formula / repo. if this is needed for VS.

ofTheo:
This looks good to me @DomAmato !

scripts/apothecary/formulas/openSSL/buildwin.cmd
```diff
@@ -0,0 +1,46 @@
+@echo off
```
LeoColomb: For consistency, I would prefer, if possible, a shell file instead. What do you think?

And can't we use lib from official OpenSSL distribution?

DomAmato: It would probably have to work like the way fmod is handled since the windows binaries come as exe files so the compiled libs would have to be copied over from the install. 

It could be done in a shell script but that would still require using cmd //c "call etc... and I was having issues calling nmake because it is not on a default path and there were issues with having a space in the URL namely the Program Files folder. The batch file was just cleaner and less hassle plus there was an issue with which version of Perl was being used since I had installed active perl to the system path but git bash was using its own version

LeoColomb: Yep, I see what you say about cmd just after my comment, and you're totally right.

About binaries, I've seen on OpenSSL site another binary compilation site (https://www.openssl.org/about/binaries.html):

> http://indy.fulgan.com/SSL/
   Pre-compiled Win32/64 libraries without external dependencies to the Microsoft Visual Studio Runtime DLLs, expect for the system provided msvcrt.dll

But it seems down currently...

DomAmato:
I am going to create separate pull requests for the script changes and the changes to the libraries just so that its easier to look through the diffs

PR #3879, opened 2015-05-31
Fixing Fontconfig memory leaks by destroying allocated FcPatterns

procedural:
Clang's Leak Sanitizer found these


arturoc:
thanks!

PR #3880, opened 2015-05-31
Closes #3878

### LINKED ISSUE #3878: Memory from FcInit() is not freed causing tiny memory leak
[FcInit()](https://github.com/openframeworks/openFrameworks/blob/df80bbd/libs/openFrameworks/graphics/ofTrueTypeFont.cpp#L332) allocates 8 kilobytes of heap memory which is not freed anywhere with [FcFini()](http://www.freedesktop.org/software/fontconfig/fontconfig-devel/fcfini.html). 

At first I wanted to add it to ofTrueTypeFont destructor, but since the library initialization is a global thing that not tied to any particular font instance, I'm thinking it would be better to add it to a function which will be called at exit of openFrameworks.

As a workaround, one can free the memory manually like so:

```cpp
#include "fontconfig/fontconfig.h"

void ofApp::exit()
{
  FcFini();
}
```

procedural:


arturoc:
thanks!

libs/openFrameworks/app/ofAppRunner.cpp
```diff
@@ -193,6 +193,10 @@ void ofExitCallback(){
 		timeEndPeriod(1);
 	#endif
 
+	//------------------------
+	// try to close font libraries
```
bilderbuchi: "try"? can this fail, and if so, what happens?

procedural: Nah, I just copied and pasted what other comments said :D Feel free to change any of my pull requests after their merge!

PR #3888, opened 2015-06-02
change valueChanged so that the sender is included

patrickFuerst:
add the ability to get access to the ofxButton in the ListenerMethod . 
Tested on Windows 8.


arturoc:
thanks!

PR #3889, opened 2015-06-02
Avoid Warning/Exception when Android video is playing and app is paused

pelayomendez:
…and resumed

By doing using reset() before releasing the media player we avoid the "mediaplayer went away with unhandled events" warning:

mediaPlayer.reset();
mediaPlayer.release();

Also by reseting the mediaplayer after creating it we avoid the internal/external state mismatch corrected exception when pausing resuming the app. Doing that forces us also to keep track of loop state when resuming the video.

06-02 18:05:37.195: E/MediaPlayer(30590): internal/external state mismatch corrected


arturoc:
thanks!

PR #3895, opened 2015-06-06
VS2015 32/64 Libs for 0.9.0

DomAmato:
These are the updated files from the apothecary script updates


ofTheo:
I think this PR needs to pull in master - its saying it can't be merged. 
As soon as it can be merged I'll merge it in. 

Thanks!!
Theo

procedural:
Why this PR duplicate tess2 source files?

https://github.com/openframeworks/openFrameworks/tree/master/libs/tess2/Sources/Source

arturoc:
it's clearly an error, we don't need the sources anymore. there's a really annoying difference in the way osx treats the leading / when using cp that makes it incompatible with other platforms, probably this is due to something like that.

i'm going to delete them and fix the script

PR #3897, opened 2015-06-07
VS2015 32/64 Scripts for 0.9.0

DomAmato:
These are just the scripting updates

### DIFF
--- a/scripts/apothecary/apothecary
+++ b/scripts/apothecary/apothecary
@@ -76,7 +76,8 @@ IOS_SDK_VER=8.1
 IOS_MIN_SDK_VER=5.1
 
 # used when building for vs
-VS_VER=11
+VS_VER=14
+VS_64_BIT_ENV='VC\bin\x86_amd64\vcvarsx86_amd64.bat'
 
 # paths to android SDK, etc
 ANDROID_NDK_ROOT=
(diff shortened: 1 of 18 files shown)


scripts/apothecary/apothecary
```diff
@@ -583,7 +584,43 @@ function downloadFormula(){
...
+			else
+				download
+			fi
+			if [ ! -e $1 ] ; then
+				echoError " It looks like downloading failed for \"$1\""
+				exit 1
+			fi
+		else
+			echo "... skipping, src dir already exists"
+		fi
```
arturoc: i think this should go in each formula, also in the case of cmake builds you can just create build32 and build64 and `cmake ..` instead of `cmake .`

arturoc:
looks great otherwise thanks so much for looking into this!

scripts/apothecary/apothecary
DomAmato: right this was only an issue in cmake builds. What do you think would be the implementation for each formula? like it does the check in each download step? not opposed to that

arturoc: yes i think in general it would be a good practice to delete and uncompress or clone everything every time. if we patch things for one platform or there are any thing remaining from a previous build it could be problematic.

also with git you do `git clone -depth=1 ...` and it'll only get the files as they are on the last commit instead of the whole history

we could add some flag so it doesn't delete the previous download if you know what you are doing but by now i would just move this code to the respective formulas

ofTheo:
maybe we could merge this in and then fix in another PR?
I'm anxious to get the VS stuff in so master is not broken for VS users.

scripts/apothecary/formulas/FreeImage/FreeImage.sh
```diff
@@ -10,20 +10,20 @@
 FORMULA_TYPES=( "osx" "vs" "win_cb" "ios" "android" )
 
 # define the version
-VER=3160 # 3.16.0
+VER=3170 # 3.16.0
```
bakercp: Are these updates required?  We already updated a bunch of the static libs and I'm a little worried about not having matching for all platforms.

Also, the comment 3.16.0 should be fixed if we stay with this.

scripts/apothecary/formulas/_depends/pixman.sh
```diff
@@ -4,7 +4,7 @@
 # http://pixman.org/
 
 # define the version
-VER=0.30.0
+VER=0.32.4
```
bakercp: Same issue here with the version bump.

scripts/apothecary/formulas/cairo.sh
```diff
@@ -30,40 +30,81 @@ GIT_TAG=$VER
 # download the source code and unpack it into LIB_NAME
 function download() {
 	curl -LO http://cairographics.org/releases/cairo-$VER.tar.xz
-	tar -xf cairo-$VER.tar.xz
-	mv cairo-$VER cairo
-	rm cairo-$VER.tar.xz
-
+	if [ "$TYPE" == "vs" ] ; then
+		7z e cairo-$VER.tar.xz
```
bakercp: On my system, using ofMingGW, 7z command line tool is not available by default.  I was able to get it to compile just fine without this custom vs section.  Is there a reason why we need a custom 7z section?

bakercp: Ah, strike that, I installed xz.exe during my research and tar xf must have been calling that.  I'll send a PR with an update.

arturoc:
yeah we should have the same versions across platforms, at least those for which the headers are directly use, so pixman should be ok but FreeImage should be recompiled

scripts/apothecary/formulas/FreeImage/FreeImage.sh
DomAmato: The reason for the update is there was no 64 bit binaries in 3.16.0

bakercp: Good reason. We should get the other platformes compiled and up to date once this is merged.

scripts/apothecary/formulas/cairo.sh
DomAmato: right I couldn't get the tar that is packaged with git bash to unpack the xz files and had 7zip on my computer since its my goto for compression/decompression. Hopefully that wont be a problem in the future

bakercp: Here's the PR with my xz fix https://github.com/DomAmato/openFrameworks/pull/1

arturoc:
i just merged this changes + done the proposed fixes but for some reason it won't so as merged so i'm closing it

PR #3912, opened 2015-06-13
remove assignment in declaration in ofNode

joshuajnoble:
I think this improves the readability and, of more interest to me, makes generating SWIG bindings for Python and Lua possible again.


ofTheo:
don't see any reason not to merge this. 
thanks @joshuajnoble

PR #3913, opened 2015-06-13
make osx 32bit c++11 compatible and begin to remove Poco

arturoc:
- Use libc++ in osx32 which brings compatibility with c++11 to the only platform was holding us back from using c++11 feature in the core.
- Removes dependencies with Poco from most of the core except for ofURLFileLoader (which already had a similar mechanism to the video and sound players, grabbers... allowing to have different implementations for different platforms) and ofXml
- Adds 3 new libraries to do what poco was doing:
   - boost::filesystem, which will be part of c++17 and it's already part of the std in visual studio. in the future we might want to have better integration with this since is going to be in the std but by now if provides the exact same api we had
   - cppuri https://github.com/cpp-netlib/uri for uri parsing, is another library which should become part of c++'s std
   - utf8cpp: http://utfcpp.sourceforge.net/  is a header only library mostly used to keep providing utf8 ofToLower and ofToUpper but also replaces the hack we where using in ofTrueTypeFont to parse utf8 strings. in the future it'll make really easy to have full support for utf8 ttf.

This PR also makes emscripten work again but now without poco, it doens't have support for ofXml but it's something that will be fixed once we port it to pugixml. 

It also brings android uptodate

This is tested on all the platforms except for ios and mingw which i'll do next if we decide to merge it before 0.9. It also doesn't break any old code except anything using ofFile::getPocoFile, or addons that had c++ binary libraries for osx which used the std (so c only libraries or source/header only libraries will be ok)


bilderbuchi:
Huge effort, but this is a huge PR whose diff doesn't even display on the GH webpage, so it's more or less unreviewable. :-(
Also, I am a bit puzzled why you put 4 different topics into one huge PR, instead of submitting several ones? libc++ on osx32, poco removal, emscriptem compat, and android compat should be 4 different PRs, shouldn't they?
for what it's worth, I tend to be against removing poco in 0.9.0 already (it's much too big a change so late in the milestone progress), so it's a bummer that the osx32 part is included here.

arturoc:
this doesn't remove poco, it's still there, just that the core is not using it in most parts. things are pretty easily divisible. c++11 support for osx32 just needs the recompiled libraries and apothecary scripts. 

in any case my goal was to get emscripten working again without patching tons of files in poco which meant i needed to use the equivalents in c++11 which meant i needed to have all platforms supporting it. that's why everything is together in one big PR.

the first 20 commits or so are the real changes to source code and have pretty explicative names. everything from there are mainly fixes, library binaries and apothecary scripts

arturoc:
btw, if we don't merge this then we need to go back to the old version of poco which means that the new version of ofSketch can't be released or drop the emscripten port which had been working for about a year and was broken after the deadline for 0.9.

bilderbuchi:
have we figured yet what exactly broke emscripten?

arturoc:
changing the version of poco

bilderbuchi:
why? was it using a deprecated poco api or something?

arturoc:
the minimal installation of poco includes a lot of stuff that is not compatible with emscripten and other platforms.

i originally patched poco to remove the things that are not compatible with emscripten (as i did at some point with android) but those patches won't work with the new version because the code has changed. i've done that too many times already and other platforms like windows RT also require that kind of patching of lots of different files of the poco source. 

if we've agreed that moving forward the correct option is to deprecate poco or at least not have dependencies with it in the core, it's better to start going in that direction rather than patching poco once again.

bakercp:
FWIW, I'm w/ @arturoc on this movement.  The promise of emscripten and better Android support is valuable -- even from the perspective of current ofSketch development.  I'm going to spend some time testing this PR and will try to report back in a couple days.

bakercp:
Also, IMHO, going back to a older (patched) version in Poco isn't an option so this seems like the way forward.

bakercp:
Also utf8cpp is what I use in [ofxUnicode](https://github.com/bakercp/ofxunicode) and some of the initial ofFont work.

arturoc:
btw the only difference in android is that while compiling the new libraries i realized that there's a new NDK. this fixes a couple of undefined symbols when compiling with that but there's no changes to the android code itself.

bilderbuchi:
ok, fair enough. I'm just concerned that, with such a big change so late in the release process, we will either delay the release even further, or introduce code which was not tested very thoroughly.

arturoc:
i know, i've done unit tests for most of what i've changed i can push them. there'll be for sure some edge case that has broken but i really think it's the best solution

tgfrerer:
When i pull this and try to compile / make the project generators (both simple and devApps) on OS X, I get:

`In file included from /Users/tim/dev/of/apps/projectGenerator/projectGeneratorSimple/src/addons/ofAddon.cpp:9:
/Users/tim/dev/of/apps/projectGenerator/projectGeneratorSimple/src/pugixmlLib/src/pugixml.hpp:97:15: error: no type named 'basic_string' in namespace 'std'; did you mean 'basic_stringbuf'?
`

Which seems really weird, because `basic_string` is defined in `std::string`.

Without the project generators, i have no way of testing or generating osx projects. 

My current debugging hypothesis is that some header files are getting shadowed / or maybe the include order got jumbled up somewhere?

bilderbuchi:
ping @arturoc regarding tim's comment.

arturoc:
i'm looking into it

arturoc:
i've managed to patch it but it's not definitive, it'll compile but i've had to disable the std extensions in pugixml. i'll look into it more later but my feeling is that our way of importing globally all of the std is clashing with a the new libraries + pugixml

tgfrerer:
now the error message has moved on to boost:

`
In file included from /Users/tim/dev/of/apps/projectGenerator/projectGeneratorSimple/src/addons/ofAddon.cpp:10:
In file included from ../../../libs/openFrameworks/utils/ofUtils.h:6:
In file included from ../../../libs/openFrameworks/utils/ofLog.h:4:
In file included from ../../../libs/openFrameworks/utils/ofFileUtils.h:11:
In file included from ../../../libs/boost/include/boost/filesystem.hpp:16:
In file included from ../../../libs/boost/include/boost/filesystem/path.hpp:25:
In file included from ../../../libs/boost/include/boost/filesystem/path_traits.hpp:23:
../../../libs/boost/include/boost/system/error_code.hpp:464:17: error: no type named 'basic_ostream' in namespace 'std'
    inline std::basic_ostream<charT,traits>&
           ~~~~~^
../../../libs/boost/include/boost/system/error_code.hpp:464:30: error: expected unqualified-id
    inline std::basic_ostream<charT,traits>&
                             ^
`

agree it's probably a namespace clash.

danoli3:
@tgfrerer Sounds like you're linking against a libstdc++ boost rather than a libc++ boost.

arturoc:
does last commit fix it for you?

tgfrerer:
hmmm. it compiles successfully, throws a ton of compiler warnings, but compiles.

When i run it, it crashes (and with quite some momentum - it takes XCode down, too.), so I can't exactly say where.

arturoc:
mmh, it's working for me fine. have you tried to recompile it?

arturoc:
i mean clean it?

tgfrerer:
yep. no dice. the 32 bit target starts and kills xcode, the 64 bit target won't link, saying that openframeworks lib is only 32 bit. strange...

arturoc:
mmh, it's working for me fine. i've just restarted in windows to check the other problem but will try in a bit

tgfrerer:
sure thing. i'm on windows on my other machine and i think i have a handle on the other problem, it's related to vs `std::filesystem::canonical` not returning the same as in boost. will send a pr in a min...

armadillu:
Is dropping POCO on the near roadmap? Hmmm I wish I knew that earlier, I have quite a lot of code that relies on POCO calls...

bakercp:
@armadillu dropping Poco from the core won't happen in the next release but will likely be spun out as an addon (perhaps a core addon) for 1.0.  I too have _LOTS_ of investment in Poco, but it was making things difficult for supporting new platforms (like Emscripten, Android, etc).  So anyway, many of my network addons that leverage Poco heavily (ofxHTTP for instance) weren't really ever going to support Emscripten for various reasons, so there will still be heavy Poco users / support as we move forward (e.g. Me :)).

armadillu:
@bakercp And me :)

I just came back from a 2 week holiday, and I suddenly see Boost is in and Poco is out :dizzy_face: So I got a bit alarmed for a sec... Until I understood what's going on :)

bakercp:
@armadillu Glad to hear it!  Hope your vacation was great.  

If you're wanting a little more background about the discussion, there is more in the of-dev mailing list logs.  @arturoc sent a proposal on 26 May with the subject "remove poco" if you're searching for it.

mrbichel:
Is there a recommended way to fix addons using getPocoFile() so they are compatible?

arturoc:
@bilderbuchi i fixed this a couple of days ago. 

@mrbichel i have just added a cast operator so you can use both ofFile and ofDirectory as if they where std paths when passing them to functions or casting them.

the reference is here: http://www.boost.org/doc/libs/1_58_0/libs/filesystem/doc/reference.html

so now you can do:

``` 
ofFile f("somefile");
std::filesystem::is_symlink(f)
```
or

```
std::filesystem::path p = f;
```

PR #3914, opened 2015-06-13
Fix customEGLWindowSettings

mruegenberg:
When trying to compile all examples, I ran into some compilation errors in customEGLWindowSettings, which I assume would also occur for users on a system where compiling this sample actually makes sense (i.e ARM).


libs/openFrameworks/app/ofAppEGLWindow.h
```diff
@@ -127,6 +129,7 @@ class ofAppEGLWindow : public ofAppBaseGLESWindow, public ofThread {
 	virtual void	setVerticalSync(bool enabled);
 	
 	struct Settings: public ofGLESWindowSettings {
+        public:
```
arturoc: this is not needed no? being a struct everything is public by default

mruegenberg:
I thought so, too, but my GCC 4.9.2 kept complaining about not finding the various struct fields.
Probably a compiler bug, or some weird detail wrt structs within classes. 

It shouldn't hurt to keep it in, hopefully.

arturoc:
ok, that's weird. thanks!

PR #3919, opened 2015-06-15
VS exports for 32/64 bit

DomAmato:
Need updated dlls for 32 and 64 bit


arturoc:
oh, thanks, i changed the osx ones but didn't realize about windows. we should remove the export folder and just pick this dll's from the libs folders at some point

bilderbuchi:
> we should remove the export folder and just pick this dll's from the libs folders at some point

could you file an issue about that, please?

PR #3920, opened 2015-06-15
Minor Script Updates

DomAmato:


danoli3:
Nice! 
Yeah OpenSSL 1.0.2c (and b) came out in the last couple of days. Lets get this script in and rebuild those libs for the release.

scripts/apothecary/formulas/cairo.sh
```diff
@@ -57,8 +70,6 @@ function prepare() {
 		cp -Rv ../pixman/* pixman
 		cp -Rv ../zlib/* zlib
 		
-		cp -v ../libpng/projects/visualc71/Win32_LIB_Release/*.lib libs
-		cp -v ../libpng/projects/visualc71/Win32_LIB_Release/ZLib/*.lib libs
 		if [ $ARCH == 32 ] ; then
```
arturoc: if this is removed...

scripts/apothecary/formulas/cairo.sh
```diff
@@ -147,13 +158,13 @@ function copy() {
...
-			cp -v Cairo-VS/projects/Release/libpng.lib $1/lib/$TYPE/Win32/libpng.lib
+			cp -v Cairo-VS/libs/libpng.lib $1/lib/$TYPE/Win32
 		elif [ $ARCH == 64 ] ; then
 			# make the libs path 
 			mkdir -p $1/lib/$TYPE/x64
 			cp -v Cairo-VS/projects/x64/Release/cairo.lib $1/lib/$TYPE/x64/cairo-static.lib
 			cp -v Cairo-VS/projects/x64/Release/pixman.lib $1/lib/$TYPE/x64/pixman-1.lib
-			cp -v Cairo-VS/projects/x64/Release/libpng.lib $1/lib/$TYPE/Win32/libpng.lib
+			cp -v Cairo-VS/libs/libpng.lib $1/lib/$TYPE/x64
 		fi
```
arturoc: shouldn't this stay as it was?

DomAmato: I changed VS-Cairo to compile libpng and zlib because the project included with libpng only compiled 32bit so the folder structure changed a little as the build step moves the libraries into the lib folder for cairo and pixman to use. I could probably have it copy them instead of move them but currently they aren't in the release folder after compiling

arturoc: ok, perfect.

PR #3930, opened 2015-06-15
ofFileUtils : Check if directory already exists before creating it

armadillu:
I was getting lots of console warnings on this; 
Not sure why you would need to createEnclosingDirectory() in ofFile:: openStream(), but if you are to create a directory, you should probably check if it already exists first.


ofTheo:
agree it is sort of strange - I would of thought that would be the job of the programmer to make the directory if the path they are writing to doesn't exist. 

happy to merge this though as it doesn't change behavior. 

can you pull from master?

arturoc:
I've fixed this before. not sure why the diff is still showing the old code. in any case this checks if the file already exists not the enclosing dir. I'm going to close it. thanks!

Issue #3926, opened 2015-06-15
Apothecary - Boost 1.58.0


danoli3:
Boost for C++17 features for filesystem:

| | osx | vs | win_cb | ios | android | emscripten |
|----|-----|-------|---------|-------------|-------------|----|--------|-----|---------|-------|
| Merged? |   ✓      | ✓  | ✓      | ✓   | ✓       |✓   |
| PR |  #3913    | #3913  |  #3913     |  #3955 |   #3913   | #3913   |

danoli3:
I'll get the iOS part going @arturoc

arturoc:
thanks so much!

danoli3:
Boost for iOS Completed with #3933

danoli3:
Tested and working! All the new iOS PR's working well

bilderbuchi:
so, close this?

danoli3:
Yeah this is completed.

PR #3933, opened 2015-06-16
Apothecary - iOS Boost 1.58.0 static library for Filesystem, System

danoli3:
This PR includes the static libraries for boost for iOS Boost 1.58.0 Filesystem, System (arm64, armv7, i386, x86_64)

https://github.com/openframeworks/openFrameworks/issues/3926


arturoc:
do you have any changes for the apothecary formula?

danoli3:
@arturoc yeah right here: https://github.com/openframeworks/openFrameworks/commit/e97ff6303538d3773d61d63eceeee69609b42f94

danoli3:
I'm just about to test all these iOS Static libs together and PR the changes to the xconfig

arturoc:
ok, perfect, feel free to merge it whenever you are done with it

Miles-Tone:
Hi there, i think this goes here...

Im getting these errors when trying to compile the of_v20150617_ios_nightly

screen shot 2015-06-17 at 09 55 01

arturoc:
yes, the ios projects are surely broken right now since they are missing this includes

Miles-Tone:
ah ok, thanks Arturo
I wish i was learned enough to help. Maybe ill be experienced enough when 1.0 is being made to be useful.
I'll keep my eyes on the nightly builds

danoli3:
The iOS Boost libs in this PR are not linking correctly. Re-looking at the script to define why it thinks they are for OSX.

arturoc:
i think i added a check for type==osx and type==ios in this scripts but i never implemented the ios part, i can look into it if you want

danoli3:
Yeah that's all working. It was just when at the symbolic link stage in the openFrameworks lib build it was saying that the simulator architectures x86_64 and i386 where built for OS X not iOS... 

I've had a look at my other iOSBoost script and there was one stage I missed where I am decomposing the built architecture in the build directory and re-structuring it using the correct Apple iPhoneOS or iPhoneSimulator located ```ar crus```. Seems to do the stick (this only seems to apply for Boost as well due to bjam).

PR #3937, opened 2015-06-16
fix for deprecation attributes in vs2015

tgfrerer:
+ vs2015RC does not yet implement the [[deprecated]] attribute:
  http://blogs.msdn.com/b/vcblog/archive/2014/11/17/c-11-14-17-features-in-vs-2015-preview.aspx

  Since our compiler-agnostic deprecation macro OF_DEPRECATED cannot add
  deprecation messages to typedefs, there appears no other option than to
  omit deprecation messages for these typedefs in visual studio for now.

### DIFF
--- a/libs/openFrameworks/types/ofTypes.h
+++ b/libs/openFrameworks/types/ofTypes.h
@@ -104,7 +104,9 @@ class ofSerialDeviceInfo{
 ///
 /// \sa http://www.cplusplus.com/reference/mutex/mutex/
 /// \sa http://www.appinf.com/docs/poco/Poco.FastMutex.html
+#ifndef _MSC_VER
 [[deprecated("Use std::mutex instead")]]
+#endif
 typedef std::mutex ofMutex;
 
 /// \brief A typedef for a cross-platform scoped mutex.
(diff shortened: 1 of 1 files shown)


bilderbuchi:
> Since our compiler-agnostic deprecation macro OF_DEPRECATED cannot add
deprecation messages to typedefs

really? have you confirmed that? I was under the impression that this works...

also, I'm not sure if it's a good idea to mix deprecation message format (i.e. C++11 and our macro) in the codebase... :-/

tgfrerer:
yes, confirmed before committing. it works for functions, but not for typedefs. with typedefs only the deprecated attribute works.

bilderbuchi:
thanks!

> also, I'm not sure if it's a good idea to mix deprecation message format (i.e. C++11 and our macro) in the codebase... :-/

to clarify: I'd rather we adapt the macro to wrap the c++11 attribute, or rip it out completely, instead of having a mish-mash of uses in the codebase.

arturoc:
the problem is that the new c++11 attributes don't work yet for this 3 things. i think it's ok to have this 2 typdefs deprecated like this and then remove our macro since the new syntax is so much nicer

PR #3938, opened 2015-06-16
update DirectShow grabber to use new syntax

tgfrerer:
+ use pixels.getData() instead of pixels.getPixels

This brings DirectShow grabber in line with the current api and removes a
deprecation warning.

### DIFF
--- a/libs/openFrameworks/video/ofDirectShowGrabber.cpp
+++ b/libs/openFrameworks/video/ofDirectShowGrabber.cpp
@@ -192,9 +192,9 @@ void ofDirectShowGrabber::update(){
 
 							int posPix = (((int)posy * inputW * 3) + ((int)posx * 3));
 
-							pixels.getPixels()[(j*width*3) + i*3    ] = viPixels[posPix  ];
-							pixels.getPixels()[(j*width*3) + i*3 + 1] = viPixels[posPix+1];
-							pixels.getPixels()[(j*width*3) + i*3 + 2] = viPixels[posPix+2];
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/video/ofDirectShowGrabber.cpp
```diff
@@ -192,9 +192,9 @@ void ofDirectShowGrabber::update(){
 
 							int posPix = (((int)posy * inputW * 3) + ((int)posx * 3));
 
-							pixels.getPixels()[(j*width*3) + i*3    ] = viPixels[posPix  ];
-							pixels.getPixels()[(j*width*3) + i*3 + 1] = viPixels[posPix+1];
-							pixels.getPixels()[(j*width*3) + i*3 + 2] = viPixels[posPix+2];
+							pixels.getData()[(j*width*3) + i*3    ] = viPixels[posPix  ];
+							pixels.getData()[(j*width*3) + i*3 + 1] = viPixels[posPix+1];
+							pixels.getData()[(j*width*3) + i*3 + 2] = viPixels[posPix+2];
```
arturoc: not a big issue but you can just do:
```
pixels[(j*width*3) + i*3    ] = viPixels[posPix  ];`
```
also looking at the code this seems to be resizing which ofPixels also can do and the implementation is for sure way faster than this:

```
ofPixels src;
src.setFromExternalPixels(viPixels,width,height,OF_PIXELS_RGB);
src.resizeTo(pixels);
```

PR #3941, opened 2015-06-16
Refactor PG submodule

arturoc:
- moves the PG submodule to `apps/projectGenerator` instead of `apps/projectGenerator/projectGeneratorSimple`
- moves the legacy PG into the PG repo instead of devApps
- moves the commandline subproject in PG simple to the root of the repo
- moves the common code to an ofxProjectGenerator addon

The whole thing was a mess, some people had fixed problems in the legacy PG and others in the simple one. There was a subproject in the PG simple for command line and an extra copy of the project and addons classes inside the libs folder in the PG simple. And the submodule was still pointing to ofZach's repo so some fixes where there.

if no one has any problem with this i'll merge it tomorrow.

i've tested to pull this from a different computer and just doing:

`git submodule sync`
`git submodule update`

pulls the correct structure without problem


arturoc:
to clarify, now we have:
```
apps
  |_projectGenerator <- this is the repo
      |_projectGeneratorSimple
      |_projectGeneratorLegacy
      |_commandLine
```

and all the duplicated code that previously was in `src/addons`, `src/projects` and `src/pugixml` is now in `addons/ofxProjectGenerator`

ofZach:
sounds good!  but I do wonder why the common code would be an addon.  Is there some way of doing this without it being an addon (it doesn't feel like an addon to me, it feels like just some shared code).  beyond that, I'm all for moving thing around and removing redundancy.  the command line stuff is more experimental so that could also be it's own branch.

bakercp:
I'd advocate for the common code being an addon.  I'm currently using it in the next version of ofSketch.

bakercp:
(or at least a version that was here https://github.com/bakercp/ofxPG).  As soon as the master branch is updated, I'll try to go back to using the "official" code.

bakercp:
Also, FWIW, I've got Pugi already broken out into an addon here https://github.com/bakercp/ofxPugiXML.  Since we're dropping Poco from the core, we might consider just having it in the core officially so other projects can use it easily.

bakercp:
Tinyxml is still there for legacy ofxXmlSettings?

ofZach:
this doesn't really fit a definition of an "addon" to me, but happy to be convinced otherwise!    To me this seems more like common code between two projects...   I do think having one place for it is important and if it's easier for folks to have it addons I won't complain, it just doesn't seem like an actual addon to me.

arturoc:
yes, i'm wasn't very convinced this should be an addon either but the makefile system doesn't handle very well yet code not in the default folders so the easiest is to have it as an addon by now. i plan to modify the makefiles so it's easier to have alternative addon folders or external source and then we can move this somewhere else. until then we can remove it from the final packages and in the meantime it might be that people find that it's useful as an addon anyway :)

as for pugixml as an addon it probably makes sense but until we don't have some kind of wrapper for it it's probably better to leave it as it is? once we change ofXml to use it we can move it to be a core library

bilderbuchi:
this sounds good, lessening the confusion about the different PGs is great!

While I agree with Zach that the ofxProjectGenerator does not sound like an addon to me, I don't have a horse in the race. However, those changes seem to _not_ be included in this PR - I can't see added code in /addons.
Also, if the makefile problems are about the folder tree depth, could we put the code in apps/projectGenerator (i.e. the PG repo root) instead of addons/ofxProjectGenerator? that we it would be out of /addons, and logically grouped with the dependent codebase (i.e. the PG).

arturoc:
> Also, if the makefile problems are about the folder tree depth, could we put the code in apps/projectGenerator (i.e. the PG repo root) instead of addons/ofxProjectGenerator?

this is exactly what we can't do :)   even with other platforms we would have to modify the projects in the gui while now generating them is automatic just by running the PG on them without even having to add the addons since they are specified in addons.make.   i think the ideal would be to be able to specify an alternative addons path in addons.make and the PG so common code and addons for projects can be in their own repo instead of having to use submodules.

arturoc:
i'm merging this then, remember to `git submodule sync` and `git submodule update` if you want the latest PG

gorkacortazar:
@arturoc I was to clone a clean version of the ofx repo after your merge, but sourcetree crashes when cloning the submodule.

I also tried a clean clone from the command line with:
git clone --recursive https://github.com/openframeworks/openFrameworks.git

And I receive this error message when  the process goes to clone the project generator:
fatal: repository 'https://github.com/openframeworks/projectGeneratorSimple.git/' not found

gorkacortazar:
And it also happens when browsing to the github page to the submodule, it throws a 404 error... it may be pointing to a private repo? The name doesn't fit the public ones in the ofx list.

arturoc:
oh, sorry that last commit should fix it. not sure why the submodule 
config was pointing to the wrong url but was working on my machine 
probably because i already had downloaded that repo

gorkacortazar:
perfect now is working

ofTheo:
for me "git submodule sync and git submodule update" doesn't do anything I still see the old submodule structure and no legacy project generator.  

every time I have to work with the PG I run into submodule issues - I wonder if we could just bring it into the main repo?

arturoc:
have you pulled the main repo first?

```
git upstream master
git submodule sync
git submodule update
```

but yes now that the main code is in an addon it probably doesn't make much sense to have the code as a submodule except for having separate issues perhaps?

tgfrerer:
it seems to work for me - although the source files for the pg simple are named testApp.h and testApp.cpp, whereas the visual studio project file wants it to be called ofApp.h, and ofApp.cpp, so the project can't compile. 

i would suggest renaming the files ofApp.cpp/.h to stay consistent with our general naming scheme.

arturoc:
ah yes i generated the projects from the command line tool in linux and didn't checked this, i'll rename them to ofApp so the automatic project generation works

ofZach:
I think the idea of having it as a separate module was to have issues and development occur out of the main OF branch (it's kind of daunting to have so many issues overall) but now that the main code is in the OF repo I don't mind having it all in one place.  I just preferred it all separate to allow for discussing to happen somewhere else.  I am still not 100% sold on the main PG code being an addon, but I understand that this has to happen for makefiles and so it kind of makes sense, but I generally prefer it all in one place.  

One of the larger discussion is how to bring both PGs back to one app (like east germany and west germany becoming germany) they share such so much common code, it's pretty silly to have two apps at this point.   I was hoping the command line tool would help on that front but it's not really being used by anyone yet (on OSX, it compiles to a .app which is kind of wonky for a command line tool...  I think some amount of work needs to be done to make it more of single file executable -- it could be just some post build script love to remove things like fmod, etc).  I made it last year and put it out to the dev list but didn't hear much feedback about it.  If the command line tool got some love and the simple and advanced pgs both used it (the advanced PG is also pretty much used as a command line tool for packaging anyway) I think it would go a long way.

arturoc:
just fixed the issue with testApp -> ofApp

ofZach:
ps: thanks arturo for moving things, it def needed to happen... ( and thanks for @bakercp for starting this work too )

ofTheo:
@arturoc yes done exactly that. 
I'm going to see if I need a newer version of git to get this to work. 

But +1Million for killing the submodule - I think bringing it into the main repo will make it easier for people who are pulling down OF via git to get up and running ( especially now that there is no PG in the main repo currently )  and also make it easier for people ( including myself ) to contribute to the PG.

ofZach:
+million?  I feel like we are in +1 bubble now... :)   but sure, putting it in the core of OF is totally cool with me now that most of the code is there in the addons folder.   I do think maybe the command line thing is kind of experimental and maybe belongs in it's own branch?

ofTheo:
@ofZach I think all that is needed to kill the old devApps PG is a Generate all examples option in the PG Simple.  Then there would be no need for both. 

I love that with the devApp PG you can generate the examples for platforms your not on. We could always leave the templates for all platforms in the download so people can easily make project files for VS / iOS etc. 

ps: 
I saw a +1000 in an other issue and sort of got carried away :)

arturoc:
in one computer i've needed to update the remote for it to work properly:

```
cd apps/projectGenerator/projectGeneratorSimple
git remote set-url https://github.com/openframeworks/projectGenerator
cd ../../..
git pull upstream master
git submodule sync
git submodule update
```

but yeah i can also bring it to the main repo.

kylemcdonald:
@ofTheo i think you need to run `git submodule init` first:

```
git submodule init
git submodule sync
git submodule update
```

if i don't run `init`, nothing happens. i'm running git 2.4.3

i also tried running `git remote set-url https://github.com/openframeworks/projectGenerator` but it gives me a usage warning, i'm not sure the syntax of the command is correct? and according to the [.gitmodules](https://github.com/openframeworks/openFrameworks/blob/master/.gitmodules) file the url is already correctly set.

ofTheo:
thanks @kylemcdonald :)

PR #3944, opened 2015-06-17
add missing include to ofEvent.h

tgfrerer:
+ header <iterator> is needed for VS2015 to find
  back_inserter, possibly other comilers might need it, too.

+ also add ofEvent to openFrameworks XCode project file, so
  that XCode shows the file in the project file outline


arturoc:
thanks!

Issue #3904, opened 2015-06-09
propose to replace `ofToString` if environments can use C++11.


2bbb:
i wanna propose to replace `ofToString` about, on environments that can use C++11.

like this:

```cpp
template <class T>
typename enable_if<!is_arithmetic<T>::value, string>::type ofToString(const T& value) {
	ostringstream out;
	out << value;
	return out.str();
}

template <class T>
typename enable_if<is_arithmetic<T>::value, string>::type ofToString(T value) {
	return to_string(value);
}
```

so, on my environment (Xcode6.3.2/clang on OSX 10.10), this code is 2-3 times faster than current version.

test with this:

```cpp
#include <iostream>
#include <string>
#include <sstream>
#include <type_traits>

struct SimpleStopWatch {
	void start() {
		_start = std::chrono::high_resolution_clock::now();
	}
	
	template <typename T = std::chrono::nanoseconds>
	typename T::rep rap() {
		_end = std::chrono::high_resolution_clock::now();
		return std::chrono::duration_cast<T>(_end - _start).count();
	}
	
	template <typename T = std::chrono::nanoseconds>
	typename T::rep getLastRap() const {
		return std::chrono::duration_cast<T>(_end - _start).count();
	}
	
	auto getLastRapMilliseconds() const 
	-> decltype(getLastRap<std::chrono::milliseconds>()) {
		return getLastRap<std::chrono::milliseconds>();
	}
	
	auto getLastRapMicroseconds() const 
	-> decltype(getLastRap<std::chrono::microseconds>()) {
		return getLastRap<std::chrono::microseconds>();
	}
	
	auto getLastRapNanoseconds() const
	-> decltype(getLastRap<std::chrono::nanoseconds>()) {
		return getLastRap<std::chrono::nanoseconds>();
	}
	
	decltype(std::chrono::high_resolution_clock::now()) _start, _end;
};

using namespace std;

template <class T>
string ofToString(const T& value){
	ostringstream out;
	out << value;
	return out.str();
}

template <class T>
typename enable_if<!is_arithmetic<T>::value, string>::type ofToString11(const T& value) {
	ostringstream out;
	out << value;
	return out.str();
}

template <class T>
typename enable_if<is_arithmetic<T>::value, string>::type ofToString11(T value) {
	return to_string(value);
}

int main(int argc, char *argv[]) {
	SimpleStopWatch w;
	long long elapsed;
	string s;
	
	w.start();
[...]

PR #3948, opened 2015-06-17
C++11 style convert string and numbers

2bbb:
discuss on #3904


arturoc:
i'm going to milestone this for 0.9.1. thinking about it a bit more i'm not sure if it might be common practice in post c++11 to overwrite to_string for custom classes apart from ovewriting the operator<<

2bbb:
@arturoc 
okay. let us considered over a period of time about this.

and so, i rewrote a bit of comments about some part.
please someone proofread if it is not fit or my english is not good.

2bbb:
i did tiny test about other point.

http://ideone.com/SImL9e

```cpp
struct vec3 {
	float x;
	float y;
	float z;
};
 
std::string toStringImpl1(const vec3 &v) {
	return std::to_string(v.x) + ", " + std::to_string(v.y) + ", " + std::to_string(v.z);
}

std::string toStringImpl2(const vec3 &v) {
	std::ostringstream ss("");
	ss << v.x << ", " << v.y << ", " << v.z;
	return ss.str();
}
```

results:
```
impl1: 255902075
impl2: 456477576
```

i think, performance is not bad.
and, maybe, we can replace implementation of operator<< in custom classes from using ostringstream to to_string.

maybe, focus of the problem is just "is to overload the std::to_string common practice?" only.

libs/openFrameworks/utils/ofUtils.cpp
```diff
@@ -469,10 +454,7 @@ bool ofToBool(const string& boolString) {
 	if(lower == "false") {
 		return false;
 	}
-	bool x = false;
-	istringstream cur(lower);
-	cur >> x;
-	return x;
+	return (bool)std::stoi(boolString);
```
bilderbuchi: does this work? Presumably (haven't tried it), if `boolString` is `False`, then `stoi` will make this into some non-zero integer, which, when cast to bool, will be `True`, which is wrong. Is that correct?

2bbb: sorry, i didn't test `ofToBool("False");`…
i will test tomorrow more cases.

but i think:
if `boolString` is `False`, then `lower = ofToLower(boolString)` is `false`.
and match the case of `lower == "false"`.
therefore, return `false`.

2bbb: argh...!!
i think `return (bool)std::stoi(lower);`, but i wrote `(bool)std::stoi(boolString);`...

and now, `ofToLower` is wrong...?
i'll check about `ofToLower`.

arturoc: I think this was correct. the only case where it'll arrive at that line is if it's a number. any variation in case of false will return earlier.

also there was a problem with oftolower but I fixed it yesterday

El 19 de junio de 2015 8:18:18 CEST, ISHII 2bit <notifications@github.com> escribió:
>> @@ -469,10 +454,7 @@ bool ofToBool(const string& boolString) {
>>  	if(lower == "false") {
>>  		return false;
>>  	}
>> -	bool x = false;
>> -	istringstream cur(lower);
>> -	cur >> x;
>> -	return x;
>> +	return (bool)std::stoi(boolString);
>
>argh...!!
>i think `return (bool)std::stoi(lower);`, but i wrote
>`(bool)std::stoi(boolString);`...
>
>and now, `ofToLower` is wrong...?
>i'll check about `ofToLower`.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/openframeworks/openFrameworks/pull/3948/files#r32803823

-- 
Sent from touchscreen

2bbb: yeah, you're right!!
i had been a misunderstanding.

but, we need a try-catch around `std::stoi` for case of `lower` is neither "true" and "false".
i fix again.

bilderbuchi: >  the only case where it'll arrive at that line is if it's a number. 

ah, you're right, in the diff the upper part of the algorithm got folded away, so I didn't realize that.

DomAmato:
Any idea of this works with funky wide string pointer types?

http://www.codeproject.com/Articles/2995/The-Complete-Guide-to-C-Strings-Part-I-Win-Chara

I'm pretty sure the current implementation takes the pointer and converts the memory address as the string but not the actual string data

Issue #3619, opened 2015-02-04
Apothecary Discussions - Optimization Levels for Statics


danoli3:
Should we standardise the optimization level for compilation across all static libraries (or specific deployments).

**GCC_OPTIMIZATION_LEVEL**
Description	

> [-O0] None: Do not optimize.
> With this setting, the compiler's goal is to reduce the cost of compilation and to make debugging produce the expected results.  Statements are independent: if you stop the program with a breakpoint between statements, you can then assign a new value to any variable or change the program counter to any other statement in the function and get exactly the results you would expect from the source code.

> [-O, -O1] Fast: Optimizing compilation takes somewhat more time, and a lot more memory for a large function. 
 > With this setting, the compiler tries to reduce code size and execution time, without performing any optimizations that take a great deal of compilation time. In Apple's compiler, strict aliasing, block reordering, and inter-block scheduling are disabled by default when optimizing.

>  [-O2] Faster: The compiler performs nearly all supported optimizations that do not involve a space-speed tradeoff.
> With this setting, the compiler does not perform loop unrolling or function inlining, or register renaming.  As compared to the 'Fast' setting, this setting increases both compilation time and the performance of the generated code.

> [-O3] Fastest: Turns on all optimizations specified by the 'Faster' setting and also turns on function inlining and register renaming options. This setting may result in a larger binary. 

> [-Os] Fastest, Smallest: Optimize for size.  This setting enables all 'Faster' optimizations that do not typically increase code size.  It also performs further optimizations designed to reduce code size.

>  [-Ofast] Fastest, Aggressive Optimizations: This setting enables 'Fastest' but also enables aggressive optimizations that may break strict standards compliance but should work well on well-behaved code. 

[GCC_OPTIMIZATION_LEVEL]

#### More detailed explanation in the GCC manual:
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Optimize-Options

Also note this discussion.
http://stackoverflow.com/questions/11546075/is-optimisation-level-o3-dangerous-in-g

---------------------

The iOS Default is ```-Os`` (fastest, Smallest), so I have setup all the build scripts in that for now.

Issue #1022, opened 2012-03-02
Optimisation consistency


elliotwoods:
Continuation of http://forum.openframeworks.cc/index.php/topic,9095.msg42378.html#new

NOTE: Things may have changed as the projects I'm using are all created using the older creation tools.
Can somebody up to date please check?

Currently XCode's default project setting is /O2 for debug projects
whilst Visual Studio's is /O0

/O0 generally makes the most sense for debug mode as it allows for effective program flow and variable tracking (because with /O2, lines of code and variables are commonly optimised away).

If we are worried that this will leave default users in a default 'slower state' then perhaps we could make release the default? Then people could be aware that they are selecting debug when they want to actually do some debugging?
But this may have other issues (such as some libraries might not perform as many checks as otherwise?)

it might be worth making users more aware of the 2 options anyway, and then start using them consistently

danoli3:
Re-igniting this @memo was this an issue with OS X project or iOS. I've just made changes to fix the iOS side. I'll have a look at the OSX settings as well...

PR #3953, opened 2015-06-17
iOS Xcode -pthread fix / iOS optimisation level default

danoli3:
## pthread issue
- Same issue as fixed in https://github.com/openframeworks/openFrameworks/pull/3939

> this was spilling a ton of compiler warnings in osx - one for nearly every .cpp file.
> with threads moving over to std::thread, pthread is no longer in use.
> this affects only the project files / templates, and the pre-generated project files for the project generator.

Silence this:
pthread

## iOS Optimisation level -Os for Release rather than -O3
Set the release optimisation to be inline with the default apple guideline for a release project for iOS (-Os).
This also brings the project in line with the built libraries for the most part as well.

## iOS Optimisation level -O0 for Debug builds.
- While the default for Xcode 5+ for debug is -Os, this was typically designed for Objective-C and Swift applications. While the later versions of Xcode to seem to get it almost correct. It would be wise to get this command in line with the rest of the framework at -O0 (no optimisations) for the Debug oF lib.
- The Absence of the GNU optimisation command so far was in the idea that the default in this case, was the -O0. See @memo's thoughts in https://github.com/openframeworks/openFrameworks/issues/1022

> Fastest, Smallest: Optimize for size. This setting enables all 'Faster' optimizations that do not typically increase code size. It also performs further optimizations designed to reduce code size. [-Os]

Screenshot of iOS+OFLib project:
optimisation

Optimisation consistency: https://github.com/openframeworks/openFrameworks/issues/1022
Optimisation levels for Statics: https://github.com/openframeworks/openFrameworks/issues/3619


danoli3:
@openframeworks/ios how does this look?

arturoc:
i think we should use -O0 for debug, it makes it's super tricky to debug with optimizations and really difficult to understand what's going on if you don't know that optimizations are on.

bilderbuchi:
Isn't that exactly what he did?

libs/openFrameworksCompiled/project/ios/Debug.xcconfig
```diff
@@ -5,8 +5,9 @@ GCC_ENABLE_SSE3_EXTENSIONS = YES
 GCC_ENABLE_SUPPLEMENTAL_SSE3_INSTRUCTIONS = YES
 GCC_INLINES_ARE_PRIVATE_EXTERN = NO
 GCC_MODEL_TUNING = G5
+GCC_OPTIMIZATION_LEVEL = 0
```
danoli3: I wonder if this is causing the issue on older mac's to crash with the simulator (too much data?)

Issue #3924, opened 2015-06-15
Apothecary - Poco with OpenSSL 1.0.2c


danoli3:
With the fixes in OpenSSL 1.0.2c it is required to rebuild Poco.
https://github.com/openframeworks/openFrameworks/issues/3923

### Poco current build status with OpenSSL 1.0.2c:

| osx | linux | linux64 | linuxarmv6l | linuxarmv7l | vs | win_cb | ios | android |
|-----|-------|---------|-------------|-------------|----|--------|-----|---------|
|   ✗ |   ✗   |   ✗     |     ✗       |       ✗     | ✓ |   ✗    |   ✗ |     ✗   |

bakercp:
I can rebuild osx, arm6/7 and linux32/64.  I'm also looking working on win_cb/mingw 32/64.

PR #3957, opened 2015-06-17
Apothecary - iOS Poco Update with OpenSSL 1.0.2c

danoli3:
Updated the Poco Libs for iOS.
All the same other than the net coding is using OpenSSL 1.0.2c now.

####  Merge this PR after OpenSSL iOS -> https://github.com/openframeworks/openFrameworks/pull/3956


ofTheo:
will these two PRs for 1.0.2.c mean new libs are needed for other platforms?
Thanks for all the iOS updates!! :)

danoli3:
Yeah it will see the issue tracking this here: https://github.com/openframeworks/openFrameworks/issues/3924

It is needed as VS64 only supported with 1.0.2c which is what we have in the repo now (1.0.2c dll for VS)

arturoc:
i've just updated osx and going to do android. the rest of the platforms use openssl from the system so that should be it

PR #3961, opened 2015-06-18
use robocopy instead of xcopy in post-build step

tgfrerer:
* xcopy overwrites all .dlls in the bin/ folder
    every time an app gets rebuilt. This leads to
    longer compile times, since about 20MB need to be
    overwritten in the bin/ folder everytime an app is
    built.

  * robocopy is similar to rsync and will only copy or
    overwrite a file if either a change was detected
    (i.e. the build target was changed from 32bit to
    64 bit) - or the directory is missing a file.

    Files which are already present and of the same
    size will not be copied.

    This speeds up app compile time. (And may also
    help towards longer SSD lifetime)

    ~~xcopy~~ robocopy is a default system tool, and available
    since Windows 7 onwards.

https://technet.microsoft.com/en-GB/library/cc733145.aspx

### DIFF
--- a/scripts/vs/template/emptyExample.vcxproj
+++ b/scripts/vs/template/emptyExample.vcxproj
@@ -106,7 +106,8 @@
       <AdditionalLibraryDirectories>%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
     <PostBuildEvent>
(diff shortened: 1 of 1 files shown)


bilderbuchi:
> xcopy is a default system tool, and available since Windows 7 onwards.

you did men robocopy here, right?

tgfrerer:
yes, indeed! and it's been around for even longer, it turns out.

PR #3964, opened 2015-06-18
fix getTimestampString app crash on windows

tgfrerer:
* on windows, `strftime` called with an invalid formatting
    string will lead to an app crash.

  * since `ofGetTimestampString()`, when called without
    parameters, uses the parameters `%Y-%m-%d-%H-%M-%S-%i`,
    this would lead to a crash upon the first call of
    `ofGetTimeStampString` because of the non-standard `%i`
    formatter.

    https://msdn.microsoft.com/en-us/library/fe06s4ak(v=vs.140).aspx

Since the `%i` parameter is non-standard over probably all
implementations of `strftime`, and we manually replace it
anyways, i've re-arranged the replacement to happen before
the call to `strftime`.

This is not super optimised (and I believe this method
could be), but it doesn't crash anymore.


arturoc:
thanks you! yeah the %i doesn't exist in any other function call, i think it's a poco only thing

PR #3967, opened 2015-06-19
fix ofParameter - extra template<>

tgfrerer:
* removes an extra template<> line which was causing
    compiler errors since then the template could not be
    specialised.

    To specialise a template which is already templated, the
    template typename has to be specified, as it is in the
    next line.


tgfrerer:
i just saw that xcode reports this only as a warning, but in visual studio it's an error.

arturoc:
thanks!

PR #3968, opened 2015-06-19
ignore linker warning 4099 for visual studio debug

tgfrerer:
* when building an app in debug mode, visual studio tries to
  find linker symbol (*.pdb) files for libraries that ship
  with openframeworks.

  Since these libraries are shipped without debug symbol files
  (these files are huge) a warning is issued for almost
  every library that ships with openFrameworks whenever an
  app is built, or re-built. For oF, this meant scores of
  warnings on debug builds.

  In the past, it has been impossible to ignore this linker
  warning message, but it had been so nagging to not being
  able to ignore it that some people even had tutorials for
  binary patching the visual studio linker executable to be
  able to get rid of it.

  It seems that with Visual Studio2015 it is finally
  officially possible to ignore this (utterly useless in our
  case) warning.

  Let's.


arturoc:
shouldn't it be better to put this in the properties files in openFrameworksCompiled/vs? that way old projects will get this option too

tgfrerer:
good point! the post build event (xcopy vs. robocopy should be there, too, i believe) will send an updated pr in a sec...

tgfrerer:
Using the .props files now for project wide settings. Tested on VS2015/32bit.

arturoc:
thanks!

PR #3970, opened 2015-06-20
speeding up compilation

tpltnt:
Hi,

i added the "-j" option to the call of make inside "compileOF.sh". It now uses all CPUs available and not just one to do the compilation, which in turn should speed things up.
Cheers,
tpltnt


pizthewiz:
The use of `nproc` will preclude OS X from being able to use `ofCompile.sh` no? Is there a more generic way to detect core count or can the script detect the host OS and do the right thing?

bakercp:
Apparently `getconf _NPROCESSORS_ONLN` is cross platform.  Worked for me on Linux / OSX.

ref: https://stackoverflow.com/questions/6481005/how-to-obtain-the-number-of-cpus-cores-in-linux-from-the-command-line

tpltnt:
@pizthewiz good call. Using `nproc` in the general build script was a stupid idea. Thank you @bakercp for pointing out a more portable way. I fixed it.

ofTheo:
this looks good to me - but maybe @arturoc should sign off on it.

arturoc:
don't like very much to default to all the cpus usually since it makes some systems really slow when it's compiling so it should be some kind of option but in this case i don't see a good way of passing the parameter since this is called from install_dependencies so i'm merging it.

thanks!

bilderbuchi:
well, we could use max(nr-of-cores - 1, 1)? that way some systems get an increase, while still preserving one core for the rest of the sys?

tpltnt:
@bilderbuchi i had thought about that too, but rejected it since quite a few embedded ARM boards have only 2 CPUs and they won't profit from it.

bilderbuchi:
> since quite a few embedded ARM boards have only 2 CPUs and they won't profit from it.

I guess that's a question of how you want to optimize:
* do you want improvements for some, without making anybody worse off, or
* do you want bigger improvements for some, while making some others much worse off? 

I personally lean toward the former, but I guess we'll just have to wait if we get reports back from people whose systems we freeze/make sluggish during compilation

PR #3971, opened 2015-06-20
Debian dependencies fixes

tpltnt:
Hi,

i added libboost-dev as an explicit dependency for Debian (x86 & ARM), since libboost-dev is missing on vanilla Debian (Jessie) installations. This also includes a minor error message fix.
Cheers
tpltnt


bilderbuchi:
only to debian? I'd guess the ubuntu script would need those changes, too? and the other archs?
aside of that, ping @arturoc - are we using the system's boost currently?

arturoc:
hey, yes would be good to have this change for all distros. the correct dependency though is libboost-filesystem-dev

tpltnt:
@bilderbuchi @arturoc good call. I added boost as an explicit dependency to Arch, Fedora and Ubuntu. I did not touch the Ubuntu OMAP4 script since it only seems to install additional OMAP4 stuff and not all dependencies.

PR #3972, opened 2015-06-21
Lib/Script Updates

DomAmato:
Fixes Glut32.lib missing for x64 builds and a logic error when running the openSSL script as well as linker errors from poco and cairo using the apothecary scripts to build the most recent libraries


arturoc:
thanks!

scripts/apothecary/formulas/openssl/openssl.sh
```diff
@@ -652,6 +652,7 @@ function copy() {
 		git checkout $1/include/openssl/opensslconf_osx.h
     	git checkout $1/include/openssl/opensslconf_android.h
     	git checkout $1/include/openssl/opensslconf_win32.h
+    	git checkout $1/include/openssl/opensslconf_vs.h
```
arturoc: this is incorrect, what we are doing is copying the original opensslconf.h as openssl_vs.h or whatever the platform is, as that file is generated in the config phase and is different for every platform. i've fixed it already but just so you know if you have to modify the script again. i'll add some comments in the script too

DomAmato: Ah ok, I was using the of master and it kept complaining that conf_vs was missing, makes more sense now

tgfrerer:
I believe this broke windows builds - I'm getting:

	PocoNetSSLmdd.lib(SSLManager.obj) : error LNK2019: unresolved external symbol _X509_STORE_CTX_get_error referenced in function __catch$??1SSLManager@Net@Poco@@AAE@XZ$0

PR #3975, opened 2015-06-22
Reallocating ofFbo on MRT, drawBuffers unable to attach.

kentaroid:
This problem occurs when reallocating ofFbo that is used on MRT. 
(such as resizing)

The fbo that attached to 1 or later , is always black.
This is probably because activeDrawBuffers has not been cleared.

When fbo is not the MRT, the attach_point is always 0.
So, it looks like there is no problem. But activeDrawBuffers will be incremented.


tgfrerer:
Great find!
+1 for merging.

arturoc:
thanks!

PR #3976, opened 2015-06-22
ofQuaternion operator<<, operator>> is wrong.

2bbb:
i fixed them with global operator overload.

reason:
now, `operator<<` and `operator>>` in `class ofQuaternion`.
but this is not call forever.

and if we call `<<` like following:
```
ofQuaternion q;
ofLogNotice() << q;
```
this code will run.

but, in fact, this code calls `ostream &operator<<(ostream &os, const ofMatrix4x4 &mat)` with implicity conversion.

### DIFF
--- a/libs/openFrameworks/math/ofQuaternion.h
+++ b/libs/openFrameworks/math/ofQuaternion.h
@@ -178,8 +178,8 @@ public:
     inline const ofQuaternion operator -(const ofQuaternion& rhs) const;    ///< Binary subtraction
     inline const ofQuaternion operator -() const;                           ///< returns the negative of the quaternion. calls operator -() on the Vec4
     
-    inline ostream& operator<<(ostream& os);
-    inline istream& operator>>(istream& is);
+    friend ostream& operator<<(ostream& os, const ofQuaternion &q);
+    friend istream& operator>>(istream& is, ofQuaternion &q);
     
     /// \}
 };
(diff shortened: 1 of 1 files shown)


2bbb:
and so, `ss >> q;` didn't work.

2bbb:
ping @arturoc

arturoc:
thanks!

arturoc:
@2bbb this has broken compiling on master on all platforms, you can check the error here https://travis-ci.org/openframeworks/openFrameworks/builds/102314298

can you check what's going on?

thanks!

libs/openFrameworks/math/ofQuaternion.h
```diff
@@ -199,21 +199,21 @@ ofQuaternion::ofQuaternion() {
 
 
 //----------------------------------------
-ostream& ofQuaternion::operator<<(ostream& os) {
-    os << _v.x << ", " << _v.y << ", " << _v.z << ", " << _v.w;
+ostream& operator<<(ostream& os, const ofQuaternion &q) {
```
tgfrerer: most likely this line leads to duplicate symbols, since it is not enclosed within a namespace

2bbb:
@arturoc 

hmmm...?

on my Mac (Latest Xcode & master branch after this was merged) got success to compile.
i do check problem soon.

libs/openFrameworks/math/ofQuaternion.h
2bbb: this `operator<<` is not class method.
i defined this as friend function in `ofQuaternion`.

maybe cause is not this. maybe...

tgfrerer: i see. then it might work if the method implementation is moved from the .h file to the .cpp file.

2bbb: yeah!
i think so too. and i post new PR #4729 with those fix.
i pray to god of ci-travis...

arturoc: or add inline to the implementation but for something like this probably better to just move it to the cpp

Issue #3916, opened 2015-06-13
opencv 32bit libs for vs2015


nykwil:
The opencv libs are compiled with vs2012 and aren't compatible with the new build. 

Related to #3910

bilderbuchi:
pinging @openframeworks/windows

DomAmato:
@elliotwoods compiled the 64 bit ones. Did you change the script to get it to compile or how did you go about doing it?

bilderbuchi:
ping @openframeworks/windows

DomAmato:
I made a new PR with only the lib updates and the apothecary script update https://github.com/openframeworks/openFrameworks/pull/4086

PR #3985, opened 2015-06-24
Addon Script and Lib fixes

DomAmato:
Updates for opencv and assimp apothecary scripts and resulting libs


DomAmato:
fixes isse: https://github.com/openframeworks/openFrameworks/issues/3916

Testing the current addon libs resulted in errors so I rebuilt them (this is including @elliotwoods libs) I copied the scripts over to the formulas just to make updating easier than typing in the path.

arturoc:
there's something strange with this PR, some files that are already there like the formulas appear as completely new. i think the problem is that the folder names have change in capitalization. we should keep everything as lower case otherwise github, linux and osx with case sensitive fs will see opencv and openCv as 2 different folders

DomAmato:
The formulas are usually located in addons/(addonName)/scripts/formulas and I moved them to apothecary/formulas so technically they are new files.

 I do see what you mean with the capitalization though specifically the openCv

arturoc:
oh i see, why not leave them in the addons folders?

bilderbuchi:
> oh i see, why not leave them in the addons folders?

I was wondering the same thing.

DomAmato:
It was just for the sake of ease of typing in the apothecary formula to run it, I'll move it back

PR #3992, opened 2015-06-24
ofTrueTypeFont: add UTF8 support

arturoc:
The PR that removed poco dependencies introduced a new library cpputf8 which makes it really easy to parse std::strings as utf8 making it relatively straightforward to add full utf8 support for ofTrueTypeFont.

This adds full UTF8 support for ofTrueTypeFont allowing to draw any glyph a font allows.Since loading every possible unicode codepoint in a texture would be impossible and since most fonts only have support for a few languages anyway the new load method receives a settings object that specifies the unicode ranges to load, among other config:

```cpp
ofTtfSettings settings("Droid Sans Japanese",24);
settings.antialiased = true;
settings.ranges = {
    ofUnicode::Latin1Supplement,
    ofUnicode::Hiragana,
    ofUnicode::Katakana,
    ofUnicode::CJKUnified
};
ttf.load(settings);
```

then ttf can be used as usual and it'll draw any glyph in the specified ranges present in the loaded font.

It seems that loading full CJK unified (jp kanji or traditional chinese or korean) + hiragana + katakana for example at 24 points is ok with one texture but for bigger font sizes loading CJK (around 20000 glyphs) might be too much so ofTrueTypeFont also adds a new method that allows to render a text into a texture, it's slower than using the cached texture atlas so it shouldn't be called every frame but for pre-creating texts and rendering them later is totally fine and even faster than drawing a text over and over every frame:

```cpp
ofTexture tex = ttf.getStringTexture("any text");
```
Internally there's also some refactoring using lambdas to remove logic replication in several functions.

The class api remains fully compatible with previous versions.


arturoc:
screenshot from 2015-06-24 11 46 14

armadillu:
ttf.getStringTexture() will be very useful to pre-cache paragraphs too. I can't remember now, does ofTrueTypeFont interpret "\n" characters?

arturoc:
yes it does

kylemcdonald:
when i try to run this on OSX i get: `ofTrueTypeFont.cpp:609:22: No member named 'accumulate' in namespace 'std'` right after the comment  `initialize character info and textures`. i'm pretty sure libc++ is being linked to correctly, but i'll try to regenerate my project file to make sure it's the most up to date.

arturoc:
there was a missing header in osx that is not needed in linux, should be
working now

kylemcdonald:
after some quick testing, this looks perfect for anything that does not require shaping (e.g., arabic and devanagari don't really work).

```c++
#include "ofMain.h"
//#define SHAPING

public:
    ofTrueTypeFont ttf;
    ofTexture tex;
    void setup() {
        ofTtfSettings settings("Arial Unicode MS",24);
        settings.antialiased = true;
        settings.ranges = {
#ifdef SHAPING
            ofUnicode::Arabic,
            ofUnicode::ArabicExtendedA,
            ofUnicode::ArabicMath,
            ofUnicode::ArabicPresFormsA,
            ofUnicode::ArabicPresFormsB,
#else
            ofUnicode::Latin1Supplement,
            ofUnicode::Hiragana,
            ofUnicode::Katakana,
            ofUnicode::CJKUnified
#endif
        };
        ttf.load(settings);
        ofBackground(255);
        ofSetColor(0);
        tex = ttf.getStringTexture(
#ifdef SHAPING
                    "عندما يريد العالم أن يتكلّم  ، فهو يتحدّث بلغة يونيكود."
#else
"ラドクリフ、マラソン五輪代表に 1万m出場にも含み."
#endif
        );
    }
    void update() {
    }
    void draw() {
        tex.draw(20, 20);
    }
};
int main() {
    ofSetupOpenGL(1024, 96, OF_WINDOW);
    ofRunApp(new ofApp());
}
```

screen shot 2015-06-24 at 13 57 06
screen shot 2015-06-24 at 13 57 15

bakercp:
@kylemcdonald I've got a lot of work with fribidi / harfbuz ready to go for shaping bi-directional once this is in.

arturoc:
yes i haven't looked into anything that is not left to right, i guess i can quickly add something basic but harbuzz is definitely the way to go and it'll even will make kerning much better than it is right now

arturoc:
i've added basic right to left rendering which just needed to invert a couple of directions here and there but that doesn't solve shaping yet. in any case for basic rendering of text as we've been doing until now i don't think shaping is an option since it would be to slow, using the new method of retrieving a texture should work with harfbuzz.

probably we want to have a small class at some point that wraps that texture and allows to position it's baseline precisely since right now all you can do is position it's left right corner but info about the baseline is lost when rendered to a texture.

PR #3994, opened 2015-06-24
Feature : ofSetMainLoop

elliotwoods:
I presumed this feature must already been in there, but I couldn't see it so here it is
let me know if it's already catered for elsewhere.

big fan of ofMainLoop. I was in the middle of making a feature request for `ofSession` today until i realized that the 0.9.0 had ofMainLoop

Elliot

### DIFF
--- a/libs/openFrameworks/app/ofAppRunner.cpp
+++ b/libs/openFrameworks/app/ofAppRunner.cpp
@@ -110,6 +110,11 @@ shared_ptr<ofMainLoop> ofGetMainLoop(){
 }
 
 //--------------------------------------
+void ofSetMainLoop(shared_ptr<ofMainLoop> newMainLoop) {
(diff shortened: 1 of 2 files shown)


elliotwoods:
btw. you need to call ofSetMainLoop on external dynamic libraries which statically link openFrameworks to sync their state with the oF which is statically linked to your application

i presume this wouldn't be an issue if oF was dynamically linked. but for file size reasons we tend not to do that right?

arturoc:
we didn't have this because for a normal use there's only one main loop. there's the possibility also to have no global main loop and instead instantiate your own (or even more than one, potentially on different threads) and call run on it/them but then you need to use the renderers explicitly as shown in devApps/explicitRendererExample

anyway i see how this can make sense if you are using OF as an external library and recreating the app more than once.

Issue #3977, opened 2015-06-23
Xcode 7.0 - Requires iOS 6.0+ as minimum. Time for iOS 7.0 Minimum step up


danoli3:
Building for a device under Xcode 7 fails due to
`clang: error: -fembed-bitcode is not supported on versions of iOS prior to 6.0`
Apple's new bitcode enhancements require at least iOS 6.0

This is due to the current Deployment Target of `5.1.1`

I think it's a good idea to push forward even further and make iOS `7.0` oF minimum.

The current statistics from Apple:
![ios8](https://cloud.githubusercontent.com/assets/830748/8297208/7791d096-199e-11e5-9d3e-3bc9e67c1431.png)

Within 2 months it is assumed 50-70% of all iOS devices used will be living in iOS `9.0`.
Coupled with iOS `9.0` being dubbed as a performance patch for older devices, as well as smaller install size, this number could be greater.
The vast majority of iOS devices still are supported for iOS 7.0 and greater.
#### Make iOS 7.0 minimum. Skip iOS 6.0 minimum.

Previous discussions: https://github.com/openframeworks/openFrameworks/issues/2568

danoli3:
@openframeworks/ios

danoli3:
Adding to this: When any static library is built without ENABLE_BITCODE for iOS 6.0 and above: The following "URGENT" warning is displayed for every object of every architecture of every static lib.

> ld: warning: URGENT: all bitcode will be dropped because '*.a(*.o)' was built without bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. Note: This will be an error in the future.

julapy:
+1
also agree that moving to iOS 7.0 minimum would be a good idea.

ofTheo:
That chart looks a bit dubious to me :)  Half my iOS devices are iOS 5 or 6. 
Any reason to not have iOS 6 as a minimum? 
Is there a huge workaround that we could then avoid? 

I know for people making games being able to support the most number of devices is really helpful.

If there is a really strong reason to make iOS 7 minimum I�m for it - but generally with OF if we can easily support more devices/OSs without much extra work we usually try and do so.

julapy:
yeah apple is pretty aggressive with their iOS upgrades. generally when working on pure iOS commercial projects (non-OF), it's safe to target the last 2 major releases, so right now it would be iOS 7 & 8.

[...]

PR #3996, opened 2015-06-24
iOS Enable Bitcode = NO for Xcode 7

danoli3:
Current fix for Xcode 7 (beta) for iOS for the BITCODE errors.

### What is Bitcode:

> Bitcode refers to to the type of code: "LLVM Bitcode" that is sent to iTunes Connect. This allows Apple to use certain calculations to re-optimize apps further and downsize executable sizes. If Apple needs to alter your executable then they can do this without a new build being uploaded.

> This differs from: Slicing which is the process of Apple optimizing your app for a user's device based on the device's resolution and architecture. Slicing does not require Bitcode. (Ex: only including @2x images on a 5s)

> App Thinning is the combination of slicing, bitcode, and on-demand resources

> Bitcode is an intermediate representation of a compiled program. Apps you upload to iTunes Connect that contain bitcode will be compiled and linked on the App Store. Including bitcode will allow Apple to re-optimize your app binary in the future without the need to submit a new version of your app to the store.
http://stackoverflow.com/questions/30722606/what-does-enable-bitcode-do-in-xcode-7

### Problem
- Under Xcode 7 it is defined by default to use the new LLVM bitcode format
- This is a iOS 6+ only requirement
- Requires all static libs to be built with the correct bitcode flag for LLVM

### Current Solution
- Disabling Bitcode in the CoreOF.xconfig 
- Disabling Bitcode in the iOS Template Project

This disables bitcode for the project explicitly and allows the project's to build under Xcode 7.
This value is meaningless in Xcode 5 and 6 so is safe to be merged straight away.

### Future Solutions
After Xcode 7 has been officially released, we are going to have to jump on this ship pretty quick as this is going to be an error for submitting apps soon after. Lucky with Apothecary, this shouldn't be a game breaker.

- Recompile all iOS static libraries with the new ```-fembed_bitcode``` flag.
- Set Minimum iOS Target to 6.0+ (iOS 7 is current next target from discussions). 
- Discussions here: https://github.com/openframeworks/openFrameworks/issues/3977


danoli3:
@openframeworks/ios

admsyn:
:+1: great summary!

PR #3997, opened 2015-06-24
iOS Template - Code Sign fix for simulator

danoli3:
Fix for Xcode Template to fix the potential issues with code signing for the iOS simulator. 

### Problem
- If the developer does not have an active or any Apple Developer account associated with their account it has issues properly signing / deploying to Simulator
- By default deploying to the simulator does not require code signing however in the current project template setup it is attempting to code sign the simulator

### Solution
- Define different targets for iPhoneSimulator and iPhoneOS for Xcode under the code signing section.
- By adding iPhoneSimulator explicitly and defining this as Do not code sign. This error is fixed.


danoli3:
@openframeworks/ios

Issue #3943, opened 2015-06-17
Xcode Crashes after finishing compiling the last master


gorkacortazar:
With a clean version of the master, Xcode 6.3.2 on Yosemite crashes and closes after when compiling and running any ofx application. It happens on all combinations of Debug/Release, 32/64 bits.

The applications compiles, but crashes when starts running. It started to happen in one of the last weeks commits, in my personal development repo I can revert the commits until a point where doesn't happen.

I will pinpoint the exact commit soon.

![screen shot 2015-06-17 at 11 48 13](https://cloud.githubusercontent.com/assets/608719/8204435/ea4f80ac-14e8-11e5-814f-c4e41ffd3ade.png)

tgfrerer:
confirmed this running the latest ProjectgeneratorSimple from apps/

gorkacortazar:
Does not happen with an empty project, but in multilight example it happens when loading an image; it may be related to the uri discovery or path translation maybe?

ofTheo:
I just tried the multiLightExample on 10.9 with Xcode 6.1 and I don't get any crash. 
Could you try building in Debug and run it from Xcode and see if Xcode can catch where it is crashing. 

Could you also post the path you are running OF from on your Hard Drive i.e.: /Users/theo/Documents/OF/

arturoc:
@ofTheo what's crashing is xcode itself not the app i think, right?

tgfrerer:
yes. xcode 6.3.2 with me. on irc, btw =)

gorkacortazar:
@arturoc  yes, xcode closes and the app never runs. But you can run the compiled binary.
@ofTheo , the main difficulty is that Xcode crashes and closes before any breakpoint in the file when you try to debug. The clean ofx resides in:
/Volumes/DataHD/Projects/ofx_090Test/openFrameworks

The problem resides in the URL parsing segment in ofImage.cpp, if I comment the full block, everything runs perfect, in line 176 to 185.

gorkacortazar:
More specifically, in ofImage.cpp, line 178, is the one that causes the crash
`uri = network::uri(fileName);`

tgfrerer:
Might be a dependency on Poco, and openssl...

Maybe this is related to what prevents me from compiling in XCode 7 beta, where it refuses to compile saying: 
`/Users/tim/dev/of/libs/poco/include/Poco/Crypto/OpenSSLInitializer.h:26:10: 'openssl/crypto.h' file not found`

[...]

PR #4002, opened 2015-06-25
Fix for iOS Xcode crash when deploying to iOS simulator

danoli3:
- Fixes iOS Simulator deployment problem which was causing Xcode to crash

Related to https://github.com/openframeworks/openFrameworks/issues/3943

### Problem
- Internal issue with lib URI exposing an issue with the Xcode debugger

### Solution
- Remove libURI for iOS for the moment until fixed.
- Revert to Poco for libURI functions.


danoli3:
@arturoc this look good?
Probably merge straight away, as I think a lot of iOS developers are running master (based on the number of people on the forum thread ;D! ).

The libURI issue will have to be deeper looked into later.

bakercp:
Tested and confirmed to fix crash issue.  Thanks @danoli3.  I'd say go ahead and merge it since it is just using known code from a past release to patch.

arturoc:
thanks!

Issue #3855, opened 2015-05-20
soundPlayerExample and soundPlayerFFTExample crashes with SIGFPE signal


procedural:
Backtrace for `soundPlayerExample`:
```cpp
Program received signal SIGFPE, Arithmetic exception.
0x0000000000428f07 in ofOpenALSoundPlayer::load (this=this@entry=0x6762b90, fileName=...,
    is_stream=is_stream@entry=false) at openFrameworks/sound/ofOpenALSoundPlayer.cpp:448
448             int numFrames = buffer.size()/channels;
(gdb) bt
#0  0x0000000000428f07 in ofOpenALSoundPlayer::load (this=this@entry=0x6762b90, fileName=...,
    is_stream=is_stream@entry=false) at openFrameworks/sound/ofOpenALSoundPlayer.cpp:448
#1  0x0000000000404665 in ofSoundPlayer::load (this=this@entry=0x6752628, fileName=..., stream=stream@entry=false)
    at openFrameworks/sound/ofSoundPlayer.cpp:79
#2  0x0000000000401c59 in ofApp::setup (this=0x6752610) at examples/sound/soundPlayerExample/src/ofApp.cpp:6
#3  0x00000000007f9d1e in ofBaseApp::setup (this=0x0, args=...) at openFrameworks/app/ofBaseApp.h:63
#4  0x00000000007d7684 in Poco::PriorityDelegate<ofBaseApp, ofEventArgs, false>::notify (this=0x6763480,
    sender=<optimized out>, arguments=...) at Poco/Foundation/include/Poco/PriorityDelegate.h:148
#5  0x00000000007d886f in Poco::PriorityStrategy<ofEventArgs, Poco::AbstractPriorityDelegate<ofEventArgs> >::notify (
    this=0x22fc10, sender=0x0, arguments=...) at Poco/Foundation/include/Poco/PriorityStrategy.h:62
#6  0x00000000007b6f97 in notify (args=..., pSender=0x0, this=0x674adb8)
    at Poco/Foundation/include/Poco/AbstractEvent.h:246
#7  ofNotifyEvent<ofEvent<ofEventArgs>, ofEventArgs> (event=..., args=...)
    at openFrameworks/events/ofEventUtils.h:172
#8  0x0000000000491902 in ofCoreEvents::notifySetup (this=0x0) at openFrameworks/events/ofEvents.cpp:193
[...]

Issue #2581, opened 2013-09-22
Remove using namespace std


abergmeier:
Including the whole C++ standard library is known to be very [bad practice](http://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice).

tgfrerer:
In an effort to show my enthusiasm for this, and after reading http://gotw.ca/gotw/053.htm
I've spent five hours this afternoon refactoring the whole of oF core so that it would work without having `using namespace std` in `ofConstants.h`.

Here's the branch: 

https://github.com/tgfrerer/openFrameworks/tree/core-explicit-std

* It's only been tested on vs2015RC, so it's well possible that other platforms still have one or the other `std::` missing.

--

I'd really love to see namespaces in 1.0, since it will allow oF to interact with a whole lot of other libraries more easily. One of the hardest things to figure out are namespace shadowing, and header file include order, things we could keep in check using namespaces. 

I ran into some collisions with pugixml recently, and debugging them was not pretty. Mostly, because the compiler errors are often meaningless when namespace clashes occur.

I've also read the [oF forum discussion from seven years ago](http://forum.openframeworks.cc/t/openframeworks-namespace/1006) and would suggest that with 1.0 oF could be at a point in maturity (and complexity) where namespaces will do more good than harm.

A change like this is makes a whole lot of older PRs un-mergeable (and will also, if un-merged itself, code-rot faster than you can say "namespace"), simply because it touches so many lines (and almost every core oF file), but I'll be happy to scrap this branch and restart the refactor feast again once we are at a point where the codebase changes less fast.

bilderbuchi:
+100 to this. I think if we merge after 0.9.0 release, we got enough time to iron out non-merging PRs and other wrinkles caused by this change.

arturoc:
i think we should merge this, we can re-import std from for example ofMain.h by now so everything outside the core keeps working as it is right now but the core or anything that doesn't include ofMain directly needs to import std manually.

tgfrerer:
Let me check this option tomorrow in my testing branch. 

I've grouped the commits in two - and I believe the better (and less intrusive way of doing this) would be to apply the changes to the header files, but in some of the cpp files add a `using namespace std;`

My hope would be that a PR would then touch less files, and merging would be simpler, and the git history less noisy. That said, I think it's good practice to add std:: to anything from the stl in core code.

tgfrerer:
[...]

PR #4012, opened 2015-06-26
fixed 'deprecated-declarations' warnings in soundBufferExample

tpltnt:
yet another letfover from #3855.
Cheers
tpltnt

### DIFF
--- a/examples/sound/soundBufferExample/src/ofApp.cpp
+++ b/examples/sound/soundBufferExample/src/ofApp.cpp
@@ -20,7 +20,7 @@ void ofApp::update(){
 	// "lastBuffer" is shared between update() and audioOut(), which are called
 	// on two different threads. This lock makes sure we don't use lastBuffer
 	// from both threads simultaneously (see the corresponding lock in audioOut())
-	ofScopedLock lock(audioMutex);
+	unique_lock<mutex> lock(audioMutex);
 
 	// this loop is building up a polyline representing the audio contained in
 	// the left channel of the buffer
@@ -88,7 +88,7 @@ void ofApp::audioOut(ofSoundBuffer &outBuffer) {
 		pulsePhase += pulsePhaseStep;
 	}
 	
-	ofScopedLock lock(audioMutex);
+	unique_lock<mutex> lock(audioMutex);
 	lastBuffer = outBuffer;
 }
 
--- a/examples/sound/soundBufferExample/src/ofApp.h
+++ b/examples/sound/soundBufferExample/src/ofApp.h
@@ -27,7 +27,7 @@ class ofApp : public ofBaseApp{
 		double pulsePhase;
 		double sampleRate;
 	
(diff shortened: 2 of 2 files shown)


ofTheo:
I think from the discussion here ( see: #2581 ) the plan is to use std:: in core files but in examples and ofApp not include std:: as it will still have `using namespace std;` in ofMain.h 

ping @openframeworks/core 
@arturoc @tgfrerer to confirm

tgfrerer:
Yes, that makes sense. We can add `using namespace std` to some core **.cpp** files, especially when it makes code (and git diffs) more readable, but would strive to remove it from any core **header** files, as it  "infects" the global namespace in counter-intuitive ways.

So any use of `STL` elements would eventually need the `std::` prefix in core header files.

tpltnt:
@ofTheo thanks for pointing it out. I will read through it.

ofTheo:
@tgfrerer thanks for the quick reply 
It might be good to get this written up somewhere ( wiki? ) so its clear what our approach will be to the std:: namespace going forward. 

@tpltnt no problem! could you update the PR without the std:: prefix? 
thanks!!!

bilderbuchi:
> It might be good to get this written up somewhere ( wiki? ) so its clear what our approach will be to the std:: namespace going forward. 

"somewhere" should probably be the [code style](https://github.com/openframeworks/openFrameworks/wiki/oF-code-style) wiki page.

tpltnt:
@ofTheo i removed the `std::` prefix as you suggested. I also updated the [oF code style page](https://github.com/openframeworks/openFrameworks/wiki/oF-code-style) with the section "Namespaces (esp. 'std')" @tgfrerer @bilderbuchi.

ofTheo:
thanks @tpltnt !!

bilderbuchi:
@tpltnt unfortunately, you got it the other way round in the wiki page edit - `using namespace std` in core cpp files is OK (as this doesn't propagate via `#include`). In core _header_ files, we should use `std::` forms so the global namespace doesn't get polluted when those headers get included elsewehere, see [this comment](https://github.com/openframeworks/openFrameworks/pull/4012#issuecomment-117204680)

fwiw, I disagree with @tgfrerer a bit in that I think we should use `std::` _everywhere_, just for consistency's sake, so that there is no confusion about what version you can/should use where, etc.

tpltnt:
@bilderbuchi good catch, but the edit was shorter than your comment ;)

arturoc:
let's keep using no std in examples for 0.9. i think for next release moving `using namespace std` from `ofConstants.h` into `ofMain.h` could make it easier to transition for beginners. we could also add group headers per folder so including different modules is easier, like ofUtils.h would include everything in the utils folder. that way it's easier to include only what you need without having to include ofMain.h that way more advanced users have a way of not having all of std imported and still have an easy way to include the parts of OF they need

PR #4023, opened 2015-06-29
fix regex matching for shader error log

tgfrerer:
std::regex uses a slightly different API as did POCO, and
regex_search has to be used instead of regex_match to find
matching tokens in a string.


arturoc:
thanks! i tried this and it was working for me i guess this just uses search cause you don't need the matches right?

tgfrerer:
Oh, you need the matches, the first capture group extracts the offending line number from the driver error message.

To get the value of the first capture group, we need to use search rather than match. 

The line number is used to print only the lines surrounding the offending line. 

If that can't be extracted or the regex search fails it will print out the full shader source along the error message.




typed on tocuh

Issue #3983, opened 2015-06-24
Separate VS Builds


DomAmato:
Changing architectures requires openframeworks to be rebuilt. Should we maybe have them separated by architecture so that it would be libs\openFrameworksCompiled\lib\vs\(Win32/x64) 

I don't think its a big deal but it will cut down on recompiling each time someone switches architectures

bakercp:
If this is possible, it would be cool.  I ran into this issue today.


---
http://christopherbaker.net

DomAmato:
Changing the output directory and intermediate directories didn't work

bilderbuchi:
closed by #4025

PR #4025, opened 2015-06-30
Bugfix - Visual Studio intermediate paths

### LINKED ISSUE #4006: Safety : Intermediate paths for Visual Studio
Since now we have Win32 and x64 support....

If you switch between Win32 and x64, VS will build the intermediate files (.obj, .lib, etc) to the same locations. Which :

1. Can cause Visual Studio to get confused (resulting in needing to rebuild solution)
2. Each time you switch platform you overwrite the previous platform's files 

I suggest that we change in `openFrameworks.vcxproj`
image

to
image

This should be transparent to the user.

cc: @DomAmato @liquidzym

elliotwoods:
Resolve #4006, #3983


arturoc:
thanks!

PR #4028, opened 2015-06-30
DON'T MERGE bugfix : warnings in ofSoundBuffer

elliotwoods:
put static casts into the function definitions to remove the warnings on building ofSoundBuffer

### DIFF
--- a/libs/openFrameworks/sound/ofSoundBuffer.h
+++ b/libs/openFrameworks/sound/ofSoundBuffer.h
@@ -100,11 +100,11 @@ public:
 	/// resample by changing the playback speed, keeping the same sampleRate
 	void resample(float speed, InterpolationAlgorithm algorithm=defaultAlgorithm);
 	/// the number of channels per frame
-	int getNumChannels() const { return channels; }
+	int getNumChannels() const { return (int) channels; }
 	/// set the number of channels. does not change the underlying data, ie causes getNumFrames() to return a different result.
 	void setNumChannels(int channels);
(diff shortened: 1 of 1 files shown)


tgfrerer:
+1

bilderbuchi:
Wouldn't it be [better](http://stackoverflow.com/questions/103512/in-c-why-use-static-castintx-instead-of-intx) to use `static_cast` instead of `()`?

bakercp:
+1 for `static_cast<T>()`

elliotwoods:
i noticed you guys tend to do that. i just thought this looks less confusing.
but best stick with the style?

bilderbuchi:
actually, `static_cast` should be less confusing, since it clarifies what kind of cast you do, as opposed to the  `()` syntax. ;-)

arturoc:
getNumFrames should probably be size_t too since size() is also size_t. can you change it? 

in most of the code we use parenthesis style cast so i don't think it matters much unless you really want to use something very specific like dynamic_cast or const_cast

bilderbuchi:
> in most of the code we use parenthesis style cast so i don't think it matters much

so, maybe we should (slowly) do something about this? We're trying to teach beginners C++, shouldn't that include doing things in the way they "should" be done?

bakercp:
Hi all, I went a little deeper on this with https://github.com/openframeworks/openFrameworks/pull/4069.  No casts, just consistent / concise data types.

bilderbuchi:
I put a don't-merge warning in the title of this one, then.

arturoc:
solved by #4069

PR #4031, opened 2015-06-30
fix hard-coded glInternalformat

tgfrerer:
when allocating an ofTexture backed by an ofBufferObject, the glInternal format would always be hardcoded to GL_RGBA32F, regardless of the glInternalFormat parameter handed over through the method parameters.

This was probably accidentally leftover from a refactor.

ping @arturoc


arturoc:
ugh, yeah thanks!

PR #4037, opened 2015-07-01
rename ofTextureData::glTypeInternal

### LINKED ISSUE #4036: texData.glTypeInternal should be glInternalFormat
In OpenGL, a texture **format** is an enum specifying the number of color components and how these are laid out in memory, whilst a **type** refers to an enum specifying a data type such as `GL_FLOAT`,` GL_UNSIGNED_BYTE` etc. 

In openFrameworks `ofTextureData::glTypeInternal` is used to refer to the internal **format**, i.e. how data is laid out, but is named after the type. Let's fix this so that we're using the same (and not the flipped) terminology as in OpenGL, and change `ofTextureData::glTypeInternal` to `ofTextureData::glInternalFormat`.

cf. our call to `glTexImage2D` in line ofTexture.app#L401

	glTexImage2D(texData.textureTarget, 
		0,
		texData.glTypeInternal,
		(GLint)texData.tex_w,
		(GLint)texData.tex_h,
		0,
		glFormat,
		pixelType,
		0);

with the specs for glTexImage2D:

	void glTexImage2D(	GLenum target,
		GLint level,
		GLint internalFormat,
		GLsizei width,
		GLsizei height,
		GLint border,
		GLenum format,
		GLenum type,
		const GLvoid * data);
 
From: https://www.opengl.org/sdk/docs/man/html/glTexImage2D.xhtml

tgfrerer:
rename ofTextureData::glTypeInternal to ::glInternalFormat
for naming consistency with the underlying OpenGL enum.

Closes #4036


arturoc:
this is going to break some code but i think we should merge it with 0.9 since we are already breaking some stuff

tgfrerer:
yes, please! it would be nice to clear these things up - if only to make maintaining ofTexture a tad simpler =)

tgfrerer:
closing this due to code rot, will re-supply.

Issue #3999, opened 2015-06-24
64bit osx crashes on close (via escape) with imageLoaderWebExample.


bakercp:
32bit osx doesn't.  

tested both http and https.

bakercp:
I believe this was fixed with one of @arturoc recent PRs.

PR #4040, opened 2015-07-02
refactor ofThread to use std::thread instead of Poco::Thread

arturoc:
Solves #3999 but introducing std::thread in ofThread this late might be problematic, i don't mind moving #3999 and this PR back to next release.

This also removes some methods in ofThread which is not possible to implement with std::thread anymore like getting the current thread as an object or checking if the calling thread is the main thread. in any case using those methods were not very good practice i think.

std thread doesn't have a numeric id so this returns now std::thread::id instead of int we could keep track of each created thread using a static variable in the class but i think is better to just expose the native id

### DIFF
--- a/libs/openFrameworks/utils/ofThread.cpp
+++ b/libs/openFrameworks/utils/ofThread.cpp
@@ -10,10 +10,10 @@
 
 //-------------------------------------------------
 ofThread::ofThread()
-:_threadRunning(false)
-,_mutexBlocks(true)
-,threadBeingWaitedFor(false){
-   thread.setName("Thread " + ofToString(thread.id()));
+:threadRunningDone(make_pair(false,true))
+,mutexBlocks(true)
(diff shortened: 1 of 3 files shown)


danoli3:
Definitely better to just track the native id. In the case where a developed application requires to send information to specific ID's in a integer layout, they can store the thread reference in a vector or some structure and still do that. 

Also get main thread is technically possible with ```pthread_main_np(void);``` however only for POSIX based OS's such as iOS/OSX/FreeBSD/OpenBSD/Android. http://www.unix.com/man-page/freebsd/3/PTHREAD_MAIN_NP/ However, the NP part meaning non portable so only for those OS. 
So best solution is to remove as you have in this PR as other operating systems do not have this.

tpltnt:
@arturoc can you resolve the conflicts with the main branch? I give it a thumbs up otherwise and would like to see it merged soon.

PR #4043, opened 2015-07-02
leave one core for OS during compilation, see #3970

tpltnt:
Since @bilderbuchi rightfully raised the point in #3970, here is a modification to leave one CPU core to the OS when running `compileOF.sh`.
Cheers,
tpltnt


pizthewiz:
What the default were left alone but one could override with an environment variable?
```sh
% CORES=16 ./compileOF.sh
```

bilderbuchi:
should you go down that route, name it something recognizably OF-specific, though, like OF_COMPILATION_CORES, though.

tpltnt:
I merged the two approaches: You can now run the script alone and it will compile on only one core (as it did before). You can run `OF_COMPILATION_CORES=23 ./compileOF.sh` to compile on 23 cores or run `OF_COMPILATION_CORES=max ./compileOF.sh` to compile use the maximum number of cores (while leaving one for the OS).

bilderbuchi:
thank you for the update. a couple of points that I noticed:
* It would probably be better to call the `max` option `opt` or `optimal`, since you're actually _not_ compiling using the *maximum* number of cores, but some *optimal* number (i.e. leaving one for the OS)
* The help text should imo only be displayed when `OF_COMPILATION_CORES` is not given (as at that point you already know about it)
* Please remove the `sleep 10`, which is arbitrary and would mean a frustrating delay that *everyone* would incur, regardless of usage. Let's just display the information, and let the user act on it at their own pace (e.g. by aborting themselves during the compilation run, which should not be a problem)

arturoc:
thanks!

bilderbuchi:
@arturoc did you read the discussion before merging?  the `sleep 10` e.g. is still in there...

arturoc:
no, sorry i thought it was correct. i just reverted it, looking into it it's probably easier to just have a -j parameter that is directly passed to make instead of an env variable. in any case this script is not used that much any more since the makefiles in the projects already compile the core as a dependency which whatever -j flag is passed to them.

Issue #4042, opened 2015-07-02
VS 2015 not compiling after ut8 commit


gorkacortazar:
Building the openframeworks project report the following error and fails:
```
1>..\..\..\openFrameworks\utils\ofUtils.cpp(682): error C2780: 'bool std::isspace(_Elem,const std::locale &)': expects 2 arguments - 1 provided
1>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\locale(240): note: see declaration of 'std::isspace'
```
It happens in 32 and 64 bits architectures, after commit: #84793fa1f767bebbcae187d3e71a917f6b18675a

tgfrerer:
just tested it, i have this too on VS2105. could it be that vs2015 has a different signature for `std::isspace(c)` - it requires the locale to be given.

arturoc:
can't take a look right now but haven't changed that at all, i think it might be a missing include or something like that

tgfrerer:
i'm working on it - seems that the trim methods need a locale specified in the second param of `std::isspace()` on VS, will send a PR in a sec - but would be good to know if other platforms have the same signature for `std::isspace`...

bilderbuchi:
god we need some CI server soonest, I get the (subjective and wrong) feeling lately that half the issues/PRs are about issues that crop up because some introduced change was not tested on all our platforms. I think that could be easily avoided with some CI solution... :-(

bilderbuchi:
closed by #4044

PR #4044, opened 2015-07-02
add locale to trim methods

tgfrerer:
visual studio requires a locale to be given for std::isspace.

it seems that there are two signatures for std::isspace available, one in `<string>` and one in `<locale>`
but Visual Studio defaults to the one in `<locale>` 

http://en.cppreference.com/w/cpp/locale/isspace 
vs.
http://en.cppreference.com/w/cpp/string/byte/isspace

+ refactor getLocale to a helper method within `ofUtils.cpp` to keep things DRY

addresses #4042

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -643,14 +643,24 @@ utf8::iterator<const char*> ofUTF8Iterator::end() const{
 }
 
 //--------------------------------------------------
+// helper method to get locale from name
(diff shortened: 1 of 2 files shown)


tgfrerer:
I had to change the character type from `int` to `w_char_t` so that XCode would correctly initialise the template in the string iterator methods when using std::isspace with a locale. 

 tested on OS X, and VS, with `ofLog() << "'" << ofTrim("      hello world      \t") << "'";` =)

Issue #4035, opened 2015-07-01
ofTexture::allocate(const ofBufferObject &, int ) is ambiguous


tgfrerer:
Allocating a texture using an `ofBufferObject` implicitly (and silently) makes the ofTexture a **Buffer Texture** which is a very specialised type of 1D texture. This confused me royally. 

**Buffer textures** are useful for sampling large amounts of memory in shaders. Their texture target must be `GL_TEXTURE_BUFFER`. In modern OpenGL this largely duplicates the functionality of SSBOs. But in GLSL, `GL_TEXTURE_BUFFER` can only be accessed via `texelFetch`, which means, almost all oF texture methods (like draw, for example) are invalid for such a texture, but it won't be reported as an oF error or warning, only the OpenGL driver will whine.

What I would expect when "allocating" a texture using an `ofBufferObject` is a regular (2D) texture, with an `ofBufferObject` for data backing. This is specially useful when using the fastest path to upload data to a texture, using an`ofBufferObject` as a Pixel Buffer Object (PBO), and currently implemented in (1).

For disambiguation, i would suggest renaming the method: 

	void ofTexture::allocate(const ofBufferObject & buffer, int glInternalFormat);

to:

	void ofTexture::allocateAsBufferTexture(const ofBufferObject & buffer, int glInternalFormat);

so that at least the intention is clear from the outset that this texture is going to be treated as a special texture.

---

(1) The method:

	void ofTexture::loadData(const ofBufferObject & buffer, int glFormat, int glType);

effectively loads data for the texture from an ofBufferObject, which is what we can use for PBO transfers.

arturoc:
my idea was that if you were allocating a texture with a buffer you want a buffertexture. when you want a normal texture then you just use the normal method and send data to it using loadData(buffer). 

a 2d texture can't backed by a buffer but you can load data to it from a buffer. a buffertexture though uses glTexBuffer which according to: https://www.opengl.org/sdk/docs/man/html/glTexBuffer.xhtml must have target GL_TEXTURE_BUFFER

but if you think it's confusing i'm totally ok changing the name to something more obvious `allocateTexBuffer`? so it's in line with the GL call?

tgfrerer:
Yes, my point is more about making the **intent** more explicit. I was originally following the red herring of believing it was a regular overloaded method.

[...]

PR #4045, opened 2015-07-02
rename ofTexture::allocate for texture buffers

tgfrerer:
When allocating a texture as a texture buffer, the texture
cannot be used as a vanilla texture anymore, but must be
used as a buffer texture.

This name change in the hope it might help disambiguation.

Addresses #4035 

+ Also updated textureBufferInstancedExample so it uses the new signature


arturoc:
thanks!

PR #4046, opened 2015-07-02
Apothecary formula for Assimp and Visual Studio 2015.

gorkacortazar:
Tested with assimpExample, 32 and 64 bits.

Important comments:
* It uses the Assimp version that was on the previous formula, but some models in the assimpExample behave weirdly; in particular the astronaut that appears really small, I think that @ofTheo  had that behavior before?
* It changes de default Assimp32.dll to assimp.dll (this was the default export name). Some of the core DLL's uses the same name in both architectures (Zlib.dll for example) and others not (fmodex64 and fmodex); should I revert to the old name? For me makes no sense a 64bit dll with a 32 in the filename.
* The formula creates now two lib folders, one per architecture, as I saw in similar core formulas, but the Project Generator adds all the libs folders to all the architectures and needs to be solved manually. Should we this the new standard for VS2015 and the support for 32/64 bits builds in addons? Or there is a way to combine them in a single lib file?


elliotwoods:
skimming through this
i'd say the standard should be

```
ofxAddon/libs/libName/lib/vs/x64
ofxAddon/libs/libName/lib/vs/Win32
ofxAddon/libs/libName/bin/vs/Win32
ofxAddon/libs/libName/bin/vs/Win32
```

Ideally the lib names / dll names should be the same in both filders, but it's not necessary (best to follow the manufacturer's defaults)

PR #4008, opened 2015-06-26
Travis Automated Testing of the Core

danoli3:
#### Automated PR / Commit and Merge testing of Core

Original Discussion here from 2012: ```Continuous integration/testing``` https://github.com/openframeworks/openFrameworks/issues/1068

- iOS Empty Example + openFrameworks Core iOS Build Testing

- OSX Empty Example + openFrameworks Core OSX Build Testing
- Adds a tick or cross to every PR for fail or success (once configured on Travis-ci website) (You may remember this happening a while ago when travis was added for documentation generation).

- Removes Documentation Automation - @HalfdanJ had some travis code to generate http://docs.openframeworks.cc automatically on merges. 
This should be done another way, travis is integration testing, even the Github API is designed for this.

Issue #1068, opened 2012-03-13
Continuous integration/testing


gabrielstuff:
Hey !

Me again. Didn't find anywhere this infos. I'm currently testing the travis-ci.org service.
As many people are working on the project and as many many issue are raised, and fixed, I thought it could be great to add some test and thus the travis service.

It would avoid the problem met here :
https://github.com/openframeworks/openFrameworks/commit/7ca7833ea1afb6bd5a6c54031e3fa688aa0c0ba8
and the discussion here :
https://github.com/openframeworks/openFrameworks/issues/804
or there :
https://github.com/openframeworks/openFrameworks/pull/921

I read that :

>feel free to make a unitTests folder at the root level of OF -
> that is where all unitTests will go.

But didn't find them.

The travis service will permit to get the little badge we know well : 
[![Build Status](https://secure.travis-ci.org/soixantecircuits/openFrameworks.png?branch=master)](http://travis-ci.org/soixantecircuits/openFrameworks)

Also, why not adding some spec test, to ensure future development of features.
I'm currently reading :
http://www.squidoo.com/cplusplus-behaviour-driven-development-tools#module124841511
http://sourceforge.net/apps/mediawiki/turtle/index.php?title=Turtle

For those who want some other nice reading :

http://gamesfromwithin.com/exploring-the-c-unit-testing-framework-jungle

The travis test is here, currently it is a total fake, but I just proposed the idea of a test driven development
http://travis-ci.org/#!/soixantecircuits/openFrameworks/builds/854259

danoli3:
For Windows testing we can use http://www.appveyor.com/

danoli3:
After seeing @bilderbuchi 's comment about automated testing.. I setup a quick test branch for this:

- Branch: https://github.com/danoli3/openFrameworks/tree/Automated/Travis-OSX
- Travis Status: https://travis-ci.org/danoli3/openFrameworks/branches
- I had to remove the documentation compiler travis code (not sure why this is a travis thing in the core).
- Testing iOS and Testing OSX.
- Using the template/emptyExample project for both

Limitations we have with Travis are the time per build (max 1 hour). 
So I thought for now the emptyExample is perfect. It validates the core has not been broken by a commit.

Testing this I created another branch from this, one which should soon fail.
- https://github.com/danoli3/openFrameworks/tree/Automated/Travis-CoreFail
- https://travis-ci.org/danoli3/openFrameworks/builds/68468345

[...]

PR #4047, opened 2015-07-02
Travis Automated Testing of the Core - 2

danoli3:
#### Automated PR / Commit and Merge testing of Core 

Original Discussion here from 2012: ```Continuous integration/testing``` https://github.com/openframeworks/openFrameworks/issues/1068

Previous PR here: https://github.com/openframeworks/openFrameworks/pull/4008

Builds using Travis's Build Matrix System:
The current automated scripts to test are:

- iOS Empty Example + openFrameworks Core iOS Build Testing with Xcode

- OSX Empty Example + openFrameworks Core OSX Build Testing with Xcode

- Documentation Automation -  included @HalfdanJ 's changes for building the docs, however split into a seperate script. (Does the same functionality as current) This will only run a docs build for the master branch and not for PR's. (normally will just return success for this matrix).

This will add a success tick or cross to every PR for fail or success (once configured on Travis-ci website) (You may remember this happening a while ago when travis was added for documentation generation).
Also gives piece of mind to new users checking out the project for the first time when we include the travis success badge to the readme.  [Build Status](https://travis-ci.org/danoli3/openFrameworks)

Something to think about adding in future as @bilderbuchi has mentioned. 
- More matrix builds for other types like make system testing.
- GUI testing (or validating an app will run).

---------------------------------

#### Travis Status: https://travis-ci.org/danoli3/openFrameworks/branches [Build Status](https://travis-ci.org/danoli3/openFrameworks) 

travis

----------------------------

So I thought for now the emptyExample is perfect. It validates the core has not been broken by a commit.

Testing this I created another branch from this PR, one which I tested making some changes that should soon fail and be caught by travis: 
#### Travis Status:  https://travis-ci.org/danoli3/openFrameworks/builds/69352852 [Build Status](https://travis-ci.org/danoli3/openFrameworks)
travis2

Clicking on the the build description in Travis gives you a full log of the console for the build. So finding an error can be very easy to track.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,20 @@
-language: node_js
+
+language: objective-c
+compiler: clang
 node_js: '0.10'
 env:
   global:
     - secure: kjz286d6rs2VHpv4WnxAOYXcdTIaTaABZ0SNXf27r5H+NDu5qx8ZSeCmj/9op/sDjTtpkLKhi2a0njUCWwak28RAWP3dNkT92MkUPVKBkevznSxWeqwLHoUos3AJnhPMB+cwrWvmRC+Ljt5K1q8SxQsS/HXM9pzKL3DG3qxM+eo=
-branches:
-  only:
-    - master
-notifications:
-  email:
-    - jonas@halfdanj.dk
+  matrix:
+    - NAME="OSX" BEFORE_THE_INSTALL="" SCRIPT_TO_RUN="scripts/osx/testTemplate.sh" AFTER_THE_SUCCESS="echo 'openFrameworks OSX Build Passing'"
+    - NAME="iOS" BEFORE_THE_INSTALL="" SCRIPT_TO_RUN="scripts/ios/testTemplate.sh" AFTER_THE_SUCCESS="echo 'openFrameworks iOS Build Passing'"
+    - NAME="Docs" BEFORE_THE_INSTALL="scripts/dev/documentation.sh before_install" SCRIPT_TO_RUN="scripts/dev/documentation.sh script" AFTER_THE_SUCCESS="scripts/dev/documentation.sh after_success"
 before_install:
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get install python-software-properties; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-add-repository -y ppa:libreoffice/libreoffice-4-2; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get update; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get install doxygen; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get install ncftp; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then gem install sass --version "=3.2.12"; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then gem install compass --version "=0.12.2"; fi'
+    $BEFORE_THE_INSTALL
 script:
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then cd scripts/; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then git clone https://github.com/halfdanJ/ofDocGenerator; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then cd ofDocGenerator; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then npm install; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then npm run start; fi'
+    $SCRIPT_TO_RUN
 after_success:
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" && $FTP_USER ]]; then ncftpput -R -v -u $FTP_USER -p $FTP_PASSWORD 104.130.212.175 / output/*; fi'
+    $AFTER_THE_SUCCESS
+git:
+    depth: 1
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,20 @@ env:
   global:
     - secure: kjz286d6rs2VHpv4WnxAOYXcdTIaTaABZ0SNXf27r5H+NDu5qx8ZSeCmj/9op/sDjTtpkLKhi2a0njUCWwak28RAWP3dNkT92MkUPVKBkevznSxWeqwLHoUos3AJnhPMB+cwrWvmRC+Ljt5K1q8SxQsS/HXM9pzKL3DG3qxM+eo=
   matrix:
-    - NAME="OSX" BEFORE_THE_INSTALL="" SCRIPT_TO_RUN="scripts/osx/testTemplate.sh" AFTER_THE_SUCCESS="echo 'openFrameworks OSX Build Passing'"
-    - NAME="iOS" BEFORE_THE_INSTALL="" SCRIPT_TO_RUN="scripts/ios/testTemplate.sh" AFTER_THE_SUCCESS="echo 'openFrameworks iOS Build Passing'"
-    - NAME="Docs" BEFORE_THE_INSTALL="scripts/dev/documentation.sh before_install" SCRIPT_TO_RUN="scripts/dev/documentation.sh script" AFTER_THE_SUCCESS="scripts/dev/documentation.sh after_success"
+    - NAME="osx" \
+        SCRIPT_PATH="scripts/$NAME" \
+        SCRIPT="$SCRIPT_PATH/testTemplate.sh"
+    - NAME="ios" \ 
+        SCRIPT_PATH="scripts/$NAME" \
+        SCRIPT="$SCRIPT_PATH/testTemplate.sh"
+    - NAME="docs" \
+        SCRIPT_PATH="scripts/dev" \
+        SCRIPT="$SCRIPT_PATH/documentation.sh"
 before_install:
-    $BEFORE_THE_INSTALL
+    "if [ \"$NAME\" == \"docs\" ]; then scripts/dev/documentation.sh before_install; fi"
 script:
-    $SCRIPT_TO_RUN
+    "$SCRIPT"
 after_success:
-    $AFTER_THE_SUCCESS
+    "if [ \"$NAME\" == \"docs\" ]; then scripts/dev/documentation.sh after_success; fi"
 git:
-    depth: 1
-
+    depth: 10
--- a/scripts/dev/documentation.sh
+++ b/scripts/dev/documentation.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
-
+set -e
 ## This script is to build the OpenFrameworks documentation automatically
-
-
 echo "Current Branch: $TRAVIS_BRANCH"
 ### Only build documentation on the master branch.
 if [ "$TRAVIS_BRANCH" == "master" ]; then
@@ -10,7 +8,7 @@ if [ "$TRAVIS_BRANCH" == "master" ]; then
         if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then 
             echo "Documentation Before Install"
             brew update 
-            brew install libreoffice;
+            brew install Caskroom/cask/libreoffice;
             brew install doxygen;
             brew install ncftp;
             gem install sass --version "=3.2.12";
@@ -26,9 +24,9 @@ if [ "$TRAVIS_BRANCH" == "master" ]; then
             npm run start; 
         fi
     elif [ "$1" == "after_success" ]; then
+        echo "Uploading to FTP"
         if [[ $TRAVIS_PULL_REQUEST == "false" && $FTP_USER ]]; then
             ncftpput -R -v -u $FTP_USER -p $FTP_PASSWORD 104.130.212.175 / output/*;
         fi
     fi
-    return 0;
 fi
--- a/scripts/dev/documentation.sh
+++ b/scripts/dev/documentation.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+## This script is to build the OpenFrameworks documentation automatically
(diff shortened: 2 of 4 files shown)


arturoc:
seems you can use several os, perhaps you can add linux support and leave the documentation part running in linux as it was? i can then add tests to compile under linux and osx with makefiles: http://docs.travis-ci.com/user/multi-os/

the most problematic is usually visual studio so it would be good to have something to test there, rackspace where we host the web has windows servers i think so we could setup one there.

danoli3:
Currently Multi-OS with Travis has been turned off which sucks.
``` We are not currently accepting multi-OS requests #

We are at capacity with Mac workers in both .org and .com environments. Consequently, we are not accepting requests for the multi-OS feature at this time. As more capacity is added, we will resume accepting requests.
```

For Windows we could look at using http://www.appveyor.com/ (Same idea as travis however for Windows).

bilderbuchi:
But, we are already using a mac worker, and it does not seem linux workers are in short supply?

The two build scripts are virtually identical, wouldn't it be better to unify them and disambiguate with an environment variable or CLI argument?

.travis.yml
```diff
@@ -1,27 +1,20 @@
...
   global:
     - secure: kjz286d6rs2VHpv4WnxAOYXcdTIaTaABZ0SNXf27r5H+NDu5qx8ZSeCmj/9op/sDjTtpkLKhi2a0njUCWwak28RAWP3dNkT92MkUPVKBkevznSxWeqwLHoUos3AJnhPMB+cwrWvmRC+Ljt5K1q8SxQsS/HXM9pzKL3DG3qxM+eo=
-branches:
-  only:
-    - master
-notifications:
-  email:
-    - jonas@halfdanj.dk
+  matrix:
+    - NAME="OSX" BEFORE_THE_INSTALL="" SCRIPT_TO_RUN="scripts/osx/testTemplate.sh" AFTER_THE_SUCCESS="echo 'openFrameworks OSX Build Passing'"
```
bilderbuchi: the after_success step seems superfluous, and additional build step just for a success message which we will get from travis anyway?

.travis.yml
```diff
@@ -1,27 +1,20 @@
...
+    - NAME="Docs" BEFORE_THE_INSTALL="scripts/dev/documentation.sh before_install" SCRIPT_TO_RUN="scripts/dev/documentation.sh script" AFTER_THE_SUCCESS="scripts/dev/documentation.sh after_success"
 before_install:
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get install python-software-properties; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-add-repository -y ppa:libreoffice/libreoffice-4-2; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get update; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get install doxygen; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then sudo apt-get install ncftp; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then gem install sass --version "=3.2.12"; fi'
-  - 'if [[ $TRAVIS_PULL_REQUEST == "false" ]]; then gem install compass --version "=0.12.2"; fi'
+    $BEFORE_THE_INSTALL
```
bilderbuchi: Instead of adding non-used variables in the `matrix` section (And having super long lines), it would be better to get the build logic into the yaml-sections like `before_install:`. for example, you could drop the `BEFORE_INSTALL` variable above, and say here (pseudocode) `if $NAME == "docs" run scripts/dev/documentation.sh before_install`. similar in the other steps. you could also, if you use lowercase `NAME`s, use that to select the correct path to the build script.

bilderbuchi:
also, couldn't we use the makefile system on OSX? That way we wouldn't have to create/maintain yet another build script?

arturoc:
we should test both, using the makefile to compile you don't test if the osx project is broken for some reason.

bilderbuchi:
good point.

also. @danoli3 I don't think the documentation script is working, looking at  [a build](https://travis-ci.org/danoli3/openFrameworks/jobs/69352855) the run durations are all 0.00s, apparently this doesn't excecute anything.

scripts/dev/documentation.sh
```diff
@@ -0,0 +1,34 @@
...
+            cd ofDocGenerator; 
+            npm install; 
+            npm run start; 
+        fi
+    elif [ "$1" == "after_success" ]; then
+        if [[ $TRAVIS_PULL_REQUEST == "false" && $FTP_USER ]]; then
+            ncftpput -R -v -u $FTP_USER -p $FTP_PASSWORD 104.130.212.175 / output/*;
+        fi
+    fi
+    return 0;
```
bilderbuchi: this script always returns 0, independent of outcome, so we never would notice a failing build...

.travis.yml
danoli3: This Variable is needed for the Documentation side of things. (which sends the data via FTP). 
Totally can cull the echo from the others as it is only shown in the log.

scripts/osx/testTemplate.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+OF_DIR=`pwd`;
+echo $OF_DIR;
+TEST_DIR="$OF_DIR/scripts/osx"
+cd $TEST_DIR
```
bilderbuchi: do we need all that directory hopping? can't you supply a path to xcodebuild (i don't know), so only would have to execute xcodebuild in the build step and be done with it (and get rid of this script)?

bilderbuchi:
sorry for the barrage of feedback @danoli3 , I'm eager to get something nice like this polished and merged :D

.travis.yml
bilderbuchi: ok, but then wouldn't you better check for the correct name in the `after_success` section, and run the relevant code?

.travis.yml
danoli3: That's true, I like the idea of using the lowercase names. The NAME was literally an afterthought to make it easier to discern the build matrix apart in the website.

This was making it easier to adapt in future if the iOS / OSX needed a before_install command run.

scripts/dev/documentation.sh
danoli3: This is due to it not being on "master", it will only build the docs on master.
Let me test this by merging this into my fork master for now.

scripts/osx/testTemplate.sh
danoli3: Yeah I can re-write this without that cd.

.travis.yml
bilderbuchi: I got the feeling this would better be explicitly handled in the yml file - then it would be immediately clear what is run when, etc, instead of hiding this in some script or Var.

scripts/dev/documentation.sh
bilderbuchi: better select this in the yml file, though, instead of having a no-op script, don't you think? see e.g. http://stackoverflow.com/a/21941526/599884

scripts/osx/testTemplate.sh
bilderbuchi: then we could even totally drop the script, and execute xcodebuild directly from the yml, no?

scripts/dev/documentation.sh
danoli3: If there is an error along the way it will exit straight away and fail. However good you pointed this one out as that should be exit 0; D:

bilderbuchi: > If there is an error along the way it will exit straight away and fail.

Oh, I thought you need to have  `set -e` in the bash script for that to happen?

danoli3:
@bilderbuchi All good :D! I found a potential issue as well with ```git depth 1``` if you make commits in fast succession and Travis is still building a the previous commit, however hasn't got to the point of checking out the repo hash for that build yet, git depth 1 doesn't allow for it to checkout the hash and it errors. (as the HEAD of that branch is now more than git 1 depth).

The idea behind setting git depth 1 is to only checkout the latest, however I think it may be wise to change this to 5-10.

.travis.yml
danoli3: Travis follows the same formula though. Check here:
https://travis-ci.org/danoli3/openFrameworks/jobs/69404953

Travis will automatically hide anything that happens in the before_install stage regardless, if you click on the variable in the travis log, it opens up to reveal anything that happened then. 
This happens regardless of the variable or not.

bilderbuchi: I'm sorry I don't get what you are saying.

As another point, check out the "description" of the builds [here](https://travis-ci.org/danoli3/openFrameworks/builds/69404952) - the stuff after the box icon - it's not really useful to anyone looking at the build matrix, only having the names there would be much better (and meaningful)

scripts/osx/testTemplate.sh
danoli3: We could yes. I can try this. I just didn't want the YML to be confusing and the added benefit that the script would work independently of travis and could be run by anyone to do a quick test.

bilderbuchi: > script would work independently of travis

ah, good point. however, afaict, this script only works when executed from the OF root dir, not from the scripts directory, due to the pwd..

danoli3: Yeah I'll fix that and should be good to keep those scripts then.

arturoc: yes let's keep the build scripts that way it's easier to test the same builds travis is doing from any machine

scripts/dev/documentation.sh
danoli3: It seems to be throwing errors without this, I'll add that just in case.

scripts/osx/testTemplate.sh
danoli3: So I have now in travis: 
``` 
    - NAME="osx" \
        SCRIPT_PATH="scripts/$NAME" \
        SCRIPT="$SCRIPT_PATH/testTemplate.sh"
    - NAME="ios" \ 
        SCRIPT_PATH="scripts/$NAME" \
        SCRIPT="$SCRIPT_PATH/testTemplate.sh"
    - NAME="docs" \
        SCRIPT_PATH="scripts/dev" \
        SCRIPT="$SCRIPT_PATH/documentation.sh"
```

This shows in the build matrix exactly where and what is being tested.
Then in each of the testTemplate scripts
```
if [ "$SCRIPT_PATH" == "" ]; then SCRIPT_PATH="."; fi
echo "Looking for iOS Template Project"
if [ ! -d "$SCRIPT_PATH/template/bin/emptyExample.app" ]; then
```

danoli3:
@arturoc I'm currently testing changes on danoli3/openframeworks master (going to reset it later)

arturoc:
ok perfect, i'm testing linux (i have an idea to make it test both platforms) and also visual studio on appveyor.

arturoc:
vs works great on appveyor: https://github.com/arturoc/openFrameworks

scripts/osx/testTemplate.sh
bilderbuchi: It can even be more clear/concise like this:

```
matrix:
  - NAME="osx" SCRIPT="testTemplate.sh"
  - NAME="ios" SCRIPT="testTemplate.sh"
  - NAME="doc" SCRIPT="documentation.sh"
before_install:
  "if [ \"$NAME\" == \"doc\" ]; then scripts/$NAME/$SCRIPT before_install; fi"
script:
  scripts/$NAME/$SCRIPT
after_success:
  "if [ \"$NAME\" == \"doc\" ]; then scripts/$NAME/$SCRIPT after_success; fi"
```
the only necessary change is to move the documentation script into a scripts/doc subfolder, which is probably a reasonable thing to do anyway.
Script is not tested since I can't do that right now. Also, I think you can get rid of most of those quotes and escaped quotes, no? (from looking at other travis.yml files)

bilderbuchi: although, under the danger or starting to bikeshed now, I think it would even make more sense to have one big script to orchestrate all the testing, which selects based on given arguments, e.g. like this:

```
matrix:
  - NAME="osx"
  - NAME="ios"
  - NAME="doc"
before_install:
  "if [ \"$NAME\" == \"doc\" ]; then ./scripts/ci_script.sh $NAME before_install; fi"
script:
  ./scripts/ci_script.sh $NAME
after_success:
  "if [ \"$NAME\" == \"doc\" ]; then ./scripts/ci_script.sh $NAME after_success; fi"
```

this is most clear, all the uninteresting bookkeeping is in the script, the script can also be called by a user locally like `ci_script.sh ios` (kinda like we use apothecary now), and the env var list is not polluted/can easily be matrixed with OSes etc. later.
@danoli3 @arturoc what do you think?

danoli3:
Okay I've added the changes to this PR for Travis.

bilderbuchi:
@arturoc nice! where is the build config, though?
@all: take care, when adding build badges, that those reflect only the *master* build status (you can select that with the URL), otherwise some random PR status will influence the badge in the readme on the GH mainpage

arturoc:
there's no build config you just need to tell it to use a certain vs project by setting a path on their config. you can also add a build config but by now with that is enough to have a badge and check the PRs

danoli3:
@arturoc change the git depth to 10 for AppVeyor for that same reason I mentioned above. If you commit  a new commit on the same branch before the last test has completed it will error.

arturoc:
yeah sure, i wonder if there's a way to tell this tools to clone a specific hash instead of depth -1 but yes just using depth 5 or 10 should be safe i guess

danoli3:
I think that is managed at least from Travis's point of view by Travis itself. Nothing we can do. I wonder if theres a bug report. I'll have a look

arturoc:
can we merge this then?

danoli3:
Yeah looks good!

danoli3:
@arturoc just change the readme to this:

```
OSX / iOS   [![Build Status](https://travis-ci.org/openframeworks/openFrameworks.svg?branch=master)](https://travis-ci.org/openframeworks/openFrameworks)

```

Current one on there points to danoli3/openframeworks travis
I've also fixed up Travis-Ci to now build for Pushes / Pull Requests. So you should see it start working on the next push to master / PR.

danoli3:
Actually I'll make this change directly so it gets a first build of master going.

kylemcdonald:
omg i can't believe it! i look away from the repo for a day and this happens! huge congratulations!

PR #4057, opened 2015-07-06
iOS Touch Event types were not set, always default to "down".

bakercp:


bakercp:
@openframeworks/ios ping

bakercp:
ping @arturoc seems my last ping didn't happen.

julapy:
that looks good to me.
will merge now.

bakercp:
Thanks @julapy.

PR #4058, opened 2015-07-08
Fixes #2605.

### LINKED ISSUE #2605: Default ofLoopMode differs between QT and GST Players.
In `ofGstUtils`, `loopMode` is inited with `OF_LOOP_NONE` whereas in `ofQuicktimePlayer` and `ofQTKitPlayer` `currentLoopState` is inited to `OF_LOOP_NORMAL`.

bakercp:


arturoc:
it would be better if only the player set this instead of ofGstUtils. ofGstUtils is used for a lot other stuff in addons... and making it default to loop could break some things.

bakercp:
Makes sense.  Closing this PR and the associated issue.

PR #4063, opened 2015-07-08
Addition of std::make_unique for unsupported platforms.

### LINKED ISSUE #4048: Add an implementation of std::make_unique<>?
Hi all, I'm really enjoying these new C++11 capabilities and thinking through teaching this in the fall.  One small syntactic frustration (shared by others e.g. [here](https://stackoverflow.com/questions/7038357/make-unique-and-perfect-forwarding)) is that unlike `std::shared_ptr`, `std::unique_ptr` has no method like `std::make_unique` in C++11.  Apparently it was an oversight and will be added in [C++14](http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique).  Anyway there are quite a few reasonable implementations out there for this helper method.  I've been using something like this:

```c++
#if __cplusplus <= 201103L

// Implementation from http://stackoverflow.com/a/13512344/1518329
namespace std {


template <typename T, typename... Args>
std::unique_ptr<T> make_unique_helper(std::false_type, Args&&... args) {
    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
}

template <typename T, typename... Args>
std::unique_ptr<T> make_unique_helper(std::true_type, Args&&... args) {
    static_assert(std::extent<T>::value == 0,
                  "make_unique<T[N]>() is forbidden, please use make_unique<T[]>().");

    typedef typename std::remove_extent<T>::type U;
    return std::unique_ptr<T>(new U[sizeof...(Args)]{std::forward<Args>(args)...});
}

template <typename T, typename... Args>
std::unique_ptr<T> make_unique(Args&&... args) {
    return make_unique_helper<T>(std::is_array<T>(), std::forward<Args>(args)...);
}


} // namespace std

#endif
```

Anyway, would it be a terrible idea to add something like that to openFrameworks for 0.9.0? @arturoc @ofTheo @ofZach

bakercp:
Running this through tests now.  Feel free to test it!

Fixes #4048

### DIFF
--- a/libs/openFrameworks/types/ofTypes.h
+++ b/libs/openFrameworks/types/ofTypes.h
@@ -284,3 +284,36 @@ public:
 //----------------------------------------------------------
 template <typename T>
 using ofPtr = std::shared_ptr<T>;
+
+
+// This is a helper method for make unique on platforms that support C++11, but not C++14.
+#if !defined(NO_OF_MAKE_UNIQUE) && (__cplusplus <= 201103L || _MSC_VER < 1800)
+
+// Implementation for C++11 platforms that do not yet have std::make_unique.
+// Implementation from http://stackoverflow.com/a/13512344/1518329
+namespace std {
+
+
+template <typename T, typename... Args>
+std::unique_ptr<T> make_unique_helper(std::false_type, Args&&... args) {
+	return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+
+template <typename T, typename... Args>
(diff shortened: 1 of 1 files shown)


bakercp:
Tested on iOS, OSX, Linux.

bakercp:
Tested successfully on VS2015.  I don't think this will be compatible with VS2012, but should be compatible with VS2013.  VS2013 / VS2015 both have make::unique_ptr implemented already.  Basically this is just for OSX and iOS.

PR #4066, opened 2015-07-08
fix #3722. Maintain global transform fixed when setting parent.

### LINKED ISSUE #3722: ofNode setParent maintaining global position doesn't work as expected

chuckleplant:
Replaces PR #3723

### DIFF
--- a/libs/openFrameworks/3d/ofNode.cpp
+++ b/libs/openFrameworks/3d/ofNode.cpp
@@ -15,12 +15,12 @@ ofNode::ofNode()
 //----------------------------------------
 void ofNode::setParent(ofNode& parent, bool bMaintainGlobalTransform) {
     if(bMaintainGlobalTransform) {
-        ofMatrix4x4 globalTransform(getGlobalTransformMatrix());
-        this->parent = &parent;
-        setTransformMatrix(globalTransform);
-    } else {
-        this->parent = &parent;
-    }
+		ofMatrix4x4 postParentGlobalTransform = getGlobalTransformMatrix() * parent.getGlobalTransformMatrix().getInverse();
+		this->parent = &parent;
+		setTransformMatrix(postParentGlobalTransform);
+	} else {
+		this->parent = &parent;
+	}
 }
 
 //----------------------------------------


libs/openFrameworks/3d/ofNode.cpp
```diff
@@ -15,12 +15,12 @@ ofNode::ofNode()
 //----------------------------------------
 void ofNode::setParent(ofNode& parent, bool bMaintainGlobalTransform) {
     if(bMaintainGlobalTransform) {
-        ofMatrix4x4 globalTransform(getGlobalTransformMatrix());
-        this->parent = &parent;
-        setTransformMatrix(globalTransform);
-    } else {
-        this->parent = &parent;
-    }
+		ofMatrix4x4 postParentGlobalTransform = getGlobalTransformMatrix() * parent.getGlobalTransformMatrix().getInverse();
```
arturoc: isn't this just the same as the else branch? multiplying the current global transformation by the inverse of the parent's global will just return the current local no?

chuckleplant: No because the incoming node is not yet the parent.

arturoc: oh, yeah you are right.

arturoc:
looks correct to me but not sure why the original implementation was like that. ping @memo who i think did it.

memo:
what seems to be the problem?

The idea was if you give a node a new parent, but tell it to maintain global transform, it needs to update the nodes *localtransform* according to its *new* parent transform to stay in the new global transform. Whereas if you tell it to NOT maintain global transform, it maintains its *local* transform, gets attached to the new parent, and hence will have a new global transform.

memo:
oh i see, the issue is in setTransformMatrix(globalTransform);
weird, I'm sure I'd tested this when I wrote it.

I think the problem is in the ambiguity of setTransformMatrix. Is it setting local transform or global? (looking at the code I can see it is setting the local).

For everything else we have setGlobalXXX vs setLocalXXX. (And we have getLocalTransform and getGlobalTransform).

perhaps we should do the same for setLocalTransformMatrix and setGlobalTransformMatrix (and keep setTransformMatrix calling setLocalTransformMatrix for backwards compatibility, but deprecate it?)

chuckleplant:
+1, there is indeed an ambiguity with that function.

arturoc:
i'm going to merge this then and we can discuss if we need setGlobalTransformMatrix in a different issue. @chuckleplant or @memo can you open it?

memo:
sure will do

PR #4067, opened 2015-07-08
Fixes #1930.  Replaces #4062. (unsigned long long -> uint64_t)

### LINKED ISSUE #1930: int ofGetFrameNum() to unsigned long long ofGetFrameNum()
Like our recently changed elapsed time functions, `ofGetFrameNum()` could roll over more quickly than expected.

As noted previously by @kylemcdonald 

"If your app is running at 1000fps then ofGetFrameNum() is effectively going to run over at the same point as ofGetElapsedTimeMillis(). so yes, i think it makes sense for it to be switched to an unsigned long long. in general i'm against adding unnecessary precision to things, but i think here it makes sense to be consistent."

This is slightly more work than the elapsed time upgrade because it requires windowing systems to be updated as well.

bakercp:
New PR to eliminate extra commits.

### DIFF
--- a/examples/graphics/imageSequenceExample/src/ofApp.cpp
+++ b/examples/graphics/imageSequenceExample/src/ofApp.cpp
@@ -77,9 +77,8 @@ void ofApp::draw() {
     // this is the total time of the animation based on fps
     //float totalTime = images.size() / sequenceFPS;
     
-    
-    int frameIndex = 0;
-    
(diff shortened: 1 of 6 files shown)


bakercp:
@arturoc I also changed the return type of `unsigned long long ofFpsCounter::getLastFrameNanos()` to match the internal member type.

PR #4071, opened 2015-07-08
URLFileLoader const corrections.

bakercp:


bakercp:
Hooray for CI!

bakercp:
Tested with image loader web example and found small logic mistake.  Fixed it with 974ab82 so it should all be working now.

PR #4075, opened 2015-07-08
Remove unused variables.

bakercp:
Removing these unused variables and using correct underlying types greatly reduces warnings.


bakercp:
Hm.  Just compiled and ran this on iOS in Xcode.  Not sure why travis error'd out.

bakercp:
Not sure why iOS error'd out.  It compiled on Xcode just fine.  Looks like the build just timed out:

```
No output has been received in the last %!f(time.Duration=3000000000000) minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

arturoc:
yes it just errored out, it should be fine, i think you should be able to ask travis to rebuild that particular job if you want to be sure

bilderbuchi:
they had problems with the mac infrastructure earlier today, maybe that's recurring now.

bakercp:
@bilderbuchi I quickly google'd how to restart a job, but couldn't find anything -- any tips?  That red X is really making my inner-engineer sad :(

bilderbuchi:
i'm on it

bilderbuchi:
sadness averted. [SO](http://stackoverflow.com/a/17624403/599884)
> If you have write access to the repo: On the build's detail screen, there is a button ↻ with the tooltip "Restart Build".

libs/openFrameworks/graphics/ofBitmapFont.cpp
```diff
@@ -174,39 +174,9 @@ static const unsigned char bmpChar_8x13_123[] = {  8,  0,  0,  0, 14, 16, 16,  8
...
-static const unsigned char bmpChar_8x13_153[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
-static const unsigned char bmpChar_8x13_154[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
-static const unsigned char bmpChar_8x13_155[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
-static const unsigned char bmpChar_8x13_156[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
-static const unsigned char bmpChar_8x13_157[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
-static const unsigned char bmpChar_8x13_158[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
-static const unsigned char bmpChar_8x13_159[] = {  9,  0,  0,  0,  0,  0,  0,170,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,130,  0,  0,  0,170,  0,  0,  0,  0,  0};
```
arturoc: can you revert this change? this adds a few characters that give full support for several languages. i'll try to add them to the texture at some point

bakercp: Sure -- can I instead add the arrays back, but then _not_ substitute the `bmpChar_8x13_032` values, so that the variables don't go unused and produce warnings?  A simple solution would be to set `bmpChar_8x13_126` through `bmpChar_8x13_160` equal to the current value of `bmpChar_8x13_032` (rather than their current values) so that it would be functionally the same, but no change to the end use until you add the textures.  Basically I'm just trying to get rid of all of the warnings.  This isn't a problem in Xcode, but when users are using ofSketch there is a lot of warning spew, which is distracting for new users.

arturoc: yeah totally, you can also just comment those out

bakercp: OK, for now, I'll just comment them out so they are there for future use, but won't create warnings.

bakercp: Done, see below.

PR #4079, opened 2015-07-09
Add isEnabled() and size() to ofBaseEvent.

bakercp:
I want to conditionally create a stack of events to be notified.  

Then I want to pass events through that stack.  Since the stack is linked in a sort of chain, I only want to include events in the stack that have subscribers and are enabled.  

Thus, it is important to be able to know if an event is enabled and if the event has subscribers.  These two accessors allow me to use the existing ofEvent system without extending ofEvent and adding these accessors to my own subclass.

### DIFF
--- a/libs/openFrameworks/events/ofEvent.h
+++ b/libs/openFrameworks/events/ofEvent.h
@@ -47,7 +47,7 @@ public:
     	enabled = false;
     }
 
-	bool isEnabled() {
+	bool isEnabled() const {
 		return enabled;
 	}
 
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/events/ofEvent.h
```diff
@@ -47,6 +47,14 @@ class ofBaseEvent{
     	enabled = false;
     }
 
+	bool isEnabled() const {
+		return enabled;
+	}
+
+	std::size_t size() const {
+		return functions.size();
```
bakercp: @arturoc I've been thinking about this and wondering -- since we're accessing the functions array -- should this actually be:

```c++
std::size_t size() const {
    return functions.size();
}

```

We'd also have to make `mtx` mutable.

PR #4086, opened 2015-07-10
VS OpenCV lib and script Updates

DomAmato:


addons/ofxOpenCv/scripts/formulas/opencv.sh
```diff
@@ -494,10 +582,10 @@ function build() {
 function copy() {
 
   # prepare headers directory if needed
-  mkdir -p $1/include
+  mkdir -p $1/../../addons/ofxOpenCv/libs/opencv/include
```
arturoc: why this? it was working in other platforms as it was. perhaps we need to fix the path passed to this function in windows?

DomAmato: It was looking in the wrong path for me. It was looking in the libraries folder instead of the addons path it should have been directed too.

Dominic Amato
Human Computer Interaction Masters Candidate
Physical Computing Lab (http://goo.gl/dguGzt)
DePaul University

-----Original Message-----

danoli3: Check out here: https://github.com/openframeworks/openFrameworks/blob/master/scripts/apothecary/apothecary#L512 
Add some output's and figure out if it is not settings LIBS_DIR to the addons path.

DomAmato:
Actually I think that was due to me changing the location of the script when I built it. I will change it back to what it was.

bilderbuchi:
can we merge this? It would resolve #3916, I think

PR #4088, opened 2015-07-11
ofxGui: improve inheritability of ofxSlider

fx-lange:
splitting generateDraw()/generateText() and render() is nicely done but you can't really build on it in derived classes in the ofxGui context because of the missing virtual modifier in the base class. render() and mouse...() are virtual functions already.

(example: for a simple gui extension I tried to override generateText() to draw some labels instead of the actual value)


bilderbuchi:
don#t see how this could hurt - ping @arturoc

arturoc:
yeah seems correct, thanks!

PR #4093, opened 2015-07-15
WIP: Multi-OS testing on Travis

bilderbuchi:
This is WIP, don't merge. As designing this will be a pretty iterative process, after it's finished, it will be rebased.
This PR will enable testing OF on Travis for both mac and linux, which has graciously been activated for us.

Probably impossible stuff:
* Switch android builds to pre-provisioned, if possible, like [this](https://github.com/forkhubs/android/blob/master/.travis.yml), cf. [the docs](http://docs.travis-ci.com/user/languages/android/)
* see if we can move single entries of the build matrix to sudo: false builds.

TODO:
* [x] Pull in Linux build
* [x] Refactor build script structure and env vars to simplify travis file
* [x] Multiline script steps seem to be [possible](https://github.com/travis-ci/travis-ci/issues/879#issuecomment-86655287):
```
  - if [ "$TRAVIS_BRANCH" == "master" ]; then
      mkdir deploy;
      cp -r !(deploy|README.md|node_modules|composer.*) deploy/;
      zip -r deploy.zip deploy;
    fi
```
* [x] before merge, remove .travis_linux.yml
* [x] ~~move docs to the after_success stage, where that should probably be, and only build for one target?~~ (leave them as their own build)
* [x] docs build on master [does not work](https://travis-ci.org/openframeworks/openFrameworks/jobs/71060266#L2172-L2173)
* [x] osx/ios scripts rely on SCRIPT_PATH existing, so those are broken.
* [x] replace superfluous $OF_ROOT by $TRAVIS_BUILD_DIR
* [x] linux build: `export` from within install script is not seen outside. See https://github.com/travis-ci/travis-ci/issues/1313. Source-ing works, but messes with Travis build log output - just pull the export out of the build script, into travis.yml, as a (temporary) exception.
* [x] scripts can have a fallback for OF_ROOT using [this](http://stackoverflow.com/a/16879155/599884): `a=${b:-$c}`, to make scripts locally runnable.
* [x] split docs build step into an additional install step and real build
* [x] check out https://github.com/travis-ci/travis-ci/issues/2882
* [x] docs build only runs on direct pushes due to encrypted vars not being exposed to PRs - filter the build and document that in the yml file.
* [x] Android often (not always) fails when extracting NDK - only shows "Ex" instead of "Everything is Ok" - Why? edit: ah, [apparently](https://github.com/travis-ci/travis-ci/issues/2946#issuecomment-62754111), this could also be a memory issue. See also [this](https://github.com/richq/android-ndk-profiler/commit/f7848920e4783192704b62e3ae9fba3057a23d6b) and [this](https://github.com/richq/android-ndk-profiler/commit/525055ad4b22258f5e5a51db45051dd358f8491f). `mt=2`could be used (`-mmt=2`?) to restrict number of CPUs to 2 instead of the detected 30 or so. See http://sevenzip.osdn.jp/chm/cmdline/switches/method.htm#MethodID
* [ ] Linux build sometimes fails with " "g++-4.9: internal compiler error: Killed (program cc1plus)"", due to lack of RAM/swap, apparently. (also happens on Arturo's build). Could be that [this](http://stackoverflow.com/a/16372032/599884) or [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56746#c7) could reduce g++ memory consumption.
* [ ] doc builds fail, see HalfdanJ/ofDocGenerator#3. set to allow failure for now.
* [x] Revert e94987515af94cd75b2faed069c02a0ef7348b6d and following tweaks
* [x] Move Android to docker builds, cf. travis-ci/travis-ci#4557
* [ ] Move docs to docker builds (all packages are whitelisted)
* [x] Cache extracted NDK for Android builds
* [x] ~~separate OF and example builds for platforms except linux?~~ not for now
* [ ] add allAddonExample to build stage


bilderbuchi:
@arturoc The Android build often (not always) fails in the `install when extracting the NDK - only shows "Ex" instead of "Everything is Ok". Do you know why? Compare a [successful](https://travis-ci.org/openframeworks/openFrameworks/jobs/71238297) with a [failing](https://travis-ci.org/openframeworks/openFrameworks/jobs/71238300) build.

bilderbuchi:
Yes!
* [x] Make it [work](https://travis-ci.org/openframeworks/openFrameworks/builds/71247675), 
* [ ] then make it beautiful! :D

bilderbuchi:
@arturoc will your linux build workaround choke or just not run if I remove `.travis.linux.yml` in this PR already?

bilderbuchi:
@arturoc is it by design that, in scripts/linux/testTemplate.sh, you build the 32bit Linux template on a 64bit Travis CI worker? Does that have an impact? [Interestingly](https://travis-ci.org/openframeworks/openFrameworks/jobs/71250664), the machines can run out of RAM or swap when compiling - "g++-4.9: internal compiler error: Killed (program cc1plus)"

arturoc:
> @arturoc The Android build often (not always) fails in the `install when extracting the NDK - only shows "Ex" instead of "Everything is Ok". 

no idea :) it might be that the process is taking too long and travis kills it. i think there's a way to tell in the travis config that you want an android toolchain, that might work better but not sure.

> @arturoc will your linux build workaround choke or just not run if I remove .travis.linux.yml in this PR already?

just remove it, i'll remove the webhook

> @arturoc is it by design that, in scripts/linux/testTemplate.sh, you build the 32bit Linux template on a 64bit Travis CI worker? Does that have an impact?

no, the makefiles are the same for every platform, you could build the osx template and it will work too

> Interestingly, the machines can run out of RAM or swap when compiling - "g++-4.9: internal compiler error: Killed (program cc1plus)"

i've seen this if i tried to compile with -j4 otherwise it's been working without problem for me

bilderbuchi:
thanks, that's helpful
> i think there's a way to tell in the travis config that you want an android toolchain, that might work better but not sure.

Yeah, see the link in the PR description. I'm just unsure what exactly to install, let's see, I'll keep fiddling with that until the end.

> i've seen this if i tried to compile with -j4 otherwise it's been working without problem for me

your personal machine probably has more RAM than those CI workers (1.5GB I think), which is why this could crop up. Can we tweak the compile settings somehow so that we use less memory? -j4 is not the default currently, but -j1, right? I remember there was a PR to change that...

bilderbuchi:
> just remove it, i'll remove the webhook

OK, I'll remove it before merging, so that all the other builds test on linux in the meantime.

arturoc:
> your personal machine probably has more RAM than those CI workers (1.5GB I think), which is why this could crop up.
i actually meant on travis,

bilderbuchi:
> i actually meant on travis, 

In this case, note that the [link](https://travis-ci.org/openframeworkstests/openFrameworks/builds/71249127) I included in my initial comment about this points to one of your openframeworkstests builds failing for this very reason, so "it's been working without problem for me" is not accurate.

bilderbuchi:
note to self: OK, this works now, but osx/ios scripts rely on SCRIPT_PATH existing, so those are broken.

danoli3:
@bilderbuchi I'll PR to your repo the fix for the SCRIPT_PATH issue. I think I found an elegant solution :)

bilderbuchi:
that looks good, thanks!
I'm still planning to continue work on this soon, but I am most of the time restricted to using the Github web interface - I "just" need to find a quiet evening with my laptop to shuffle scripts around and tweak stuff. ^.^

danoli3:
Nice! Oh BTW, with the documentation script. You can remove the Brew code and change it back to apt-get if it is back on the Linux box. Want me to fix that back up again? (may fix whatever issue is currently happening as mentioned above)

bilderbuchi:
> Oh BTW, with the documentation script. You can remove the Brew code and change it back to apt-get if it is back on the Linux box. Want me to fix that back up again? (may fix whatever issue is currently happening as mentioned above)

yeah that would be great! maybe also worth it to look at the dependencies - why do we need to pull in the  (huge) libreoffice? It's not noted as dependency at https://github.com/halfdanJ/ofDocGenerator

Will be hard to test, though, as it only runs on the master branch - maybe worth it to, at least temporarily, move the `if [ "$TRAVIS_BRANCH" == "master" ]` check in documentation.sh to only include the after_success stage (as that is the only part that can only run on the master branch).  abit wasteful because we generate, but don't upload docs, but we can always move it back later.

danoli3:
I found a way to test by merging into my Forks/Master. I'll try that. Yeah wtf libreoffice. I'll remove that see if it plays nice.

danoli3:
Also note: https://travis-ci.org/danoli3/openFrameworks/builds/71900193
My Fork is tracking Travis, however does not have Multi-OS permissions (Obviously as I have not requested), however it failed under the OSX / iOS due to this. Might need to add another check in those OSX / iOS scripts specifically to let the user know what it failed if not OSX due to not having Travis / Multi-OS permission. (It defaults to Linux).

bilderbuchi:
I think that the single-OS restriction is temporary until Travis adds more hardware. also, non-experienced devs will typically only check/see the travis tests on our fork, which works with multi-os; the experienced devs will know about this shortcoming. in any case, as soon as you PR the change, you can work multi-os anyway (which is why I am developing this bit-by-bit as a PR to the OF repo, too).

bilderbuchi:
btw, if you have more time and are experienced with Android, one area where I could use input is the failing android builds. 
Basically, the NDK extraction sometimes [fails silently](https://travis-ci.org/openframeworks/openFrameworks/jobs/71904417) (i.e. no "Everything is Ok" message at the end of the install step), which of course makes the subsequent compilation fail. See checkmark nr 4 in the PR description for some details how this maybe can be fixed/circumvented.

danoli3:
I like the idea of the CI folder, however I think the testTemplate script is still valid you can use that manually before pushing PR's to validate everything working.

bilderbuchi:
> I think the testTemplate script is still valid you can use that manually before pushing PR's to validate everything working.

yes, I agree, and in the final version I intend users to be able to run these scripts locally if they want, by adding logic around the expected OF_ROOT, and maybe having one general script in /scripts/ci itself which runs the scripts appropriate to your platform.

bilderbuchi:
@arturoc for the linux build, in the build script, can I export/append CXXFLAGS and make will take it into account? It seems this [does not get picked up](https://travis-ci.org/openframeworks/openFrameworks/jobs/72296764#L1324) correctly, do you know why? and how can I set this?

arturoc:
it's possible that it just doesn't work, we should fix it but meanwhile you can do:

```bash
echo "PROJECT_CFLAGS = -ftrack-macro-expansion=0" >> config.make
```

bilderbuchi:
Looking good, Linux build worked, let's see if this is a permanent state.
Now I need to shave down the NDK extraction a bit more, the build is still failing sometimes due to too few resources (err 137)

bilderbuchi:
so, apparently neither the linux build flag nor whittling down the NDK extracted size were enough - both builds still fail sometimes, and often work when restarted. :sadface:

bilderbuchi:
OK I just squashed the commit mess into one commit in preparation for merge review.
Android and Linux builds still spuriously fail, doc build fails because of mysterious error that I have reported upstream. Overall, I think the situation is not worse than before.
@arturoc @danoli3

danoli3:
Yeah the Android part is really really bad right now on master. 
I say comment out the actual testing of Android at the very least on the master branch ASAP until we figure out a way to get it to work all the time. Would it be too much to comment out the android part from this PR changes and merge straight away?

bilderbuchi:
> Would it be too much to comment out the android part from this PR changes and merge straight away?

that's not really necessary, I can switch this to allow_failure, like I did with the docs buidl - that way the builds still run and we can continue to fiddle/observe, but the PR indicators won't go red.

danoli3:
Ah awesome! Yeah do that and I think this is good to go.

arturoc:
isn't it possible to cache those files? perhaps we can have the config that runs the android tests on a docker instance?

arturoc:
i mean cache the ndk. otherwise we can try to host the ndk ourselves somewhere else compressed with something more light to uncompress or something like that

bilderbuchi:
Currently, I'm working on switching the android builds to docker.
Caching the NDK: no idea. the approach I found somewhere else is put the NDK into a git repo and pull that in. We could probably do that, a slimmed-down NDK is extracted ~1GB on my machine. Let's see if the container approach fixes things, first.

arturoc:
once you have it on an android docker you can cache any directory so it should be easy to do.

danoli3:
Google just triaged the issue of making the NDK < 1024MB https://code.google.com/p/android/issues/detail?id=96997

bilderbuchi:
@arturoc I think I have the linux builds working now. As I haven't yet heard back from Travis (see travis-ci/travis-ci#4557), I will revert the Android docker part, and put this and the rest of the TODO items into a follow-up PR to get this one merged and running. Is it OK to merge from your side?

arturoc:
yeah do it :)  about the android problem acording to the comments in https://code.google.com/p/android/issues/detail?id=96997 just recompressing as tar.xz or bz2 and hosting ourselves should solve it

bilderbuchi:
> just recompressing as tar.xz or bz2 and hosting ourselves should solve it
yeah, but if our storage is not "near" (network-wise) to Travis, it will be very slow, and don't forget that our server would be hit with ~400MB traffic on every CI build!

arturoc:
> it will be very slow, and don't forget that our server would be hit with ~400MB traffic on every CI build!

i don't think it would be much slower than downloading it from the google servers, the ideal would be to cache that directory but meanwhile i think it's totally ok to download it. the linux gateway has been downloading the whole repo with depth 1 in our servers which is more or less the same i think.

here's a link in case you want to use it i'll take a look at trying to get it to run on a docker instance too.

http://192.237.185.151/android-ndk-r10e.tar.bz2

or perhaps we can put it in a repo

bilderbuchi:
> here's a link in case you want to use it i'll take a look at trying to get it to run on a docker instance too.

all right, I'll try that. could you also upload a minified version using [this](https://github.com/bilderbuchi/openFrameworks/blob/bb605ba67d68ce6ac7f92f53e52cda6339d62a89/scripts/ci/android/install.sh#L11) to rip out all the stuff we don't need, before recompressing? Would make it even smaller, and we wouldn't need to do the ripping-out on Travis.

bilderbuchi:
> run on a docker instance too.

the docker instance is not the problem, our combination of multi-os, docker on some builds, sudo/VM on some others, osx son some others, seems to be what trips up the travis parsing machine.

arturoc:
just uploaded it, same link

bilderbuchi:
OK, I'll merge this now. @arturoc can you disable the triggering of the openframeworkstests CI runs?
The linux builds still fail from time to time, I have set them to allow_failures for now so that they don't make red noise in the CI runs.

Remaining issues (to be dealt with in a follow-up PR):
* [ ] Linux build sometimes fails with " "g++-4.9: internal compiler error: Killed (program cc1plus)"", due to lack of RAM/swap, apparently. (also happens on Arturo's build). Could be that [this](http://stackoverflow.com/a/16372032/599884) or [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56746#c7) could reduce g++ memory consumption.
* [ ] doc builds fail, see HalfdanJ/ofDocGenerator#3. set to allow failure for now.
* [ ] Move docs to docker builds (all packages are whitelisted), cf. travis-ci/travis-ci#4557
* [ ] Confirm cache works correctly, maybe have to remove ever-changing files (.awk?)
* [ ] add allAddonExample to build stage

arturoc:
great, just disabled it.

bilderbuchi:
btw @arturoc is there anything about the emptyExample compilation which would trigger huge memory usage? the OF core compilation always succeeds, it's only the emptyExample compile where it fails.

arturoc:
it's probably that linking uses way more memory than compiling

HalfdanJ:
@arturoc the link to the NDK doesn't work any more.

arturoc:
i had to recreate the server some days ago and forgot to add this. it's now in:

http://ci.openframeworks.cc/android-ndk-r10e.tar.gz

Issue #3629, opened 2015-02-08
ofAVFoundationPlayer can't dance backwards


kylemcdonald:
(Accidentally submitted before I filled out the text.)

The brand new ofAVFoundationPlayer can step one frame back at a time, but can't handle a negative play rate. Instead it stays on the current frame, or toggles back and forth between the previous and current frame. Haven't looked at possible causes yet.

To reproduce, build videoPlayerExample and move the mouse left of center.

julapy:
ah ok, ill investigate.

admsyn:
Did a quick bit of research on this, as far as I can tell as-is ofAVFoundationPlayer is struck with whatever AVFoundation reports with regards to backwards playback capabilities. 

When loading an asset, you get 3 read-only bool properties back: [canPlayReverse](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVPlayerItem_Class/index.html#//apple_ref/occ/instp/AVPlayerItem/canPlayReverse), [canPlaySlowReverse](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVPlayerItem_Class/index.html#//apple_ref/occ/instp/AVPlayerItem/canPlaySlowReverse), and [canPlayFastReverse](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVPlayerItem_Class/index.html#//apple_ref/occ/instp/AVPlayerItem/canPlayFastReverse) corresponding to rates of -1, 0 to -1 and less than -1 respectively (presumably codec / hardware dependent). The fingers vid gives me false for all of these on a MacBook Pro, 10.10.

I doubt it's a case of Cannot Be Done but it may be that a straight-up AVPlayer isn't going to be able to do it.

urshofer:
I've posted the same question to the forum [http://forum.openframeworks.cc/t/avfoundationplayer-and-negative-speed/18684] - It's probably also dependent on the media, sometimes it works with i-frame only formats like prores…

i-n-g-o:
ofAVFoundationPlayer uses AVAssetReader to get the audio- and videodata.
An AVAssetReader is not capable of reading an asset backwards.

after playing around with the implementation i have backwards playing running, but it is really slow... doing a seekToTime for every frame.
(reducing the timeRange to one frame, so the reader would not unnecessarily preload frames)

[...]

PR #4094, opened 2015-07-17
AVFoundationPlayer dance backwards - fixes

i-n-g-o:
as explained here:
https://github.com/openframeworks/openFrameworks/issues/3629

it also fixes frame-stepping issues for pre OSX 10.8 and improves accuracy when seeking for time


i-n-g-o:
hm.
for pre OSX 10.8 the old code is able to step every frame. it has the problem though that it does not start at the first frame.
i'll investigate...

i-n-g-o:
ok. this was just a matter of precision.
all good now :)

apps/projectGenerator
```diff
@@ -1 +1 @@
-Subproject commit b5fafc830128915cef181ae5c1cb76fbf75a9aee
```
ofTheo: should this change be included in the PR?

libs/openFrameworksCompiled/project/osx/openFrameworksLib.xcodeproj/project.pbxproj
```diff
@@ -188,7 +188,7 @@
...
 		6766729E1A749D1900400051 /* ofAVFoundationPlayer.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ofAVFoundationPlayer.h; sourceTree = "<group>"; };
 		6766729F1A749D1900400051 /* ofAVFoundationVideoPlayer.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ofAVFoundationVideoPlayer.h; sourceTree = "<group>"; };
-		676672A21A749D1900400051 /* ofAVFoundationPlayer.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; path = ofAVFoundationPlayer.mm; sourceTree = "<group>"; };
+		676672A21A749D1900400051 /* ofAVFoundationPlayer.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; path = ofAVFoundationPlayer.mm; sourceTree = "<group>"; usesTabs = 0; };
 		67D96B941651AF6D00D5242D /* ofGLUtils.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = ofGLUtils.cpp; path = gl/ofGLUtils.cpp; sourceTree = "<group>"; };
```
ofTheo: also prob shouldn't be committing this file either  :)

ofTheo:
Thanks so much @i-n-g-o ! 
just noted a couple of small files that shouldn't be committed with this PR. 

So with this PR does the videoPlayerExample now work as it used to with backwards playback and scrubbing working well?

i-n-g-o:
removed some stuff in seekToTime:withTolerance:
stuff i thought would help improve accuracy, but it was the float/double thing anyways.

strange for the additional files in this PR... should i revert? or is it fine like this?

i did not test iOS...
would be great if somebody fluent with iOS could do some testing...

ofTheo:
ping @danoli3 to maybe give a quick test on iOS

ofTheo:
@i-n-g-o if you can remove those files from being committed without redoing the PR that would be ideal

apps/projectGenerator
bilderbuchi: I'm guessing this comes from the git submodule sync - probably someone committed something to the PG, and forgot to update the submodule within OF, so this gets done now.

i-n-g-o:
hm. tried to change branching-point with `git branch -f`
not sure, if this resolves the issues...

i could also ditch this and create a new PR

ofTheo:
looks good to me. 
I only see the three files relating to the AV Video stuff. 
Thanks @i-n-g-o

i-n-g-o:
cool

PR #4099, opened 2015-07-19
ofxOsc and Visual Studio fix

charlesveasey:
Resolves two issues with Visual Studio and ofxOsc addon


arturoc:
mmh, the ofx namespace is quite hacky, this should be modified in the original oscpack code with a different namesapce probably osc? also isn't there a way to solve it without adding the namespace? perhaps with forward declaration so that file is not included from the .h?

bilderbuchi:
agreed, that's also something I don't really like about this solution.

charlesveasey:
Replaced with https://github.com/openframeworks/openFrameworks/pull/4108

PR #4102, opened 2015-07-20
fix state leak when recursively drawing MSAA fbos

### LINKED ISSUE #4096: multisampled ofFbo issues when drawing inside another ofFbo
Hi, I am finding weird issues when drawing one multisample FBO inside another one.

	int numSamples = 8;
	fbo1.allocate(300,300, GL_RGBA, numSamples);
	fbo2.allocate(300,300, GL_RGBA, numSamples);

	fbo1.begin();
	ofClear(0,0,55,255);
	ofDrawCircle(mouseX, mouseY, 33);
	fbo1.end();

	fbo2.begin();
	fbo1.draw(0,0);
	fbo2.end();

This should draw nothing to screen, but if numSamples is > 0, it does. 
In my setup (today's github master, OS X 10.10.4), if numSamples > 0, it draws fbo1 at the bottom-left of my window, flipped vertically.

It works as expected with numSamples == 0.

tgfrerer:
+ fixes #4096 

When resolving MSAA fbos, global fbo binding state must be
changed, because the resolve is a blit from one fbo into
another.

This PR makes sure to correctly keep track of the global
renderer FBO binding state and to restore it after the
blitting operation has completed.

+ Also includes some minor typo fixes for the doxygen
documentation of the updateTexture() method.

### DIFF
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -934,16 +934,21 @@ void ofFbo::updateTexture(int attachmentPoint) {
 			glPushAttrib(GL_COLOR_BUFFER_BIT);
 		}
 
-		bind();
+		// Because we temporarily change global Fbo binding state
+		// when blitting from one framebuffer to the other, we first have 
+		// to query the renderer for its binding state, and store it.
+		GLuint rendererFboBindingState = ofGetGLRenderer()->getCurrentFramebufferId();
+		
+		glBindFramebuffer(GL_READ_FRAMEBUFFER, fbo);
 		glReadBuffer(GL_COLOR_ATTACHMENT0 + attachmentPoint);
 		glBindFramebuffer(GL_DRAW_FRAMEBUFFER, fboTextures);
 		glDrawBuffer(GL_COLOR_ATTACHMENT0 + attachmentPoint); 
 		glBlitFramebuffer(0, 0, settings.width, settings.height, 0, 0, settings.width, settings.height, GL_COLOR_BUFFER_BIT, GL_NEAREST);
 
-		glBindFramebuffer(GL_READ_FRAMEBUFFER, 0); // reset to defaults
-		glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0); 
-		unbind(); // this will restore GL_FRAMEBUFFER to previousFramebufferBinding
-		
+		// Restore renderer Fbo binding state
+		glBindFramebuffer(GL_FRAMEBUFFER, rendererFboBindingState);
+
+		// Set ReadBuffer state to GL_BACK (default)
 		glReadBuffer(GL_BACK);
 
 		if(!ofIsGLProgrammableRenderer()){
--- a/libs/openFrameworks/gl/ofFbo.h
+++ b/libs/openFrameworks/gl/ofFbo.h
@@ -94,11 +94,11 @@ public:
 
 	void flagDirty() const; ///< check whether attached MSAA buffers need updating
 
-	/// \brief    Explicityl resolve MSAA render buffers into textures 
-	/// \note     if using MSAA, we will have rendered into a colorbuffer, not directly 
-	///           into the texture call this to blit from the colorbuffer into the texture 
-	///           so we can use the results for rendering, or input to a shader etc.
-	/// \note     This will get called implicitly upon getTexture();
+	/// \brief    Explicitly resolve MSAA render buffers into textures 
+	/// \note     If using MSAA, we will have rendered into a colorbuffer, not directly 
+	///           into the texture. Call this to blit from the colorbuffer into the texture 
+	///           so you can use the results for rendering, or input to a shader etc.
+	/// \note     This is called implicitly upon getTexture();
 	void updateTexture(int attachmentPoint);
 
 


bilderbuchi:
hey @tgfrerer I appreciate you wanting to get this PR merged, but the 0.9.0 milestone is not appropriate anymore for such a bug (i.e. non-crashing, non-regression) ;-). This does not mean that it can't be merged before/right now after review, only that it can't delay the 0.9.0 milestone completion.

procedural:
@bilderbuchi how would fixing a bug delay the release?

bilderbuchi:
> @bilderbuchi how would fixing a bug delay the release?

fixing a bug would not, but _not_ fixing the bug (and the bug still being milestoned) would delay a release. Therefore, we have a [policy](https://github.com/openframeworks/openFrameworks/wiki/Issue-and-pull-request-management-details#milestones) of locking a milestone for everything except critical and regression bugs when it nears completion (or intended completion, at least).

procedural:
@bilderbuchi still doesn't make much sense to me personally for this particular bug, when it's a matter of pushing merge button to fix it for current release.

bilderbuchi:
again, it's not about pushing merge buttons, it's about reviewing and testing the code before pushing the merge button, not worsening feature creep, and not accumulating the time debt incurred when needing to test, review and merge dozens of "little" PRs where you "only" have to push the merge button, when the milestone is already months delayed, and you have more important (i.e. critical/regression) bugs to fix, and are dealing with significant shortages of (experienced) developer manpower and time. I can't make it more clear than that, sorry.

procedural:
It's also a matter of trust to the people of your community. Like, if you know that @tgfrerer is a highly skilled OpenGL specialist and most certainly he tested his own pull request already, either you waste time rechecking his work or fix a bug quickly without imaginary constraints.

bilderbuchi:
that point is really orthogonal to the milestoning question. We still consider a two-man rule to be good practice, and this has changed from a more strict state several years ago. We have a [formal procedure](https://github.com/openframeworks/openFrameworks/wiki/Pull-Request-Review-Procedure) in place, but this is in practice followed only loosely (precisely because we trust our expert members).

bilderbuchi:
Also, let me repeat *again* (because I'm not sure you picked this up): We could merge this PR right now, after a short review, and it would be totally ok (that's what will probably happen, actually). We just won't milestone this for the upcoming, delayed, release, so _that_ can't be delayed by, say, Tim firing off this PR, leaving on summer vacation (with a couple other 2D experts), and @kylemcdonald wanting to get 0.9.0 wrapped up and shipped, and nobody being around to verify the changes.

tgfrerer:
Heyheyhey, peace, guys... 

I trust @bilderbuchi & everyone are doing their best in getting 0.9 ready & shipped, and i support this 110%. 

And i have trust in both of you =)

I didn't know we were so close to finally ship 0.9, which is a welcome surprise.

And whether we merge this fix now or a little later (or find another way to fix the bug) does not matter so much to me personally. It's only a few lines and won't code rot until 0.9.1 =)

libs/openFrameworks/gl/ofFbo.cpp
```diff
@@ -934,16 +934,21 @@ void ofFbo::updateTexture(int attachmentPoint) {
 			glPushAttrib(GL_COLOR_BUFFER_BIT);
 		}
 
-		bind();
+		// Because we temporarily change global Fbo binding state
+		// when blitting from one framebuffer to the other, we first have 
+		// to query the renderer for its binding state, and store it.
+		GLuint rendererFboBindingState = ofGetGLRenderer()->getCurrentFramebufferId();
```
arturoc: this isn't already in ofGLRenderer, no?  in any case this call to a singleton makes any dependent call thread unsafe. i can take a look later to see if there's any way to do this without using the singleton

bilderbuchi:
> I didn't know we were so close to finally ship 0.9, which is a welcome surprise.

well, we've been "close", in scare-quotes, (and in freeze) for months now, but finally the effective open issue count is at 10, which imo puts it in the region where I think it feasible we can get to a release (candidate) with a concerted push on a weekend or somesuch. OTOH, it's summer where most of the devs are.. :-P

libs/openFrameworks/gl/ofFbo.cpp
tgfrerer: `getCurrentFramebufferId()` is available in both GL renderers, yes.

Ideally, i'd like to get rid of the getRenderer() calls as much as possible, too. 

But here, we need a way to query the renderer state, since only the renderer keeps track of & knows the currently bound FBO...

bilderbuchi: -> @procedural and this is exactly why it is sensible to review even small contributions by experts.

arturoc: yeah it makes sense that the renderer keeps track of the bound buffers... but the way it's implemented is kind of weird, i'm trying to change it to remove the state from the fbo and completely move it to the renderer. one question, in this change you are never setting back the read framebuffer and draw framebuffer to the previous value, is that correct?

tgfrerer: Both read and write framebuffers are set in one instruction when you call `glBindFramebuffer` with the target `GL_FRAMEBUFFER`.

So, first, we have to bind read and write to different framebuffers in lines 943 and 944. Then, in line 949, **both** targets are restored to previous state =)

https://www.opengl.org/sdk/docs/man/html/glBindFramebuffer.xhtml

armadillu:
@tgfrerer Thx so much for looking into this so quickly! I will test in my setup very soon

libs/openFrameworks/gl/ofFbo.cpp
arturoc: oh, i see, man opengl state is such a pain... i've refactored this a bit, it's doing the same as your solution but the renderer is now keeping track of everything. if you can check that i'm not doing something wrong:

https://github.com/openframeworks/openFrameworks/compare/master...arturoc:refactor-fbo?expand=1

armadillu:
Both @tgfrerer and @arturoc PRs fix the original issue #4096 for me.

PR #4107, opened 2015-07-20
fix fbo ids call

armadillu:
This won't compile for me, I think you meant getIdDrawBuffer();


arturoc:
yeah soory, i just fixed it in master and also changed getFbo -> getId since getFbo is deprecated now

armadillu:
cool thx so much!

PR #4112, opened 2015-07-21
OpenGLES / iOS Fixes for recent Renderer changes.

danoli3:
Fixes some OpenGLES (iOS) errors caused by recent: 9988a637545e05d9c39f4d94710194be70c3bfde

@arturoc give this a quick look over. This just enables to compile again.

Xcode was very strict with needing to see an explicit definition of the GLuint which is why the following:

```
	GLint defaultFramebufferBinding = defaultFramebufferId;
	glGetIntegerv(GL_FRAMEBUFFER_BINDING, &defaultFramebufferBinding);
```

Aside from that, defined out the non-compatible OpenGL commands for OpenGLES in the new ```bindForBlitting``` function. I did a quick check and it seems those commands are only OpenGLES 3.0+.
Not sure what we can do for an alternative. Anyway this fixes it to compile.


arturoc:
sorry, only saw this after fixing it

danoli3:
All good dude.

Fixed by: https://github.com/openframeworks/openFrameworks/commit/5c138a69df6873be6edb57bdbd432eb201a7f4d3

PR #4114, opened 2015-07-21
Bugfix ofArduino constructor

tpltnt:
Hi,

i fixed some missing initializations in the ofArduino class.
Cheers,
tpltnt


bilderbuchi:
@openframeworks/core do we really do copyright notices? Isn't that kinda absurd, as every contributor would have to add her name to any files she edited?

libs/openFrameworks/communication/ofArduino.h
```diff
@@ -458,31 +460,31 @@ class ofArduino{
...
 	// whether pin reporting is enabled / disabled
 	
-	int _analogPinReporting[ARD_TOTAL_ANALOG_PINS];
+	int _analogPinReporting[ARD_TOTAL_ANALOG_PINS]={INT_MAX};
 	// whether pin reporting is enabled / disabled
 	
 	bool bUseDelay;
 	
-	bool connected;
+	bool connected;  // true if serial connection to Arduino exists
```
bakercp: Please add doxygen compatible documentation according to the documentation style guide https://github.com/openframeworks/openFrameworks/wiki/openFrameworks-doxygen-documentation-style-guidelines#document-the-member-variables

bakercp:
Regarding copyright, some of the older files still have it, but individually documenting copyright on each commit is not a style we've been following.  AFAIK, the last major discussion was on the of-dev mailing list on or around 8/26/12 (@bilderbuchi -- you were a part of that discussion).  Unfortunately, this is not something we discussed much during the roadmap conference @ YCAM, and it's not mentioned in the roadmap.

bakercp:
For reference ... http://dev.openframeworks.cc/pipermail/of-dev-openframeworks.cc/2012-August/030570.html

libs/openFrameworks/communication/ofArduino.h
tpltnt: @bakercp I added the doxygen commands for *my* changes. I would leave the rest to when someone (I?) goes over the whole class and does style fixes etc.

bakercp:
Hey @tpltnt sorry to be picky about documentation comments, but could you use complete sentences, punctuation, etc (it's not me, it's the style guide ;)).  

Also, I'm not sure what the policy is for copyright, but regardless, your host http://tpltnt.dropcut.net doesn't resolve to a website ... (or at least the subdomain doesn't).

Thanks for your work on all of this!

tpltnt:
@bakercp good call. The domain/hosting should be fixed soon. (It is not a top priority currently.)

bilderbuchi:
from #4117: 
>  if copyright notices [be] cleaned up/removed [...] This would require clearance with the respective copyright holders.

If requiring clearance for this is really true (I would have assumed that the git history is sufficient proof of individual contributions), I am even more against adding still more copyright notices (as those create basically unremovable baggage down the line). 
@tpltnt, I would ask you to write an email to the list about this (referencing #4117), as I think a PR discussion is not a great venue for the discussion we need to have here.

bakercp:
I tend to agree -- I think of-dev is the place to have the #4117 discussion (though I do support #4117 as an issue!).

arturoc:
we don't do copyright headers and in any case not for such a small contribution as this. there's some headers here and there because we've taken pieces of code verbatim from the original source, like the noise implementation, the bitmap font or this class. in those cases the original license forces us to keep the license but we don't add headers other than that.

@tpltnt if you can remove it?

also we don't have any other initialization in headers so if you can move those to the constructor in the cpp with the rest of the variables?

tpltnt:
@arturoc done

tpltnt:
@arturoc I resolved the merge conflicts.

PR #4119, opened 2015-07-21
Extending ofxBaseGui and ofxGuiGroup

frauzufall:
Just a few small things:

 - the `parent` attribute is moved from `ofxGuiGroup` to `ofxBaseGui`: when a button or something else derived from ofxBaseGui changes size, the parent control should be updated, therefore ofxBaseGui needs a parent
 - `sizeChangedCB` moved to `ofxBaseGui`, gets called every time an element gets resized, is public to enable calling it even if the child has a different class (or is there another way without adding every possible child class as a friend class to `ofxBaseGui`?)

 - the `add`-functions of `ofxGuiGroup` return the new crontrol: allows visual changes to control element returned by the function:
```
ofxBaseGui* control = add(param.set(..));
control->setFillColor(..);
```

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -5,8 +5,8 @@
 using namespace std;
 
 
-void ofxGuiSetFont(const string & fontPath,int fontsize, bool _bAntiAliased, bool _bFullCharacterSet, int dpi){
-	ofxBaseGui::loadFont(fontPath,fontsize,_bAntiAliased,_bFullCharacterSet,dpi);
+void ofxGuiSetFont(const string & fontPath, int fontsize, bool _bAntiAliased, bool _bFullCharacterSet, int dpi){
+	ofxBaseGui::loadFont(fontPath, fontsize, _bAntiAliased, _bFullCharacterSet, dpi);
 }
 
 void ofxGuiSetBitmapFont(){
@@ -49,7 +49,7 @@ void ofxGuiSetDefaultHeight(int height){
 ofColor
 ofxBaseGui::headerBackgroundColor(64),
 ofxBaseGui::backgroundColor(0),
-ofxBaseGui::borderColor(120,100),
+ofxBaseGui::borderColor(120, 100),
 ofxBaseGui::textColor(255),
 ofxBaseGui::fillColor(128);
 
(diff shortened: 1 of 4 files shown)


arturoc:
the style fixes make it really difficult to parse, also there's some things like:

```
if(...){
}
else if{
.....
```

that are not the official style, is this done with some tool? i know @bilderbuchi started something but i'm not sure if it's completely finished.

anyway, the first 2 proposed changes seem correct to me, but returning pointers to the gui elements is something i've been trying to avoid when designing ofxGui. Raw pointers are really difficult to work with, if the original goes out of scope for any reason and you keep that pointer, it'll become invalid and will crash the application if you try to use it. also the syntax is kind of weird for beginners having to use the ->.

an option would be to return a shared_ptr which doesn't have the first problem or implement the same pattern we use for ofParameter where all the variables of the gui would be contained in a Data inner class with a shared_ptr making the gui element have shallow copy semantics, it'll behave as if it was a pointer.

perhaps the easiest is if you can send the first 2 changes separately and remove the style and pointer returns, i would merge that right away but the other change need more thought and the style seems kind of broken

bilderbuchi:
If with "parse", you mean reviewing the PR, you can append a `?w=!` to the URL, to see the files without the whitespaces changes, [like so](https://github.com/openframeworks/openFrameworks/pull/4119/files?w=1).

The commit message implies that it was done with ofStyler, which lives in /scripts/dev/style. I considered it finished, and even started an [automated tool](https://github.com/bilderbuchi/ofCodeStyleGuard) to check style conformance. If there are inaccuracies in the style config, let's collect and fix them.

I love that people use that tool to check code style, and I think in no case should we discourage people from submitting style-conforming code, so I think the style changes should stay in (maybe with corrections).

@arturoc I don't see which style-violations you mean, could you point to the specific code, and the part in the [style doc](https://github.com/openframeworks/openFrameworks/wiki/oF-code-style) that it violates? maybe the wording of the style needs to be more precised.

arturoc:
the if else curly braces are using stroustroup style but we use K&R. not sure if it's in the style guide but it should since it's what we use everywhere in the code:

```cpp
if(...) {
    ...
} else {
    ....
}

**Comment by bilderbuchi** [comment]
Sorry, but the style doc is the definitive source, and it is modeled after/adapted from the Qt style (and says so), and has been for years. specifically, we say:
* Never use a space between a keyword and an opening curly brace or between a closing parenthesis and an opening brace.

Else rules & curly brace position are not well specified, are you saying the closing bracket should be on the same line as the else, i.e.
`} else{...` ? (where `else{...` conforms to the existing style doc.

**Comment by arturoc** [comment]
yeah, i'm not talking about the space but about the extra breakline between the closing brace for the if and the else:

```cpp
// correct K&R
if(...){
    ...
}else{
    ....
}

//wrong stroustroup
if(...){
    ...
}
else{
    ....
}
```

bilderbuchi:
well, that's easy to fix. @frauzufall, could you change `nl_brace_else=add` to `nl_brace_else=remove` in [this place](https://github.com/openframeworks/openFrameworks/blob/507784e862eeb6b6ba533f85e3ebcdb21a60af10/scripts/dev/style/openFrameworks_style.cfg#L214) and run the style again, commit and push the changes?

arturoc:
just fixed it in master.

@frauzufall sorry for going so offtopic : ) in any case, the changes in parent and callbacks are ok but if you can remove the pointer to the gui elements return value for the add functions by now?

frauzufall:
No problem, it's interesting. Yes I will do this, can I just add this change as a new commit or should I go back, do the other changes again and create a new PR?

arturoc:
i think you can probably just revert that commit: 0efa7a7 but not sure if you'll have merge conflicts, whatever is easier for you

bilderbuchi:
also, you'll need to pull in latest master to see Arturo's fixes for the style, before running the styler again.

bilderbuchi:
easiest will probably be to git cherry-pick bead0d6902c26f5d85ab8d4c08c8bc8ed02ac394 over to a new branch off of current master, and make a new PR, that way the style back-and-forth changes will not end up in the git history.

frauzufall:
That's what I thought. I will do that.

Just a short question I have when I remove the return statements of the add function: 
What's the best way to change the style of a gui element without saving this control in the header?
Is something like this fine? It works, but it's always difficult for me to understand in which cases I get memory leaks with pointers.
```
setup() {
    ofxMinimalToggle* t = new ofxToggle(toggle_param.set("toggle", true));
    panel.add(t);
    t->setBackgroundColor(ofColor::white);
}
```

arturoc:
no that will leak memory since ofxGuiGroup doesn't delete the elements it holds, they are supposed to be temporary references only. we should also replace that add method, the one that receives a pointer, with something less ambiguous but by now this should work:

```cpp
panel.add(bool_parameter.set("name"));
panel.getToggle("name").setBackground();
```

or 

```cpp
//.h
ofxToggle t;

//setup
panel.add(t.setup(...));
t.setBackground(...);
```

frauzufall:
Got it, thank you. 
When I extend ofxGui with other controls, I need the `add(ofxBaseGui*)` function do add a certain control like 
```
add(new ofxRotarySlider(param.set(...));
```
there is no other way to select which type of slider I want to create, is there?

arturoc:
no, that's the only way, i'm actually realizing that all the controls created internally that way are being leaked so we need something to solve this, i'll try to take a look later

PR #4123, opened 2015-07-22
Extending ofxBaseGui #2

frauzufall:
Corrected version of https://github.com/openframeworks/openFrameworks/pull/4119:
 - the parent attribute is moved from ofxGuiGroup to ofxBaseGui: when a button or something else derived from ofxBaseGui changes size, the parent control should be updated, therefore ofxBaseGui needs a parent
 - sizeChangedCB moved to ofxBaseGui, gets called every time an element gets resized, is public to enable calling it even if the child has a different class


bilderbuchi:
thank you! I'm closing the other one.

arturoc:
thanks!

PR #4125, opened 2015-07-23
ofEvent cleanup

tpltnt:
Hi,

i fixed some style issues, added documentation and fixed a missing assignment in ofEvent.h
Cheers,
tpltnt

### DIFF
--- a/libs/openFrameworks/events/ofEvent.h
+++ b/libs/openFrameworks/events/ofEvent.h
@@ -41,6 +41,7 @@ public:
             [&](std::unique_ptr<Function>&f){
                 return std::unique_ptr<Function>(new Function(*f));
             });
+        enabled = mom.enabled;
         return *this;
     }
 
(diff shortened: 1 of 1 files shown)


ofZach:
thanks for these changes...  just a quick note that we are using tabs as per https://github.com/openframeworks/openFrameworks/wiki/oF-code-style

tpltnt:
@ofZach so the spacing is done with tabs which are 4 colums wide, and not 4 spaces instead of tabs?

ofZach:
In your editors preferences you can set tabs to be 4 spaces wide.  We just don't use spaces for indentation in the code since keeping it as tabs means its configurable for the end user.

tpltnt:
Ok, i also adjusted the style guide to make it more clear, so others aren't as mistaken as i was.

bilderbuchi:
@tpltnt I don't know if you know, but we have a tool that can do style changes automatically, using uncrustify: https://github.com/openframeworks/openFrameworks/tree/master/scripts/dev/style

tpltnt:
@ofZach i fixed the tab issue.
@bilderbuchi thank you for the hint.

PR #4128, opened 2015-07-23
Fix videoInput unresolved externals

DomAmato:


bilderbuchi:
I think we can merge this

PR #4130, opened 2015-07-23
Fix #4122 - properly resolve projects outside of apps directory

### LINKED ISSUE #4122: Project generator issues for projects outside ..\..\.. level, VS2015
Hi all. I'm having issues getting generated projects to load in VS2015 using both the legacy & simple PGs when creating projects outside the $OF_ROOT\apps\myApps level. The solution opens, but the project (in the sidebar) reads as "Project Name (load failed)" and another line (when clicking the project's disclosure triangle) reads "The project file cannot be loaded."

I dug into the .vcxproj file a little bit but didn't see anything obvious. There's a line which references the oF lib project, but correcting the path (I changed the beginning to ..\..\libs... since I put the app directly into the apps folder) didn't fix it. Curiously, oF will compile, even with an incorrect path to the ofLib.vcxproj file. 
<ProjectReference Include="..\..\..\libs\openFrameworksCompiled\project\vs\openframeworksLib.vcxproj">

Pinging @arturoc and @tgfrerer who seemed to be involved in these related issues/PRs:
https://github.com/openframeworks/openFrameworks/pull/3941
https://github.com/openframeworks/openFrameworks/pull/3929
https://github.com/openframeworks/openFrameworks/issues/3928

tgfrerer:
I thought tracking down compile issues in a makefile project was painful, but a trip down the Visual Studio Compiler toolchain made me feel in touch with The Heart of Darkness at times. Here's what I bring back from that trip:

The main reason I could find for #4122 was that when compiling a user project against openFrameworks, the openFrameworks headers will be resolved using the `.props` files we provide in `libs\openFrameworksCompiled\project\vs\`.

These "Property Sheets" contain "additional include paths" - i.e. where the compiler goes on safari to hunt down the openFrameworks headers. But these were using relative paths.

With Visual Studio property sheets the problem is that relative paths always resolve relative to *the project being compiled*. A classic openFrameworks Visual Studio Solution (`.sln`) has two project files: 

1. A shared project file for openFrameworksLib (`openFrameworksLib.vcxproj`)
2. An individual project file for the prohect at hand, i.e. `mySketch.vcxproj`

So when the solution compiles, the header search paths would resolve great for openFrameworksLib, since the header paths are relative to it's project file. But for `mySketch`, the header search paths would be resolved again, **relative to mySketch this time**. It's like blindly retracing the left and right turns of your safari, but from your current location, assuming everything will be in the same place as it was in Africa. No wonder if no elephants can be seen.

## Proposed solution

The solution I'm suggesting after some tinkering is to introduce a new `openFramworksCommon.props` file, which defines a new Visual Studio Macro,  `$(OF_ROOT)`, which can be used throughout the project files as a reference to the openFrameworks root directory.

The file itself acts as a beacon, by making the declaration of `$(OF_ROOT)` referring to the actual absolute path of the file itself. 

We may also use this file in the future to add any other global build variables and parameters.

Having `$(OF_ROOT)` should also make it simpler to edit the project files manually, for example when adding addons. 

Since the new property file is included by our existing property files, which gain access to `$(OF_ROOT)` in the process, there is no need to update or re-create project files on the user-side. 

Existing project files should be able to get access to `$(OF_ROOT)` out of the box =)

It also fixes an issue with the icon file now resolving properly when sketches are outside `apps`


mattfelsen:
Brilliant, this is working for me with both the simple & legacy PG's (I know they should be the same under the hood now but just wanted to make sure :smile: )

I'm also seeing crashes when adding addons, but I think it's unrelated to the ..\..\.. path problem. I'll do some additional testing and make a new ticket.

ofZach:
```It's like blindly retracing the left and right turns of your safari, but from your current location, assuming everything will be in the same place as it was in Africa. No wonder if no Elephants can be seen.```

beautiful analogy :) 

this sounds like a very good solution (And somewhat similar to how on xcode where we have CoreOF.xcconfig and Project.xcconfig)

tgfrerer:
Sweet! Seems it got the green light from the windows CI server. I'd consider it ready to merge.

arturoc:
thanks!!

Issue #618, opened 2011-07-05
call ofEasyCam::setPosition in testApp::setup() is forgotten


roxlu:
Setting the position of an ofEasyCam instance in ofSetup() does not work.  What does work in testApp::setup() is calling ofEasyCam::setDistance().  When calling ofEasyCam::setPosition() in for example testApp::draw() does work.

bilderbuchi:
@arturoc thinks we should rather fix this bug instead of adding documentation.

charlesveasey:
Is this really a bug? You can call setPosition() in setup() as long as you first setAutoDistance(false). This makes sense to me. You can call setPosition() in draw() or wherever you like.

arturoc:
perhaps it should show an error then and say that auto distance is enabled so you can't use setPosition

oxillo:
Hi,

We are trying to clean out the many issues that we have.
Many versions of OF have been released since this was issued.

Can you help us by checking if this issue is still present in the latest version ?

In case of inactivity, we'll close this issue by end of january, 2020.

Thanks

roymacdonald:
This got fixed.  https://github.com/openframeworks/openFrameworks/blob/patch-release/libs/openFrameworks/3d/ofEasyCam.cpp#L513-L518
By calling setPosition, the onPositionChanged() callback gets triggered and it disables updating the position using the autodistance.
Please close

PR #4131, opened 2015-07-23
added note to ofEasyCam

tpltnt:
Hi,
I added a note to ofEasyCam in order to address #618.
Cheers,
tpltnt


bilderbuchi:
shouldn't this also be documented for setPosition() (which is the function that does not work everywhere)?

arturoc:
this sounds more like a bug rather than something we want to document

tpltnt:
@arturoc fair call. My proposal just followed @bilderbuchi statement.

bilderbuchi:
The relevant bug is #618. I'm going to close this, but leave the branch alive for now.

Issue #4103, opened 2015-07-20
Unicode support for MinGW


procedural:
[Commenting on this comment in code](https://github.com/openframeworks/openFrameworks/blob/ec83be4/libs/openFrameworks/utils/ofSystemUtils.cpp#L633):

It's true that MinGW uses `char` by default, but if unicode flags are defined it switches to `wchar` and wide strings as MSVC, thus there's no need for separate code logic for these two compilers.

Here's a patch that makes MinGW compile openFrameworks with unicode characters:

https://github.com/ofnode/of/blob/4798f6c/dev/patches/openframeworks_mingw_unicode.patch

It's not a pull request because in order to make these changes work you should define `UNICODE` and `_UNICODE` flags in Code::Blocks projects and I don't know how to do that.

Pinging @ofTheo I suppose :)

bilderbuchi:
> It's not a pull request because in order to make these changes work you should define UNICODE and _UNICODE flags in Code::Blocks projects and I don't know how to do that.

You can always make a PR, indicate what is missing, and people more knowledgeable in the area you need can send _you_ a PR to your originating branch, which when merged will automatically appear in the PR.

procedural:
> can send you a PR to your originating branch, which when merged will automatically appear in the PR.

Ah, I didn't think of that! Thanks, Chris, I'll PR then.

procedural:
The branch is merged, but `UNICODE` and `_UNICODE` flags should be defined in Code::Blocks projects

arturoc:
can you add it in the template in scripts/win_cb/template ?

procedural:
I don't know where to add, nor how to check if it will work, I don't use Code::Blocks. Guide me!

procedural:
[Compiler options](https://github.com/openframeworks/openFrameworks/blob/master/scripts/win_cb/template/emptyExample.cbp#L20) maybe?

arturoc:
what do you use? how do you usually pass this option?

procedural:
@arturoc raw cmake & command line.

Ok, I've generated CB project and definitions indeed in <Compiler> section, each line looks like:

```xml
         <Compiler>
            <Add option="-DUNICODE" />
            <Add option="-D_UNICODE" />
            ...
```

arturoc:
yes that looks correct

bilderbuchi:
[...]

PR #4139, opened 2015-07-24
Unicode support for MinGW

procedural:
See #4103

Requires `UNICODE` and `_UNICODE` flags in Code::Blocks projects!

### DIFF
--- a/libs/openFrameworks/utils/ofSystemUtils.cpp
+++ b/libs/openFrameworks/utils/ofSystemUtils.cpp
@@ -374,16 +374,7 @@ ofFileDialogResult ofSystemLoadDialog(string windowTitle, bool bFolderSelection,
 		ofn.lStructSize = sizeof(ofn);
 		HWND hwnd = WindowFromDC(wglGetCurrentDC());
 		ofn.hwndOwner = hwnd;
-#ifdef __MINGW32__
-		char szFileName[MAX_PATH];
-        memset(szFileName,0,260);
-		if(defaultPath!=""){
-            strcpy(szFileName,ofToDataPath(defaultPath).c_str());
-		}
 
-		ofn.lpstrFilter = "All\0";
-		ofn.lpstrFile = szFileName;
-#else // Visual Studio
 		wchar_t szFileName[MAX_PATH];
 		wchar_t szTitle[MAX_PATH];
 		if(defaultPath!=""){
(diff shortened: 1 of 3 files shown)


libs/openFrameworks/utils/ofSystemUtils.cpp
```diff
@@ -753,131 +730,6 @@ string ofSystemTextBoxDialog(string question, string text){
 		 DestroyWindow(dialog);
 		 EnableWindow(WindowFromDC(wglGetCurrentDC()), TRUE);
 
-	#else
-
```
arturoc: why does all this code go away? i get that everything else is related to wide strings differences but this?

procedural: The code I deleted here is _exactly_ repeats what the code under `#ifdef _MSC_VER` do, except it was targeting chars instead of wide chars.

procedural: It's even says in comment:

```
//we have to do this because mingw wants non wide strings and vs wants wide strings
```

But it's not true. MinGW wants non wide strings when `UNICODE` and `_UNICODE` are not defined, if you define them it perfectly works with the code from `#ifdef _MSC_VER` block because it's the same Windows calls except it uses wide strings.

procedural: Oh, and by the way, don't let the indentation fool you: this `#else` and removed code below belongs to `#ifdef _MSC_VER` line, the indentation mixes tabs and spaces.

arturoc: oh, i see i saw  such a big chunk of code go away and freaked out :)

yeah i agree on tabs and i don't think there's really a strong preference for them it's just been like that, at some point if we run the styler script on all the code base we can discuss about changing to 4 spaces

arturoc:
can you fix the merge conflicts? it won't merge anymore

procedural:
@arturoc sure

libs/openFrameworks/utils/ofSystemUtils.cpp
bilderbuchi: > P.S. [no offense to anyone, ideological IMO] tabs are really one of the stupid ideas of openFrameworks...

well, it's only a problem if people mix spaces and tabs.. ;-)

procedural: Not only, there's another subtle problem that one can notice: dependency on local width size, a variable that can't be synchronized with preferences of others. The simplest example is here, before our eyes: a guy who wrote this code set tabs to be 4 spaces, the Github interpreted them as 8. One may say "it is a problem of Github, it should change 8 spaces to 4", the other may say "it is a problem of this guy who wrote the code, he should change 4 spaces to 8", the third one will say "don't mix spaces with tabs", and so on, but no one can say anything about spaces -- their width is strictly defined and set in stone, everyone interpret them and see equally.

procedural:
@arturoc Done! That was my first rebase, took little longer to figure out... :)

arturoc:
thanks!

procedural:
As a reminder, someone need to add `UNICODE` and `_UNICODE` flags in Code::Blocks projects, otherwise it will compile with errors on Windows when MinGW will be used!

Issue #2113, opened 2013-06-17
Deprecate ofxXmlSettings


bilderbuchi:
Now that ofXml is here, we should deprecate ofxXmlSettings asap so that people get more time to transition.

tpltnt:
I checked the codebase as of 45ad9ef287fdc764eaf8a2fa66625158dbb71e4d. The only remaining code which uses ofxXmlSettings is examples/addons/allAddonsExample, examples/addons/xmlSettingsExample,  examples/ios/xmlSettingsExample and, well, ofxXmlSettings. So it looks feasible to deprecate ofxXmlSettings for 0.9.

bilderbuchi:
deprecation is more about usage "in the wild" by users, than usage in the OF repo. I think this is what @ofTheo was driving at.

bilderbuchi:
also, -1 to deprecating this for 0.9, no features will be added any more, only critical bugfixes.

PR #4141, opened 2015-07-24
deprecating ofxXmlsettings

tpltnt:
Hi,

this PR removes the 'ofxXmlSettings' addon (source and references) as suggested in #2113.
Cheers,
tpltnt


bakercp:
Should we actually deprecate it before we remove it completely?  I'm not sure if we ever formally deprecated it in favor of ofXml, and now that Poco is no longer part of the core, ofXml is a bit in limbo until we choose another 3rd part xml lib (I'd vote for Pugi as that's what we're using for project generator, etc).

ofZach:
+1 to deprecating ofxXmlSettings and note removing -- I also think that ofxXmlSettings is kind of a staple of so many projects that including it in the core for at least some release cycles would be useful / warranted with some advice about how to transition.  It will definitely break thousands of projects and dozens of addons if it is removed completely and I feel like some capability layer (like how ofDirectory took on some of the key features of ofxDirList) would also be worthwhile to consider.

bakercp:
By the way @tpltnt you are rocking it with all the code contributions lately.  Keep up the good work.  I hope you don't read push-back on a PR like this as criticism of your extremely helpful contributions! :+1:  I just closed one of my old PRs (https://github.com/openframeworks/openFrameworks/pull/2773), so I know how it goes!

tpltnt:
@bakercp thank you. I don't regard the push-back(s) as criticism. Your points are valid and you are the ones steering the project. I am just contributing where i see fit.

bilderbuchi:
so, close this then? @tpltnt "_deprecation_" of code means adding some kind of warning to it, so that users using it will be made aware that they should nor use it any more, and what they should use instead (and how). after a couple of releases the users should be aware of that and have migrated, at which time we can _remove_ it. you can search the codebase for `OF_DEPRECATE` if you want to know how this is handled currently.

tpltnt:
@bilderbuchi I am closing. I opened #4152 to handle the deprecation and i am going to open this PR again with the explicit intention of removing ofxXmlSettings. This could be merged after 0.9 when appropriate.

PR #4142, opened 2015-07-24
fixing compiler warnings of 45ad9ef

tpltnt:
Hi,

i fixed the code due to three compiler warnings in 45ad9ef287fdc764eaf8a2fa66625158dbb71e4d and reworked it in accordance with the oF style guide.
Cheers,
tpltnt


bakercp:
Hey there, I'm a bit confused about the across the board indenting on this one.  Why is the entire file indented?  Also, the QT Style is spaces not tabs ... yet for some reason our code style guide says tabs ... but then says to set your tabs == to 4 spaces ... anyway.  Just a bit confused ...

bilderbuchi:
@bakercp the QT style is only the baseline, which we adapted and modified where appropriate to make the most number of contributors happy. That is also the reason why we have tabs. It says to set your tab *width* to 4 spaces (which some editors have a setting for, that's purely for display, though), but it's still tabs for indentation. The [wording](https://github.com/openframeworks/openFrameworks/wiki/oF-code-style#indentation) in the code style was recently clarified about that.

tpltnt:
I tried to sync with master, but it only got worse. Closing it now since i doubt the warning warrant that much work.

PR #4145, opened 2015-07-25
Fix gst_element_query_duration argument type (int64_t to gint64).

PetrosKataras:
On Yosemite 10.10.4 with Clang 6.1, projects that use GStreamer fail to compile because of this type mismatch.


arturoc:
thanks!

PR #4148, opened 2015-07-25
Fix warnings and typos

bakercp:


bakercp:
@arturoc ping.  Small updates.

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
```diff
@@ -600,10 +600,10 @@ bool ofTrueTypeFont::load(string _filename, int _fontSize, bool _bAntiAliased, b
 			// 1-bit format, hella funky
 			// here we unpack it:
 			unsigned char *src =  bitmap.buffer;
-			for(unsigned int j=0; j <bitmap.rows;j++) {
```
arturoc: i have unsigned int in linux, a possible solution that would work on every version would be to use:

```
for(decltype(bitmap.rows) j=0; j <bitmap.rows;j++) {
```

arturoc: also we have width and height as local variables a some lines before in that function so we can do:

```
for(int j=0; j <height;j++) {
```

bakercp:
I used your second idea, as it's a little clearer.  I also the other places so we stay a bit more consistent.

bakercp:
@bilderbuchi -- indeed it should be an unsigned int, but on OSX, the lib is defined like this:

```c++
  /*************************************************************************/
  /*                                                                       */
  /* <Struct>                                                              */
  /*    FT_Bitmap                                                          */
  /*                                                                       */
  /* <Description>                                                         */
  /*    A structure used to describe a bitmap or pixmap to the raster.     */
  /*    Note that we now manage pixmaps of various depths through the      */
  /*    `pixel_mode' field.                                                */
  /*                                                                       */
  /* <Fields>                                                              */
  /*    rows         :: The number of bitmap rows.                         */
  /*                                                                       */
  /*    width        :: The number of pixels in bitmap row.                */
  /*                                                                       */
  /*    pitch        :: The pitch's absolute value is the number of bytes  */
  /*                    taken by one bitmap row, including padding.        */
  /*                    However, the pitch is positive when the bitmap has */
  /*                    a `down' flow, and negative when it has an `up'    */
  /*                    flow.  In all cases, the pitch is an offset to add */
  /*                    to a bitmap pointer in order to go down one row.   */
  /*                                                                       */
  /*                    Note that `padding' means the alignment of a       */
  /*                    bitmap to a byte border, and FreeType functions    */
  /*                    normally align to the smallest possible integer    */
  /*                    value.                                             */
  /*                                                                       */
  /*                    For the B/W rasterizer, `pitch' is always an even  */
  /*                    number.                                            */
  /*                                                                       */
  /*                    To change the pitch of a bitmap (say, to make it a */
  /*                    multiple of 4), use @FT_Bitmap_Convert.            */
  /*                    Alternatively, you might use callback functions to */
  /*                    directly render to the application's surface; see  */
  /*                    the file `example2.cpp' in the tutorial for a      */
  /*                    demonstration.                                     */
  /*                                                                       */
  /*    buffer       :: A typeless pointer to the bitmap buffer.  This     */
  /*                    value should be aligned on 32-bit boundaries in    */
  /*                    most cases.                                        */
  /*                                                                       */
  /*    num_grays    :: This field is only used with                       */
  /*                    @FT_PIXEL_MODE_GRAY; it gives the number of gray   */
  /*                    levels used in the bitmap.                         */
  /*                                                                       */
  /*    pixel_mode   :: The pixel mode, i.e., how pixel bits are stored.   */
  /*                    See @FT_Pixel_Mode for possible values.            */
  /*                                                                       */
  /*    palette_mode :: This field is intended for paletted pixel modes;   */
  /*                    it indicates how the palette is stored.  Not       */
  /*                    used currently.                                    */
  /*                                                                       */
  /*    palette      :: A typeless pointer to the bitmap palette; this     */
  /*                    field is intended for paletted pixel modes.  Not   */
  /*                    used currently.                                    */
  /*                                                                       */
  /* <Note>                                                                */
  /*   For now, the only pixel modes supported by FreeType are mono and    */
  /*   grays.  However, drivers might be added in the future to support    */
  /*   more `colorful' options.                                            */
  /*                                                                       */
  typedef struct  FT_Bitmap_
  {
    int             rows;
    int             width;
    int             pitch;
    unsigned char*  buffer;
    short           num_grays;
    char            pixel_mode;
    char            palette_mode;
    void*           palette;

  } FT_Bitmap;
```

bakercp:
@bilderbuchi though I do find it odd that it is unsigned here:

http://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#FT_Bitmap

hmmm are our osx headers up to date?

bakercp:
@arturoc @bilderbuchi Strange but true, the master branch does not have the latest freetype version installed.  I just recompiled with the master branch apothecary scripts and got an updated version of freetype with unsigned ints, etc.

Did we simply miss uploading the freetype headers / libs?

ping @danoli3, etc.

bilderbuchi:
ha, strange, nicely spotted! Interesting though that the lib works without problems with the headers and lib out of sync...

arturoc:
this won't merge anymore, can you fix it?

bakercp:
@arturoc Fixed.

PR #4150, opened 2015-07-26
Refactor gui

arturoc:
this solves some problems with the current architecture of ofxGui:

- when ofxGuiGroup created a gui element it was leaked on it's destruction, now elements created internally are destroyed.
- elements passed from outside are now passed by reference to make clear that they need to live until the gui is destroyed, this will probably be deprecated later and leave only the option to add an element using a parameter.
- all elements now have a Config struct that allows to configure the gui element on creation instead of having to keep a reference later, the config structs follow the same inheritance hierarchy as it's class so we can pass a base gui config to any element like:

```cpp
ofxBaseGui::Config config;
config.backgroundColor = ofColor::blue;
gui.add(floatParameter, config);
gui.add(boolParameter, config);
```
- this PR also makes it easy to add new elements that are not in the official ofxGui: any new element needs to have an inner Config class and a constructor that receives an ofParameter or ofParameterGroup + a config:

```cpp
class CircularSlider: public ofxBaseGui{
    struct Config: public ofxBaseGui::Config{
        Config(){}
        Config(const ofxBaseGui::Config & c):ofxBaseGui::Config(c){}
        float innerRadius = 5;
        float outterRadius = 10;
    }
    CircularSlider(ofParameter<float> & p, const Config & config);
    ....
}
```
any element with such characteristics can be added to the gui using:

```cpp
gui.add<CircluarSlider>(floatParameter);
```
or

```cpp
CircularSlider::Config config;
config.backgroundColor = ofColor::blue;
gui.add<CircluarSlider>(floatParameter, config);
```

or

```cpp
ofxBaseGui::Config config;
config.backgroundColor = ofColor::blue;
gui.add<CircluarSlider>(floatParameter, config);
```

- before we had a parent pointer on every element which being a raw pointer could be problematic if the objects were put on vectors... now every object has a sizeChangedE ofEvent through which it can notify that it's size has changed (this is what the parent pointer was mostly used for before)

ping @frauzufall, this solves most of the problems you found i think, allows to config elements without needing to get a pointer to them after being added, frees the elements, removes the use of most raw pointers which makes the ownership of the elements less ambiguous, and allows to add custom elements from parameters. does this fit with your extensions? we could add more events like sizeChangedE to notify changes upwards if you are using something else.

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -63,7 +63,6 @@ bool ofxBaseGui::useTTF = false;
 ofBitmapFont ofxBaseGui::bitmapFont;
 
 ofxBaseGui::ofxBaseGui(){
-	parent = NULL;
 	currentFrame = ofGetFrameNum();
 	serializer = std::shared_ptr <ofBaseFileSerializer>(new ofXml);
 
@@ -212,17 +211,20 @@ void ofxBaseGui::setPosition(float x, float y){
 void ofxBaseGui::setSize(float w, float h){
 	b.width = w;
 	b.height = h;
-	sizeChangedCB();
+	sizeChangedE.notify(this);
+	setNeedsRedraw();
 }
 
 void ofxBaseGui::setShape(ofRectangle r){
 	b = r;
-	sizeChangedCB();
+	sizeChangedE.notify(this);
+	setNeedsRedraw();
 }
 
 void ofxBaseGui::setShape(float x, float y, float w, float h){
 	b.set(x, y, w, h);
-	sizeChangedCB();
+	sizeChangedE.notify(this);
+	setNeedsRedraw();
 }
 
 ofPoint ofxBaseGui::getPosition(){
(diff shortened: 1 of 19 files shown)


frauzufall:
That looks awesome! I will adapt my code as soon as possible and tell you how it works out.

I used the parent object for ofxGuiSpacer to check if the parent is horizontal or vertical. I liked that I could just set `ofxGuiSpacer::setup(30)` and it would automatically set the right space instead of calling `ofxGuiSpacer::setup(30, 0)` or `ofxGuiSpacer::setup(0, 30)`. I would therefore favor to keep it in. And I just remember another more important reason: I wanted implement the ability of a toggle to be exclusive in a group into ofxToggle (https://github.com/frauzufall/ofxGuiExtended/issues/16). To do that, I need to know the parent group of a toggle.

arturoc:
> I wanted implement the ability of a toggle to be exclusive in a group into ofxToggle

you can have a ofxToggleGroup which can be multioption or exclusive in the second case it would listen the toggles events and set everyone else to false when one of the toggles changes

> I used the parent object for ofxGuiSpacer to check if the parent is horizontal or vertical. 

this is slightly trickier but i think we can just have a layout property in the base ofxBaseGui that the groups set to whatever correspond then any gui element can render differently depending on the layout.

i see how the pointer to the parent is really flexible because it allows to query properties that we haven't account for yet but i think it's problematic to rely on a raw pointer to do it, let's see if we can make what you have until now with the new Config objects and using events to notify upwards if it complicates things to much i don't mind recovering the parent pointer.

PR #4152, opened 2015-07-26
added deprecation warnings to ofxXmlSettings

tpltnt:
Hi,

this pull request add deprecation warnings to ofxXmlSettings in preparation of #4141.
Cheers,
tpltnt

### DIFF
--- a/addons/ofxXmlSettings/src/ofxXmlSettings.h
+++ b/addons/ofxXmlSettings/src/ofxXmlSettings.h
@@ -47,36 +47,36 @@ using namespace std;
 class ofxXmlSettings: public ofBaseFileSerializer{
 
 	public:
-        ofxXmlSettings();
-        ofxXmlSettings(const string& xmlFile);
+        OF_DEPRECATED_MSG("Use ofXml instead.", ofxXmlSettings() );
+        OF_DEPRECATED_MSG("Use ofXml instead.", ofxXmlSettings(const string& xmlFile) );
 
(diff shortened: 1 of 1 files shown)


bilderbuchi:
relevant issue: #2113

bilderbuchi:
@tpltnt can you test if you correctly get the deprecation warnings when using ofxXmlSettings, and test if the remaining uses [in the examples you found](https://github.com/openframeworks/openFrameworks/issues/2113#issuecomment-124740344) in the OF codebase show the warnings?
@openframeworks/core when we deprecate, should we also remove the examples, as we want to discourage further use of deprecated code?

tpltnt:
@bilderbuchi yes, i checked them (& fixed an issue) except "examples/ios/xmlSettingsExample". That one fails probably because i did not ran it on an Apple device.

tpltnt:
ping

bilderbuchi:
fine with me. pinging @openframeworks/core  again.

tpltnt:
what about this @openframeworks/core ?

arturoc:
@tpltnt we are working in 0.9.1 which is a patch version, bug fixes only, no api changes so this won't get merged until at least 0.10.

i don't think we should deprecate ofxXmlSettings just yet though. the api for ofXML is not very mature yet and depends on Poco which is going away so and until we have something that works well and is not dependenc on poco we can just keep ofxXmlSettings  around

PR #4153, opened 2015-07-26
Feature remove ofxXmlSettings

tpltnt:
Hi,

as a result of #4141, this pull request actually removes ofxXmlSettings since we have ofXml in place now. #4152 should be merged first to give everyone a heads up.
Cheers,
tpltnt


bilderbuchi:
I'm not sure if it makes sense to keep this PR around - as soon as #4152 is merged, this will become, I expect, unmergeable because of conflicts, and even if that doesn't happen this will probably code-rot as I don't realistically expect ofxXmlsettings to be removed in less than a year. 
Much better to make an issue about ofxXmlsettings removal (not deprecation), and milestone it to the next major milestone (i.e. 0.10.0), and decide how we should proceed when that release comes around.

ofTheo:
Yeah - I don't think we should remove ofxXmlSettings its probably one of the most widely used addons.

tpltnt:
Ok, no problem.

PR #4159, opened 2015-07-27
Fixes iterator types ofXml.

bakercp:
Replaces #3794


arturoc:
one thing i realized yesterday about this changes is that we should be better using:

```
for(auto & e: collection)
```

which is safer than using any index and won't break if the api changes

bakercp:
Indeed, that would be better in most cases and would solve a lot of these problems.

PR #4164, opened 2015-07-28
Update freetype headers to 2.5.5

bakercp:
Update freetype headers to match versions currently in apothecary scripts (v 2.5.5).


arturoc:
i merged this yesterday as par of a fix for emscripten

PR #4166, opened 2015-07-28
Unicode flags for Code::Blocks projects

procedural:
See #4103


procedural:
Anything else? I feel like OF has more than one CB project

arturoc:
i think this: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/win_cb/openFrameworksLib.cbp would need the same flags

procedural:
@arturoc certainly!

procedural:
@arturoc by the way, shouldn't emptyExample have `<Add option="-std=gnu++11" />` as well? It's missing it right now, add it too?

arturoc:
yeah it should, if you want to add it too go for it...

arturoc:
thanks!

Issue #4162, opened 2015-07-27
Fix remaining CI issues


bilderbuchi:
The new CI build has a couple of remaining TODOs/problems, I'll try to deal with those in further PRs

* [x] Linux build sometimes fails with " "g++-4.9: internal compiler error: Killed (program cc1plus)"", due to lack of RAM/swap, apparently. (also happens on Arturo's build). Could be that [this](http://stackoverflow.com/a/16372032/599884) or [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56746#c7) could reduce g++ memory consumption.
* [x] Sometimes (currently often) the apt-get step in linux builds fails with Error 137 when installing g++-4.9, reason currently unknown.
* [ ] doc builds fail, see HalfdanJ/ofDocGenerator#3. set to allow failure for now.
* [x] Move docs to docker builds (all packages are whitelisted), cf. travis-ci/travis-ci#4557
* [x] Confirm cache works correctly, maybe have to remove ever-changing files (.awk?)
* [x] add allAddonExample to build stage
   * ~~Ask for 14.04 beta access (done but doesn't work when using macos builds, too)~~ Done - needs a `dist: trusty` and `sudo: required` (so does not work on the docker builds)
   * `dist: trusty` flag to activate 14.04 workers does collide with `os: osx`, making those builds run on Linux. Therefore, this has been postponed until this is fixed by travis.
   * With 14.04, we (hopefully) get recent enough versions of the addons' dependencies (e.g. opencv, assimp) to activate compiling the allAddonsExample, which would cover more of the OF codebase.

bilderbuchi:
so,  g++ and apt-get being killed seems to have disappeared now.
I have no idea how to fix the docs build.

danoli3:
@arturoc Want to just add Linux to Allowed Failures for the moment till that is sorted?

arturoc:
it's now fixed

HalfdanJ:
Regarding the doc generator. Please just remove it form the travis file for now. I'm working on an update, but there is no reason for now to letting it fail all the time

bilderbuchi:
@HalfdanJ I just did, [looks godd so far](https://travis-ci.org/openframeworks/openFrameworks/builds/91895109).

PR #4167, opened 2015-07-28
WIP: CI improvements 2

bilderbuchi:
Don't merge yet. The fun continues. Related issue #4162

* [ ] Linux build sometimes fails with " "g++-4.9: internal compiler error: Killed (program cc1plus)"", due to lack of RAM/swap, apparently. (also happens on Arturo's build). Could be that [this](http://stackoverflow.com/a/16372032/599884) or [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56746#c7) could reduce g++ memory consumption.
* [x] ~~doc builds fail, see HalfdanJ/ofDocGenerator#3. set to allow failure for now.~~ this has been documented in scripts/ci/Readme.md
* [x] Move docs to docker builds (all packages are whitelisted), ~~doesn't work, see travis-ci/travis-ci#4592~~ worked around this by removing the root-level `os` key.
* [x] Confirm cache works correctly, maybe have to remove ever-changing files (.awk?)
* [x] add allAddonExample to build stage
   * ~~Ask for 14.04 beta access (done but doesn't work when using macos builds, too)~~ Done - needs a `dist: trusty` and `sudo: required` (so does not work on the docker builds)
   * `dist: trusty` flag to activate 14.04 workers does collide with `os: osx`, making those builds run on Linux. Therefore, this has been postponed.
   * With 14.04, we (hopefully) get recent enough versions of the addons' dependencies (e.g. opencv, assimp) to activate compiling the allAddonsExample, which would cover more of the OF codebase.
* [x] Re-add ofxOpenCv to the allAddonsExample
* [x] revert the os:level restructuring (960e4ebf09905126bf45e1954d0e5b9e339e464d and ec3750efe01a4d5fce24189d91310af84f1095e7), as those create an unwanted build without env vars.
* [x] stopping db process with `service` does not consistently suceed.
* [x] Change NDK excludes to reflect that we need toolchain 19 now, cf. #4214 Notes:
   * 19 and 21 take 360 of 790MB in the /platforms folder (21 most)
   * there is no 20
   * the other entries are 3,4,5,8,9,12-18


bilderbuchi:
hey @arturoc I just tried compiling the allAddonsExample on Travis, but it [fails here](https://travis-ci.org/openframeworks/openFrameworks/jobs/73946414#L1586-L1592), not finding cv.h when trying to compile ofxOpencv. Do you know what's going wrong?

arturoc:
probably the version of opencv in ubuntu 12.04 is too old, there might be a ppa for opencv but we should ask for access to the 14.04 beta

bilderbuchi:
Yeah I can do that.

bilderbuchi:
argh, even shutting down DB processes to save memory doesn't help with the Linux builds, they keep getting killed. >:-(

bilderbuchi:
rebased to solve conflicts

bilderbuchi:
argh the dist: trusty flags are pretty problematic/not working. :-( With our setup, we are obviously on the cutting edge of what Travis is capable or not capable of doing. :-/

arturoc:
i changed the .travis.yml today to run the tests instead of the emptyExample which won't use ofMain.h and seems to be better

arturoc:
have you tried removing the debug symbols? using g0 or removing it completely that could make it use less memory and we don't need them at all in the test build

arturoc:
also have you seen this: https://github.com/travis-ci/travis-ci/issues/1972  seems it's a problem that others are having and no apparent solution, the op says that using clang solves it, we might try that, it might skip some errors or trigger others that are not in gcc but it would be mostly ok i think.

arturoc:
and btw, this is the NDK excludes that i used to generate the current ndk package:

```
llvm-*
stlport
x86-4.8
docs
android-3
android-4
android-5
android-8
android-9
android-12
android-13
android-14
android-15
android-16
android-17
android-18
android-21
*mips*
*4.8*
*4.6*
python2.7
perl5
```

arturoc:
i'm testing and compiling with g0 seems to make a difference in the memory usage, if you want to test it just add:

```
PROJECT_OPTIMIZATION_CFLAGS_DEBUG = -g0
```

at the end of config.make

arturoc:
i've tested this with your changes to stop the databases + the -g0 here: https://github.com/openframeworks/openFrameworks/pull/4266 i've restarted the build about 15 times and hasn't failed once. if you want to integrate it with your changes? would be good to get it in master so the linux builds are reliable

bilderbuchi:
> also have you seen this: https://github.com/travis-ci/travis-ci/issues/1972  seems it's a problem that others are having and no apparent solution, the op says that using clang solves it, we might try that, it might skip some errors or trigger others that are not in gcc but it would be mostly ok i think.

yes I've seen this. Clang sounds good, it's also what I would have tried. How do I do this, and what's the minimum clang version? just export CXX=clang or somesuch before running the build?

I'll take your changes from #4266 - would be really great if that fixed the problems!

arturoc:
let's try this first if it doesn't work then we move it to clang. and yes. you would need to install clang 3.6 instead of gcc4.9 i think it's in the same ppa we are using: https://github.com/neovim/neovim/commit/b456d084a2c500d6522b45a42d03c5a36f57732d

then just call 

```
make CXX=clang Debug
```

in any case i would switch -g0 apart from seemingly solving the crashes the builds seem to be taking around 2 minutes less

bilderbuchi:
alright. Btw, have you seen in any builds that the install script fails at the apt stage (as that gets killed by Travis after ~30s)? This currently happens sometimes, and it's pretty annoying as it does not fail the build automatically, so it only fails later during compilation.

bilderbuchi:
argh, it has [just happened again](https://travis-ci.org/openframeworks/openFrameworks/jobs/76065449#L723). has our set of dependencies grown recently?

arturoc:
yes i've seen it once only i think. guess once we move to docker instances all this will be solved or at least better, let's add the -g0 flag and merge it as it is since seems to be better than what we have now, perhaps try allAddonsExample. if it keeps failing in master then we can make it compile with clang

bilderbuchi:
> yes i've seen it once only i think. guess once we move to docker instances all this will be solved or at least better

I have tried some things that maybe fix this. The problem with docker instances is that we will not be able to test our install_dependencies script (no sudo there). Also, I'm not sure if we can get recent enough versions of all our dependencies for alladdonsexample to compile on 12.04, and 14.04 is not available on docker. I've documented over at #4162

> let's add the -g0 flag and merge it as it is since seems to be better than what we have now, 

yes, that's already added. I'm trying some more things, then ask for a quick review, and merge.

> perhaps try allAddonsExample. 

see above .

> if it keeps failing in master then we can make it compile with clang

yeah, agreed.

bilderbuchi:
OK, the failures in the install stage still happens, but now it at least [errors out](https://travis-ci.org/openframeworks/openFrameworks/jobs/76081046) instead of hiding the failure.

arturoc:
ok let's merge it as it is then. i think instead of the allAddonsExamples we can add individual tests for each addon that can be just some application that compiles every supported addon by now

bilderbuchi:
@arturoc so, [apparently](https://travis-ci.org/openframeworks/openFrameworks/jobs/76086799) it's _only_ the g++-4.9 installation that makes the troubles (pretty consistently, too). although why it should take too much memory to install this I have no idea.

bilderbuchi:
OK, I've written to Travis support (again), and in the meantime have switched the linux build to `allow_failures` so that the builds at least stay green until this is fixed.

bilderbuchi:
So, I think we can merge this now, review please @arturoc @danoli3. 
Linux builds are very flaky/unstable currently, so they have been added to allow_failures for now. MacOS builds run and compile emptyexample and also run tests.

arturoc:
let's switch to clang then, i'd rather prefer having this fail some times than have tests that don't pass in a platform and don't notice in which commit they were introduced

bilderbuchi:
> let's switch to clang then, i'd rather prefer having this fail some times than have tests that don't pass in a platform and don't notice in which commit they were introduced

That won't fix it I fear, as install_dependencies will still install g++-4.9.
I could try to set up an additional Linux build using docker (no testing install_dependencies, and I don't know if we can get all dependencies) to see if that improves things?

bilderbuchi:
edit: restarting builds seems to often fix the problems, but we can't do that all the time. I'll work on getting a linux docker build to run.

arturoc:
i mean, make install_dependencies install clang instead of gcc we can pass an argument to the script to choose between the 2.

arturoc:
can you move linux out of allow fail and merge this? let's see how much it fails in master first and then we can decide if we should use clang

kylemcdonald:
out of curiosity: is gcc only failing on travis, or will some linux users also have intermittent problems building OF with gcc?

arturoc:
only on travis

arturoc:
it's a memory limitation issue and gcc provide things that clang still doesn't

bilderbuchi:
> can you move linux out of allow fail and merge this? let's see how much it fails in master first and then we can decide if we should use clang

alright, done.

arturoc:
great!

PR #4170, opened 2015-07-29
fixes for adding frameworks to osx projects

ofZach:
this is a fix for #3774, adding non system frameworks to osx projects.   It works for syphon which is the test case. 

I haven't tested system level frameworks (side note, is there a good addon that adds other system frameworks to test on?)  fixed also an issue where the framework was getting parsed recursively (since it is a folder) and added to header search paths. 

ping @bangnoise @martialgallorini

### DIFF
--- a/addons/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/addons/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -1104,12 +1104,22 @@ void xcodeProject::addAddon(ofAddon & addon){
              addFramework( addon.frameworks[i] + ".framework", "/System/Library/Frameworks/" + addon.frameworks[i] + ".framework", "");
         } else {
             
-            vector < string > pathSplit = ofSplitString(addon.frameworks[i], "/");
+            if (ofIsStringInString(addon.frameworks[i], "/System/Library")){
+                
+                vector < string > pathSplit = ofSplitString(addon.frameworks[i], "/");
+                
+                addFramework(pathSplit[pathSplit.size()-1],
+                             addon.frameworks[i],
+                             "");
+                
+            } else {
             
-            addFramework(pathSplit[pathSplit.size()-1],
-                         addon.frameworks[i],
-                         addon.filesToFolders[addon.frameworks[i]]);
-            
+                vector < string > pathSplit = ofSplitString(addon.frameworks[i], "/");
+                
+                addFramework(pathSplit[pathSplit.size()-1],
+                             addon.frameworks[i],
+                             addon.filesToFolders[addon.frameworks[i]]);
+            }
         }
                                                                                             
                                                                                             //
--- a/addons/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/addons/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -72,6 +72,26 @@ STRINGIFY(
 
 );
 
+//-----------------------------------------------------------------
+const char PBXFileReferenceWithoutEncoding[] =
+STRINGIFY(
+          
+          <key>FILEUUID</key>
+          <dict>
+          <key>explicitFileType</key>
+          <string>FILETYPE</string>
+          <key>isa</key>
+          <string>PBXFileReference</string>
+          <key>name</key>
+          <string>FILENAME</string>
+          <key>path</key>
+          <string>FILEPATH</string>
+          <key>sourceTree</key>
+          <string>SOURCE_ROOT</string>
+          </dict>
+          
+          );
+
 
 //-----------------------------------------------------------------
 const char PBXFileReferenceXib[] =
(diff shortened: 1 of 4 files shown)


ofZach:
also, is there a need for this for ios?  (if so, is there a potential test case addon to check?)

bakercp:
@ofZach https://github.com/bakercp/ofxTouchPad/tree/develop uses a private system framework.  You'll also need https://github.com/bakercp/ofxPointer/tree/deveop and https://github.com/jefftimesten/ofxJSON

bakercp:
Also, all above, except ofxJSON use the develop branch with 0.9.0 for the latest.

ofZach:
@bakercp cool will give it a try... 

to test it with the PG the .mk file needs a line like: 

ADDON_FRAMEWORKS = ...

I think right now, there's an assumption that a framework with no path is a system framework but we might need to be smarter and check, like seeing if it's a path starting with "/system"....

the reason for the differentiation is that non system frameworks get added to the sidebar and embedded in the .app itself but system frameworks (I think) will skip some of those steps.

ofZach:
@bakercp can confirm that this doing something *close* to the right thing...  there might be subtle things to tweak.  made an addon that looks like this: 

	meta:
		ADDON_NAME = ofxFake2
		ADDON_DESCRIPTION = 
		ADDON_AUTHOR = 
		ADDON_TAGS = 
		ADDON_URL = 

	common:
		ADDON_FRAMEWORKS = /System/Library/PrivateFrameworks/MultitouchSupport.framework

then when I create the file, the framework is added, but it's potentially a little wonky -- some quick notes.  It actually appears in the source folder, since there was no folder such as an addon suggesting where to put it: 

screen shot 2015-07-28 at 11 33 59 pm

and I've noticed that the path to the framework is added to the framework search paths (which might not be necessary?).   there isn't a build step (such as copying the framework into the .app) for system frameworks the way there is for non system frameworks which need the dylib added.

bakercp:
@ofZach hey -- I'd like to test this, but am a little confused about the current location of the app :)  I understand the core code has been moved to the ofxProjectGenerator addon -- but where is the latest app?

ofZach:
it's a submodule here: 
https://github.com/openframeworks/openFrameworks/tree/master/apps

bangnoise:
Big 👍 as a fix for #3774 

Re System frameworks - behaviour should probably be the same for System frameworks and user-installed frameworks in /Library (not /System/Library) - link them but don't copy them into the app bundle.

addons/ofxProjectGenerator/src/projects/xcodeProject.cpp
```diff
@@ -1035,11 +1101,28 @@ void xcodeProject::addAddon(ofAddon & addon){
         
         size_t found=addon.frameworks[i].find('/');
         if (found==std::string::npos){
-             addFramework( addon.frameworks[i] + ".framework", "/System/Library/Frameworks/" + addon.frameworks[i] + ".framework");
+             addFramework( addon.frameworks[i] + ".framework", "/System/Library/Frameworks/" + addon.frameworks[i] + ".framework", "");
```
bangnoise: might also be in `/Library/Frameworks` (user-installed Frameworks)

ofZach:
thanks -- I will try both (a) not adding the framework to the left sidebar and (b) checking on user installed frameworks for the system level things. 

I do think before this PR is merged, I'd love to hear from @openframeworks/ios and to check any addons that might need similar help on the ios side.  Are there any ios addons that either include a framework ala ofxSyphon or reference a system level framework ala ofxTouchPad ?

ofZach:
@admsyn sorry that version doesn't have the changes in this PR, which wasn't pulled in yet (still looking at how to handle OS level frameworks and ios): 

https://github.com/openframeworks/openFrameworks/pull/4170

if you grab that PR, compile the command line and swap it out it (in the package of the august hackpad agenda or the later one I posted to the same issue thread) should give you better luck.

ofZach:
@admsyn here's a zip of that command line tool if you are feeling lazy, swap it and you should see improvements (I just tried it and for ofxSyphon, before I tried I got your error but after it worked): 

https://www.dropbox.com/s/qo0y2sxcjfqm8zh/commandLinePG.zip?dl=0

admsyn:
Ah, that worked! :)

ofZach:
I think this is good to merge now, tested IOS and it adds / links system frameworks ok now.   I also put non system frameworks in the addon folder itself in the sidebar to it's easier to track.  There may still be edge cases to figure out but this should work for the majority of cases.    This PR also has some work to quell the number of errors / warnings on making projects but there's still some more work to be done.

julapy:
hi @ofZach, (cc @ofTheo / @openframeworks/macos)
in reply to your comment in #4226,
i've just started digging around in the osx xcode template project and was really surprised to see the frameworks removed from the side bar, had no idea that was possible! :) but i can now see they've all been moved to `CoreOF.xcconfig`

yeah i think it would be a good idea to mirror the osx setup in the ios template project too.
osx and ios projects are very similar and we should make them as consistent as possible.

apart from the framework changes, has there been anything else changed in the osx template project?

ofZach:
I think @ofTheo will know the best, I think the scripts might have been altered (the run script build phase) but I'm not sure what else....

arturoc:
should we merge this?

ofZach:
yes!  thanks

ofZach:
just a ping to remind us to add this PR, I don't know what the conflict is but I can take a look later today.  I'm trying to tighten up the pg gui now so all pg related changes happening soon help.

arturoc:
just merged this in the PG repo since now the addon is there

ofZach:
thank you !!!!

PR #4173, opened 2015-07-29
Const corrections and std::make_shared use.

bakercp:
Getting rid of the "new" keyword in favor of std::make_shared when possible.

Clarify and document a complex ofParameter function that is efficient but difficult to grasp at first glance.


arturoc:
thanks!

Issue #3506, opened 2014-12-29
Document 0.9.0 Platform Support Coverage


bakercp:
Please edit this list directly if you have more information :)

OS X
- ~~10.6~~ _Removed_
  - _No support for libc++ / c++11, etc_
- 10.7 - 10.10 (_FAT Bundle_)
  - Note: QTKitVideo pipeline will be deprecated, but essential for 10.7 support.
  - i386 (with libc++)
  - x86_64 (with libc++)

iOS (using 10.9+ OSX)
- ~~iOS 4~~ _Removed_
  - _see #2568 for more discussion_
- iOS 5.1 - 9.1 (_FAT Bundle_ with libc++ for all archs)
  - armv7
  - ~~armv7s~~ _Removed_
    - _potential performance gains not worth the binary size increase, see #3423 for more discussion_
  - arm64
  - i386 
  - x86_64

Android
- OS?
  - ~~armv5~~ _Removed_
    - _No new hardware for this._ 
  - armeabi-v7a
  - i386

Windows
- Win7 - Win8
  - Visual Studio 2012
    -  i386
  - MinGW / Codeblocks 12.11
    -  i386

Linux
  - i386
  - x86_64
  - armv6l
  - armv7l

Emscripten
  - VM

pizthewiz:
It might also be useful to document our tooling support as well.

On the iOS side, `arm64` was not supported until Xcode 5 and Xcode 5 requires OS X 10.8 to run, so should 10.8 become our minimum for OS X, that would be a natural fit.

(Feel free to remove this comment)

danoli3:
For iOS Apple are really stepping up that you have to be at the latest in order to submit to the AppStore.
So if a developer wants to use oF with iOS, they are going to have to be at the latest Xcode and SDK in order to do that.

However there are probably a lot of developers who just want to make OS X apps... and if they are on 10.8, shouldn't stop them unless they want to make iOS apps specifically, then it's apple stopping them not us.

notjosh:
fwiw, Apple make it fairly easy/cheap (read: free) to be on the latest OS/dev tools, and they run on hardware from as far back as 2007.

As [64-bit is required for submission now](http://9to5mac.com/2014/12/17/apple-reminds-developers-about-64-bit-ios-app-store-requirement-app-updates-now-must-also-comply-by-june/), it'd seem wise to bump the minimum to what'll cause the fewest unexpected issues throughout dev/submission imo.

[...]

PR #4178, opened 2015-08-01
current platform list

tpltnt:
Hi,

i created a document to address (and solve?) #3506.
Cheers,
tpltnt


bilderbuchi:
I think this should be called "supported_platforms.md" and be moved into the `/docs` directory (with a link in the table of contents).
do we need to update its contents? (it's been a while since it was created) @bakercp @kylemcdonald @arturoc @openframeworks/macos

PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
+As of 0.9, the following platforms are supported:
```
bilderbuchi: I think we should remove the reference to a specific platform (So just say "Currently, the following.." or so), otherwise this is yet another place where we have to update a version string, and the version should be clear from the OF release this document is delivered with.

bilderbuchi:
@tpltnt any chance to update this with the suggested changes?

tpltnt:
@bilderbuchi done

bilderbuchi:
thanks!
I still think this:
> I think this should be called "supported_platforms.md" and be moved into the /docs directory (with a link in the table of contents).

should still be done (trying to keep the root directory as clean as possible), then we're probably good to go. @kylemcdonald ?

tpltnt:
@bilderbuchi @kylemcdonald I moved it to docs/

bilderbuchi:
thanks! could you add a link to the file in the [TOC document](https://github.com/openframeworks/openFrameworks/blob/master/docs/table_of_contents.md), Getting started section?

danoli3:
(this is very exciting!!!!!!!!!!!!!!!!!!!!! :D)

docs/PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
+openFrameworks currently supports the following platforms:
+
+# Android #
+
+versions: ?
```
arturoc: this is android 2.2 or +

docs/PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
...
+
+architectures:
+  * armeabi-v7a
+  * i386
```
arturoc: this is i686

arturoc: or better perhaps change i386 to x86 it's wat we use in most cases in the folders

docs/PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
...
+  * x86_64
+  * armv6l
+  * armv7l
+	
+# Microsoft Windows #
+
+versions: Win7 - Win8
+
+architectures: ?
```
arturoc: x86 and x86_64

docs/PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
...
+
+notes:
+  * Visual Studio 2012 for i386
```
tgfrerer: Visual Studio 2015, + Windows 10

tpltnt:
Ok, my git repo seems to be profoundly out of sync. I will create another PR.

PR #4189, opened 2015-08-05
Implicit type conversion warnings.

bakercp:


libs/openFrameworks/types/ofColor.cpp
```diff
@@ -169,7 +169,7 @@ A clampedMultiply(const A& a, const B& b) {
 }
 
 template<typename PixelType>
-float ofColor_<PixelType>::limit() {
+PixelType ofColor_<PixelType>::limit() {
```
arturoc: this is kind of dangerous, i'm pretty sure in some places we are doing things like:
```
value/limit()
```

to normalize a value and this change will break those places for integer numbers, casting this to float would avoid that and the warning.

in general i'm a little nervous about this changes of the types being used, if it's to solve warnings i guess we need to do it but otherwise i would hold until 0.9 is released so we have more time to test. it's usually really big commits and even if most changes are innocuous, things like this can really easily slip through.

bakercp: Yeah, I'll revert this one and we can cast where needed and test after 0.9.0.

bakercp:
@arturoc I reverted the return type.

libs/openFrameworks/utils/ofUtils.h
```diff
@@ -735,11 +735,11 @@ char ofToChar(const string& charString);
 template <class T>
 string ofToBinary(const T& value) {
 	ostringstream out;
-	const char* data = (const char*) &value;
+	const unsigned long long* data = static_cast<unsigned long long*>(&value);
```
arturoc: is this change well tested? don't wont to merge something that might break at this point but if this fixes warnings i'll do it. also can you change unsigned long long with uint64_t?

bakercp: Yeah, it's the default argument for the bitset, so it definitely gets rid of warnings.  I'll update to the more compact type.

bakercp:
@arturoc this should be ready to go now.

PR #4196, opened 2015-08-05
Fix unhandled enum in switch statement.

bakercp:


bakercp:
@arturoc This is still causing warnings in various platforms.

arturoc:
thanks!

PR #4200, opened 2015-08-06
MinGW related fixes

procedural:


libs/openFrameworks/utils/ofSystemUtils.cpp
```diff
@@ -378,14 +378,14 @@ ofFileDialogResult ofSystemLoadDialog(string windowTitle, bool bFolderSelection,
 		wchar_t szFileName[MAX_PATH];
 		wchar_t szTitle[MAX_PATH];
 		if(defaultPath!=""){
-			wcscpy_s(szFileName,convertNarrowToWide(ofToDataPath(defaultPath)).c_str());
```
ofTheo: Did you check if this change also compiles on VS? 
I remember from a while back VS would require one version of a wide string related function and CB would require another.

procedural: Yea, I've changed that in https://github.com/openframeworks/openFrameworks/pull/4139 and https://github.com/openframeworks/openFrameworks/pull/4166 so MinGW supports wide strings as well.

Here I'm replacing MSVC-specific `wcscpy_s` with `wcscpy` since they're not different except `wcscpy_s` ensures the string will be null terminated, but in this case `ofToDataPath(defaultPath)).c_str()` returns null terminated string anyway, so no need for that.

ofTheo:
great thanks!!

PR #4208, opened 2015-08-09
mingw working on msys2.

arturoc:
This adds a scripts/win_cb/msys2/install_dependencies.sh script which installs
the necessary dependencies under msys2.
    
The libraries which are not avalible through msys2 are compiled through apothecary:
      - poco: which is on the msys2 repos but doesn't link properly
      - tess2
      - glfw3
    
This also adds Makefile and config.make to the mingw template

For this to work under codeblocks we would need to add a MSYS2_ROOT variable to the project file and use that to add the corresponding include paths. I would suggest to just drop codeblocks, it hasn't really changed much in the last five years and it's really weird to use, we can have visual studio as the only IDE and provide a download with makefiles so ofSketch keeps working but also people who want to use OF on a different IDE can do it through the makefiles


procedural:
MinGW-w64 (and soon Clang) is a solid alternative to MSVC worth keeping. 

+1 for dropping Code::Blocks. Would be cool just to link on ofSketch on the main site for alternative. There really no alternatives to Visual Studio on Windows aside of Qt Creator or JetBrains CLion which are, unfortunately, doesn't support Makefiles projects directly.

procedural:
Qt Creator supports GNU Autotools, I don't have experience with it, but does it relate to Make somehow?

arturoc:
this is actually working with clang. and no autotools generate makefiles but it's a previous step, it's similar to cmake in the sense that you provide it with some config files and it detects the platform... and generates makefiles.

procedural:
>this is actually working with clang

What works with clang? :)

arturoc:
this branch, it uses clang under msysw2 instead of gcc since gcc was crashing

arturoc:
i'm going to merge this then and see how we can generate the packages them

procedural:
Well, currently not all of OF works with Clang from MSYS2:

* OF doesn't draw circles with `ofDrawCircle` for some reason
* OF crashes if programmable renderer is used, see [this patch](https://github.com/ofnode/of/blob/master/dev/patches/openframeworks_clang_windows.patch) that replaces unordered_map that currently doesn't work properly in Clang with standard map

arturoc:
i'm testing in master now and both ofDrawCircle and the programmable renderer work without problem.

procedural:
@arturoc huh, I should retest it now, it was before GCC's 5 standard library which Clang uses...

procedural:
@arturoc Indeed, ofDrawCircle works with Clang now :) But I still get this weird crash from unordered_map call in `ofShader.cpp` and libstdc++-6 for some reason...

```
rogram received signal SIGSEGV, Segmentation fault.
0x000000006fc9a07b in ?? () from C:\msys64\mingw64\bin\libstdc++-6.dll
(gdb) bt
#0  0x000000006fc9a07b in ?? () from C:\msys64\mingw64\bin\libstdc++-6.dll
#1  0x00000000006eba06 in _M_insert_unique_node ()
    at C:\msys64\mingw64\bin/../include/c++/5.2.0\bits/hashtable.h:1594
#2  0x00000000006d2703 in operator[] ()
    at C:\msys64\mingw64\bin/../include/c++/5.2.0\bits/hashtable_policy.h:600
#3  operator[] ()
    at C:\msys64\mingw64\bin/../include/c++/5.2.0\bits/unordered_map.h:668
#4  setupShaderFromSource () at ../src/openframeworks/gl\ofShader.cpp:183
#5  0x00000000006be5a6 in setup ()
    at ../src/openframeworks/gl\ofGLProgrammableRenderer.cpp:2309
#6  0x000000000069470b in __fu779_glewExperimental ()
    at ../src/openframeworks/app\ofAppGLFWWindow.cpp:262
#7  0x0000000000693bbd in setup ()
    at ../src/openframeworks/app\ofAppGLFWWindow.cpp:114
#8  0x000000000069b343 in createWindow ()
    at ../src/openframeworks/app\ofMainLoop.cpp:64
#9  0x0000000000698fc8 in ofCreateWindow ()
    at ../src/openframeworks/app\ofAppRunner.cpp:182
#10 0x0000000000401618 in main () at ../src\main.cpp:7
```

arturoc:
can you post an example that crashes? i've tested the shader tutorials which use the programmable renderer and they were working fine

procedural:
@arturoc 

```cpp
#include "ofMain.h"

class ofApp : public ofBaseApp
{
  public:
    void draw();
};

int main()
{
  ofGLWindowSettings settings;
  settings.setGLVersion(3,2);
  auto app = make_shared<ofApp>();
  ofRunApp(window, app);

  return ofRunMainLoop();
}

void ofApp::draw()
{
  ofSetColor(ofColor::red);
  ofDrawCircle(100, 100, 50);
}

**Comment by procedural** [comment]
I mean, it totally weird that libstdc++, *the* standard library of C++, causes *a segfault*. That's messed up :) It's fixable by replacing unordered_map with plain map in my case, but overall non-deterministic behavior of C++ is killing me lately...

**Comment by procedural** [comment]
Haha :D

```cpp
#include <unordered_map>
#include <string>

int main()
{
  std::unordered_map<std::string, int> test;
  test["cool"] = 5;
}
```

```
Program received signal SIGSEGV, Segmentation fault.
0x000000006fc9a07b in ?? () from C:\msys64\mingw64\bin\libstdc++-6.dll
(gdb) bt
#0  0x000000006fc9a07b in ?? () from C:\msys64\mingw64\bin\libstdc++-6.dll
#1  0x00000000004019c6 in _M_insert_unique_node ()
    at C:\msys64\mingw64\bin/../include/c++/5.2.0\bits/hashtable.h:1594
#2  0x0000000000401962 in operator[] ()
    at C:\msys64\mingw64\bin/../include/c++/5.2.0\bits/hashtable_policy.h:625
#3  0x0000000000401622 in operator[] ()
    at C:\msys64\mingw64\bin/../include/c++/5.2.0\bits/unordered_map.h:672
#4  main () at ../src\main.cpp:7
```

procedural:
It crashes even with default constructor:

```cpp
#include <unordered_map>

int main()
{
  std::unordered_map<int, int> test { {0, 1} };
}
```

procedural:
Opened an issue in MSYS2, kinda wrong place for that but I don't where else: https://github.com/msys2/msys2.github.io/issues/7

arturoc:
this works for me without problem with latest msys2 on master.

arturoc:
i mean the programmable renderer example that you posted

procedural:
@arturoc yeah, I faced some kind of bug in Clang and standard c++ library I guess, reported it here: https://llvm.org/bugs/show_bug.cgi?id=24408

What's about undefined constants from `climits`?

arturoc:
i fixed that already in master

procedural:
Cool, thanks!

BananasGoMoo:
Hi,

I see you dropped support for code::blocks, but I was wondering, you mentioned that to make it work it code::blocks it requires an MSYS2_ROOT variable. How would I set this up? do I just add the variable to code::blocks and make it point to the root directory of MSYS2? are there any other steps I need?

PR #4211, opened 2015-08-09
using latest version of fmodx

i-n-g-o:
using latest version of fmodx in apothecary script


bakercp:
@i-n-g-o is there a reason this needs to be updated for 0.9.0?  If we are updating the script then we should probably also update the dynamic libs as well?

arturoc:
we were testing yesterday to fix a problem that happens between fmod and rtaudio. this version was working better but we decided not to include it since it didn't fix the problem completely and we found a better workaround.

going to close this

PR #4222, opened 2015-08-10
Update version string to rc1

bakercp:


arturoc:
i don't think this should ever go into master. ideally we would create a branch for the release and only have the rc there. since rc1 is already published anyway i don't think it makes sense to change it now but let's talk about how we do it for rc2 and future releases.

bakercp:
Yeah, that makes sense.  If we want to keep branches under control and stick with tags, perhaps it should basically be a double commit when we tag it -- update the string, tag and then the very next commit is changed back to `master`.

bilderbuchi:
> update the string, tag and then the very next commit is changed back to master.

Yeah, +1, that's a method I've seen in several other projects.

PR #4226, opened 2015-08-11
removed blank lines from shellScript in ios xcode template project

julapy:
this was tripping up the compiler for some reason and it was giving errors in xcode 6.4


ofZach:
@julapy  (cc @ofTheo / @openframeworks/macos) just was wondering if the changes to the OSX side (moving frameworks to the config file, etc) should be reflected in the IOS template?   I was digging into IOS a bit on this issue 

https://github.com/openframeworks/openFrameworks/pull/4170

and I saw that the templates look a little different from each other because of those config changes, etc.   It's not a big deal but I thought I would flag it.

PR #4228, opened 2015-08-11
ofThreadChannel documentation.

### LINKED ISSUE #4225: Document ofThreadChannel
Hey all, I'm updated and formatting documentation for the ofThreadChannel, just in case someone else was thinking about it (or doing it?).  It's a very nice class.  Will submit a PR shortly.

bakercp:
Fixes #4225

@kylemcdonald as one recently using ofThreadChannel, would you mind giving this a proof-read?

### DIFF
--- a/libs/openFrameworks/utils/ofThreadChannel.h
+++ b/libs/openFrameworks/utils/ofThreadChannel.h
@@ -9,16 +9,14 @@
 
 /// \brief Safely send data between threads without additional synchronization.
 ///
-/// ofThreadChannel is a powerful class that makes it easy to safely pass data
-/// between threads. When data is sent form one thread to another via the
-/// ofThreadChannel, the user does not need to use additonal mutexes, condition
-/// variables, critical sections, semaphores, monitors or other synchronization
-/// primitives or techniques. Data sent with ofThreadChannel is synchronized
-/// by ofThreadChannel's internal `std::mutex` and notifies receiving threads when
-/// data is available from the sending thread.
+/// ofThreadChannel makes it easy to safely and efficiently share data between
+/// threads without the need for shared memory, mutexes, or other
+/// synchronization techniques. Additionally, ofThreadChannel employs a
+/// signalling system that allows receiving threads to sleep until new data
+/// arrives or the ofThreadChannel is closed.
 ///
-/// A single ofThreadChannel class implements uni-directional communication. In
-/// most cases an additional ofThreadChannel can be used for bi-directional
+/// A single ofThreadChannel class is desgined for one-way communication. In
+/// most cases an additional ofThreadChannel can be used for two-way
 /// communication.
 ///
 /// Data is sent through the ofThreadChannel in a FIFO (first in, first out)
@@ -26,32 +24,10 @@
 /// it was sent.
 ///
 /// If multiple threads attempt to send data using the same ofThreadChannel, the
-/// send method will block the calling thread until it is free. ofThreadChannel
-/// does not use recursive mutexes and is therefore not designed to send and
-/// receive data from the same thread. Doing will result in deadlocks.
+/// send method will block the calling thread until it is free. Sending and
+/// receiving data from the same thread may result in deadlocks.
 ///
-/// ofThreadChannel is a templated class, meaning that the value type must be
-/// set as a template argument during declaration.
-///
-/// ofThreadChannel is ideal for sharing most complex data structures between
-/// threads. For example, if you want to share `ofPixels` between threads, one
-/// would declare `ofThreadChannel<ofPixels> myChannel`. When used
-/// with traditional C++ syntax, the data type must be copyable and one copy of
-/// the data will be made when it is sent. This copy (and any associated
-/// performance penalties) can be avoided using modern C++ `std::move`
-/// semantics, but remember that source data will be invalidated.
-///
-/// While ofThreadChannel will certainly work for "plain old datatypes" (e.g.
-/// `int`, `char`, `float`, etc), users are encouraged to use the new
-/// `std::atomic` data types available with C++11 and later. These "lock-free"
-/// data types require no special data protection for shared use between threads
-/// and are guaranteed to be "thread-safe".
-///
-/// \sa http://en.cppreference.com/w/cpp/thread/mutex
-/// \sa http://en.cppreference.com/w/cpp/utility/move
-/// \sa http://en.cppreference.com/w/cpp/atomic/atomic
 /// \sa https://github.com/openframeworks/ofBook/blob/master/chapters/threads/chapter.md
-/// \sa https://en.wikipedia.org/wiki/Thread_(computing)
 /// \tparam T The data type sent by the ofThreadChannel.
 template<typename T>
 class ofThreadChannel{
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/utils/ofThreadChannel.h
```diff
@@ -1,28 +1,97 @@
...
+/// A single ofThreadChannel class implements uni-directional communication. In
+/// most cases an additional ofThreadChannel can be used for bi-directional
+/// communication.
+///
+/// Data is sent through the ofThreadChannel in a FIFO (first in, first out)
+/// order, guaranteeing that that data will be received in the same order that
+/// it was sent.
+///
+/// If multiple threads attempt to send data using the same ofThreadChannel, the
+/// send method will block the calling thread untill it is free. ofThreadChannel
```
kylemcdonald: untill => until

libs/openFrameworks/utils/ofThreadChannel.h
```diff
@@ -1,28 +1,97 @@
...
+/// does not use recursive mutexes and is therefore not designed to send and
+/// receive data from the same thread. Doing will result in deadlocks.
+///
+/// ofThreadChannel is a templated class, meaning that the value type must be
+/// set as a template argument during declaration.
+///
+/// ofThreadChannel is ideal for sharing most complex data structures between
```
kylemcdonald: might move this paragraph closer to the top. for me, it's the main takeaway.

kylemcdonald:
looks great to me! :+1:

libs/openFrameworks/utils/ofThreadChannel.h
bakercp: Which paragraph -- sorry github ambiguity ...

bakercp: Wait I think it was a mobile phone induced ambiguity.  Is it line 36?

libs/openFrameworks/utils/ofThreadChannel.h
```diff
@@ -1,28 +1,97 @@
...
-/// Usually a second channel can be used to send the
-/// processed data back to the first thread.
+
+/// \brief Safely send data between threads without additional synchronization.
+///
+/// ofThreadChannel is a powerful class that makes it easy to safely pass data
+/// between threads. When data is sent form one thread to another via the
+/// ofThreadChannel, the user does not need to use additonal mutexes, condition
+/// variables, critical sections, semaphores, monitors or other synchronization
+/// primitives or techniques. Data sent with ofThreadChannel is synchronized
```
arturoc: i don't think it's necessary to mention every possible sync primitive here :) doesn't add to the explanation and it might be confusing i think

arturoc:
hey @bakercp i was adding notes in several places but there's too many things that i don't think are correct. i think in general it's too long and gives to many implementation details that are not really necessary to know how the class should be used. For example the fact that there's a mutex and specifically an std one doesn't really mean anything for anyone trying to use this class and is not even important since it could perfectly be implemented using a lock free queue (and possibly will be in the future). And the key is really in the condition variable that allows to lock the receiving thread while there's no data to be processed but i wouldn't really mention the particular sync primitive either.

In general the most important from this class is that it allows 2 threads (or more but 2 is the most common) to communicate without using explicit locks. That it has a receive method that locks until there's data which allows to completely sleep an auxiliary thread until it has anything to do. The idea behind it is to not use shared memory even behind mutexes and instead send messages from one thread to another which avoids any possible race conditions.

There's also details that are not accurate like:
```
+/// While ofThreadChannel will certainly work for "plain old datatypes" (e.g.
+/// `int`, `char`, `float`, etc), users are encouraged to use the new
+/// `std::atomic` data types available with C++11 and later. These "lock-free"
+/// data types require no special data protection for shared use between threads
+/// and are guaranteed to be "thread-safe".
```

sending primitive types through a channel can be a perfectly normal case. reading and even writing to a primitive number has always been "thread safe" in the sense of not crashing the application which is what i think most people would understand by thread safety, what can be avoided by atomics is wrong states of a variable when written by 2 threads but that case is not really what a channel aims to solve.

I was planning to write something about channels in the blog article but meanwhile if you want to take a look the whole idea is taken from Rust's channels: http://doc.rust-lang.org/std/sync/mpsc/index.html with the main difference that ours doesn't separate receiver and sender and that it could be used as multiple sender multiple receiver but doesn't really make sense and it's better seen as mpsc too. Clojure http://clojure.com/blog/2013/06/28/clojure-core-async-channels.html  and Go http://guzalexander.com/2013/12/06/golang-channels-tutorial.html  have their own versions too although the way those work is slightly different mainly because they can block on the sender while ours won't right now.

bakercp:
Sounds good @arturo I'll slim this down and remove implementation specific
details and other extras.

bakercp:
Hi @arturoc take a look -- I changed the wording a little bit and got rid of some of the implementation details.  After removing the extra stuff, I think the most important features you note are better emphasized.

libs/openFrameworks/utils/ofThreadChannel.h
```diff
@@ -1,28 +1,73 @@
...
+/// communication.
+///
+/// Data is sent through the ofThreadChannel in a FIFO (first in, first out)
+/// order, guaranteeing that that data will be received in the same order that
+/// it was sent.
+///
+/// If multiple threads attempt to send data using the same ofThreadChannel, the
+/// send method will block the calling thread until it is free. Sending and
+/// receiving data from the same thread may result in deadlocks.
+///
```
arturoc: i don't think this could happen, a thread could get blocked for ever waiting on data if there's no data available but i think a deadlock is impossible?

libs/openFrameworks/utils/ofThreadChannel.h
```diff
@@ -39,16 +84,44 @@ class ofThreadChannel{
...
-	/// and returns true or returns false if there
-	/// is no data available or the channel was closed
-	bool tryReceive(T & ret){
+	/// \brief If available, receive a new sent value without blocking.
+	///
+	/// ofThreadChannel::tryReceive is similar to ofThreadChannel::receive,
+	/// except that it will block the receiving thread as it waits for sent
+	/// data.  If no data is available, it will return false and continue
+	/// immediately.
+	///
```
arturoc: tryReceive won't block the receiving thread and instead returns immediately if there's no data.

arturoc:
if you can fix that couple of things i think it's ready to merge

libs/openFrameworks/utils/ofThreadChannel.h
bakercp: I assumed this would be possible because std::mutex is not recursive -- but I suppose if a thread is blocking on receive, with its mutex locked, then it won't be able to send, thus a deadlock wouldn't be possible.

libs/openFrameworks/utils/ofThreadChannel.h
bakercp: Yeah, that was a typo.  The next line says the opposite.  I think it was a cut-n-paste error but it's fixed now.

bakercp:
OK.  I think that's fixed!  Thanks for all of your feedback @arturoc and @kylemcdonald.  I suppose documenting code is kind of like teaching it ... you learn a lot in the process.

arturoc:
thanks!

PR #4232, opened 2015-08-12
bugfix for flickering window on creation and destruction on OS X. closes  #4229 

### LINKED ISSUE #4229: screen flicker on app launch
have only see this on osx,
doesn't seem to be happening on ios.

when launching an app, the screen is black for (one or more) frame and then changes to the default background color. when setting `ofBackground(ofColor::red);` in `setup` or `draw`, it shows the same behaviour where the app launches with a black screen and then changes to red.

ofTheo:


libs/openFrameworks/app/ofAppGLFWWindow.cpp
```diff
@@ -322,6 +321,11 @@ shared_ptr<ofBaseRenderer> & ofAppGLFWWindow::renderer(){
 //--------------------------------------------
 void ofAppGLFWWindow::update(){
 	events().notifyUpdate();
+	
+	//show the window right before the first draw call.
+	if( settings.visible && windowP && ofGetFrameNum() == 0 ){
```
arturoc: can we better use the local frame number? that way it'll work to for windows opened after OF is running and doesn't use global calls which are not thread safe

ofTheo: good catch!! totally makes sense. 
going to update that.

ofTheo:
@arturoc I switched it to use a bool as there wasn't an internal frame counter for the window and it seemed to be better to rely on a dedicated variable.  I think this should be good to merge now.  Thanks for the catch!

arturoc:
thinking of it, wouldn't it be enough to do it at the end of setup? there shouldn't be anything going on between the end of setup and the first update and it would remove the need for that extra flag and check

ofTheo:
With a basic app I see a bit less of a flicker putting it after update instead of end of setup. 

Also if your update loop is slow ( or if you are dynamically allocating stuff on first frame in update which is not uncommon ), you'll see the black window much more visibly then. 

So this way the time between the window being shown and the time you can see content rendered to it is as minimal as it can be.

arturoc:
ok, it just feels super hacky but if it fixes it... will check when we update glfw since it'll probably be fixed then

PR #4259, opened 2015-08-16
unit tests

arturoc:
some tests i've been using mainly for the changes in ofFileSystem + an OF addon for unit tests.

right now is pretty simple and could be integrated in the future with some already existing unit tests framework.

- The addon has a ofxUnitTestsApp that is an ofBaseApp so tests can implement it to be called with the proper OF initialization. Most tests only need to implement a `run()` method in that class
- It provides, test and test_eq methods by now that will fail if the result is not true or the passed parameters are not equal
- It also provides a colored log channel so the output from tests is easily checked

This is ready to be merged but mostly uploading it to test: #4285 which is working fine for me

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,6 @@ matrix:
   allow_failures:
     # temporary until we find out how to prevent (spurious) build failures
     - env: TARGET="docs"
-    - env: TARGET="linux"
 before_install:
     # should source before_install.sh so we can export CXX in the linux script.
     # instead, export CXX here, because sourcing messes up Travis log output
--- a/addons/ofxUnitTests/src/ofxUnitTests.h
+++ b/addons/ofxUnitTests/src/ofxUnitTests.h
@@ -93,6 +93,8 @@ protected:
 			numTestsPassed++;
 		}else{
 			ofLogError() << testName << " failed " << msg;
+			ofLogError() << "value1 " << t1 << msg;
+			ofLogError() << "value2 " << t2 << msg;
 			numTestsFailed++;
 		}
 	}
--- a/addons/ofxUnitTests/src/ofxUnitTests.h
+++ b/addons/ofxUnitTests/src/ofxUnitTests.h
@@ -7,6 +7,9 @@
 #pragma once
 
 #include "ofConstants.h"
+#include "ofLog.h"
+#include "ofBaseApp.h"
+#include "ofAppRunner.h"
 #include <string>
 
 
--- a/addons/ofxUnitTests/src/ofxUnitTests.h
+++ b/addons/ofxUnitTests/src/ofxUnitTests.h
@@ -1,0 +1,104 @@
+/*
+ * ofxUnitTestsApp.h
+ *
+ *  Created on: Jul 8, 2015
+ *      Author: arturo
+ */
+#pragma once
+
+#include "ofConstants.h"
+#include <string>
+
+
+class ofColorsLoggerChannel: public ofBaseLoggerChannel{
+	std::string CON_DEFAULT="\033[0m";
+	std::string CON_BOLD="\033[1m";
+	std::string CON_RED="\033[31m";
+	std::string CON_YELLOW="\033[33m";
+	std::string CON_GREEN="\033[32m";
+	std::string getColor(ofLogLevel level) const{
+		switch(level){
+		case OF_LOG_FATAL_ERROR:
+		case OF_LOG_ERROR:
+			return CON_RED;
+		case OF_LOG_WARNING:
+			return CON_YELLOW;
+		case OF_LOG_NOTICE:
+			return CON_GREEN;
+		default:
+			return CON_DEFAULT;
+		}
+	}
+public:
+	void log(ofLogLevel level, const std::string & module, const std::string & message){
+		// print to cerr for OF_LOG_ERROR and OF_LOG_FATAL_ERROR, everything else to cout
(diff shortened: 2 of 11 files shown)


scripts/ci/linux/build.sh
```diff
@@ -13,11 +13,18 @@ echo "PLATFORM_CFLAGS += $CUSTOMFLAGS" >> libs/openFrameworksCompiled/project/li
 cd libs/openFrameworksCompiled/project
 make Debug
 
-echo "Building emptyExample"
```
bilderbuchi: @arturoc why did you delete this part? I don't think a CI build is really sensible if we can't even build the emptyexample. Even if we don't want to build that temporarily, why delete this and not simply comment it out and adding some explanation?

.travis.yml
```diff
@@ -31,7 +31,6 @@ matrix:
   allow_failures:
     # temporary until we find out how to prevent (spurious) build failures
     - env: TARGET="docs"
-    - env: TARGET="linux"
```
bilderbuchi: @arturoc I very much would prefer if we left the linux build in allow_failures, so that we can still get a feeling how often the build fails due to memory concerns, and if whatever measures we implement have a meaningful impact.

bilderbuchi:
@arturoc how nice of you to just merge these changes to the CI setup without waiting for any review or input. You must have been aware that I'm currently trying to improve that setup.
I know that you are part of @openframeworks/core but sometimes it would be great to wait a little to invite comments/feedback. :disappointed: 
For example, the way I envision the setup is to separate builds and test runs into separate scripts, so it will be more easily recognizable if travis fails because of the former or the latter. Also, if you plan to just delete the emptyexample build from the CI process, it would be great to say so - I wouldn't have bothered spending hours/days on tinkering with gcc flags and travis tweaks to get this to run. Sometimes I wonder why I even bother...

scripts/ci/osx/build.sh
```diff
@@ -3,3 +3,19 @@ set -ev
...
+echo "Unit tests"
+cd $ROOT/tests
+for group in *; do
+	if [ -d $group ]; then
+		for test in $group/*; do
+			if [ -d $test ]; then
+				cd $test
+				cp ../../../scripts/osx/template/Makefile .
+				cp ../../../scripts/osx/template/config.make . 
+				make Debug
```
bilderbuchi: @arturoc wouldn't it be better to also use xcodebuild for this (or make for the main compilation) to have homogenity in the build system per platform?

addons/ofxUnitTests/src/ofxUnitTests.h
```diff
@@ -0,0 +1,122 @@
+/*
+ * ofxUnitTestsApp.h
```
bilderbuchi: this header doesn't match the filename?

addons/ofxUnitTests/src/ofxUnitTests.h
```diff
@@ -0,0 +1,122 @@
...
+#pragma once
+
+#include "ofConstants.h"
+#include "ofLog.h"
+#include "ofBaseApp.h"
+#include "ofAppRunner.h"
+#include <string>
+
+class ofColorsLoggerChannel: public ofBaseLoggerChannel{
+	std::string CON_DEFAULT="\033[0m";
```
bilderbuchi: do these flags work on non-unixes? if not, might be a good idea to formulate this in a cross-platform way.

scripts/ci/osx/build.sh
bilderbuchi: btw, this compiles OF library _twice_, once with xcode, and once with make, see https://travis-ci.org/openframeworks/openFrameworks/jobs/75829657#L984

.travis.yml
bilderbuchi: also, this stil has some builds failing because of memory issues: https://travis-ci.org/openframeworks/openFrameworks/jobs/75839351#L1494

scripts/ci/linux/build.sh
arturoc: building empty example makes the build crash and my suspicion is that this is happening because empty example is including ofMain.h which is huge. i've changed it to build the tests which are much lighter because they only include the headers that are needed and it seems to work better.

the tests are also normal applications so if those are building emptyExample should build too. 

empty example is also building in osx so we will catch most possible errors affecting emptyExample but not the tests there

scripts/ci/osx/build.sh
arturoc: i know, i'll change it when i have time but if i use xcodebuild i need to compile the PG, generate the project for the tests and then compile them so it's not trivial. i'll probably move the PG generation here if i finally do it like that

arturoc:
the tests are using the same flags you've been adding, i've only removed
the empty example and added a test, i don't think is such a big deal but
feel free to change the travis file as you wish.

i'm doing a ton crap of work to get the release out and it helps a lot
to be able to tests things on each platform and generate the PG without
having to switch between 5 platforms manually in my computer.

scripts/ci/linux/build.sh
bilderbuchi: > building empty example makes the build crash

unit tests [also make the build crash](https://travis-ci.org/openframeworks/openFrameworks/jobs/75839351#L1501), so you just replaced one build-crashing application with another...

kylemcdonald:
@tpltnt you expressed interest in working on unit tests in OF, it looks like @arturoc has decided this is the direction they will be going so please feel free to jump in once this is merged :)

PR #4260, opened 2015-08-16
Reworked Android life cycle

tallavi:
# Summary

This PR reworks the android life cycle handling. Unlike current version, it allows the destruction and recreation of the activity in the same process without causing crashes or errors. It also allows the initialization of the native side without having the first activity be an OFActivity. My project needed this and I think it can be useful in more projects. It also addresses all sorts of errors that pops up in Logcat (see below). The ofApp class is not affected at all by the change - fully backwards compatible. The only affected files in projects are main.cpp and OFAndroid.java, that in most projects just have boot strapping code. I provided two examples for demonstration:

## Changes to OF core:

1) OFAndroid became too large. Moved some code away from OFAndroid to OFAndroidLifeCycleHelper.

2) Since initialization/destruction process  happens as a result of Android’s callbacks, but should happen asynchronously (on GL’s thread), we needed a mechanism to synchronize the steps, wait for one step to finish before starting the next. This is done in OFAndroidLifeCycle.java.

3) While refactoring OFAndroid, I also removed the parts that were related to GPS to a different class - OFAndroidGPS, so this addon is included in it’s entirety in this PR.

4) bug fix in ofBitmapFont - the event was not correctly released (causing “something is wrong here” errors and crashes)

5) bug fix in ofTexture,ofVbo - don’t release textures/vbos after if surface is already destroyed. They are released automatically. Releasing without a context causes the black rectangles issues + “no GL context” errors.

### AndroidEmptyExample 

Was modified to include the updated boot strapping code:

**Changes in OFAndroid**: Moved code to base OFAndroid class, instead of having all events in the project and routed manually into OFAndroid.

**Changes in main.cpp**: Old version had the method Java_cc_openframeworks_OFAndroid_init( JNIEnv*  env, jobject  thiz ) in main to trigger the init process. If someone used an example without this method it would have failed in runtime instead of compile time. We moved this method into the core ofxAndroid and not in main, and we exposed two easy methods that needs to be implemented in main.cpp: OFAndroidInit() and OFAndroidCreate(). That way, if the user forgets these methods, he’ll get linking errors instead of run time. Further more, now there are two methods in order to support app scope initialization vs. activity scope initialization. Most simple projects will just use OFAndroidCreate to initialize the ofApp when the activity is initialized (like done in this example).

### AndroidNativeActivityExample 

I created another example to demonstrate a new use case: **having the first activity be a non-OFActivity**, that opens a OFActivity with a button. The OFActivity may be closed (with the back button) and reopened without errors. The native side is available to JNI calls even before the OFActivity is created, by initializing it manually using two lines:

OFAndroidObject.setActivity(this);
OFAndroidLifeCycle.init(this, this);

Simple projects doesn’t need to do anything, since base OFActivity handles initialization.


tallavi:
@openframeworks/android

julapy:
this is great.
ofxiOS went through the same change a couple years back,
and it really opens up the way the OF can be used inside mobile apps,
where you don't always want a single opengl screen experience,
but have the option of embedding OF within a native UX.

tallavi:
Exactly. That is how my OF-powered iOS/Android app is structured. It has lots of UI and I wanted to retain that native feel to it.

iOS always worked for me off the bat, but I always had strange Android errors and crashes that took some while to track down. This endevour is pretty old, I've been working on it on and off for 6 months. And today I decided to cleanup, and prepare some condensed commits to have it (possibly) merged already. I regret missing the 0.9 line, I was really hoping to include it there.

The next logical step would be to somehow allow MULTIPLE ofApps to be stacked one on the other, and the right one be active depending on which OFActivity is currently in the foreground. But it may take a while, so I decided to suggest this stable state to the core, and have some other people enjoy from it (and for me to enjoy my contribution, of course).

tallavi:
Made another commit to resolve the latest conflict. BTW, forgot to ask before, I fixed androidEmptyExample, but there's also the rest of the examples. Do I need to update them all or do the project generator somehow takes care of it..? (I'm not sure how it works).

HalfdanJ:
I really like this!

tallavi:
Thanks @HalfdanJ!

Eagerly awaiting the merge. I added some more commits a while back to fix all the android examples. About 50% are working, the rest either have compilation errors that are not related to this change or have issues when minimizing the app, that I believe that been there before..

arturoc:
@tallavi now that 0.9.1 is released i'd like to merge this. it won't merge anymore i guess it's related with the event listeners to restore the textures when the gl context is recreated. i added copy and move constructors during the 0.9.1 release cycle to fix some problems so it's probably just a matter of merging those.

can you take a look? also i've quickly checked through your code and once thing that i find strange is that OFAndroidLifeCycle keeps a stack of states, why do we need that? can't we just have the last state?

anyway thanks so much for looking into this and sorry it took so long

tallavi:
Hi! No problem at all, I'll have a look and update here once it's done :-) 

Regarding the stack of states, we use it because of the asynchronic nature of the initialization process. Sometimes we have an edge case that we already get a pause but the resume has not started yet, so we just remove the two from the stack.

And one last thing - the examples. We converted the boiler plate code on all examples and wanted to test them, but only about half of the examples work, and for reasons that I'm very sure are unrelated to the changes. I made a list if you want to go over them.

tallavi:
Ok, @arturoc. We should be good to go.

arturoc:
thanks for looking into this, some doubts and suggestions:

- About the stack changes, i've never found any issue related to that, this are just callbacks so they'll get called in the same order they happen no? i mean the same callbacks already form a kind of stack so i'm not sure why we need a explicit stack in that class. it makes that class way too complex and doesn't seem really necessary, unless i'm missing something. is it something that you've found in a real use case or just a precaution? 

- In that same class there's a worker thread that seems to deal with the lifecycle? do we need it? seems like we might want to do GL operations that will crash if done from a thread other than the GL one

- OFAndroidInit() and OFAndroidCreate() have a weird capitalization we don't use capital letters for the first letter of a function, it should probably be ofAndroidCreate and ofAndroidInit. Also the names are kind of arbitrary, there's even some comments below the functions to clarify what they do. wouldn't it be better to choose names that explain better what the functions do? ofActivityStart or ofActivityInit and ofApplicationStart/Init?  or even onAndroidActivityStart / onAndroidApplicationStart (which will differentiate them from real OF functions since they belong to the app not to OF itself) perhaps there's even some android standard naming for those events that we could adapt, not sure

- The andorid native activity example is confusingly named too, NativeActivity in android refers to an activity that is pure c++ with no java but this example is actually the opposite right? i think it would be better to name it javaOnlyActivityExample or something like that.

- There's an iOS implementation for the gps addon which might be really nice to have but would better go in a different PR so the iOS people can review it easily

Other than that this is looking really great, i think once this things are solved/clarified it's ready to be merged

tallavi:
ok, here goes:

- Yes, it is needed as we want to perform the cpp initialization in a non-UI thread (decompress the zip for example. That happens on a worker thread even before this PR), and we have no control on when the activity changes status. Sometimes the OS signals us to start the "resume", but we are still in the middle of the create, and we can't call "resume" in the cpp until the "create" part is done (otherwise it crashes due to race conditions). So we need to remember which OS events were already received and handle them synchronously, but in the background. 

- We can't do this work on the GL thread, since there's no GL thread yet. The first activity doesn't have to be a GL activity, that's exactly what the soon to be renamed example nativeActivityExample comes to show. The first loaded activity is not OFActivity, but still you have the cpp loaded and available for use with JNI calls. The setup(), update() and draw() are called on a GL thread, as before, and you shouldn't be drawing anywhere else..

- You right. Let’s go with **ofAndroidApplicationInit** and **ofAndroidActivityInit**. I would still leave the comments there…

- Yeah, it’s confusing. For a lack of a better name, we’ll rename it to **androidJavaOnlyActivityExample**

- Ok, I’ll push a commit that deletes the iOS related files, effectively making ofxGPS addon be Android only for this PR. I’ll create an iOS PR **after** this one is merged. Hopefully deleting the files will fix the PR so that these files are not shown?

arturoc:
ok that sounds great if you can change the names of those 2 functions and the java only example i think it's ready

tallavi:
Ok, done and working, but I had an issue verifying with the examples now. I verified with the javaOnlyExample, which compiles and works, but every other example has a weird build error:

`/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-g++" not found in PATH
`
Any ideas? I think it should be  “darwin-x86_64” instead of  “linux-x86_64" and then it would work, right?

I checked in “config.android.default.mk” and found this:

```
ifeq ($(shell uname),Darwin)
ifneq ($(wildcard $(NDK_ROOT)/toolchains/$(TOOLCHAIN)/prebuilt/darwin-x86_64),)
	HOST_PLATFORM = darwin-x86_64
```

Like it's not detecting that it's a Darwin project or something? I checked on two build machines, to make sure it's not an environment thing. Any ideas?

arturoc:
i changed the compiler to clang recently since the std that gcc uses was missing some symbols required for a library we added some weeks ago. not sure why is still trying to use g++ for you, have you pulled the very latest master? i might have missed something for osx though, i only tested it in linux

libs/openFrameworks/gl/ofTexture.cpp
```diff
@@ -137,12 +141,21 @@ static void release(GLuint id){
 		if(getTexturesIndex().find(id)!=getTexturesIndex().end()){
 			getTexturesIndex()[id]--;
 			if(getTexturesIndex()[id]==0){
-				glDeleteTextures(1, (GLuint *)&id);
+
+#ifdef TARGET_ANDROID
+				if (!ofAppAndroidWindow::isSurfaceDestoryed())
```
arturoc: small typo here: should be isSurfaceDestroyed()

tallavi:
Yeah, I have a complete blindspot for this specific typo... I actually needed to read it a few times in order to understand what you mean :) Anyway, I committed the change.

Regarding the compiler change, it sounds something like that (and both my machines are indeed OSX), but I have no idea why our new example runs and the others not. Or where to start with a fix.

arturoc:
The empty example is compiling fine on the test server so i guess it's ok, going to merge it so people start to work with this and can spot any potential problem during the 0.10 release cycle.

thanks so much for doing all this work! it's going to make android dev so much more flexible

tallavi:
(champagne bottle popping)

First of all, thank **you** for all the patience and advice! 

Second of all - my pleasure! Our vectors have aligned in such a way that allowed us to take some time to achieve this, and help both of our causes on the way..

And lastly - it was a joint effort! My Android guru, ArtemP, does allot of the heavy lifting on this. He doesn't write much, but boy does he code.

See you on the next PRs/issues :)

arturoc:
no not yet, most people use the download from the web i guess and this is not published yet, so probably not many people have used it. i'll try to port a project i did with 0.8.4 and i recently updated to 0.9 will let you know how it goes

PR #4261, opened 2015-08-16
Of arduino updates

DomAmato:
Updates to the newer protocol and includes support for stepper motors, encoders, i2c, and onewire devices. Handles hardware abstraction and pin handling better than with predefined values which adds support for boards other than just the uno variants


DomAmato:
@arturoc @bilderbuchi  @bakercp  @ofTheo  sorry to ping you guys but any chance this gets into the 0.9.0 release? The old ofArduino classes are actually setup to only work with uno style boards and doesn't utilize the hardware abstraction of the "newer" protocol. Plus this handles the pin capabilities and checks against them so that people can only query and set pins based on their capabilities

bilderbuchi:
looking at the diff alone, it's a huge amount of changes. I lean towards postponing this after 0.9.0 release.

arturoc:
yes, i think it's too late to merge something like this. i've just milestoned for the next release so we can merge it right away after 0.9 is out

ofTheo:
@DomAmato - apologies for not looking at this earlier! 
I agree we should do this in the next release, only as we are in the no new features part of the release process. 

I do see a lot of naming things that might need adjusting - there are a lot of structs and enums that might not need to be in the global namespace.  

I'll put some comments in the diff 
cheers!!

ofTheo:
Could the structs used be named in a way that makes them more tied to ofArduino - or could they be inside of the ofArduino class? ( ping @openframeworks/core on this as we don't have this in the wiki ) .

Also curious if you have done heavy testing with the events - as they can always be a little tricky to get right.  

Thanks!!
Theo

DomAmato:
Yeah the structs aren't very uniquely named and changing them is easy enough, though I would like to have some kind of agreed upon paradigm before changing them twice.

I tested the stepper, encoder, and i2c events and did experience a few issues which is why it checks the byte length of the sysex message and then flushes the buffer when things get out of sync. Really it wasn't the events fault but the buffer. 

I dont own any onewire devices to check but I just wrote it based on the protocol and other clients implementations that have tested responses.

@ofTheo its fine that this doesn't make 0.9.0 as the arduino communication doesn't seem to come up super often in the forums. I feel like i use it more often than anyone haha

arturoc:
I agree that there's probably too many constants in the global space + the structs naming.

It's really great to see some work happening in this area after several years of having been almost unmaintained.

Perhaps we should do some refactoring of how this class works, it's in general too low level and the event based system is perhaps not the best idea. We could look into some kind of polling based system?

Also we could perhaps make it higher level or we can leave this class as a low level interface with firmata but add more objects that are higher level interfaces with the hardware?

DomAmato:
Making a handler for the events can be sorta finicky though all they need to do is copy and paste an example function which is what I think people probably do pretty often (open the example and copy the  code from the example or just copy the example)

What kind of polling system are you imagining? technically direct pin polling does exist in firmata but thats really not the best way to handle checking pin states, I haven't included the handler in this PR either.

I would leave this class be and any higher level objects should be built on top of its functionality since this class really only interacts with firmata and splitting of features will only make things more confusing. What kind of objects were you thinking to split off as higher level?

DomAmato:
There is currently a bug in the code where certain pins are not reported correctly so while I don't see this being pulled in right away don't pull this until the bug fix is applied =)

DomAmato:
Goona make a little checklist for tested and passing modules

- [x] Digital Output
- [x] Digital Input & Input Pullup
- [x] Analog Output
- [x] PWM Output - note pwm on megas for 44, 45, and 46 are not reported by firmata as having pwm capabilities
- [ ] Servo
- [ ] I2C
- [ ] Onewire
- [x] Stepper - only AdvancedFirmata and RobustFirmata support Steppers
- [x] Encoders - Only ConfigurableFirmata and RobustFirmata support Encoders
- [ ] Serial

tpltnt:
What is the current state of things @DomAmato ?

DomAmato:
I mean its stable and the features that are checked above should be fine. i2c should be easy enough to test for i just haven't had a need yet so its been pushed back on my list. I don't actually own any onewire devices or servos but the latter can be acquired pretty easily and tested, though they should work since the communication protocol is pretty simplistic.

Realistically the communication from of -> firmata should be fine since its taken straight from the protocol, its the returned data that really needs to be tested more thoroughly.

DomAmato:
since this PR is so old should I just submit a new one or whats the recommended solution here. Sorry school and research got really busy so I sort of had no time but now its summer and I want to do more OF work

arturoc:
rebasing it with master as i think you just did is totally fine. it's failing to compile right now in most platforms though

DomAmato:
I saw those, I will have to update it which isn't a problem. I did that last night before going to bed so I didn't actually check it very closely haha

DomAmato:
looking at the problem is there a reason ofSerial needs two writeByte functions? the error is due to it being ambiguous which writeByte is being called but the ofSerial thats a regular char just gets cast to an unsigned char and passed to the other function:

```
bool ofSerial::writeByte(char singleByte){
	return writeByte(static_cast<unsigned char>(singleByte));
}
```

arturoc:
yes it's useful to send char or unsigned char, which are not the same, without casting. 

also it's not an intended behaviour but an int is 4 bytes so if writeByte(65459) worked it would compile but not work since such a high number would be auto casted to char and overflow.

it's safer if you explicitly cast to char or unsigned char. or even better declare the variable as char or uint_8 to avoid values higher than what a char can hold in the future.

DomAmato:
Ok I fixed the errors and warnings, its still a very large PR but at least its building on all systems now

arturoc:
thanks!

PR #4262, opened 2015-08-16
Of random updates

DomAmato:
moves functions over to c++11 uniform real distribution. also includes random integer functions and random gaussian noise functions


arturoc:
i wonder if the gaussian distribution call makes sense at all as a global call. if there's 2 functions or objects calling it with different frequency the distribution they'll get won't be gaussian, no?

DomAmato:
I mean the distribution should still be the same over the course of the sampling period unless you are taking very few samples. taking a base of a million samples and dividing it by 2 should give you 2 similar distributions of 500,000 data points each so i don't know that having functions called at a different frequency would change the distribution dramatically.

kylemcdonald:
if one function / object calls twice as much, they both still have a gaussian distribution:

```c++
#include "ofMain.h"

#include <random>

static std::default_random_engine generator;
static std::normal_distribution<float> nDistribution(0, 1);

ofMesh plot(vector<float>& v, float mean = 0, float width = 3) {
    int binCount = 100;
    vector<int> bins(binCount, 0);
    for(auto& x : v) {
        int b = ((x - mean) / width) * (binCount / 2) + (binCount / 2);
        if(b > 0 && b < binCount) {
            bins[b]++;
        }
    }
    ofMesh mesh;
    mesh.setMode(OF_PRIMITIVE_LINE_STRIP);
    for(int i = 0; i < bins.size(); i++) {
        mesh.addVertex(ofVec2f(i, bins[i]));
    }
    return mesh;
}

public:
    ofMesh mesha, meshb;
    void setup() {
        vector<float> a, b;
        for(int i = 0; i < 10e6; i++) {
            a.push_back(nDistribution(generator));
            a.push_back(nDistribution(generator));
            b.push_back(nDistribution(generator));
        }
        mesha = plot(a);
        meshb = plot(b);
        ofBackground(0);
    }
    void update() {
    }
    void draw() {
        ofScale(12, 12 * ofGetHeight() / 10e6);
        mesha.draw();
        meshb.draw();
    }
};
int main() {
    ofRunApp(new ofApp());
}
```
screen shot 2015-08-17 at 14 31 22

if you sample less, the shape is still the same, just noisier:

screen shot 2015-08-17 at 14 33 10

DomAmato:
Thanks Kyle for the visual proof it was a better example than the one i had cooked up haha

kylemcdonald:
going to merge this as there were no other questions/complaints.

arturoc:
mmh i still don't think is very correct, we've been trying to avoid using static variables everywhere in the core and this introduces several more. the fact that  this is an object in the std instead of a function call instead of a function call i think is enough indication.

in any case i think this needs more though and shouldn't be merged right before 0.9. i'm going to revert it and we can see after 0.9 how to proceed.

i think perhaps an addon that exposes the different distributions in the std with an easier syntax makes more sense than having a global function with several statics in the core

DomAmato:
I get your reasoning but I don't see why we should continue to use the legacy random functions with all those alterations to make it actually random. I'm open to other implementations but leaving the current random functions doesn't make a ton of sense either

kylemcdonald:
i'm ok with using our very carefully crafted wrappers around `rand()`, i'm more confused about avoiding static variables. `rand()` itself is a function, but it is also basically a static variable: it has internal state, and is not thread safe, and it hasn't caused any problems so far. so it seems like replacing it with the c++11 equivalent would be a welcome upgrade, and it would have all the same qualities. if someone needs thread-safe random numbers they should also be advanced enough to use c++11 directly, i think, i don't see that as a place for us to try innovating :)

arturoc:
>  I don't see why we should continue to use the legacy random functions with all those alterations to make it actually random.

the only changes are not really to make them random but to fix the ranges to be in [a,b).

Introducing changes like this after we've already realeased an RC doesn't seem very sensible. Among other things:

- it adds new global functions to the current API not only tries to fix it. 
- The current API is working well and this is not fixing any bug.
- It uses statics directly in the cpp which can crash in certain cases
- This `if (ufDistribution.max() != max || uiDistribution.min() != 0) {` comparing floats like that is probably going to break in some cases.


And in general i'd prefer a change like this to be introduced after the release with proper tests...

ofTheo:
@DomAmato its great to get some attention on this issue, but as @arturoc said we really try to avoid doing big changes to widely used functions so close to a release.  We've been burned a few times already and I think its a good policy to be a little conservative when a PR has the potential to introduce new bugs or changing behavior. 

I would say though that this is something we could do fairly quickly after 0.9.0 releases, but we should do some fairly heavy testing first.  Maybe add a test to the new tests/ folder?

DomAmato:
I'll make up a test though obviously suggestions for test cases is welcome. I wasn't insisting this had to make the release I was worried it would fall to the side. We put in the effort of upgrading to c++11 compatibility we may as well utilize the updates rather than relying on c89/99 functions.

DomAmato:
Reviving this now that 0.9.0 has been released and this was not actually merged

PR #4266, opened 2015-08-17
trying to fix linux travis crashes

arturoc:
just checking if this fixes the problems with linux builds in travis

- using gcc5 instead of 4.9 which supposedly uses less memory
- using -g0 to compile
- stopping services (from @bilderbuchi's setup)

### DIFF
--- a/libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
+++ b/libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
@@ -196,7 +196,7 @@ endif
 
 ifndef PROJECT_OPTIMIZATION_CFLAGS_DEBUG
 	# DEBUG Debugging options (http://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html)
-	PLATFORM_OPTIMIZATION_CFLAGS_DEBUG = -g3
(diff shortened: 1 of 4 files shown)


libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
```diff
@@ -196,7 +196,7 @@ endif
 
 ifndef PROJECT_OPTIMIZATION_CFLAGS_DEBUG
 	# DEBUG Debugging options (http://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html)
-	PLATFORM_OPTIMIZATION_CFLAGS_DEBUG = -g3
+	PLATFORM_OPTIMIZATION_CFLAGS_DEBUG = -g0
```
bilderbuchi: hm, I guess this change slipped in during commit, as that is exactly what you do with the `sed` below?

bilderbuchi:
looks great, I'll integrate this into my PR!

libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
arturoc: yes that shouldn't be there, i'm going to close this anyway.

PR #4269, opened 2015-08-17
fix for spurious warning

ofZach:
this fixes an annoying warning on OSX where this call: 

    std::filesystem::create_directories(dirPath);

return false when *any* of the folders in the path already exist.  this means that although the folder is created, we get a warning that it wasn't created. 

the solution is to search for the directories existence before attempting to create it and using that to drive the warning. 

I am cleaning up the text based output from the project generator and this will go along way to making it much cleaner.


ofZach:
yay!  the output from the PG is getting cleaner and cleaner...

Issue #4242, opened 2015-08-13
0.9.0 RC1 Failing to compile on RPI1


jvcleave:
Haven't dug in much yet but here is what I am getting out of the box

```
Assembler messages:
Error: unknown architecture `native'

Error: unrecognized option -march=native
Compiling /home/pi/openFrameworks/libs/openFrameworks/graphics/ofCairoRenderer.cpp
Assembler messages:
Error: unknown architecture `native'

Error: unrecognized option -march=native
cc1plus: error: bad value (native) for -march switch
cc1plus: error: bad value (native) for -mtune switch
cc1plus: error: unrecognized command line option ‘-std=c++11’
cc1plus: error: bad value (native) for -march switch
cc1plus: error: bad value (native) for -mtune switch
cc1plus: error: unrecognized command line option ‘-std=c++11’
```

jvcleave:
looks like g++ 4.8 is required but the install_dependencies script will fail to find it as it is searching for 

`g++4.8`

this works for manual install (regex needs dash)

`sudo apt-get install g++-4.8`

arturoc:
can you fix the install_dependencies script and send a PR?

jvcleave:
I just updated it - I thought it would generate a PR but it just applied it directly - let me know if that causes any issues.

I'm still trying to get it to compile

jvcleave:
ok, a few things

I tried to compile it natively on the RPI overnight but it was still going today so I canceled it. I'm not sure if this would be the same situation on the latest RPI Image

I did get it to compile via my cross-compiler setup using [Virtual Box but had to upgrade that to Jessie](https://gist.github.com/jvcleave/8489b3ec5897eb1a26a2)

The [lines with -mtune=native](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk#L189) were problematic with everything I tried. I was able to finally cross-compile with

```
ifeq ($(CROSS_COMPILING),1)
    PLATFORM_OPTIMIZATION_CFLAGS_RELEASE = -O3
    PLATFORM_OPTIMIZATION_CFLAGS_DEBUG = -g3
else
    PLATFORM_OPTIMIZATION_CFLAGS_RELEASE = -O3 -march=native -mtune=native
    PLATFORM_OPTIMIZATION_CFLAGS_DEBUG = -g3 -march=native -mtune=native
endif
```

[...]

PR #4271, opened 2015-08-18
Remove unnecessary armv6l install scripts.

bakercp:
Related to https://github.com/openframeworks/openFrameworks/issues/4242


arturoc:
won't we need to modify the normal ones to install gstreamer-omx?

bakercp:
Hmm. Will double check.

arturoc:
also set g++ 4.8 as default if we are on wheezy:

```
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.6 20
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 50
sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.6 20
sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 50
```

if you can add that in ubuntu that'd be great but i can do it later too

PR #4272, opened 2015-08-18
Fixes #4224, makes rpi2 default armv7l variant.

### LINKED ISSUE #4224: Make default linuxarmv7l makefile configuration work for Raspberry Pi / Raspberry Pi 2
Currently in order to deal with all of the strange configuration issues associated with arm linux distros, we have a variable called `PLATFORM_VARIANT` in our makefiles that allows us to choose which configuration file to use.  For example, on linuxarmv7l, we have three configurations:

https://github.com/openframeworks/openFrameworks/tree/master/libs/openFrameworksCompiled/project/linuxarmv7l

I can't recall what target hardware / distro the default.  Thus in order to compile for Raspberry Pi 2 (which is linuxarmv7l) we need to call the makefile with:

`make PLATFORM_VARIANT=rpi2` etc.

To compile for udoo, we use

`make PLATFORM_VARIANT=udoo` etc.

For reference, the only known linuxarmv6l platform we've supported is raspberry pi 1, so we don't need to specify the PLATFORM_VARIANT, as rpi is our default.

https://github.com/openframeworks/openFrameworks/tree/master/libs/openFrameworksCompiled/project/linuxarmv6l

Anyway, the goal with this suggestion is to make it super easy to use the raspberry pi 2 because it is so popular.

P.S. In an ideal world we'd be able to detect exactly what hardware / distro is being used and pick the right distro ... but that's a very tricky business.

P.P.S I'd also like to update the documentation to clarify how this PLATFORM_VARIANT thing works for people working with platforms that we haven't tested.

bakercp:
Makes rpi2 default platform variant for armv7l based on discussion here #4224 .


bakercp:
@ofTheo / @arturoc thoughts on this one?

arturoc:
i wonder if we should just check if /opt/vchost.h exists and decide that we are on raspberry (that's not the exact path but i guess you get the idea)

bakercp:
@arturoc That would be fine, but based on the forums and other feedback, the people are primarily using raspberry pi, so it seems that we should be able to just treat RPI devices as default for arm6/7 (for consistency) and use the platform variant system for udoo and others.

jvcleave:
+1 on making RPI2 default for arm7

We should edit the GCC version to be 4.9 as well for cross-compiling

Should be able to use the same block as in the arm6 variant
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/linuxarmv6l/config.linuxarmv6l.default.mk#L192

bakercp:
So, can we merge this for now?  Then make the GCC changes to armv6/7?

arturoc:
ok, i don't like it very much to assume that an specific brand is the default but if most people is using it let's do it...

PR #4273, opened 2015-08-18
Fixes #4217

### LINKED ISSUE #4217: ofMap divide by zero warning causes RPI to slow to an unusable speed under some conditions
In some cases, such as with the 3DPrimitivesExample, if the ofMap params lead to a divide by zero ofLog warning, which, in the case of RPI2 runs in the background and slows the app down so much that you can't interact with the app anymore.

I'd recommend that we document the fact that we explicitly avoid divide by zero errors and remove the logged warning, or make it verbose.

bakercp:
Removes warning for ofMap and adds documentation about default safety behavior.


tgfrerer:
+1

bakercp:
@ofTheo merge?

ofTheo:
awesome!

PR #4276, opened 2015-08-19
Fix ofToBinary

bakercp:
I introduced a regression into ofToBinary when attempting to get rid of warnings.  The current version in the repo is wrong and doesn't make sense.  This is a fix that matches the old functionality.

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.h
+++ b/libs/openFrameworks/utils/ofUtils.h
@@ -3,7 +3,6 @@
 #include "ofConstants.h"
 #include "utf8.h"
 #include <bitset> // For ofToBinary.
-#include <limits.h>
 
 #include "ofLog.h"
 
@@ -735,7 +734,7 @@ char ofToChar(const string& charString);
 /// \returns a binary string.
 template <class T>
 string ofToBinary(const T& value) {
-	return std::bitset<CHAR_BIT * sizeof(T)>(*reinterpret_cast<const uint64_t*>(&value)).to_string();
+	return std::bitset<8 * sizeof(T)>(*reinterpret_cast<const uint64_t*>(&value)).to_string();
 }
 
 /// \brief Converts a string value to a string of only 1s and 0s.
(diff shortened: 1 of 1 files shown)


bakercp:
Alternatively, we could go back to the old way, but we would get the warnings back.

bilderbuchi:
fails on Travis linux with
```
In file included from /home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/3d/ofMesh.h:10:0,

                 from /home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/types/ofBaseTypes.h:7,

                 from /home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/gl/ofTexture.h:5,

                 from /home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/video/ofVideoPlayer.h:4,

                 from /home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/video/ofVideoPlayer.cpp:1:

/home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/utils/ofUtils.h: In function ‘std::string ofToBinary(const T&)’:

/home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/utils/ofUtils.h:737:21: error: ‘CHAR_BIT’ was not declared in this scope

  return std::bitset<CHAR_BIT * sizeof(T)>(*reinterpret_cast<const uint64_t*>(&value)).to_string();

                     ^

/home/travis/build/openframeworks/openFrameworks/libs/openFrameworks/utils/ofUtils.h:737:41: error: template argument 1 is invalid

  return std::bitset<CHAR_BIT * sizeof(T)>(*reinterpret_cast<const uint64_t*>(&value)).to_string();
```

bakercp:
OK, that seems to compile on Linux now w/ the limits.h included.  We'll see what travis says.

bakercp:
Also, I'm seeing that we usually just write a literal `8` rather than `CHAR_BIT` in the various other functions ... so perhaps we should just do that?

arturoc:
sizeof(char) ?

arturoc:
oh yeah it's in bits, yes probably just using 8 makes sense

bakercp:
I'll change it to 8 bits to match the rest of them.

arturoc:
ok

bakercp:
@arturoc I've been using this for the last week and it's working well.  The existing version is broken.

PR #4277, opened 2015-08-19
Fix rtaudio gnome-shell

underdoeg:
On my arch system running gnome shell, I get some warnings and the audioReceived callback does not get triggered when using the default rtaudio API. This PR will check for GNOME and tries to use pulseaudio which integrates the app better into the os. 
This also fixes some inconsistencies with the tab spacing.

### DIFF
--- a/libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
+++ b/libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
@@ -63,34 +63,34 @@ vector<ofSoundDevice> ofRtAudioSoundStream::getDeviceList() const{
 }
 
 //------------------------------------------------------------------------------
-void ofRtAudioSoundStream::setDeviceID(int _deviceID) {
+void ofRtAudioSoundStream::setDeviceID(int _deviceID){
 	inDeviceID = outDeviceID = _deviceID;
 }
 
-int ofRtAudioSoundStream::getDeviceID()  const {
+int ofRtAudioSoundStream::getDeviceID()  const{
 	return inDeviceID;
 }
 
-void ofRtAudioSoundStream::setInDeviceID(int _deviceID) {
+void ofRtAudioSoundStream::setInDeviceID(int _deviceID){
 	inDeviceID = _deviceID;
 }
 
-void ofRtAudioSoundStream::setOutDeviceID(int _deviceID) {
+void ofRtAudioSoundStream::setOutDeviceID(int _deviceID){
 	outDeviceID = _deviceID;
 }
 
 //------------------------------------------------------------------------------
-void ofRtAudioSoundStream::setInput(ofBaseSoundInput * soundInput) {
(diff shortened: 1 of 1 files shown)


underdoeg:
Looks like travis has some older version of rtaudio installed. I'll try to implement a version check.

arturoc:
i've tried to use pulse audio before but the latency is terrible for some things it might be ok, but if you are doing anything interactive or with some minimum real time requirements it's too much

underdoeg:
yes pulseaudio can be slow. Although it seems to work fine for me. I did configure my system for low latency though...

Still there is the issue of ofApp::audioReceived not being called when I don't use pulseaudio. I'll try to figure out what the problem is...

underdoeg:
Just found this: http://forum.openframeworks.cc/t/of-audio-in-ubuntu-14-04-not-work/19682

But event after rebooting and only running the of app, I can't make it work. What about using a config switch for advanced users to optionally enable pulseaudio? Maybe something like ofRtAudioSoundStream::useLinuxLowLatency(true / false) ?

arturoc:
can it be that it's going through the hdmi out? it's happened to me some times recently when i had an external monitor connected through hdmi or (mini)display port even if it didn't had speakers.

and yes having an option to specify the api on setup makes more sense than trying to fix it for a very specific case, we should have some kind of config object like we do with fbos, windows...  so the setup function doesn't have a million parameters :)

underdoeg:
Ok, I can try to layout a base for some sort of config file. Since rtaudio does not get set up manually should I add some sort of ```static ofRtAudioSoundStream::setDefaults(rtAudioSettings)``` function?

At the moment it would only have one setting, the api, though?

arturoc:
i'd create a settings object that has all the parameters we have until now + any other that we can add in the future:

```
Settings{
    deviceId
    bufferSize
    channelsOut
    channelsIn
    numBuffers
    api
}
```

and the add a new setup function that takes a const reference to such object

underdoeg:
Makes sense yes, but where do we set the api then? Since OF calls ofRtAudioSoundStream::setup automaticall? Have a static  ofRtAudioSoundStream::defaultSettings? Doesn't make sense though since bufferSize etc is set externally?

arturoc:
not sure what you mean, we usually do:

```
soundStream.setup(2,0,256,44100,4);
```

with this we would do something like:

```
//ofApp::setup
ofRtAudioSoundStream::Settings settings;
settings.bufferSize = 256;
settings.channelsOut = 2;
settings.api = ofRtAudioSoundStream::API_PULSE_AUDIO;
soundStream.setup(settings);
```

where there's an ofRtAudioSoundStream::Settings that extends ofSoundStream::Settings

underdoeg:
Ah yes. of course. nevermind then :)

arturoc:
i'm going to close this then, if you want to open an issue or just send a new PR?

underdoeg:
Yes fine. I'll try to send a PR when I get to it

Issue #4291, opened 2015-08-24
ofPopStyle() applies the wrong style


armadillu:
I've been observing some ofPushStyle() and ofPopStyle() weirdness, and I tracked it back into ofPopStyle(); 

```
void ofGLRenderer::popStyle(){
	if( styleHistory.size() ){
		setStyle(styleHistory.front()); ///!!!!
		styleHistory.pop_back();
	}
}
```
Why are we applying the very first style pushed (styleHistory.front()) in popStyle()? 
Shouldn't it apply the most recent style (which will be at the end of the deque, styleHistory.back())?
Especially bc that's the one you are deleting afterwards...

This has been in since October 2014, I'm surprised nobody noticed so far; This happens in all renderers:
ofGLRenderer, ofGLProgrammableRenderer and ofCairoRenderer.

armadillu:
Prepared a PR #4292

kylemcdonald:
wow. this would explain some really weird bugs i've seen in the past.

armadillu:
yeah I was going really insane; I had some additive blending sneak into the deque which made really hard to understand what what going on with the colors.

PR #4292, opened 2015-08-24
fix popStyle applying the wrong style in all renderers

armadillu:
As explained in #4291


arturoc:
i guess most people just use 1 level of push/pop style at most. thanks!

PR #4298, opened 2015-08-24
fix 4282

### LINKED ISSUE #4282: ofAppEGLWindow/ES2 crashing with OF 0.9 RC1
(actually currently using the 2015 08 20 nightly but saw this previously)

Test case
```
#include "ofMain.h"
#include "ofApp.h"
#if (OF_VERSION_MINOR != 9)
#include "ofGLProgrammableRenderer.h"
#endif
int main( ){

	ofSetLogLevel(OF_LOG_VERBOSE);
	
	#if (OF_VERSION_MINOR == 9)
		ofGLESWindowSettings settings;
    	settings.width = 1280;
   	 	settings.height = 720;
		settings.setGLESVersion(2);
		ofCreateWindow(settings);
	#else
		ofSetLogLevel("ofThread", OF_LOG_ERROR);
		ofSetCurrentRenderer(ofGLProgrammableRenderer::TYPE);
	#endif
	ofRunApp( new ofApp());

}

```

Crashes with

```
[notice ] ofAppEGLWindow: setupRPiNativeWindow(): screenRect: 1360x768
[notice ] ofAppEGLWindow: setupRPiNativeWindow(): windowRect: 1280x720
[notice ] ofAppEGLWindow: createSurface(): setting up EGL Display
[notice ] ofAppEGLWindow: createSurface(): EGL Display correctly set 0x1
[notice ] ofAppEGLWindow: createSurface(): no current renderer selected
[notice ] ofAppEGLWindow: createSurface(): GLES2 renderer detected
[notice ] ofAppEGLWindow: createSurface(): surface created correctly
[notice ] ofAppEGLWindow: createSurface(): API bound correctly
[notice ] ofAppEGLWindow: createSurface(): -----EGL-----
[notice ] ofAppEGLWindow: createSurface(): EGL_VERSION_MAJOR = 1
[notice ] ofAppEGLWindow: createSurface(): EGL_VERSION_MINOR = 4
[notice ] ofAppEGLWindow: createSurface(): EGL_CLIENT_APIS = OpenGL_ES OpenVG
[notice ] ofAppEGLWindow: createSurface(): EGL_VENDOR = Broadcom
[notice ] ofAppEGLWindow: createSurface(): EGL_VERSION = 1.4
[notice ] ofAppEGLWindow: createSurface(): EGL_EXTENSIONS = EGL_KHR_image EGL_KHR_image_base EGL_KHR_image_pixmap EGL_KHR_vg_parent_image EGL_KHR_gl_texture_2D_image EGL_KHR_gl_texture_cubemap_image EGL_KHR_lock_surface 
[notice ] ofAppEGLWindow: createSurface(): GL_RENDERER = VideoCore IV HW
[notice ] ofAppEGLWindow: createSurface(): GL_VERSION  = OpenGL ES 2.0
[notice ] ofAppEGLWindow: createSurface(): GL_VENDOR   = Broadcom
[notice ] ofAppEGLWindow: createSurface(): -------------
[notice ] ofAppEGLWindow: setupPeripherals(): peripheral setup complete
[notice ] ofAppEGLWindow: setupNativeUDev(): created udev object
[notice ] ofAppEGLWindow: setupMouse(): unabled to find mouse
[ error ] ofAppEGLWindow: setupMouse(): did not open mouse
[ error ] ofAppEGLWindow: setupMouse(): did not open mouse, mouse_fd < 0
[warning] ofAppEGLWindow: setupKeyboard(): unabled to find keyboard
[ error ] ofAppEGLWindow: setupKeyboard(): did not open keyboard
[ error ] ofAppEGLWindow: setupKeyboard(): did not open keyboard, keyboard_fd < 0
[notice ] ofAppEGLWindow: setupPeripherals(): native event setup complete
[verbose] ofShader: checkAndCreateProgram(): creating GLSL program
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
[verbose] ofSignalHandler: Aborted
[notice ] ofAppEGLWindow: destroyNativeKeyboard()
[notice ] ofAppEGLWindow: destroyNativeKeyboard(): unable to reset terminal
[notice ] ofAppEGLWindow: destroySurface(): destroying EGL surface
```

Using latest Raspbian Wheezy/ RPi 1/g++ 4.8.2

tgfrerer:
* implements "#pragma include" parsing using stream ops to be independent of `std::regex` which is not yet implemented in gcc < 4.9. 

* remove raspberry pi specific shader linker error parsing until we have found the right compiler switches - it seems that the gcc version needs to be set in a -D flag in the makefiles - @arturoc could this affect the android target, too? The main thing I couldn't get my head around with the RPi error log parser is that it seemed to assume (hardcoded) that any error found must have occurred in the fragment shader - which may or may not be the case, and so the listing printed out might be for the wrong file if a compile error happened in the vertex shader for example

* exclude raspberry from targets where shader compiler error logs are parsed for error line number. It might be worth checking on other linux targets whether gcc >=4.9, so that regex was properly supported. 

I've written a test case as a main.cpp to test the includes parser, you can find it here: https://gist.github.com/tgfrerer/61c352b8fe8d2269ede6

### DIFF
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -382,7 +382,7 @@ void ofShader::checkShaderInfoLog(GLuint shader, GLenum type, ofLogLevel logLeve
 		GLchar* infoBuffer = new GLchar[infoLength];
 		glGetShaderInfoLog(shader, infoLength, &infoLength, infoBuffer);
 		ofLog(logLevel, "ofShader: %s shader reports:\n%s", nameForType(type).c_str(), infoBuffer);
-#ifndef TARGET_RASPBERRY_PI
+#if (!defined(TARGET_LINUX) || defined(GCC_HAS_REGEX))
 		if (shaderSource.find(type) != shaderSource.end()) {
 			// The following regexp should match shader compiler error messages by Nvidia and ATI.
 			// Unfortunately, each vendor's driver formats error messages slightly different.
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -236,18 +236,55 @@ string ofShader::parseForIncludes( const string& source, vector<string>& include
 	stringstream output;
 	stringstream input;
 	input << source;
+
+	auto match_pragma_include = [](const std::string& s_, std::string& filename_) -> bool {
+		filename_ = "";
+		std::istringstream s(s_);
+		s >> std::ws; // eat up any leading whitespace.
+		
+		if (s.peek() != '#') return false;
+		// -----| invariant: found '#'
+		s.seekg(1, std::ios::cur); // move forward one character
+		
+		std::string p, i, f;
+		
+		// while skipping whitespace, read in tokens for: pragma, include, and filename
+		s >> std::skipws >> p >> i >> f;
+		
+		if (p.empty() || i.empty() || (f.size() < 2) ) return false;
+		// -----| invariant: all tokens have values
+		
+		if (p != "pragma") return false;
+		if (i != "include") return false;
+		
+		// first and last character of filename token must match and be either
+		// '<' and '>', or '"
+		
+		if (f[0] == '<' && f[f.size()-1] != '>') return false; //< mismatching brackets
+		
+		if ((f[0] == '"' || f[0] == '\'') && (f[0] != f[f.size()-1])) return false; // mismatching quotes
+		
+		// invariant: filename properly quoted.
+		
+		filename_ = f.substr(1,f.size()-2);
+		
+		return true;
+	};
 	
-	std::regex re("^\\s*#\\s*pragma\\s+include\\s+[\"<](.*)[\">].*");
-	std::smatch matches;
+	// once std::regex is available across the board, use this regex in favour of the above lambda:
+	// std::regex re("^\\s*#\\s*pragma\\s+include\\s+[\"<](.*)[\">].*");
+	
 	string line;
 	while( std::getline( input, line ) ) {
-		std::regex_match( line, matches, re );
-		if ( matches.size() < 2 ) {
+
+		string include;
+		
+		if (!match_pragma_include(line, include)){
 			output << line << endl;
 			continue;
-		}
+		};
 		
-		string include = matches[1];
+		// --------| invariant: '#pragma include' has been requested
 		
 		if ( std::find( included.begin(), included.end(), include ) != included.end() ) {
 			ofLogVerbose("ofShader") << include << " already included";
@@ -258,7 +295,6 @@ string ofShader::parseForIncludes( const string& source, vector<string>& include
 		
 		include = ofFile(ofFilePath::join(sourceDirectoryPath, include)).getAbsolutePath();
 		included.push_back( include );
-		
 		
 		ofBuffer buffer = ofBufferFromFile( include );
 		if ( !buffer.size() ) {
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/gl/ofShader.cpp
```diff
@@ -381,29 +427,13 @@ void ofShader::checkProgramInfoLog(GLuint program) {
...
 		GLchar* infoBuffer = new GLchar[infoLength];
 		glGetProgramInfoLog(program, infoLength, &infoLength, infoBuffer);
+		// TODO: it appears that Raspberry Pi only reports shader errors whilst linking,
+		// but then it becomes hard to figure out whether the fragment or the
+		// vertex shader caused the error.
+		// We need to find a robust way of extracing this information from
+		// the log, and unfortunately can't use regex whilst gcc on RPi is assumed to
+		// be < 4.9, which is the first version fully supporting this c++11 feature.
 		string msg = "ofShader: program reports:\n";
-#ifdef TARGET_RASPBERRYPI
```
arturoc: let's leave this in but ifdef'd for HAS_REGEX. you can run arch linux in the pi which already is on gcc 4.9 or 5

libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
```diff
@@ -131,7 +131,7 @@ ifeq ($(CXX),g++)
 			PLATFORM_CFLAGS = -Wall -std=c++0x -DHAS_TLS=0
 		else
 			ifeq ("$(GCC_MINOR_GTEQ_9)","1")
-				PLATFORM_CFLAGS = -Wall -std=c++14
+				PLATFORM_CFLAGS = -Wall -std=c++14 -DGCC_HAS_REGEX
```
arturoc: there's another line where this should be defined below this, where the gcc version comes from the compiler name

arturoc:
apart from those 2 comments it looks great!  thanks!!

libs/openFrameworks/gl/ofShader.cpp
tgfrerer: Yes, it would be great to have something in, but i'm not too sure about how we'd catch errors from within the vertex shader - the removed code assumed hardcoded that any error must have come from the fragment shader source. 
I figure it's better to print the straight error log rather than maybe confusing listings from the wrong file.

PetrosKataras:
It would be great to merge this as soon as possible :) . I m hitting on this right now on xubuntu 14.04

tgfrerer:
@PetrosKataras if you wanted to checkout the branch from the PR locally, you could help speeding up the process by reporting back if the PR solves the  issue on your config =) 

Here's how to check out a PR locally: https://help.github.com/articles/checking-out-pull-requests-locally/

libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
```diff
@@ -142,10 +142,10 @@ ifeq ($(CXX),g++)
 	endif
 else
 	ifeq ($(CXX),g++-5)
-		PLATFORM_CFLAGS = -Wall -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0
+		PLATFORM_CFLAGS = -Wall -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0 -DGCC_HAS_REGEX
 	else
 	    ifeq ($(CXX),g++-4.9)
-		    PLATFORM_CFLAGS = -Wall -std=c++14
+		    PLATFORM_CFLAGS = -Wall -std=c++14 -DGCC_HAS_REGEX
```
tgfrerer: I've added the defines to these two checks, assuming gcc 5 must support `std::regex`, too.

arturoc:
done :)   i'll check at some point if t's possible to find out if the error in the RPI is happening on the vertex or the fragment but from what i remember it wasn't, so i set it to the fragment which is usually the most complex. But yes it can be really confusing.

PetrosKataras:
Just went away for half an hour and everything is set already... :) Just tried it and I can confirm that it works for me. Many thanks to both of you..

PR #4299, opened 2015-08-25
Fixes osx ofToDataPath. See openframeworks/projectGenerator#22

arturoc:
In osx when OF starts it sets the current directory to the directory
where the exe is located and then uses a relative path from there to
correctly find the data folder.

This has several problems:

- The current working directory is not correct anymore, mostly for command
line applications is useful to know where the command was run from. This is
for example making the command line PG fail in osx and would require a hack
like https://github.com/openframeworks/projectGenerator/pull/22, in general
people would need to know that OF internally changes the cwd but even with
that the actual cwd is totally lost so there's no way to recover it except
by queryng $PWD which is not multiplatform

- If someone changes the current working directory using setcwd or std::current_path
the data folder won't be found anymore

Instead of that this commit changes the data path to EXE_DIR/../../..data
which always works no matter if the cwd is changed by the application or
where the application is run.

This also changes `ofFilePath::getCurrentWorkingDir` which right now is returning
the exe dir instead of the cwd

### DIFF
--- a/libs/openFrameworks/app/ofAppGlutWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGlutWindow.cpp
@@ -325,6 +325,12 @@ void ofAppGlutWindow::setup(const ofGLWindowSettings & settings){
 	if (settings.isPositionSet()) {
 		setWindowPosition(settings.getPosition().x,settings.getPosition().y);
 	}
+
+#ifdef TARGET_OSX
+	// The osx implementation of glut changes the cwd, this restores it
+	// to wherever it was when the app was started
+	ofRestoreWorkingDirectoryToDefault();
+#endif
 }
 
 #ifdef TARGET_LINUX
--- a/libs/openFrameworks/app/ofAppRunner.cpp
+++ b/libs/openFrameworks/app/ofAppRunner.cpp
@@ -44,6 +44,7 @@ static bool & initialized(){
 
 void ofExitCallback();
 void ofURLFileLoaderShutdown();
+void ofSetWorkingDirectoryToDefault();
 
 #if defined(TARGET_LINUX) || defined(TARGET_OSX)
 	#include <signal.h>
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -1455,20 +1455,7 @@ bool ofFilePath::isAbsolute(const std::string& path){
 
 //------------------------------------------------------------------------------------------------------------
 string ofFilePath::getCurrentWorkingDirectory(){
-	#ifdef TARGET_OSX
-		char pathOSX[FILENAME_MAX];
-		uint32_t size = sizeof(pathOSX);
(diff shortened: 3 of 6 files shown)


ofZach:
one option would be to add this code we remove in this PR just to ofAppGlutWindow since as I recall this is specific to glut on osx trying to automagically change the CWD and having wonkiness (like the data folder not being found) w/ glut apps because of the app's propensity to guess where you want the CWD to be and change it.  It's also a setting in the app itself: 

screen shot 2015-08-25 at 4 56 43 am

I remember this code was specifically to override what glut is doing.

arturoc:
i don't think we should have the current fix anywhere. this would work for glut too, and the current way is really strange since it even overrides the menu option and sets a cwd which is not even the correct one

ofZach:
sorry there's two conversations happening.  I think IMO the real offender is glut for changing the cwd auto-magically (and often times spuriously), I think even just adding a function in ofAppGlutWindow that has this behavior documented might be enough, so we don't do this cwd fix by default, but if you are using glut and you want to normalize your cwd you can.    I feel like at least having the option around might be good.

arturoc:
:) yes i remember osx being problematic with version changes and ofToDataPath. I've added an ofRestoreWorkingDirectory which restores the working directory to whatever was when the app was started before creating any window so that should solve the problem without explicitly setting it to the exe dir.

I'm not calling it from anywhere but we could call it from the glut window if you think it's better.

arturoc:
i'm merging this, since it fixes a problem with the command line PG and is behaving now the same as before with ofAppGlutWindow

armadillu:
Hey this broke relative paths?

Before this, ofToDataPath("file.txt", false) would return a relative path (i.e. ../../data/file.txt). Now it's always absolute regardless of the provided "bool makeAbsolute" argument...

arturoc:
in which case do you need relative paths? it's more robust this way since it'll always do the right thing no matter where you call the application from

bakercp:
For me, I use relative paths when saving settings and doing configurations, etc.

arturoc:
what do you mean? you can still specify relative paths is just ofToDataPath which would turn the passed path into a canonical absoulte one

armadillu:
yeah loading assets, etc... let's say I have a project with asses specified on the XML... having relative really makes it easy to move binaries from dev to deployment machine, etc. I think it's pretty crucial.

arturoc:
oh you mean when you store the path returned from ofToDataPath?

bakercp:
@arturoc yes

armadillu:
yep, here for example https://github.com/armadillu/ofxTextureAtlas

arturoc:
mmh, ok, i think you should be storing the path before passing it to ofToDataPath(), then when you read it pass it through ofToDataPath that way it'll still work and would be even multiplatform

but let me check if i can make the final path into a relative to data

bakercp:
@arturoc you're probably right.  just backward compat, etc.  MOST of the time I always ask for absolute paths anyway.

PR #4301, opened 2015-08-25
Update projectGenerator submodule commit

mattfelsen:
This commit simply updates the submodule commit to the HEAD of openframeworks/projectGenerator. From what I can tell, that's the current/working version, while the commit in the main oF repo points to openframeworks/projectGenerator@1bb1f78, which is about 2 weeks old and none of the "flavors" of the PG want to compile. Hopefully it's not being held back for a particular reason -- I know the PG is super in flux so apologies if that's the case!


ofZach:
+1 this seems right (sorry I've been forgetting to push the submodule forward)

PR #4309, opened 2015-08-27
Feature ofGetSoundStream

roymacdonald:
added function to get the ofSoundStream used when ofSoundStreamSetup is used


arturoc:
mmh, there was something really weird with those variables, soundStreamInput is not really needed and not even used. I've fixed it now.

what do you need this for? The idea with the global soundStream was that it would go out and relying on an static object will only be problematic. We've done a big effort to remove statics from the graphics system i don't think we should maintain them anywhere else. Can't you make the addon only usable with a sound stream object instead? ofBaseSoundObject api should be even completely independent from any sound stream.

roymacdonald:
Hi,
I also noticed the unused soundStreamInput variable.
I just needed it to make the addon work just in the way that we left it working, but of course it can work a sound stream object.
Actually I dont need to change the addon itself, just the examples. It seemed to be cleaner/clearer to use the already declared soundstream objects.

BTW,  Why are the statics so problematic?

Best!

PR #4313, opened 2015-08-28
bugfix. stores circle resolution in current style. closes #4312

### LINKED ISSUE #4312: ofPushStyle not storing / resetting circle resolution.
This is with the current master. The circle will draw with the default circle resolution. 
If the push/pop style lines are not there it will draw with a resolution of 4. 

Looking into the issue. 

To reproduce. 
```
ofApp::draw(){
	if( ofGetFrameNum() == 0 ){
    	ofSetCircleResolution(4);
    }
    
	ofPushStyle();
        ofNoFill();
        ofDrawCircle(200, 200, 100);
    ofPopStyle();
}
```

ofTheo:
see the issue #4312 
@arturoc I think this should be good to merge - but maybe you could give the okay, just incase there was a reason the circle resolution wasn't being stored.


arturoc:
thanks! yes i think i just forgot to add that : )

PR #4315, opened 2015-08-29
Modernize for 09

jvcleave:
removing ofSetCurrentRenderer(ofGLProgrammableRenderer::TYPE);

### DIFF
--- a/examples/gl/shaderExample/src/main.cpp
+++ b/examples/gl/shaderExample/src/main.cpp
@@ -1,20 +1,23 @@
 #include "ofMain.h"
 #include "ofApp.h"
-#ifdef TARGET_OPENGLES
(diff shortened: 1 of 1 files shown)


bilderbuchi:
is this fixing a deprecation warning?

bilderbuchi:
also @jvcleave would you (in the future) mind creating the PR branches in your own fork of OF, instead of the main repo? Thanks :-)

jvcleave:
`ofSetCurrentRenderer(ofGLProgrammableRenderer::TYPE)` actually causes a compiler error now

I would typically do the PR via branches but working with git on the RPi is too slow (I use the nightlies). I just realized I can commit to my branch on github so I'll do that

Issue #2791, opened 2014-02-04
RFC: .gitignore template file for project generator


bilderbuchi:
Preparing the OFbook chapter on version control, I realized that it would be useful for our users if our project generator would offer to add a `.gitignore` example file to newly generated projects, so users who want to use Git don't have to come up with their own solution.
Failing that (or until that is implemented), I would at least put such a file at an appropriate place in the OF repo so that interested users can get it manually.

Thoughts on that? Would this be welcomed? Where should I put this `example-.gitignore` file? I'm thinking `/scripts/` would be a good place, since all the project templates that the PG uses are deeper in that tree (correct?), in OS-specific directories, but `example-.gitignore` is not really OS-dependent. Or do we have a place where PG-related, OS/IDE-independent files reside?

@ofZach @arturoc @kylemcdonald @ofTheo

bilderbuchi:
@ofZach @arturoc  would this be something you'd find useful in the PG/project templates?

arturoc:
yes i think it would be useful, it could even be a new template that can be optionally applied to a project. do you want to send a PR?

bilderbuchi:
can do, the proposed gitignore file is already [here](https://github.com/openframeworks/openFrameworks/issues/2791#issuecomment-35990153). question: where should I put it in the filetree?

arturoc:
in scripts/templates/gitignore/.gitignore along with a template.config file, you can check the format in any of the none platform templates (any of the gl ones for example)

bilderbuchi:
thank you, done: #4623

PR #4321, opened 2015-08-30
PG custom templates

arturoc:
needed for this PR: https://github.com/openframeworks/projectGenerator/pull/26#issuecomment-136132617


bilderbuchi:
what's the purpose of this additional "standard" directory level in the template directories?

arturoc:
there can be other templates check the PG discussion, but mostly for osx for example, there's now a `nofmod` template which creates a project without fmod. there's some templates that are common to every platform which live in scripts/pgtemplates like gl3.2, gl3.3... but then every platform can have it's own templates too.

ofZach:
@bilderbuchi see also this conversation https://github.com/openframeworks/projectGenerator/pull/26 I had made changes to the command line project to remove fmod, this makes a template out of those changes.  (I think it def. may be helpful down the line for different types of apps)

bilderbuchi:
I see, thank you. I must have read over that part (the link arturo posted above pointed to a comment by mattfelsen, so that created some confusion).
While modifying the templates, might be good to look at #2791 again.

PR #4323, opened 2015-08-31
Re-added TARGET_WIN32 guards around logger calls.

mhashmi:
The TARGET_WIN32 guard went missing in ofxUDPManager.cpp -- re-adding (otherwise compile fails in DEBUG mode). 

Tested on Android.

### DIFF
--- a/addons/ofxNetwork/src/ofxUDPManager.cpp
+++ b/addons/ofxNetwork/src/ofxUDPManager.cpp
@@ -160,7 +160,9 @@ bool ofxUDPManager::ConnectMcast(char* pMcast, unsigned short usPort)
 	if (!Bind(usPort))
 	{
 #ifdef _DEBUG
+#ifdef TARGET_WIN32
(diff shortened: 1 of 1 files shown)


arturoc:
Thanks so much for catching this. We have a multiplatform way of catching socket errors. can you change the defines from something like:

```cpp
 #ifdef _DEBUG
#ifdef TARGET_WIN32
 		ofLogError("ofxUDPManager") << "ConnectMcast(): couldn't bind to " << usPort<< ", err " << WSAGetLastError();
 #endif
#endif
```

to:

```cpp
 #ifdef _DEBUG
 		ofLogError("ofxUDPManager") << "ConnectMcast(): couldn't bind to " << usPort;
                ofxNetworkCheckErrno();
#endif
```

mhashmi:
@arturoc Done! (and tested on Android).

arturoc:
thanks!

PR #4326, opened 2015-08-31
UDPManager updated network error log function

mhashmi:
Hi @arturoc, the log function to call is ofxNetworkCheckError() (wrapper for ofxNetworkCheckErrno()). 

Unfortunately eclipse didn't clean the project correctly and I didn't catch the error on testing. This patch should be correct however.

Thanks -- apologies about the trouble!


arturoc:
thanks!

PR #4333, opened 2015-09-01
changed templates fs so they are 3 levels below root

arturoc:
that way projects can be directly compiled where they are, will send PR for corresponding changes in PG once this passes tests

ping @julapy


arturoc:
this also adds a template.config file which is necesary for the template to be shown by the command line tool and which includes a description for the template and the supported platforms so when the PG is run with the github distro it only shows the correct templates for the specified platform

ofZach:
@arturoc do you imagine these configs would need to be shared in the gui PG or can that come later?  (I have no problem to add this, just not sure how common this would be).  I will take a look at updating the gui to this new layout / system shortly.

arturoc:
we can leave it for later if you want, it'll probably need a little bit of thought to get the gui right and we should release 0.9. but if you think it can be integrated easily go for it. 

i think the main usage right now would be to create projects with the programmable renderer without having to copy and paste the syntax from some other project.

arturoc:
this is the output right now:
```
gl4.1		OpenGL 4.1 application with programmable renderer
gl3.1		OpenGL 3.1 application with programmable renderer
gl4.0		OpenGL 4.0 application with programmable renderer
gl3.3		OpenGL 3.3 application with programmable renderer
nofmod		OSX application with no FMOD linking
gl3.2		OpenGL 3.2 application with programmable renderer
```

with `\t\t` between the name of the template and the description so it's easier to parse it automatically, so it would be relatively straight forward to add a select list or similar for the template

ofZach:
thanks that's parseable !! :)   If we can expose it on the gui we will... it's seems a little complicated because the UI lets you choose multiple platforms, so it's not clear what that means in terms of templates or how to show them.   (some templates would conflict too, since they change main.cpp). 

I feel like maybe just getting the tool we have at the moment shipped for 0.9 is enough but happy to look at this down the line...

arturoc:
the command line tool already returns this:

```
$ projectGenerator -l -p linuxarmv6l,osx

Templates for target linuxarmv6l
gles2		OpenGL ES 2 application with programmable renderer
Templates for target osx
gl4.1		OpenGL 4.1 application with programmable renderer
gl3.1		OpenGL 3.1 application with programmable renderer
gl4.0		OpenGL 4.0 application with programmable renderer
gl3.3		OpenGL 3.3 application with programmable renderer
nofmod		OSX application with no FMOD linking
gl3.2		OpenGL 3.2 application with programmable renderer
```

but yes that's probably not that easy to parse, a possibility would be to not show any templates by now if there's more than one platform selected

ofZach:
yeah I just meant what happens when you ask for a template that only works on osx, but with multiple platforms (since you can pass in multiple platforms....)...  ie, does it give you an error?   I think that approach of just showing template if one platfrom is suggested is a possibility.   I think this is complicated for 0.9 but I'll take a look...

arturoc:
actually i just changed the output of the commandLine tool to:

```
$ projectGenerator -l -p linux,osx

Templates available for all targets
gl3.1		OpenGL 3.1 application with programmable renderer
gl3.2		OpenGL 3.2 application with programmable renderer
gl3.3		OpenGL 3.3 application with programmable renderer
gl4.0		OpenGL 4.0 application with programmable renderer
gl4.1		OpenGL 4.1 application with programmable renderer
```

and:

```
$ projectGenerator -l -p linuxarmv6l,osx

No templates available for all targets
```

and the process will return 0 if there's available templates or something else if there aren't so you just need to call that with the selected platforms every time the selected platforms change to get the list of supported platforms for all selected targets

Issue #4241, opened 2015-08-13
ofxOpenCv compilation fails  - RC0.9 


mrzl:
This occured on Win8.1, when the ofxOpenCv addon is added.

opencv_calib3d249.lib(fundam.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in main.obj
2>opencv_calib3d249.lib(fundam.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in main.obj
2>opencv_calib3d249.lib(calibration.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in main.obj
2>opencv_calib3d249.lib(calibration.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in main.obj
2>opencv_calib3d249.lib(calibinit.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in main.obj
2>opencv_calib3d249.lib(calibinit.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in main.obj
2>opencv_calib3d249.lib(opencv_calib3d_pch.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in main.obj
2>opencv_calib3d249.lib(opencv_calib3d_pch.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in main.obj
2>opencv_calib3d249.lib(modelest.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in main.obj
2>opencv_calib3d249.lib(modelest.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in main.obj
[...]

PR #4336, opened 2015-09-03
Swap vs x64 debug/release libs for OpenCV back to proper place

mattfelsen:
Looks like there was a mixup where the vs x64 debug libs were moved to the release folder, and the release libs were moved to the debug folder.

I was getting errors when compiling with ofxOpenCV in either debug/release in x64 but not Win32. I saw issue https://github.com/openframeworks/openFrameworks/issues/4241, fixed by commit https://github.com/openframeworks/openFrameworks/commit/f9f47b946a1045877693191e07025b54c16182e1 and noticed the paths got switched. This PR puts them in their proper homes :smile:  I can now compile & run on debug/release on both Win32/x64.

libs


bilderbuchi:
ping @openframeworks/windows

LeoColomb:
Seems true! :smile: :+1:

arturoc:
thanks!

PR #4338, opened 2015-09-04
added option for loading jpeg

2bbb:
this PR is proposal.

juggie_lenna

few days ago, motoishmz met problem about ofImage.
he do only load jpg image with ofImage and call draw, but he got result with juggies.
then, we resolved with using https://github.com/satoruhiga/ofxTurboJpeg 

i found options in FreeImage:
https://github.com/openframeworks/openFrameworks/blob/590bc7d0b9306be35e33c06f96370ff6cd2e0762/libs/FreeImage/include/FreeImage.h#L693-L695
and i added parameter, getter and setter make we can set this options.

i have no confidence in naming. if there are more better naming, please tell me.


kylemcdonald:
very interesting! do you know what "default" means? is it fast, accurate, or a mix of both?

2bbb:
hi kyle.

default means fast.
`#define JPEG_DEFAULT        0		//! loading (see JPEG_FAST); saving (see JPEG_QUALITYGOOD|JPEG_SUBSAMPLING_420)`
https://github.com/openframeworks/openFrameworks/blob/590bc7d0b9306be35e33c06f96370ff6cd2e0762/libs/FreeImage/include/FreeImage.h#L693

kylemcdonald:
wow, so right now the default for OF is to load fast, not accurate... very interesting. i'm curious to know what other developers think about this.

jvcleave:
I like having the option exposed. How about just making a static settable `int` as (opposed to the `ofImageJpegLoadingOption` enum)? That way you could use the other flags easily.

2bbb:
@kylemcdonald 
exactly! default is not accurate now!

@jvcleave 
you're right.
i thought type information is helpfull for user.
i noticed enum is not good solution. (because can't use some bit operator, maybe as you say...)
i'd rewrite to add all options about loading jpeg with FreeImage.

i think this has a tiny problem from FreeImage,
maybe options excluding `JPEG_FAST` and `JPEG_ACCURATE` as bit flag, i.e., those can be synthesized.
but if we give `(JPEG_FAST | JPEG_ACCURATE)` then we get result with `JPEG_ACCURATE`.

okay, maybe this is little problem. but, but, i'll be hazy... 😞

arturoc:
i don't like very much having global settings or functions, perhaps we could have some kind of advanced version of the load and save functions that gets an extra parameter for passing flags? if could even be a settings struct like we do in other places:

```
class ofImage{
struct LoadSettings{
    bool accurate = false;
    bool exifRotate = false;
    bool separateCMYK = false;
    bool grayscale = false;
}

void load(string path, LoadSettings settings = LoadSettings());
void load(ofFile file, LoadSettings settings = LoadSettings());
void load(ofBuffer buffer, LoadSettings settings = LoadSettings());
```

in the implementation of those functions we would check if any of the settings is true and set the corresponding freeimage flags.

2bbb:
@arturoc 
thanks for good idiom.
i agree your interface. this is very simple and descriptive.
and if this style is standard in oF now, i will follow.

but, i think, i want a system that sets default setting once to change all loading after that.
because, if many module load images, then we write same setting in all places.
do you have a good solution about this hope?
or, is correct style "write explicitly all places"?

arturoc:
yes, also most of the times you'll be loading images from the same class so you can do something like:

```
ofImage::LoadSettings settings;
settings.accurate = true;
settings.exifRotate = true;
img1.load("someimge.jpg",settings);
img2.load("someimge2.jpg",settings);
img3.load("someimge3.jpg",settings);
...
```

if you really need to load images from several classes and want to use settings from all of them you can always create a small loader class that sets the settings you want to use and loads an image like:

```
class ImageLoader{
ofImage::LoadSettings loadSettings;
public:
    ImageLoader(){
        loadSettings.accurate = true;
        loadSettings.rotateExif = true;
    }
    ofImage loadImage(string path){
        return ofImage(path,loadSettings);
    }
};

// in other places:
img1 = ImageLoader().loadImage("someImage.jpg");
```

or something similar which is almost no code.

2bbb:
@arturoc 
okay. you're right!
i'll follow!
:+1:

2bbb:
i implemented with arturo's advice.

but i don't implement as inner class.
this setting is unaffected with pixel types and that causes complexity about template parameter.

arturoc:
thanks!

bilderbuchi:
@arturoc @2bbb wow, [this huge commit](https://github.com/openframeworks/openFrameworks/commit/ae629a9f346bd79d386a87a2f60383780f0c81e8) hidden in this PR really touched a lot of OF files unnecessarily...

arturoc:
mmh yeah that's strange, the final diff doesn't show it though so i guess it's ok, seems like a weird merge commit that wasn't let with the automatic message.

@2bbb please be more careful when sending PRs it's really difficult to go thorugh all the commits, if you get some weird merge or similar while working on a branch don't send a PR on that branch and instead copy the modified files to a new branch to start from scratch

2bbb:
@arutorc @bilderbuchi 

i've noticed this comments now.
and, sorry for my stupid git operation...
i will be careful in the future.

Issue #2475, opened 2013-08-16
0.8.0 Light and Material problem


Geistyp:
http://imgur.com/GiiGBx3
Left is 0.7.4 , right is 0.8.0.

same code

``` c++
float no_mat[] = {0.0f, 0.0f, 0.0f, 1.0f};
float mat_ambient[] = {0.0215,  0.1745,   0.0215,  1.0};
float mat_diffuse[] = {0.07568, 0.61424,  0.07568, 1.0};
float mat_specular[] = {0.633,   0.727811, 0.633,   1.0};
float low_shininess = 15.0f;

//light properties
float ambient[] = {0.5f, 0.5f, 0.5f, 1.0f};
float diffuse[] = {1.0f, 1.0f, 1.0f, 1.0f};
float specular[] = {1.0f, 1.0f, 1.0f, 1.0f};
float position[] = {100.0f, 100.0f, 100.0f, 0.0f};

void ofApp::setup(){

    glLightfv(GL_LIGHT0, GL_AMBIENT, ambient);
    glLightfv(GL_LIGHT0, GL_DIFFUSE, diffuse);
    glLightfv(GL_LIGHT0, GL_POSITION, position);

    glEnable(GL_LIGHT0);
    glEnable(GL_LIGHTING);

    glEnable(GL_DEPTH_TEST);
    glShadeModel(GL_SMOOTH); 
}

void ofApp::draw(){

    cam.begin();

    ofPushMatrix();
    ofTranslate(0, 0);

    glMaterialfv(GL_FRONT, GL_AMBIENT, mat_ambient);
    glMaterialfv(GL_FRONT, GL_DIFFUSE, mat_diffuse);
    glMaterialfv(GL_FRONT, GL_SPECULAR, mat_specular);
    glMaterialf(GL_FRONT, GL_SHININESS, low_shininess);
    glMaterialfv(GL_FRONT, GL_EMISSION, no_mat);
    ofSphere(0, 0, 50);

    ofPopMatrix();


    cam.end();
}
```

ofMaterial not work? I try to load a model with material, material data loaded, but show nothing.
http://imgur.com/yMZda6w

``` c++
ofxAssimpModelLoader loader;
ofxAssimpMeshHelper mesh;

ofLight light;

void ofApp::setup(){

    ofDisableArbTex(); 
    loader.loadModel("mini.dae");

    ofEnableSeparateSpecularLight();
    ofEnableDepthTest();
[...]

PR #4355, opened 2015-09-10
Making sure GL_COLOR_MATERIAL is disabled when using an ofMaterial

prisonerjohn:
I noticed that ofMaterial colours were not working when using the fixed pipeline. The idea for the fix is from [a comment](https://github.com/openframeworks/openFrameworks/issues/2475#issuecomment-26028340) by @julapy in #2475. It seems to be working fine now.

### DIFF
--- a/libs/openFrameworks/gl/ofGLRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLRenderer.cpp
@@ -24,6 +24,7 @@ ofGLRenderer::ofGLRenderer(const ofAppBaseWindow * _window)
 	linePoints.resize(2);
 	rectPoints.resize(4);
 	triPoints.resize(3);
+	colorMaterialEnabled = false;
 	normalsEnabled = false;
 	lightingEnabled = false;
 	alphaMaskTextureTarget = GL_TEXTURE_2D;
@@ -535,6 +536,9 @@ void ofGLRenderer::unbind(const ofFbo & fbo){
 
 //----------------------------------------------------------
 void ofGLRenderer::bind(const ofBaseMaterial & material){
+	colorMaterialEnabled = glIsEnabled(GL_COLOR_MATERIAL);
+	glDisable(GL_COLOR_MATERIAL);
+	
 	ofFloatColor diffuse = material.getDiffuseColor();
 	ofFloatColor specular = material.getSpecularColor();
 	ofFloatColor ambient = material.getAmbientColor();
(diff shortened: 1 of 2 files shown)


arturoc:
the problem i think is that we've always considered that ofMaterial and ofLight have to be used together. ofEnableLighting already enables GL_COLOR_MATERIAL.

i fixed something similar some days ago:

https://github.com/openframeworks/openFrameworks/commit/dd07291948c46f8f90a90ed20ef5d8825677d3a4

the problem was that we were using:
```
glColorMaterial(GL_FRONT_AND_BACK, GL_AMBIENT_AND_DIFFUSE);
```

instead of:

```
glColorMaterial(GL_FRONT, GL_AMBIENT_AND_DIFFUSE);
glColorMaterial(GL_BACK, GL_AMBIENT_AND_DIFFUSE);
```

but it you enable the lights it'll work.

prisonerjohn:
That seems to have fixed it for me. I guess we can drop this PR, right?

arturoc:
The problem is, it seems to have broken some workarounds people were using to apply colors withtout materials like: https://github.com/openframeworks/openFrameworks/issues/4397#issuecomment-143475011

prisonerjohn:
Ah right, I just saw your comment on the dev list too...

leozimmerman:
With OF 0.9.0 stable version I had problems when enabling lights using ofSetColor & coloured meshes with no materials. I solved it replacing back 
`glColorMaterial(GL_FRONT_AND_BACK, GL_AMBIENT_AND_DIFFUSE);`

instead of  
```
glColorMaterial(GL_FRONT, GL_AMBIENT_AND_DIFFUSE);
glColorMaterial(GL_BACK, GL_AMBIENT_AND_DIFFUSE);
```

arturoc:
where are those lines?

leozimmerman:
ofGLRenderer.cpp 
func: void ofGLRenderer::setColor(int r, int g, int b, int a){} (line 1050 aprox.)

arturoc:
can you send a PR with this fix to the 0.9.3 branch? we'll do a quick release with this and a couple other critical fixes.

take into account that the current syntax is the correct one for opengles so you'll need to add an ifdef for that. also the same change probably needs to be done in the programmable renderer.

also there's an issue for this problem here already: https://github.com/openframeworks/openFrameworks/issues/4684 can you post there about this from now on?

thanks!

arturoc:
in which platform is this happening for you? i'm checking it out but it seems correct for desktop and broken in GLES

leozimmerman:
I'm using an OSX 10.10.5 desktop pc. I'll follow this issue here #4684

PR #4359, opened 2015-09-11
ofPixel: Correct bits for pixel format. Fixes #4358

### LINKED ISSUE #4358: ofGstVideoPlayer loading with OF_PIXELS_NATIVE not working properly.
It seems that trying to load a video on the default format for avoiding the cpu colorspace conversion is broken somehow.  

The video will load but half of it is displaying correctly and the other half is just glitch. I will try to have a look but it would be great if someone else can confirm this also. For me it happens on both the programmable and the legacy renderer.

Just call :  

```
 player.setPixelFormat(OF_PIXELS_NATIVE)
```  
before loading.

I m testing with GStreamer 1.4 on OS X 10.10.4 and Linux 14.04.

PetrosKataras:


arturoc:
thanks!

PR #4365, opened 2015-09-12
enable bound check flags in debug and disable asserts in release

arturoc:
Both `gcc` and `clang` or better libstdc++ and libc++ don't do bound checks when accessing collections using the [] operator, this enables such checks in debug mode, it might slow down debug a bit but i think it'll help a lot caching certain errors.

This also disables asserts in release by defining -DNDEBUG

Both changes are the default in visual studio so it doesn't need to be changed


kylemcdonald:
+1 this should definitely help catch lots of errors

bilderbuchi:
mac and ios builds fail for some reason.

arturoc:
yes, seems llvm doesn't have full support for this yet: http://libcxx.llvm.org/debug_mode.html

i've removed the parts affecting clang and will merge it after the tests pass, this won't work in osx or ios then

PR #4369, opened 2015-09-14
Fix tcpserver deadlock w/ recursive mutex

### LINKED ISSUE #3580: deadlock in ofxTCPServer
This is really strange:

isClientSetup tries to hold the mConnectionsLock lock:

//--------------------------
bool ofxTCPServer::isClientSetup(int clientID){
	ofMutex::ScopedLock Lock( mConnectionsLock );
	return TCPConnections.find(clientID)!=TCPConnections.end();
}

while this lock is already held by another function that calls isClientSetup. For example:

//--------------------------
bool ofxTCPServer::isClientConnected(int clientID){
	ofMutex::ScopedLock Lock( mConnectionsLock );
	return isClientSetup(clientID) && getClient(clientID).isConnected();
}

This happens all over ofxTCPServer, not sure how no one caught it before.
Or maybe I'm missing something?

### LINKED ISSUE #4329: Bug ofxTCPServer::close
I cannot close a tcp server. Seems like there is a deadlock, caused by: https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxNetwork/src/ofxTCPServer.cpp#L56

I am on ubuntu 15.04. 

```c++
#include "ofMain.h"
#include "ofxNetwork.h"

        void setup()
        {
                _tcp.setup(11111);
        }
        void exit()
        {
                _tcp.close();
        }

        ofxTCPServer _tcp;
};
int main(  int argc, char *argv[]  ){
        ofSetupOpenGL(1024,768, OF_WINDOW);
        ofRunApp( new ofApp());
}

```

admsyn:
This is a bit of a band-aid fix to work around multiple member functions calling each other and trying to lock the same mutex, without reworking the call structure. It swaps the regular `std::mutex` for a `std::recursive_mutex`. If you're not familiar, it allows the same mutex to be locked multiple times from the same thread, blocking other threads until it's complete.

Also switched the `std::unique_lock` for the simpler `std::lock_guard`, since in all cases it's just being used as a simple automatic scoped lock.

There's a [bit of related weird logic in `ofxTCPServer::close()`](https://github.com/openframeworks/openFrameworks/blob/621876270b585498e4557eacddf8afc60ffdb8b5/addons/ofxNetwork/src/ofxTCPServer.cpp#L54-L66) that this addresses. There are redundant calls to `stopThread` and a manual `mConnectionsLock.unlock()` that was possible to be hit without a corresponding `mConnectionsLock.lock()`

Fixes #4329 and fixes #3580

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPServer.cpp
+++ b/addons/ofxNetwork/src/ofxTCPServer.cpp
@@ -53,7 +53,7 @@ bool ofxTCPServer::close(){
 
 	if(connected)
 	{
-		std::lock_guard<std::mutex> lock( mConnectionsLock );
+		std::lock_guard<std::recursive_mutex> lock( mConnectionsLock );
 		map<int,ofPtr<ofxTCPClient> >::iterator it;
 		for(it=TCPConnections.begin(); it!=TCPConnections.end(); it++){
 			it->second->close();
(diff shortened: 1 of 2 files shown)


admsyn:
looks like @arturoc fixed this simultaneously :)

arturoc:
:)

PR #4370, opened 2015-09-14
Compile boost using apothecary on a 10.8 machine.

bakercp:


arturoc:
can you also recompile boost for ios?

bakercp:
@arturoc sure -- does 1.58 (used in apothecary) include the patch mentioned here: https://svn.boost.org/trac/boost/attachment/ticket/10591/boost-fs-fchmodat-darwin.patch  or do we need to patch something manually?

libs/boost/include/boost/config/compiler/visualc.hpp
```diff
@@ -285,7 +285,7 @@
 
 //
 // last known and checked version is 19.00.22129 (VC14 Preview):
-#if (_MSC_VER > 1900 && _MSC_FULL_VER > 190022310)
+#if (_MSC_VER > 1800 && _MSC_FULL_VER > 190022310)
```
arturoc: i patched this manually to avoid a warning in visual studio, but it seems i forgot to add it in the apothecary script or probably i added but only to the visual studio target. if you are going to look into the ios thing, can you patch it back?

bakercp: Yes I can.

arturoc:
and, no we need to patch it manually. that patch is only from 2 weeks ago

libs/boost/include/boost/config/compiler/visualc.hpp
arturoc: thanks :)

ofTheo:
its in there current master: https://github.com/boostorg/filesystem/blob/9910eccc13efc7b6aa8498a27d6bef2e6c937a10/src/operations.cpp#L1433 

not sure if we can safely just pull from that?

bakercp:
I'd tend to prefer to pull from a correct upstream rather than patching it ourselves ... but since it isn't released yet, I'm not sure.  We can go either way.  I haven't taken the time to look into the api/abi changes, but it looks like next release will probably be 1.60?  Let me know what you all think we should do and I can do either.

ofTheo:
hmm - we could always clone their current master in a seperate repo and apply the VS patch. 
then point apothecary to that url. that is what we are doing with glfw. 

when they do a 1.6 release we could just change the url to that?

arturoc:
let's just apply that patch over the current version. we are using 1.58 but 1.59 is already released, if we used current master we would need to recompile for all platforms,

bakercp:
OK.

bakercp:
Done here.  https://github.com/openframeworks/openFrameworks/pull/4373

PR #4371, opened 2015-09-14
Compile freetype using apothecary on a 10.8 machine.

bakercp:


bakercp:
These haven't been tested extensively, so I hope they work :)

PR #4373, opened 2015-09-15
Boost filesystem compiled on 10.8 ios + apothecary updates w/ patches.

bakercp:
@ofTheo @arturoc check it out, maybe should test first?  But if the old one wasn't working, then probably just merge and test.


ofTheo:
ping @NickHardeman who was able to reproduce the crash.

thanks! @bakercp  :)

NickHardeman:
I just tested building the empty example in the 7.1 simulator and it worked as expected. 
I have iOS 8 on my phone and can't download XCode 6 to test because it requires >= 10.9.4

PR #4374, opened 2015-09-16
Added info for compiling with OF 0.9

PTS93:
Added extra steps that I had to take to get Android Studio compile with of_v20150915_android_nightly e.g. 0.9.0

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -28,7 +28,7 @@ In the event that Android Studio does not come with an SDK, you can install it f
 
 ### Install the Android NDK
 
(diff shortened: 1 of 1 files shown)


arturoc:
thanks, it's actually better to change the readme to only mention what's required to run 0.9.

the latest sdk will change overtime so it's better to mention the specific version that it's compatible right now since we can only warranty compatibility with that one.

also better to not mention 0.9 and leave the old requirements but instead mention only the new requirements for the current version which in github is 0.9.

In the web page we'll keep the requirements for 0.8.4 until we make 0.9 offical and then move this ones to the web.

Also the current instructions only refer to installing eclipse so the paragraph on how to copy the gradle file should probably go on a new section on how to use android studio

PTS93:
Sure makes sense, I'll close this PR and make a new one.

PR #4375, opened 2015-09-17
Fixed Android Studio tutorial for 0.9

PTS93:
The old NDK info doesn't work with 0.9 and some settings need to be changed with current releases of Android Studio in order to compile.

Changed according to https://github.com/openframeworks/openFrameworks/pull/4374#issuecomment-141001075

I only mentioned that 0.9 is a minimum requirement so people don't try it right now with 0.8.4

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -28,16 +28,8 @@ In the event that Android Studio does not come with an SDK, you can install it f
 
 ### Install the Android NDK
 
-For the current version of openFrameworks, you need to install Android NDK revision r10b or earlier. r10c breaks compatibility with some libraries, and should (at this time) be avoided.
-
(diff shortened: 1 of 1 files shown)


arturoc:
not sure why this PR won't merge can you take a look?

PTS93:
Yea its weird. I can also not modify the file anymore directly in GitHub. Same issue with the last one, thats why I closed it again. I close and open up another..

arturoc:
ok, i'v just divided the readme in 2, one for android studio and one for eclipse and the gradle properties is now in the repository so there's probably no need for this anymore but thanks!

PTS93:
Oh cool. I just made a new one for the studio one. Thing that should still be changed though is the NDK download which is listing 0.9 which doesn't work anymore.

PR #4377, opened 2015-09-18
Adding gradle.properties file

edap:
As reported in this issue https://github.com/openframeworks/ofSite/issues/315 a `gradle.properties` file is needed in the `/libs/openFrameworksCompiled/project/android` folder


arturoc:
is this working for you? i'm trying it but it doesn't work, is there something else that needs to be added?

edap:
@arturoc actually I still didn't get the `libs/openFrameworksCompiled/project/android` running on the emulator. I will try this evening with a real device.

PR #4379, opened 2015-09-18
Fixed Android Studio tutorial for 0.9

PTS93:
The old NDK info doesn't work with 0.9 and some settings need to be changed with current releases of Android Studio in order to compile.

Changed according to https://github.com/openframeworks/openFrameworks/pull/4374#issuecomment-141001075

I only mentioned that 0.9 is a minimum requirement so people don't try it right now with 0.8.4


arturoc:
thanks!

Issue #1477, opened 2012-08-10
fix iOS orientation


julapy:
currently when setting orientation on ios,
the glView remains in portrait mode but everything drawn in the glView is flipped or rotated depending on the specified orientation.

a more elegant method would be to rotate and resize the glView directly to a orientation.
and as far as OF is concerned, its always in OF_ORIENTATION_DEFAULT mode but the glView is adjusted.
this way you never have to flip or rotate any coordinates.... which will cut out a lot of orientation bugs which currently exist.

PR #4381, opened 2015-09-19
Fix for iOS rotation issue + ARC example

borg:
Ping: @julapy @danoli3 
+Added glView frame resize support in ofxiOSViewController in viewWillTransitionToSize
+ofGetWidth()/ofGetHeight() now return value according to orientation
+Added ARC example with native view and resize on rotation

Note that settings.enableHardwareOrientation seems to only affect whether ofSetOrientation takes effect, not whether the view actually rotates according to device (which it does anyhow, unless set in root.plist). It appears to me ofSetOrientation is pretty useless and should be deprecated. 

Also, eg. ofGetWidth() will return 2048 as opposed to 1024 by [[UIScreen mainScreen] bounds].size.width). This makes sense but can be confusing when mixing with native methods. Perhaps some util functions eg. ofGetNativeWidth() would be useful?

#1477


ofTheo:
@borg  thanks so much for this PR! 
Usually though we prefer to get separate PR for different issues / features. 

Did you get to test any of the rotation / orientation changes on actual devices?
Any iOS 7 devices ( as there is a lot of iOS 8 fixes )? 
We've had several iOS orientation fixes in the past and it seems that are always a lot of edge cases. 

btw: ofSetOrientation isn't just used for iOS its also used for Desktop ( so you can easily run OF apps on screens that are orientated differently ). 

going to ping @julapy and @danoli3 

one last small thing is the dive.jpg image license CC 0 ? or free to use?
we have some OF images in the directory lister example if you need images that have clear ownership.

borg:
@ofTheo The example is tied to the issue being fixed in that it shows how to handle some orientation cases, only it also happens to be ARC enabled which is probably what users unaware of the issue would focus on I would have thought. Seeing many examples are 4 yrs old and there is no ARC example that seems a bit urgent.

I am sure there are many PR since there are many ways to fix this issue. I think it is a critical issue though since as long as the view gets cropped on rotation you can't really use oF for iOS development unless you only support one orientation. Is there a place where the official strategy for handling the relationship between ofGetOrientation/iOS roation methods/ofSetOrientation/ofGetWidth is discussed, as it isn't obvious?

I have only tested on iPad Air iOS8.1.2, iPad2 8.4.1, & iPhone6 8.4.1 but seeing it's really hard not to update these days we should probably focus on iOS9 rather than iOS7 (not that there is any reason it should have to break on old device).

This fix essentially only adds a view size transform to the iOS8+ call in ofxiOSViewController.mm
```
- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator {
...
self.glView.center = center;
self.glView.transform = CGAffineTransformMakeRotation(0);
self.glView.frame = CGRectMake(0, 0, size.width,size.height);//this
```
(It's worth noting that this function only seems to receive correct dimensions if the view is currently visible.)

And also ignores the old way of handling rotation for iOS8+
```
- (void)rotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation
                            animated:(BOOL)animated {
	bAnimated = animated;
	
	//borg
	//if 8+ ignore, else gets delayed crop
	if( [[[UIDevice currentDevice] systemVersion] compare:@"8.0" options:NSNumericSearch] != NSOrderedAscending ) {
		return;
	}
```

Finally I removed the check for if rotation had happened inside ofxiOSAppDelegate.mm
```
- (void)receivedRotate:(NSNotification*)notification {
	UIDeviceOrientation deviceOrientation = [[UIDevice currentDevice] orientation];
	//borg
	//subscribe to deviceOrientationChanged(int newOrientation) by calling ofxiOSAlerts.addListener(this); in your app setup
	ofxiOSAlerts.deviceOrientationChanged( deviceOrientation );
}

```
so that any app can receive those events. Better to have duplicate notifications of rotation, rather than none. 

I assumed ofSetOrientation was iOS because it depends on ofiOSWindowSettings settings.enableHardwareOrientation being false to take effect.

```
void ofAppiOSWindow::setOrientation(ofOrientation toOrientation){

...
    if(settings.enableHardwareOrientation == true) {
        [glViewController rotateToInterfaceOrientation:interfaceOrientation animated:settings.enableHardwareOrientationAnimation];
    } else {
        [[UIApplication sharedApplication] setStatusBarOrientation:interfaceOrientation animated:settings.enableHardwareOrientationAnimation];
        if(bResized == true) {
            [glView layoutSubviews]; // calling layoutSubviews so window resize notification is fired.
        }
    }

```
Thx for note about ofSetOrientation. That's awesome and useful, but if it is a oF wide function I'm not sure I understand what the relationship between ofOrientation and device orientation is, as they seem to be tangled (eg. in ofxiOSAppDelegate). I would suggest separating them as you might want to rotate the app view separately from the device view. Isn't iosOrientationExample an exampel of that? (Incidentally I think it is also broken currently unless settings.enableHardwareOrientation = false)

Finally, the image was already in the old example, so I defer to @julapy on that.

borg:
The issue I was trying to fix with making rotateToInterfaceOrientation iOS7- specific was that on iOS8 on init in landscape mode initial layout was correct, but after a split second it would crop to portrait size. I have traced this down to what appear to be a race condition hack (bReadyToRotate) where it was waiting for things to be initialised until rotating first time. I am guessing Apple has changed something there and it is no longer needed, and instead causes this issue. 

In order to make ofAppiOSWindow::setOrientation work when enableHardwareOrientation is true, I have now disabled that race condition fix for iOS8. rotateToInterfaceOrientation is now also enabled for iOS8, but I had to adjust the center point for the rotation to prevent offset in landscape mode. It works, but I am not sure about the code. 
```

- (void)rotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation
                            animated:(BOOL)animated {
	bAnimated = animated;

	
    if(bReadyToRotate == NO) {
        pendingInterfaceOrientation = interfaceOrientation;
        
        // we need to update the dimensions here, so if ofSetOrientation is called in setup,
        // then it will return the correct width and height
        CGSize screenSize = [UIScreen mainScreen].bounds.size;
        CGRect bounds = CGRectMake(0, 0, screenSize.width, screenSize.height);
        if(UIInterfaceOrientationIsLandscape(interfaceOrientation)) {
            bounds.size.width   = screenSize.height;
            bounds.size.height  = screenSize.width;
        }
        self.glView.bounds = bounds;
        [self.glView updateDimensions];
        
        return;
    }
 
	
    if(currentInterfaceOrientation == interfaceOrientation && !bFirstUpdate) {
        return;
    }
	
	if(pendingInterfaceOrientation != interfaceOrientation) {
		CGSize screenSize = [UIScreen mainScreen].bounds.size;
		CGRect bounds = CGRectMake(0, 0, screenSize.width, screenSize.height);
		if(UIInterfaceOrientationIsLandscape(interfaceOrientation)) {
			bounds.size.width   = screenSize.height;
			bounds.size.height  = screenSize.width;
		}
		self.glView.bounds = bounds;
		[self.glView updateDimensions];
	}
	
    CGSize screenSize = [UIScreen mainScreen].bounds.size;
    CGPoint center;
    CGRect bounds = CGRectMake(0, 0, screenSize.width, screenSize.height);
	
	if(UIInterfaceOrientationIsLandscape(interfaceOrientation)) {
		center.x = screenSize.height * 0.5;
		center.y = screenSize.width * 0.5;
	} else {
		center.x = screenSize.width * 0.5;
		center.y = screenSize.height * 0.5;
	}
	
	// Is the iOS version less than 8?
	if( [[[UIDevice currentDevice] systemVersion] compare:@"8.0" options:NSNumericSearch] == NSOrderedAscending ) {
		if(UIInterfaceOrientationIsLandscape(interfaceOrientation)) {
			bounds.size.width = screenSize.height;
			bounds.size.height = screenSize.width;
		}
	} else {
		// Fixes for iOS 8 Portrait to Landscape issues
		if((UIInterfaceOrientationIsPortrait(interfaceOrientation) && screenSize.width >= screenSize.height) ||
		   (UIInterfaceOrientationIsLandscape(interfaceOrientation) && screenSize.height >= screenSize.width)) {
			bounds.size.width = screenSize.height;
			bounds.size.height = screenSize.width;
		} else {
			bounds.size.width = screenSize.width;
			bounds.size.height = screenSize.height;
		}
		//borg
		//NSLog(@"w %f h %f",[UIScreen mainScreen].bounds.size.width,[UIScreen mainScreen].bounds.size.height);
		//assumes Portrait orientation
		if(screenSize.width>screenSize.height){
			center.x = screenSize.height * 0.5;
			center.y = screenSize.width * 0.5;
		}else{
			center.x = screenSize.width * 0.5;
			center.y = screenSize.height * 0.5;
		}
		//NSLog(@"rotating to portrait %i, is portrait %i, currentInterfaceOrientation %i, bound: w %f h %f",UIInterfaceOrientationIsPortrait(interfaceOrientation),UIInterfaceOrientationIsPortrait(self.interfaceOrientation),UIInterfaceOrientationIsPortrait(currentInterfaceOrientation),bounds.size.width,bounds.size.height);
	}
	
    float rot1 = [self rotationForOrientation:currentInterfaceOrientation];
    float rot2 = [self rotationForOrientation:interfaceOrientation];
    float rot3 = rot2 - rot1;
    CGAffineTransform rotate = CGAffineTransformMakeRotation(rot3);
    rotate = CGAffineTransformConcat(rotate, self.glView.transform);
	
    if(animated) {
        NSTimeInterval duration = 0.3;
        if((UIInterfaceOrientationIsLandscape(currentInterfaceOrientation) && UIInterfaceOrientationIsLandscape(interfaceOrientation)) ||
           (UIInterfaceOrientationIsPortrait(currentInterfaceOrientation) && UIInterfaceOrientationIsPortrait(interfaceOrientation))) {
               duration = 0.6;
           }
        [self.glView.layer removeAllAnimations];
        [UIView animateWithDuration:duration animations:^{
            self.glView.center = center;
            self.glView.transform = rotate;
            self.glView.bounds = bounds;
        }];
    } else {
        self.glView.center = center;
        self.glView.transform = rotate;
        self.glView.bounds = bounds;
    }
    
    currentInterfaceOrientation = interfaceOrientation;
    bFirstUpdate = NO;
    
    [self.glView updateDimensions];
}

```

Because oForientation is bundled with device orientation I don't quite understand how it is supposed to work (it has no less thant 4 different orientation variables self.interfaceOrientation, interfaceOrientation, currentInterfaceOrientation, pendingInterfaceOrientation). Getting @julapy's eyes on this would be fantastic. I have a feeling it can all be simplified considerably for iOS8+.

ofTheo:
@julapy - any objections if we merge this in? 
I think getting it in the next RC could be the most useful way to test any possible regressions.

julapy:
hey, will try to review this today... sorry, just been super busy recently.

ofTheo:
@julapy no worries! 
maybe merge it in and we can test the examples? might be easier to evaluate that way?

bilderbuchi:
> maybe merge it in and we can test the examples? might be easier to evaluate that way?

please don't merge into master, and _then_ review, that's the wrong order. ;-) Better to check out and test the merge commit separately to review: `git fetch  upstream pull/4381/merge:review-4381`

bilderbuchi:
Btw, stuff like this is in our wiki: https://github.com/openframeworks/openFrameworks/wiki/openFrameworks-git-workflow#tips-and-tricks (I just clarified that you can pull either the PRs HEAD or the merge commit that way, in case that was unclear).

danoli3:
This breaks landscape projects (set only landscape mode in Xcode). 
I noticed a strange off centring of the window at one point, can't replicate, when rotating a lot very fast.

I'll PR a fix for the landscape mode, it's an issue with the setup code.

ofTheo:
@danoli3 do we need to merge both PRs or does yours include both?

danoli3:
Mine includes both yes.

ofTheo:
thanks @borg ! - merged this in via @danoli3 's PR

borg:
@ofTheo Sweetness
@danoli3 
>I noticed a strange off centring of the window at one point, can't replicate, when rotating a lot very fast.

Why this part of the rotation transform assumes portrait is exactly what I found counterintuitive. But without it there was a consistent offset. Current behaviour clearly better than before but it sounds like rotating quickly creates another race condition in that case.
```
(void)rotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation
                            animated:(BOOL)animated 
...
		//assumes Portrait orientation
		if(screenSize.width>screenSize.height){
			center.x = screenSize.height * 0.5;
			center.y = screenSize.width * 0.5;
		}else{
			center.x = screenSize.width * 0.5;
			center.y = screenSize.height * 0.5;
		}
```

PR #4382, opened 2015-09-19
Correct wiki adding the working NDK version

edap:
As reported in this thread http://forum.openframeworks.cc/t/error-building-the-androidemptyexample/20759/4, the old versions of the NDK are not working, the
r10e works


arturoc:
thanks!

PR #4387, opened 2015-09-21
FreeImage compiled on 10.8 for ios.

bakercp:


bakercp:
Hmm..  Not sure why this failed.  I ran the apothecary script and committed.  Looks like it's missing a bunch of symbols.

bakercp:
@danoli3  Can you take a look at the travis output and let me know if you have any tips?

ofTheo:
looks like missing libpng and libjpeg symbols. 
not sure why they are not getting picked up.

danoli3:
Try and rebuild, clean your build directory.

danoli3:
Just reviewing this, we shouldn't be building the iOS libs with 10.8.
No iOS development for any devices is possible since February this year with Xcode 5.1.1 and that is the latest Xcode available for 10.8. Also building against 10.8 may cause serious errors due to not being compiled with latest LLVM.

As stated: 
> Starting February 1, 2015, new iOS apps uploaded to the App Store must include 64-bit support and be built with the iOS 8 SDK, included in Xcode 6 or later. To enable 64-bit in your project, we recommend using the default Xcode build setting of “Standard architectures” to build a single binary with both 32-bit and 64-bit code.

https://developer.apple.com/news/?id=10202014a

10.9.4 / 10.10 should be minimum OSX build for iOS for 0.9.
This is probably why people on 10.8 are having so much trouble. 
It's not supposed to work with Xcode 5.1.1. LLVM changed so much since.

ofTheo:
@danoli3 I think that sounds good to me. 
I think we can still support OF OS X on 10.8 right? 
If 10.9 is min for iOS I think that is reasonable - Apple is usually much tougher about needing the latest OS for development.

danoli3:
Yeah OSX is fine at 10.8 min, since the iOS Libs are all seperate now to the OSX libs (mandatory since Xcode 7 they be compiled with dedicated iOS targets now).

danoli3:
Can we close this with a wont fix

Issue #4380, opened 2015-09-19
RTAudio cannot find devices when in ASIO 


DomAmato:
I could not open a soundstream to any input device as it never finds any available devices. This was a problem before that stemmed from plugging in a Blue Snowball mic and caused me to recompile RTAudio libraries to compile without ASIO support. Alternatively I added this into the setup and now I can see and use sound devices

#ifdef TARGET_WIN32
		audio = shared_ptr<RtAudio>(new RtAudio(RtAudio::Api::WINDOWS_WASAPI));
#else
		audio = shared_ptr<RtAudio>(new RtAudio());
#endif

However, we can probably make a nicer enumerator or something so that the libraries don't have to be recompiled and can still have ASIO support but currently not specifying the API breaks soundstream on my machines

PR #4388, opened 2015-09-21
Enable specifying RTAudio Api

DomAmato:
This PR is addressing this issue https://github.com/openframeworks/openFrameworks/issues/4380

Im not sure this is the best method but it does work


libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
```diff
@@ -91,39 +91,58 @@ void ofRtAudioSoundStream::setOutput(ofBaseSoundOutput * soundOutput){
...
-	bufferSize			= ofNextPow2(_bufferSize);	// must be pow2
+	sampleRate = _sampleRate;
+	tickCount = 0;
+	bufferSize = ofNextPow2(_bufferSize);	// must be pow2
 
 	try {
-		audio = shared_ptr<RtAudio>(new RtAudio());
-	} catch (std::exception &error) {
-   		ofLogError() << error.what();
+		if (api == OF_RT_API_UNSPECIFIED)
```
bakercp: Seems to be an inconsistent use of brackets here + style guide says to always use brackets for one-liners.

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
```diff
@@ -91,39 +91,58 @@ void ofRtAudioSoundStream::setOutput(ofBaseSoundOutput * soundOutput){
...
+			audio = shared_ptr<RtAudio>(new RtAudio());
```
bakercp: audio = std::make_shared<RtAudio>();

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
```diff
@@ -91,39 +91,58 @@ void ofRtAudioSoundStream::setOutput(ofBaseSoundOutput * soundOutput){
...
+		else {
+			audio = shared_ptr<RtAudio>(new RtAudio((RtAudio::Api)api));
```
bakercp: audio = std::make_shared<RtAudio>((RtAudio::Api)api);

libs/openFrameworks/sound/ofRtAudioSoundStream.h
```diff
@@ -9,6 +9,18 @@
 class RtAudio;
 typedef unsigned int RtAudioStreamStatus;
 
+enum ofRtAPI {
+	OF_RT_API_UNSPECIFIED,	
+	OF_LINUX_ALSA, //The Advanced Linux Sound Architecture API.
```
bakercp: For doxygen purposes, this should be:

```c++
OF_LINUX_ALSA, /// \brief The Advanced Linux Sound Architecture API.
```

etc.

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
DomAmato: shouldn't it be 
audio = std::make_shared(RtAudio(RtAudio::Api)api));

I actually didn't write the shared_ptr I just added the api stuff but should that be changed over in other files too?

bakercp: no, should be 
```
audio = std::make_shared<RtAudio>((RtAudio::Api)api);
```

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
bakercp: somehow it got code unformatted.

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
bakercp: Sorry, github unformatted my code.

libs/openFrameworks/sound/ofRtAudioSoundStream.h
```diff
@@ -9,6 +9,18 @@
 class RtAudio;
 typedef unsigned int RtAudioStreamStatus;
 
+enum ofRtAPI {
```
bakercp: Also, I'm wondering if it's necessary to duplicate an enum already in the underlying code, especially since it's not part of the base class.  For people needing this functionality, couldn't they just call `RtAudio:: LINUX_ALSA` directly?  At the very least, if an oF enum is going to ghost an existing one, it would probably be specified like:

```c++
enum ofRtAPI {
OF_RT_API_UNSPECIFIED = RtAudio:: UNSPECIFIED,
OF_LINUX_ALSA = RtAudio:: LINUX_ALSA,
...

```

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
DomAmato: happened on the website to me to when I was trying to explain a vector lol alright ill fix what you specified here

bakercp: Also, this make_shared approach isn't the law, but we've been slowly migrating over now that we have c++11 and make_shared is available on all of our platforms (basically, it creates the shared_ptr directly w/o copy, etc).

libs/openFrameworks/sound/ofRtAudioSoundStream.h
DomAmato: I tried that but RtAudio::Api is not accessable unless you include rtAudio.h, otherwise it complains about an incomplete type

libs/openFrameworks/sound/ofRtAudioSoundStream.h
```diff
@@ -5,6 +5,7 @@
 #include "ofBaseSoundStream.h"
 #include "ofTypes.h"
 #include "ofSoundBuffer.h"
+#include "RtAudio.h"
 
 class RtAudio;
```
bakercp: If you've got the header now, you can get rid of the forward declaration here.

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
```diff
@@ -91,39 +91,58 @@ void ofRtAudioSoundStream::setOutput(ofBaseSoundOutput * soundOutput){
...
+}
+
+//------------------------------------------------------------------------------
+bool ofRtAudioSoundStream::setup(ofBaseApp * app, int outChannels, int inChannels, int sampleRate, int bufferSize, int nBuffers){
+	setInput(app);
+	setOutput(app);
+	return setup(outChannels,inChannels,sampleRate,bufferSize,nBuffers, OF_RT_API_UNSPECIFIED);
+}
+
+bool ofRtAudioSoundStream::setup(int outChannels, int inChannels, int _sampleRate, int _bufferSize, int nBuffers, ofRtAPI api)
```
bakercp: `ofRtApi` no longer exists.  I think it should be `RtAudio::Api` ?

libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
bakercp: Then if api is of type `RtAudio::Api`, the cast is no longer needed.

libs/openFrameworks/sound/ofRtAudioSoundStream.h
DomAmato: I moved this over to the .h as it was originally in the cpp file, removing it makes accessing the RtAudio::Api enum throw an error

bakercp: Removing the forward declaration `class RtAudio;` causes compile time errors with the header specified above?

DomAmato: ah my bad I can remove that :laughing: my multitasking during class is atrocious

arturoc:
this methods with a ton of arguments are really hard to use, we should have some Settings class like we are doing in other places. That way also the settings can be passed to a plain `ofSoundStream` object.

i'd have something like:

ofSoundStream::Settings
    |_ofRtAudioSoundStream::Settings
    |_ofxAndroidSoundStream::Settings
    |_....

where the hierarchy indicates inheritance. then we could even deprecate most of the setup methods and only have a simplified version perhaps with number of channels of in/out and a pointer to ofApp and then one with the settings object.

take a look at the window classes to see how it's done there

DomAmato:
So something like this?
https://github.com/DomAmato/openFrameworks/blob/OFTinkering/libs/openFrameworks/sound/ofSoundStreamSettings.h

underdoeg:
That's cool. This was a related discussion of the topic but regarding linux https://github.com/openframeworks/openFrameworks/pull/4277

arturoc:
@DomAmato yes something like that but using inner classes:

```cpp
class ofSoundStream{
public:
    class Settings{
    public:
        int outChannels;
        ....
    }
};

class ofRtAudioStream{
public:
    class Settings: public ofSoundStream::Settings{
    public:
        RtAudio::Api api
    }
    ....   
};
```

arturoc:
also since we have ofSoundDevice it might be useful to have some way to specify the device using that object directly instead of the deviceID.

DomAmato:
So I tried that method and when I try to compile it gives me errors saying that ofSoundStream doesn't name a call or namespace and that the base class is undefined. Any ideas, maybe I'm not seeing it

arturoc:
it might be a recursive include issue?

DomAmato:
So I changed around the structures a bit but if I try a simple test of loading an RtAudioSettings it casts into a NULL container.

In setup doing this

    ofRtAudioSoundStream::ofRtSoundStreamSettings settings;
    settings.setApi(settings.MS_WASAPI);
    soundStream.setup(settings);

causes this path -

    ofSoundStream.h
      bool setup(const ofBaseSoundStream::ofBaseStreamSettings & settings) {
		const ofSoundStreamSettings * soundSettings = dynamic_cast<const ofSoundStreamSettings*>(&settings);
		return setup(*soundSettings);
	  }

which then gets passed along to the cpp:

      bool setup(const ofBaseSoundStream::ofBaseStreamSettings & settings) {
		const ofSoundStreamSettings * soundSettings = dynamic_cast<const ofSoundStreamSettings*>(&settings);
		return setup(*soundSettings);
	  }

and then to the specific sound stream class where it goes through the cast again from base to rtAudio settings

I was hoping to just have it inherit like base->abstract->specific class but it kept throwing errors that the ofSoundStream was an undeclared identifier. I am curious if it was the nested classes that was causing it or if thats actually the best method to try.

DomAmato:
So I altered somethings and now it works though its a bit of a departure from the current scheme. It looks a lot more like the ofWindowSettings file and preparing the settings file is a little different

    ofRtSoundStreamSettings settings;

	settings.setApi(settings.MS_WASAPI);
	settings.setNumInputChannels(2);
	settings.setBufferSize(bufferSize);
	settings.setNumBuffers(4);
	settings.setInDeviceID(3);

	soundStream.setup(settings);

	soundStream.printDeviceList();

The input and output pointers was a slight problem and could probably be implemented better

elliotwoods:
Ah great. I was just having this problem myself and posted about it in the oF dev list

My suggestion was to:

1. Have `getDeviceList()` support different API's (either through using an argument, enemerating all api's always, or using the `setApi` pattern)
2. `ofSoundStream::Device` having an Api variable
3. Calling `soundStream.setup(const ofSoundStream::Device &)` will check the Api variable 

There is a question in general:
Is there any disadvantage to setting WASAPI as default API on Windows? From my knowledge, it seems like the best option.

arturoc:
i've changed the api for this a little bit before merging to use something similar to what was discussed in the developers list recently. the examples are also updated to show the new api.

PR #4386, opened 2015-09-21
FreeImage compiled on 10.8 for osx.

bakercp:
Fixes #4385

PR #4390, opened 2015-09-22
Compiling Freeimage 3.17.0 against 10.8 SDK Fixes #4385

### LINKED ISSUE #4385: Freeimage 3.17.0 'Undefined symbols for architecture x86_64'
On latest master 0.9.0 Xcode 4.x 10.8 SDK I can't compile x64 projects because of:

>Undefined symbols for architecture x86_64:
  "___exp10", referenced from:
      _opj_tcd_rateallocate in freeimage.a(tcd.o-x86_64)
ld: symbol(s) not found for architecture x86_64

Recompiling a universal 32/64bit lib with c++11 support against the 10.8 SDK fixes the problem on my machine. 

@danoli3 maybe the build in #3936 and mentioned in #3925 done against 10.9, 10.10 SDK?

Anyway, I've been out of the game for a while so maybe I just checked out the wrong 0.9.0 or perhaps we dropped support for 10.8 - I'm not sure - but in any case I had to recompile :(

gameoverhack:
This is an i386 x86_64 universal library compiled against 10.8 SDK

Fixes issue #4385 where freeimage was not able to link when compiling 64bit apps on 10.8

Probably should be checked on 10.9/10.10 machines - I don't have one nearby at the moment.


bakercp:
Potential duplicate of https://github.com/openframeworks/openFrameworks/pull/4386 ... but this one doesn't have the permissions errors, so I'd go for this one.  Milestone-wise, I think this should be 0.9.0 as it's a regression.

arturoc:
thanks!

PR #4392, opened 2015-09-24
events: add suport for std::function listeners

arturoc:
ofEvent now supports adding std::function as a listener, that allows
among other things:

- pass a lambda function as a listener
- pass a function with a different signature as a listener.

For example to pass a lambda as a listener now you can do:

```cpp
ofEvents().mouseMoved.newListener([](ofMouseEventArgs & mouse){
	cout << mouse << endl;
});
```

Since std::function doesn't have an equality operator, there's no way to
unregister an event the old way calling `ofRemoveListener(function)` since
that required to compare the passed function with every other registered in
the event and remove it if it was the same.

`ofRemoveListener`also has some other problems, for example, the typical pattern:

```cpp
class MyClass{
public:
    MyClass(){
         ofAddListener(ofEvents().mouseMoved,this,&MyClass::mouseMoved);
    }

    ~MyClass(){
         ofRemoveListener(ofEvents().mouseMoved,this,&MyClass::mouseMoved);
    }

    void mouseMoved(ofMouseMoved & mouse){
    ....
    }
}
```

Has a problem with copy construction:

```
MyClass obj;
vector<MyClass> objs;
objs.push_back(obj);
```

Won't be registered as listener and there won't be any compiler error.
Because of the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_%28C%2B%2B_programming%29)
Any class that adds a listener in it's constructor and unregisters
in it's destructor. Needs also a copy constructor, operator=, and in
c++11 to be completely correct also a move constructor and an operator=
with move parameter.

With the new way of registering listeners one does:

```cpp
class MyClass{
    ofEventListener mouseMovedListener;
public:
    MyClass(){
         mouseMovedListener = ofEvents().mouseMoved.newListener(this,&MyClass::mouseMoved);
    }

    void mouseMoved(ofMouseMoved & mouse){
    ....
    }
}
```

The ofEventListener class unregisters the event automatically when it goes out
of scope avoiding the need for a destructor and because of that avoiding the
need for copy constructor...

Also since the class is not copyable if someone tries to copy an instance
of a class that contains an ofEventListener the compiler will give an error.

In any case the old method still works.

### DIFF
--- a/libs/openFrameworks/events/ofEvent.h
+++ b/libs/openFrameworks/events/ofEvent.h
@@ -7,107 +7,6 @@
 #include <memory>
 #include <iterator>
 
-class ofEventAttendedException: public std::exception{};
-
-
-template<typename Function, typename Mutex=std::mutex>
-class ofBaseEvent{
-public:
-	/// \brief Basic constructor enabling an ofBaseEvent.
-	///
-	/// \see ofBaseEvent::ofBaseEvent(const ofBaseEvent & mom)
-	/// \see ofBaseEvent::enable()
-	/// \see ofBaseEvent::disable()
-	/// \see ofBaseEvent::isEnabled()
-	ofBaseEvent()
-	:enabled(true)
-    ,notifying(false){
-	}
-
-	/// \brief Copy-constructor for ofBaseEvent.
-	///
-	/// \see ofBaseEvent::ofBaseEvent()
-	ofBaseEvent(const ofBaseEvent & mom)
-	:enabled(mom.enabled)
-	,notifying(false){
-		std::unique_lock<Mutex> lck(const_cast<ofBaseEvent&>(mom).mtx);
-		functions = mom.functions;
-	}
-
-	/// \brief Overloading the assignment operator.
-	ofBaseEvent & operator=(const ofBaseEvent & mom){
-		if(&mom==this){
-			return *this;
-		}
-		std::unique_lock<Mutex> lck(const_cast<ofBaseEvent&>(mom).mtx);
-		functions = mom.functions;
-		enabled = mom.enabled;
-        notifying = mom.notifying;
-		return *this;
-	}
-
-	/// \brief Enable an event.
-	///
-	/// \see ofBaseEvent::disable()
-	/// \see ofBaseEvent::isEnabled()
-	void enable() {
-		enabled = true;
-	}
-
-	/// \brief Disable an event.
-	///
-	/// \see ofBaseEvent::enable()
-	/// \see ofBaseEvent::isEnabled()
-	void disable() {
-		enabled = false;
-	}
-
-	/// \brief Check whether an event is enabled or not.
-	///
-	/// \returns true if enables
-	/// \see ofBaseEvent::enable()
-	/// \see ofBaseEvent::disable()
-	bool isEnabled() const {
-		return enabled;
-	}
-
-	std::size_t size() const {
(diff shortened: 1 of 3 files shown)


libs/openFrameworks/events/ofEvent.h
```diff
@@ -192,111 +280,206 @@ class ofEvent: public ofBaseEvent<of::priv::Function<T>,Mutex>{
...
+
+	of::priv::Function<T> make_function(std::function<void(const void*,T&)> & f, int priority  = OF_EVENT_ORDER_APP){
+		return of::priv::Function<T>(priority, [f](const void*s,T&t){f(s,t); return false;}, of::priv::make_function_id());
+	}
+
 	template<class TObj, typename TMethod>
-	std::unique_ptr<FunctionId<TObj,TMethod>> make_function_id(TObj * listener, TMethod method){
-		return std::unique_ptr<FunctionId<TObj,TMethod>>(new FunctionId<TObj,TMethod>(listener,method));
+	std::unique_ptr<FunctionId<TObj,TMethod>> make_function_id(TObj * listener, TMethod method, int priority  = OF_EVENT_ORDER_APP){
+		return std::unique_ptr<FunctionId<TObj,TMethod>>(new FunctionId<TObj,TMethod>(listener,method,priority));
```
bakercp: can we use our `std::make_unique` here? and similar places?

elliotwoods:
This looks nice.

The way this differs from ofxLiquidEvent:

* It doesn't look like you can add an order index when you add the listener
* It's quite common to want to notify listeners in reverse
* I love the `+=` syntax stolen from C# : https://github.com/elliotwoods/ofxLiquidEvent/blob/master/Example_simpleExample/src/ofApp.cpp#L6 Any chance it might make it over to here?

Other than that, I mostly use an arbitrary pointer to flag the events owner, e.g. if i know that i want to remove the listeners when a class is destroyed:

```c++
otherClass.onEventHappened.addListener([]() { ...}, this);

...

otherClass.onEventHappened.removeListeners(this); // remove any listeners associated with this
```

i've been using that pattern for a while and haven't found a situation where it doesn't work yet.
but.. i can see that it's not very tidy in some respects so prefer your `ofEventListener` idea for the core

arturoc:
yeah ii saw that in ofxLiquidEvents, i think if we can keep things free of raw pointers we'll get rid of a lot of problems. an object address changes for example when it's stored in a vector and the vector moves in memory invalidating the event registration. with this method that problem is solved.

another problem we had with the add in constructor remove in destructor pattern is that by the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_%28C%2B%2B_programming%29) or the rule of 5 in c++11 you need to implement copy constructors, assignment operator and move constructor and assignment operator whenever you have a destructor which makes it super tedious and error prone to implement and maintain any class that listens to events right now.

that last problem is not fully solved by this since although there's no  destructor anymore whenever you copy an object the copy won't autoregister to the event if the original was. one possible solution is to delete the copy constructor of this token objects so objects listening to events can't be copied or moved

arturoc:
also there's already priorities in the core which will also work for this. we don't have notifyInReverse but since there's an ordering by the priority it would be trivial to implement

arturoc:
eg: you can do
```
mouseMovedListener = ofEvents().mouseMoved.newListener(this,&MyClass::mouseMoved,priority);
```

PR #4393, opened 2015-09-24
ofParameter: add ofParameter<void>

arturoc:
which it's pretty useless by itself but allows to add a parameter to
an `ofParameterGroup` that will end up represented as a button in an
ofxGui.

an `ofParameter<void>` can't be set or have operators... but has a
`trigger()` method that calls the corresponding event.

This still doesn't add the posibility to add an `ofParameter<void>` to
an `ofxPanel`.

ping @frauzufall

### DIFF
--- a/libs/openFrameworks/types/ofParameter.cpp
+++ b/libs/openFrameworks/types/ofParameter.cpp
@@ -54,3 +54,67 @@ istream& operator>>(istream& is, ofAbstractParameter& p){
 	p.fromString(str);
 	return is;
 }
+
+
+
+ofParameter<void>::ofParameter()
+:obj(new Value){
+
+}
+
+ofParameter<void>::ofParameter(const string& name)
+:obj(new Value(name)){
+
+}
+
+void ofParameter<void>::setName(const string & name){
+	obj->name = name;
+}
+
+string ofParameter<void>::getName() const{
+	return obj->name;
+}
+
+std::string ofParameter<void>::toString() const{
+	return "";
+}
+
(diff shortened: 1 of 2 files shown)


frauzufall:
I am implementing the functions to add the void parameter to the gui as a button and the example would look like this:
```
ofParameter<void> ringParameter; 
gui.add(ringParameter.set("ring"));
```
But it says `error: 'class ofParameter<void>' has no member named 'set'` . It should be possible to set the void parameter using a name. I can't find the point where I could add this to ofParameter.h.

arturoc:
you should call setName("...") since it has now value, set doesn't make much sense but we could include it to be consistent with other parameters. it should go here: https://github.com/openframeworks/openFrameworks/pull/4393/files#diff-ff74a8400c9c3e21fadf9c2b75f311b7R875

frauzufall:
It would be just about consistency, because it could look like this:
```
gui.add(circleResolution.set("circleRes", 5, 3, 90));
gui.add(screenSize.set("screenSize", ""));
gui.add(ringParameter.set("ring"));
```
instead of
```
gui.add(circleResolution.set("circleRes", 5, 3, 90));
gui.add(screenSize.set("screenSize", ""));
ringParameter.setName("ring");
gui.add(ringParameter);

```

arturoc:
yeah makes sense, if you want to send me a PR? otherwise i'll change it later

frauzufall:
I sent you a PR. I updated my refactored ofxGui-branch accordingly and it works so far.

PR #4394, opened 2015-09-25
Made the deprecation macro work with gcc 5.0

Foaly:
Hi there! 
I work on the SFML library and we are currently implementing a deprecation macro. While looking at existing solutions someone on the forum noticed a bug in your implementation. `OF_DEPRECATED_MSG` won't work with GCC 5.0, because the minor version is no longer less than 5. I thought I should report this bug to the team.

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -35,7 +35,7 @@ enum ofTargetPlatform{
 // Cross-platform deprecation warning
 #ifdef __GNUC__
 	// clang also has this defined. deprecated(message) is only for gcc>=4.5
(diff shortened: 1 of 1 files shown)


bilderbuchi:
I think that has been fixed in the meantime, because in current master it looks like [this](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofConstants.h#L38). ping @arturoc 
Thank you anyway for spotting this, and a fix!

arturoc:
yes i fixed this yesterday already, thanks anyway!

Foaly:
Haha dang! One day to late. Anyway good to see it is fixed. Keep up the amazing work!

Issue #4401, opened 2015-09-28
ofAVFoundationVideoPlayer memory leak


trentbrooks:
Just noticed a large memory leak in ofAVFoundationVideoPlayer's -(void) updateFromVideoOutput method. The CVPixelBufferRef buffer does not get released. If you load any HD movie and constantly adjust the playhead (setPosition() or setFrame()) - it will eat through all your memory very quick (8gb of memory in less than 2 min for me).

Fix: https://github.com/trentbrooks/openFrameworks/commit/7312aff0f7667df379c4b3bd91857d51e9ec25d0

bilderbuchi:
ping @openframeworks/video @openframeworks/macos

arturoc:
can you send a PR?

i-n-g-o:
oops.
PR here: https://github.com/openframeworks/openFrameworks/pull/4404

i-n-g-o:
sorry, did not see that there was a PR already. closed mine.

Issue #4384, opened 2015-09-21
ofAVFoundationPlayer previousFrame() then play() texture cache not updating


trentbrooks:
When running the standard videoPlayerExample and rewinding via previousFrame() (in frame by frame mode), once you hit play again (toggle f key) the texture stays frozen until the playhead catches up to the time it was initially paused. The timecode updates, and sound plays as normal during this frozen texture state (tested with other videos). Note- this does not occur with nextFrame().

May be related to https://github.com/openframeworks/openFrameworks/issues/3629

trentbrooks:
Basically the -seekToTime:toleranceBefore:toleranceAfter: values need to be set to kCMTimeZero for higher precision seeking instead of the default kCMTimePositiveInfinity (https://developer.apple.com/library/mac/documentation/AVFoundation/Reference/AVPlayer_Class/index.html#//apple_ref/occ/instm/AVPlayer/seekToTime:toleranceBefore:toleranceAfter:). 

In OF, this means changing line 881 in ofAVFoundationVideoPlayer.m... (https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m#L881) to the following...

[self seekToTime:CMTimeMakeWithSeconds(timeSec, NSEC_PER_SEC) withTolerance:kCMTimeZero];

The fallout for using this method is the slight 'additional decoding delay', as mentioned in the apple docs. So not sure if this is something worth submitting a PR for?

Thanks,
Trent

P.S. without modifying the core, i used this ofApp snippet instead of previousFrame() https://gist.github.com/trentbrooks/f8a0bd42334c12890083

trentbrooks:
Here's the commit if a future PR is required- https://github.com/trentbrooks/openFrameworks/commit/eaca4ca36967143619c4356fa69ac93970df57ed

i-n-g-o:
while PR #4402 fixed reported issue another issue is related to the same problem: https://github.com/openframeworks/openFrameworks/issues/4651

closing this ticket and continue discussion in https://github.com/openframeworks/openFrameworks/issues/4651

PR #4402, opened 2015-09-28
Bugfix avfvideo memleak

trentbrooks:
Two fixes for avfoundationvideplayer:

1. memory leak: https://github.com/openframeworks/openFrameworks/issues/4401
2. texture cache freeze: https://github.com/openframeworks/openFrameworks/issues/4384

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -663,6 +663,9 @@ static const NSString * ItemStatusContext;
 			return;
 		}
 		
+		// release temp buffer
+		CFRelease(buffer);
+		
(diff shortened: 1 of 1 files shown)


i-n-g-o:
hi.
thanks for this PR.

could you please use `CVBufferRelease` instead of `CFRelease` on the `CVPixelBufferRef`?
it may be that `copyPixelBufferForItemTime` returns NULL, then `CFRelease` would crash.

from the docs:
`Unlike CFRelease, CVBufferRelease can be passed NULL without causing a crash.`

ofTheo:
thanks for the tip @i-n-g-o 
these both look good @trentbrooks - could you switch it to CVBufferRelease and we can merge this in?

ofTheo:
Thanks!!

PR #4407, opened 2015-10-02
globalLocalEventsExample

borg:
I put together an example that shows a pretty simple event system that I think feels familiar for people from a web dev background. It shows how you can easily manage local and global events and a few other tricks. Very basic, but I found myself struggling to get my head around the oF event system back in the day. I don't think the other event examples cover these things.

Include if you like. Reject if your don't. 
Best

### DIFF
--- a/addons/ofxiOS/src/core/ofxiOSEAGLView.mm
+++ b/addons/ofxiOS/src/core/ofxiOSEAGLView.mm
@@ -10,6 +10,7 @@
 #include "ofAppiOSWindow.h"
 #include "ofGLRenderer.h"
 #include "ofGLProgrammableRenderer.h"
+#include "ofxiOS3dTouch.h"
 
 static ofxiOSEAGLView * _instanceRef = nil;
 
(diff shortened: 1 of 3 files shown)


examples/events/globalLocalEventsExample/src/MyButton.cpp
```diff
@@ -0,0 +1,86 @@
+/*
+ *  MyButton.cpp
+ *  globalLocalEventsExample
+ *
+ *  Created by Andreas Borg on 10/2/15
+ *  Copyright 2015 Local Projects. All rights reserved.
```
bilderbuchi: I think "all rights reserved" is not an appropriate copyright line, as we then couldn't use/include the code in OF, which is MIT-licensed.

borg: Of course...auto generated stuff

bilderbuchi: thanks :-)

PR #4413, opened 2015-10-06
tvOS - Apple TV OS Feature

danoli3:
Just an initial working version of oF 0.9 for Apple TV tvOS.

Requirements: 
------
- Xcode 7.1+ 

Changes:
---------

- Added seperate static project file for tvOS (as it is quite different from iOS)
- Added initial tvOS template project
- Shares the iOS fat libs (only uses the arm64 slices).

Changes made include modifications to ofxiOS with some pre-processor defines to not expose certain functions that are not in the tvOS SDK (such as video grabber, mapkit, orientation code etc).


julapy:
ah wow thats cool!
i haven't really looked into tvOS too much,
but great to know we can get OF running on it.

will have a look through and leave comments and questions about anything that stands out.

if all good, we should merge this after 0.9 release.

addons/ofxiOS/src/video/AVFoundationVideoPlayer.m
```diff
@@ -326,7 +326,7 @@ - (BOOL)createAssetReaderWithTimeRange:(CMTimeRange)timeRange {
     }
     
     //------------------------------------------------------------ add audio output.
-    double preferredHardwareSampleRate = [[AVAudioSession sharedInstance] currentHardwareSampleRate];
+    double preferredHardwareSampleRate = [[AVAudioSession sharedInstance] sampleRate];
```
julapy: does `sampleRate` compile on older iOS versions - 5.0 and up?

julapy:
do you think that ofxtvOS might be too general as a name?
i mean, theres probably going to be quite a few tv operating systems out there over the years...
maybe something like ofxAppleTVOS or ofxAppleTV might be more future proof?

julapy:
do you think it will be cleaner to have a new app delegate and view controller for Apple TV?
instead of #ifdef'ing the existing ofxiOSAppDelegate and ofxiOSViewController?

addons/ofxiOS/src/video/AVFoundationVideoPlayer.m
danoli3: iOS 6.0 and up. Since oF 9.1 is iOS 7.0+ Should be all good.

danoli3:
> do you think that ofxtvOS might be too general as a name?
i mean, theres probably going to be quite a few tv operating systems out there over the years...
maybe something like ofxAppleTVOS or ofxAppleTV might be more future proof?

I think Apple have already made their stake on the name tvOS so it is unlikely to be used by other companies as that. Also by keeping with their name standard too, it will make it easier for those looking to build for it to find openFrameworks as an option too.

ofxAppletvOS could work though if it is too ambiguous. Post release though, tvOS will be standard lingo to devs though, like iOS is for iPhone / iPad.

danoli3:
>do you think it will be cleaner to have a new app delegate and view controller for Apple TV?
instead of #ifdef'ing the existing ofxiOSAppDelegate and ofxiOSViewController?

Yes it might be cleaner as we can focus more on the specifics for tvOS 9.0, however there is a lot of overlap, since tvOS is basically the same as iOS however restricted in it's SDK, it doesn't allow for older SDK references that don't exist in tvOS and deprecated functions become a real issue.

All the ifdefs currently are to get it to build and work.

addons/ofxiOS/src/video/AVFoundationVideoPlayer.m
julapy: ok, i think the first thing we need to do is move the minimum ofxiOS support from the current 5.0 to 6.0. i agree its time to move it up since iOS 9.0 has been released.
we should do this in a separate PR and link it to a relevant issue.
i'll open up the issue now...

julapy: issue for moving up ofxiOS to 6.0 is here #4431
@danoli3 is it possible to take elements from this PR to address this issue in a separate PR?

julapy:
> tvOS is basically the same as iOS however restricted in it's SDK
yeah good point. in which case it makes sense to include it in ofxiOS

how about we create a folder inside ofxiOS for tvOS specific code,
`addons/ofxiOS/src/tvOS`

and in there we would have,
`addons/ofxiOS/src/tvOS/ofxtvOSAppDelegate`
`addons/ofxiOS/src/tvOS/ofxtvOSViewController`

it will be worth making ofxtvOSAppDelegate & ofxtvOSViewController as separate from ofxiOS.
even though they are close in functionality, it will be cleaner and easier to maintain in the long run.

julapy:
i was also thinking...

with all those #ifdefs you had to add to a lot of the,
`addons/ofxiOS/src/utils` classes...

its currently an issue with ofxiOS that it references all those classes as soon as you include `ofxiOS.h`
none of those utils classes are core and should be completely optional.

so if this is cleaned up properly in ofxiOS, you then shouldn't need to add all those #ifdefs

i'll try to do this clean up soon and send a PR through.

julapy:
hey @danoli3, ping me once you've made any updates to tvOS and i'll review.
keen to help you get it into ofxiOS.

danoli3:
Awesome yeah looking at doing it later today

Issue #4414, opened 2015-10-06
ofAVFoundationPlayer Thread count leaks


armadillu:
I'm working on a project that needs to load, playback, and unload an unlimited number of videos over time. The current ofAVFoundationPlayer (OS X) creates a great number of Threads when a video is loaded / played, but it doesn't make them disappear when the video is unloaded / destroyed. This make my app reach a freeze when the thread count reaches >1024 threads... After a couple hundred videos have been played.

Digging inside the ofAVFoundationPlayer implementation, I see the assets are not released on deallocate... I don't really know what I'm doing as I lost track of Apple's video technology quite a while ago but common sense dictates to release things you are not using any more. I might be missing something and it might not be be the best order in which to release things, but this seems to fix the thread leak for me. PR coming soon.

armadillu:
On my quick test that consists of doing:

-loading a video
-playing a video for 1 sec
-deallocating it 

2000 times, the app stays at a healthy ~30 threads now. 

But I do see memory is 1.3Gb, so there's still a leak in there somewhere.

armadillu:
I'm also seeing some errors every now and then, but it doesn't seem to affect performance:

```
[12:10:23.148] <<<< VMC >>>> vmc2PostDecodeError: posting DecodeError (-12903) -- PTS was 0.160 = 4004/25025
````

**Comment by ofTheo** [comment]
I've seen the -12903 error as well. 
I believe it stands for kVTInvalidSessionErr  see: https://github.com/lionheart/openradar-mirror/issues/2619

**Comment by armadillu** [comment]
I'm going to leave the issue open for a while as it still leaks memory. I'll have a look when I have a bit of time.

**Comment by i-n-g-o** [comment]
@armadillu are you creating and destroying `ofAVFoundationPlayer` objects or are you loading/closing videos on the same object?
i am testing with one `ofAVFoundationPlayer` object loading/closing the same video-file. as you proposed (load, play for  one second, close)
the app is on stable on ~32 threads, i cannot see an increase in memory consumption.

**Comment by i-n-g-o** [comment]
also tested creating and destroying `ofAVFoundationPlayer` objects, but could not detect a memory-leak.
tested on 10.10.5.

doing it this (pseudo-code):
```
shared_ptr<ofAVFoundationPlayer> myMovie;

setup(){
    myMovie = shared_ptr<ofAVFoundationPlayer>(new ofAVFoundationPlayer());
    myMovie->load("mymovie.mov");
[...]

PR #4415, opened 2015-10-06
ofAVFoundationPlayer release used resources on dealloc

armadillu:
#4414

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -122,19 +122,33 @@ static const NSString * ItemStatusContext;
 	// destroy player
 	if(self.player != nil) {
 		[_player removeObserver:self forKeyPath:kRateKey];
-		
 		self.player = nil;
 		[_player release];
 	}
-	
+
 #if USE_VIDEO_OUTPUT
 	if (self.videoOutput != nil) {
 		self.videoOutput = nil;
 	}
-	
 	dispatch_release(_myVideoOutputQueue);
+
 #endif
-	
+
+	if(self.assetReaderVideoTrackOutput != nil) {
+		self.assetReaderVideoTrackOutput = nil;
+	}
+
+	if(self.assetReader != nil) {
+		[self.assetReader cancelReading];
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
```diff
@@ -122,19 +122,33 @@ - (void)dealloc
...
+	if(self.assetReader != nil) {
+		[self.assetReader cancelReading];
+		self.assetReader = nil;
+		self.assetReaderVideoTrackOutput = nil;
+		self.assetReaderAudioTrackOutput = nil;
+	}
+
+	if(self.asset != nil) {
+		self.asset = nil;
+	}
```
ofTheo: is this really releasing the asset? - 
I think setting it to nil is not releasing it. 

ping @julapy @danoli3  @admsyn

2bbb:
i think, if ARC is enabled then this code will release assets.
but it can call `[super dealloc];` at the last.
so, ARC is not enabled on this code.
maybe we must do call release for each objs.

PS.
why do you check nil?
i think this can be more simple like:

```
self.assetReaderVideoTrackOutput = nil;
[self.assetReader cancelReading];
self.assetReader = nil;
self.assetReaderVideoTrackOutput = nil;
self.assetReaderAudioTrackOutput = nil;
self.asset = nil;
```

armadillu:
I left those in bc I was checking for Release Counts in my debug code, but yeah we can drop most of these; not the one with self.assetReader though.

@ofTheo with ARC it will get auto-released. I really don't like the pattern and I prefer the usual alloc - release strategy, but those object haven't been allocated at creation, so they don't need a release. The release pool will release them if their RefCount is 0

ofTheo:
ah okay - I was a little confused as I didn't realize setting a strong reference to nil decreased the refcount and essentially releases the reference.  

so this is a ton of stuff which isn't getting released!
crazy. going to merge.

i-n-g-o:
`assetReader, assetReaderVideoTrackOutput, assetReaderAudioTrackOutput` and `asset` are defined as properties retaining the object. that makes the object being release when they are set to `nil`.
setting them to nil should be enough.

PR #4420, opened 2015-10-09
Updated arm6 makefile

jvcleave:
I added `$(RPI_ROOT)` to the paths to reduce redundancy. If this isn't set in the user's environment it will be blank (common case)

- made  `TOOLCHAIN_ROOT` and `GCC_PREFIX` also able to be set via environment variables


jvcleave:
I tested this with a cross-compiler I built for Jessie as well as natively on the RPi (v1) with latest Raspbian Jessie. 

I plan on writing up the cross-compiler setup as it is fairly complicated (but scriptable). The former raspberrypi/tools repo is stuck on 4.8.3 and is incompatible with Jessie.

arturoc:
mmh, this seems to be overwriting the previous settings for arch linux?

jvcleave:
I really wanted to require `TOOLCHAIN_ROOT` and `GCC_PREFIX ` being set by the user and throw an error if absent. (`RPI_ROOT ` as well but no error until the CROSS_COMPILING block) 

I tried using `$(error Please specify a TOOLCHAIN_ROOT)` but trying to print anything at this stage was giving me `recipe commences before first target. stop`

The values I did put in there are really more of a hint of what they should be.

jvcleave:
@arturoc any thoughts to move forward?

arturoc:
oh sorry completely forgot about this, i see now how it works, i'll merge it but we should include instructions on how to use it in the docs

jvcleave:
Great!

Here are some notes/instructions on how I built one for a VM on my Mac
https://gist.github.com/jvcleave/ed342fb2d41564bcdab6

I need to see if it needs adjustments for arm7

PR #4422, opened 2015-10-11
Feature/new regex example

edap:
I've added a small example of usage of `std::regex`


edap:
I've changed the song, this one should be ok

edap:
@tgfrerer the song is a popular song used in kindergarten, let me know what do you think. If you are not convinced, i can add an example with normal text.

tgfrerer:
That's great @edap, thanks! 

I've been looking at the code again, and i'm wondering why the image extractor is gone - was it broken in the first place? It's well possible that it was.

Also, but that might just me being a bit nitpicky, although it can make sense to put parsers into their own classes, for the sake of the example to be short and self-contained, it might be worth considering keeping it all within ofApp.h and ofApp.cpp. 

My thinking here is that object oriented programming is a concept worth its own example, and introducing two concepts at the same time (regex + object oriented programming) might be more than a beginner might be wishing for.

Although more advanced programmers will undoubtedly also look at these examples, I believe it's best to keep them as focused and as clear as possible.

That said, do you think there might be a chance to add a bit of mischief to the example to give it a bit of spice? Like, for example, changing the words of the songs into some simple form of [pig latin](Https://En.Wikipedia.Org/Wiki/Pig_Latin)? 

Or maybe you have another idea for something that might be easy & interesting to do with regex and these song texts? 

I'd be happy to help!

edap:
Yes, the image example was not working. 
To be honest, I find sometimes difficult to read some ofApp.cpp file in the example folder, there is too much stuff and it’s not easy to isolate what the example is about, which part of the code should be important in the context of the example. I would, for example, find complicate to explain a particle system without using a Particle class, and probably would also be more difficult to understand for someone without any object oriented programming experience.
Said that, I do not think nor that this example was so complicated to deserve a class, like in a particle system, neither that this code is “exemplar” as an example should probably be.
So I've added the methods that were in a separate class into the ofApp.cpp ;) https://github.com/edap/openFrameworks/commit/5fe1335eb78e651aac78b9831b64746fe1431a64
My priority was simply to substitute the example that was using the Poco library with one that is using the standard library

edap:
@tgfrerer any news on this front?

edap:
@tgfrerer  I don't want to put pressure on you, but it would be nice to have an example that shows how to use regex in OF without having to fall back to the poco libs. Waiting for feeback

arturoc:
@edap this won't merge anymore because of the changes people did to examples during the docsprint last week. also, right now there's some platforms in which std::regex won't work well yet. it's just old versions of gcc which should happen only on old versions of linux but perhaps it's better to move the example as a new one instead of overwriting the old one. other than that it looks good to me so if you could do that change i think it's ready to merge

Issue #4427, opened 2015-10-13
Argument order for apothecary in install-dependencies.sh


ni-cc:
Options need to be before command

PR #4428, opened 2015-10-13
fixing error in install-dependencies.sh of cdee8cb

ni-cc:
Hi,

i fixed the error #4427 in install-dependecies.sh due to the argument order for apothecary in cdee8cb. 
Cheers,
ni-cc


arturoc:
thanks!

PR #4429, opened 2015-10-13
QtCreator project files and wizard

arturoc:
This adds some project files to use openFrameworks with qtcreator under windows with mingw and linux. It also adds a qtcreator plugin which can be installed by copying the files under scripts/qtcreator/templates to the qt install:

in windows to:

```
c:\Qt\qtcreator-3.5.0\share\qtcreator\templates
```

in linux you can just run the `install_linux_template.sh` in scripts/qtcreator

That will add a new openframeworks project type so to create a new project one can just go to file > new project in qtcreator.

The project wizard only allows to add official addons by now but adding any addon afterwards from the project file config is really straight forward:

https://vimeo.com/142272907

In windows qtcreator comes by default with the clang plugin which can be enabled by doing:

```
Configuring Clang Code Model Plugin

    - Select Help > About Plugins > C++ > ClangCodeModel to enable the plugin.
    - Restart Qt Creator to be able to use the plugin.
    - Select Tools > Options > C++ > Code Model, and select the parser to use for files of each type.
    - To process pre-compiled headers before processing any project files, deselect the Ignore pre-compiled headers check box.
```

and gives full static analysis so you get errors as you type instead of having to compile.


scripts/qtcreator/templates/wizards/openframeworks/src/main.cpp
```diff
@@ -0,0 +1,13 @@
+#include "ofMain.h"
```
bilderbuchi: @arturoc can't you reuse the existing templates in `scripts/templates`? This change introduces another set of template files, in a non-obvious location (i.e. not within `scripts/templates`), which is just confusing imo.

arturoc: yes i know but the wizard needs to have the template in it's own folder to be able to find it once it's copied to the qtcreator install. 

i plan to extend this further being able to specify the OF path in a config option in qtcreator... and then it'll be possible to get the template from the standard location, use custom templates... but by now we need this here

bilderbuchi: I see, thanks.

maxruby: @arturoc.  This is brilliant. Are you planning to extend this to Qt on OSX?  It would be great to be able to use oF in both Windows and Mac Qt.

arturoc: i wasn't but it should be pretty straight forward so i'll take a look when i have a moment

maxruby: Have you had a chance to look this?  I am trying to get oF 0.9.0 running on OSX (works fine on Windows already).  Thanks.

Issue #2882, opened 2014-03-21
oF static lib product named differently when built in Xcode than from make


ghost:
I've run into this a few times and been confused but when building the oF static lib, Xcode uses a different build product name than when building from `make` :frowning:

Xcode: `libs/openFrameworksCompiled/lib/osx/openFrameworksDebug.a`
make: `libs/openFrameworksCompiled/lib/osx/libopenFrameworksDebug.a`

Should these product names be the same?

pizthewiz:
Still a bummer that `make` and the Xcode project create and use different OF static libs.

flvi0:
I would like to give this a try in order to make the oF cmake config for external projects not having to account for this, as mentioned in #4430. 

@pizthewiz @ofTheo Regarding the ordering, the ld manpage says "Both dynamic and static libraries are searched as they appear on the command line.", and also
```
-Ldir       Add dir to the list of directories in which to search for libraries.  Directories specified with -L are searched in the order they appear on the command line and before the default search path.
```
So I think we should be fine if the project/application gets the libdir ordering right.

Regarding the invasiveness towards users, we might solve this via symlinking the old name to the new one for the time being, announcing deprecation of that library naming and maybe get rid of it with the 0.9 release? I don't know if there's any scheme on this, but I guess an upcoming major release might be a point in time? What do you think?

flvi0:
Ok got it renamed and the post-build symlinking worked out. Why is the lib deployed in the bin/ folder actually?

ofTheo:
I agree with @pizthewiz here - we shouldn't rename the Xcode / IDE library names, but fix it so make generates openFrameworks.a and openFrameworksDebug.a 

There is far less potential for causing new issues with this approach.  
And although the way we are not using the lib prefix on OS X is non-standard, I think if we are going to not have libFreetype.a but FreeType.a for 3rd party libs we should also have it be the same for openFrameworks.

flvi0:
[...]

PR #4430, opened 2015-10-13
Add first version of oF CMake config (for use in external CMake proje…

flv0:
Works for me, tried adding VS/osx configs, but things need to be tested by someone else, don't know if the actual library names are identical for example.

### DIFF
--- a/cmake/openFrameworksConfig.cmake
+++ b/cmake/openFrameworksConfig.cmake
@@ -1,0 +1,24 @@
+get_filename_component(openFrameworksRoot "${CMAKE_CURRENT_LIST_DIR}/../" ABSOLUTE)
+
+# Having to make the following distinction is a result of
+# openFrameworks being not properly deployed into standard lib/include
+# paths as of yet, but this is subject to change. This does not
+# support 32 bit linux as of now.
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    if(${CMAKE_SIZEOF_VOID_P} EQUAL 8)
+        set(PLATFORM "linux64")
+    else()
+        set(PLATFORM "linux")
+    endif()
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(PLATFORM "osx")
+elseif(${WIN32})
+    set(PLATFORM "vs")
+endif()
+
+set(openFrameworks_INCLUDES
+    ${openFrameworksRoot}/libs/openFrameworks)
+set(openFrameworks_LIBRARIES
+    ${openFrameworksRoot}/libs/openFrameworksCompiled/lib/${PLATFORM}/libopenFrameworks.a)
+
+message("-- Found openFrameworks in ${openFrameworksRoot}")
--- a/cmake/openFrameworksConfig.cmake
+++ b/cmake/openFrameworksConfig.cmake
@@ -2,8 +2,7 @@ get_filename_component(openFrameworksRoot "${CMAKE_CURRENT_LIST_DIR}/../" ABSOLU
 
 # Having to make the following distinction is a result of
 # openFrameworks being not properly deployed into standard lib/include
-# paths as of yet, but this is subject to change. This does not
-# support 32 bit linux as of now.
+# paths as of yet, but this is subject to change.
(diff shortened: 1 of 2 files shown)


pizthewiz:
Building via `make` on OS X does use the _libopenFrameworksDebug.a_ naming scheme, #2882 has a few details.

procedural:
Shouldn't it search for other libraries on which openFrameworks depends? The project from your forum post:

```
set(PROJECT etudes)

project(${PROJECT})
cmake_minimum_required(VERSION 3.0)

set(openFrameworks_DIR $ENV{OPENFRAMEWORKS_ROOT}/cmake)
find_package(openFrameworks REQUIRED)

add_executable(${PROJECT} src/main.cpp)

target_include_directories(${PROJECT}
    PUBLIC ${openFrameworks_INCLUDES})

target_link_libraries(${PROJECT}
    PUBLIC ${openFrameworks_LIBRARIES})
```

Won't build the application since no OF dependencies is given to the linker...

flvi0:
@pizthewiz Ok expected that somehow, will look into getting stuff straightened out tomorrow.
@procedural Well, as of now, libopenFrameworks is linked as a static lib which simplifies stuff. It will link against the static versions of the dependant libraries, so no problem with that. When using shared objects, LD_LIBRARY_PATH will need to be set accordingly before building in order for the linker to find the dynamically linked objects. Providing a convenience script or something for that is another issue, which we might look into when it becomes necessary. The application builds and runs just fine here.

procedural:
>It will link against the static versions of the dependant libraries

No it won't. How the linker supposed to know that the application needs to be linked against libGL.so, for example?

procedural:
@flvi0 are you on Linux? Could you post here a link to a ofApp binary that runs fine for you?

flvi0:
I haven't got to that point yet, I was concerning with getting the build set up until now and it was a long day. I will keep you updated!

flvi0:
@procedural You are of course right regarding the static libraries. All resolution of symbols is performed by the loader.. Hence my initial question about the shared object build, is there a reason we are using static libraries at all? Since with a shared object build you don't need this kind of information since symbols only required as transitive dependencies (and not the application itself) will be resolved at link time of the library.

flvi0:
Either way it is possible to just pass the respective paths via the CMake config. I will see tomorrow if I find time during the day to look further at this!

flvi0:
Updated, linux only now. Have a mac available an i'm improving, just for people to test it on Linux (@procedural)

bilderbuchi:
Is the `cmake` base directory where this resides mandatory somehow? Otherwise I think it would be better _not_ to add another OF-root-level directory, and put this somewhere more appropriate. E.g. most of the makefile stuff lives in `libs/openFrameworksCompiled/project`.

flvi0:
by no means mandatory. it is placed commonly in that cmake toplevel dir by many projects, but I see it kinda makes sense to put in in libs/openFrameworksCompiled/project. will adapt accordingly, thanks.

flvi0:
updated. other wishes/complaints/ideas?

bilderbuchi:
ping @arturoc

flvi0:
where do i get libkiss from on osx?

flvi0:
Ok, works on mac os now. Not runtime tested since I'm remotely working on my mac system at work and have trouble with X11 forwarding (which I'm about to fix), don't see a reason why it should not run however. Maybe you want to give this some testing. I don't have a windows platform at my disposal.

flvi0:
It seems that the name of the openFrameworks library on mac os depends on wether I type make in the project directory of go to osx/ subfolder and build using the xcode project. Typing make names the library libopenFrameworks.so, while the xcode config misses the lib prefix. I could account for that in the config file, however I would prefer to have this straightened in the xcode project to use lib prefix as well.

pizthewiz:
Yup, that's tracked by the issue #2882 I [linked](https://github.com/openframeworks/openFrameworks/pull/4430#issuecomment-147803579) above ;0)

flvi0:
So pull request for #2882 is open as well, with a (temporary) symlink solution to account for the name change, which should be removed in some upcoming release. is there still interest in this?

PR #4432, opened 2015-10-14
removing ofxMultiTouch

julapy:
just realised that ofxiOS no longer requires the ofxMultiTouch,
as touch events were integrated into the core a while back.

ive quickly searched all OF source files and i couldn't see anything else using ofxMultiTouch apart from ofxiOS.

think it should be all good to remove.
unless anyone else has any objections.


danoli3:
Yeah nice!! :+1:

julapy:
since ofxMultiTouch was integrated into the core.
is it worth thinking about integrating ofxAccelerometer into core also?
cc @arturoc

arturoc:
thanks! didn't remember this wasn't being used anymore. 

not so sure about ofxAccelerometer, it needs a refactoring for sure and would be good to have more common code between android an ios in general but it can also be an addon without problem. 

the main reason to move multitouch into the core was that more OSs not only mobiles have mutitouch events now but i'm not so sure that the accelerometer is such a common feature accross platform so perhaps it makes sense to leave it as an addon.

i would also rename it to something like ofxOrientation or ofxSensors since accelerometer is not representative anymore of what's going on.

also @tallavi started some addons to uniformize the API between iOS and android.

julapy:
@arturoc, sounds good, thanks for clarifying.

PR #4438, opened 2015-10-15
Removing declarations without definition

### LINKED ISSUE #4426: Missing definition for functions
Using these functions results into a linker error, as there is no definition:

```
ofMesh  ofTexture::getMeshForSubsection(float, float, float, float, float, float, float, float, float) const;
ofVec2f& ofVec2f::getLimited(float max); // not const
```

chaosct:
closes #4426

### DIFF
--- a/libs/openFrameworks/gl/ofTexture.h
+++ b/libs/openFrameworks/gl/ofTexture.h
@@ -649,7 +649,6 @@ class ofTexture : public ofBaseDraws {
 	/// \param sh Subsection height within the texture.
 	void drawSubsection(float x, float y, float z, float w, float h, float sx, float sy, float sw, float sh) const;
 
-	ofMesh getMeshForSubsection(float x, float y, float z, float w, float h, float sx, float sy, float sw, float sh, bool vflipped, ofRectMode rectMode) const;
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/gl/ofTexture.h
```diff
@@ -664,7 +663,7 @@ class ofTexture : public ofBaseDraws {
 	/// \param sy Subsection y axis offset within the texture.
 	/// \param sw Subsection width within the texture.
 	/// \param sh Subsection height within the texture.
-	ofMesh getMeshForSubsection(float x, float y, float z, float w, float h, float sx, float sy, float sw, float sh) const;
+	ofMesh getMeshForSubsection(float x, float y, float z, float w, float h, float sx, float sy, float sw, float sh, bool vflipped, ofRectMode rectMode) const;
```
bilderbuchi: do we need documentation for those two parameters in the doxygen code above? @arturoc

arturoc:
would be good to have them but i'm ok merging it as it is, this is mostly used internally by the renderers when rendering a texture.

@chaosct: if you want to add the comments, vflipped returns a mesh that takes into account the flipped state in OF and rectMode creates the mesh taking x,y as the center or the top left corner

chaosct:
here you go

arturoc:
thanks!

PR #4440, opened 2015-10-15
Fix assimp anim load

### LINKED ISSUE #4337: ofxAssimpModelLoader not cleaning up properly
When running the assimpExample, I noticed that it crashes whenever changing models (press keys 1-5).

It looks like ofxAssimpModelLoader is not clearing its resources properly in clean():
 * The `scene` shared pointer is not `reset()`, so calls like [this](https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp#L48-L50) won't work.
 * I might be wrong here, but I don't think testing for `if (scene != NULL) { ... }` is valid since it's a shared pointer, I believe the [correct syntax](http://www.cplusplus.com/reference/memory/shared_ptr/operator%20bool/) is simply `if (scene) { ... }`.
 * Quickly glancing at the [assimp documentation](http://assimp.sourceforge.net/lib_html/cimport_8h.html#ab6085941982ef6158f3aa8be87470602), it looks like `aiPropertyStore` needs to be "manually destroyed using the [aiReleasePropertyStore](http://assimp.sourceforge.net/lib_html/cimport_8h.html#ac7e29d2f3bd4366bead505ed53f07108) API function", which is not happening.

I tried fixing these issues but I'm still getting crashes in the example, so maybe someone with better assimp knowledge can take a look?

My setup is OF master / Windows 10 / VS 2015.

tgfrerer:
Addresses a few minor issues with the assimp loader. 

* Mainly fixes an crashing issue with animations not being counted properly
* Brings forward declarations in line with assimp library objects
* More explicit deletion of scene object, to secure assimp has freed the old scene when new scene is loaded.

More details in individual commits. 

Closes #4337

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpMeshHelper.h
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpMeshHelper.h
@@ -11,7 +11,7 @@
 #include <assimp/postprocess.h>
 #include "ofxAssimpTexture.h"
 
-class aiMesh;
+struct aiMesh;
 
 class ofxAssimpMeshHelper {
 
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
@@ -280,6 +280,12 @@ void ofxAssimpModelLoader::loadGLResources(){
         meshHelper.validCache = true;
         meshHelper.hasChanged = false;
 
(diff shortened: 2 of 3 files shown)


tgfrerer:
Seems that Travis times out when not receiving input for an `apt-get`.

arturoc:
yes, i've seen that before but not sure why it's happening since nothing has changed in the scripts

PR #4442, opened 2015-10-15
iOS Fixes Bug with AVSoundPlayer 

### LINKED ISSUE #4441: iOS AVSoundPlayer prevented from playing/killed by AVAssetReader
I was testing the examples for any issues for iOS and noticed this one.

The AVSoundPlayer class (AVAudioSession) is very sensitive to if the AVAssetReader is loading something (say AVFoundationVideoPlayer or something else on the system). 
It basically kills the sound player.

I found some documentation online of the same issue:
http://thebesthacker.com/question/avassetreader-kills-playback-in-avaudioplayer.html

With a easy mixer solution (iOS 3+)
Tested and it is a good fix. 
Similar code was in the older iOS Sound Stream class (likely for the same reason).

Keen for 0.9 to be honest.

danoli3:
Fixes #4441

### DIFF
--- a/addons/ofxiOS/src/sound/AVSoundPlayer.m
+++ b/addons/ofxiOS/src/sound/AVSoundPlayer.m
@@ -25,6 +25,18 @@
     return self;
 }
 
+- (void) setupSharedSession {
+	static BOOL audioSessionSetup = NO;
+	if(audioSessionSetup) {
+		return;
+	}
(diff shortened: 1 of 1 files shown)


julapy:
ok cool, one thing...
inside the `setupSharedSession` method you created,
can you add some comments in there to explain why that code was added.

danoli3:
Sure pushing now

Issue #4290, opened 2015-08-24
ios issues with 0.9.0 examples


julapy:
ive gone through a fixed up the smaller issues for all the ios examples.
below are remaining issues which are a little bit more involved,
including outstanding TODO items that will need to be fixed for 0.9.0 release.

* iosCustomSizeExample is buggy.
* iosExternalDisplayExample needs to be tested properly with an external display.
* ~~iosNativeExample reveals a bug with frame number not resetting~~ FIXED
* iosOrientationExample is broken.
* opencvExample needs AssetsLibrary.framework added for example to compile
* opencvFaceExample needs AssetsLibrary.framework added for example to compile
* oscReceiverExample runs but needs testing with OSC data
* oscSenderExample runs but needs testing with OSC data
* videoGrabberExample probably need some kind of message that it doesn’t run on the simulator.

julapy:
@ofZach,
the `opencvExample` and `opencvFaceExample` examples require `AssetsLibrary.framework` to be added by the PG to the xcode project.
i seem to remember this being fixed for osx...
can you comment on how/if this works the same for ios.
cheers.

arturoc:
@julapy if you add:

```make
ios:
    ADDON_FRAMEWORKS = AssetsLibrary
```

not sure now if you need to add the .framework

to addon_config.mk in the ofxOpenCv addon it should do the right thing, can you give it a try? i think there's a specific hack for this or something else related to the opencv addon in the PG code, but better to solve it through the config file.

ofZach:
@julapy I'll take a look, frameworks (system level) should work via the method arturo is describing now with the latest PG (I tested it against a different addon, I think ofxMidi, but I'll check this with opencv shortly)

arturoc:
> iosNativeExample reveals a bug with frame number not resetting.

in 0.9 the frame number is not hold by the window any more but by the main loop so recreating the window won't reset it. you can do the initialization more explicit by using a main loop like:

```
ofGLESSettings settings;
auto window = std::make_shared<ofAppiOSWindow>();
window->setup(settings);
auto mainloop = std::make_shared<ofMainLoop>();
auto app = std::make_shared<ofApp>();
mainloop->run(window, app);
ofSetMainLoop(mainloop);
ofRunMainLoop();
```

julapy:
thanks @arturoc, ill look into implementing ofxiOS in this way.

julapy:
hey @arturoc,
im having

[...]

PR #4445, opened 2015-10-17
Fix for iOS OpenCV missing AssetsFramework

danoli3:
Fix for AssetsFramework work not being built into the project generator builds for iOS OpenCV examples


danoli3:
https://github.com/openframeworks/openFrameworks/issues/4290

https://github.com/openframeworks/projectGenerator/issues/58

PR #4450, opened 2015-10-18
Fix for iOS OpenCV missing AssetsLibrary*

danoli3:
#4290

openframeworks/projectGenerator#58


danoli3:
Okay tested and working. Thanks @ofZach

PR #4453, opened 2015-10-19
First iteration in converting to uniform indentation/formatting

flvi0:


bilderbuchi:
I will close this, as we would format our code with uncrustify, and all style changes at once. I would however welcome a tested PR with the relevant changes for the uncrustify [config file](https://github.com/openframeworks/openFrameworks/blob/master/scripts/dev/style/openFrameworks_style.cfg) (without the formatting changes to the rest of the code).

PR #4454, opened 2015-10-19
Fix an issue with ofShader compilation error reports

tgfrerer:
Changes in rev. b2a33da2d prevented shader source code to be
quoted when GLSL compilation was not successful.

Control would return false before the shader source code was
cached internally. This gave `checkShaderInfoLog()` no
source file to compare the error log line numbers to.

This also fixes two other issues:

+ makes sure to retain the shader GPU id if a shader program
object was allocated successfully, before returning control

+ fixes an offset-by-one issue with the reported GLSL shader
compilation error line number


arturoc:
i was wondering today why this wasn't working. thanks!!

Issue #3404, opened 2014-11-08
ofVideoPlayer linux / Error on buffer size for some videos when setting OF_PIXELS_NATIVE


PetrosKataras:
Hi,

I am with latest master under xubuntu 14.04.

If I set the player to use the default pixel format I get this error with a video captured from a GoPro camera :

[notice ] ofGstVideoPlayer: allocating with 860x646 I420
[ error ] ofGstVideoUtils: buffer_cb(): error on new buffer, buffer size: 834632!= init size: 833340
[ error ] ofGstUtils: gstHandleMessage(): embedded video playback halted for plugin, module qtdemux0  reported: GStreamer encountered a general stream error.
[warning] ofGstUtils: didn't received EOS in 5s, closing pipeline anyway
[ error ] ofGstVideoUtils: buffer_cb(): error on new buffer, buffer size: 834632!= init size: 833340
[ error ] ofGstUtils: gstHandleMessage(): embedded video playback halted for plugin, module qtdemux1  reported: GStreamer encountered a general stream error.

This error does not appear if don't set the pixels to the default format and everything renders fine.

It also does not appear for the fingers.mov for example, even if I set the default format.

In addition it seems to me that this error is actually misleading, since if I force the player to render on the default format I will get a result that seems to me correct i.e the movie seems to render properly on the default colorspace .

EDIT: Not sure about the above comment anymore since I just tried the programmable renderer to see if the colorspace conversion would work properly but it doesn't.. The video is being rendered like this:

![testvideo](https://cloud.githubusercontent.com/assets/1078053/4964010/b12ca042-6742-11e4-9684-a8396f77125e.png)

I am using the legacy gl renderer and not the programmable one . 

Here are some info of the video file that might be helpful:

General
Complete name                            : test.mp4
Format                                   : MPEG-4
Format profile                           : Base Media / Version 2
Codec ID                                 : mp42
File size                                : 265 MiB
Duration                                 : 1mn 13s
Overall bit rate                         : 30.1 Mbps
Encoded date                             : UTC 2014-10-09 10:19:24
Tagged date                              : UTC 2014-10-09 10:19:28
©TIM                                     : 00;00;00;00
©TSC                                     : 2997
©TSZ                                     : 50

Video
ID                                       : 1
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High@L5.1
Format settings, CABAC                   : Yes
Format settings, ReFrames                : 3 frames
[...]

PR #4456, opened 2015-10-20
dirty fix for #3404

aspeteRakete:
new setFromAlignedPixels gets taken when PixelFormat == OF_PIXELS_I420.

### DIFF
--- a/libs/openFrameworks/graphics/ofPixels.cpp
+++ b/libs/openFrameworks/graphics/ofPixels.cpp
@@ -336,7 +336,7 @@ void ofPixels_<PixelType>::setFromAlignedPixels(const PixelType * newPixels, int
 	int channels = channelsFromPixelFormat(_pixelFormat);
 	if(channels==0) return;
 
-	if(width*channels==strides[0]){
+	if(width==strides[0] && width/2==strides[1] && width/2==strides[2]){
 		setFromPixels(newPixels,width,height,_pixelFormat);
 		return;
 	}
--- a/libs/openFrameworks/graphics/ofPixels.cpp
+++ b/libs/openFrameworks/graphics/ofPixels.cpp
@@ -332,6 +332,40 @@ void ofPixels_<PixelType>::setFromAlignedPixels(const PixelType * newPixels, int
 }
 
 template<typename PixelType>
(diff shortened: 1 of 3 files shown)


arturoc:
thanks, can you add a check for the format and number of samples, something like:

```
switch(pixelFormat){
case OF_PIXELS_I420:
if(samples.size() != 3){
    ofLogError("ofPixels") << "number of planes for I420 should be 3";
}
...
break;
default:
    ofLogError("ofPixels") << "setFromAlignedPixels with planes strides: pixel format not supported yet";
break;
}
```

also if we could have an option to send anything without planes to the simpler setFromAligned liek:

```
case OF_PIXELS_RGB:
case OF_PIXELS_RGBA:
case OF_PIXELS_GRAY:
case OF_PIXELS_GRAY_ALPHA:

File: libs/openFrameworks/graphics/ofPixels.cpp [hunk 1]
```diff
@@ -332,6 +332,40 @@ void ofPixels_<PixelType>::setFromAlignedPixels(const PixelType * newPixels, int
 }
 
 template<typename PixelType>
+void ofPixels_<PixelType>::setFromAlignedPixels(const PixelType * newPixels, int width, int height, ofPixelFormat _pixelFormat, std::vector<int> strides) {
+	int channels = channelsFromPixelFormat(_pixelFormat);
+	if(channels==0) return;
+
+	if(width*channels==strides[0]){
```
arturoc: this doesn't make much sense for I420 since it doesn't have channels. something like:

```
if(width==strides[0] && width/2==strides[1] && width/2==strides[2])
```

would be a better check

aspeteRakete:
i added the stride size checks. Should the format and number checks go into ofPixels::setFromAlignedPixels or ofGstUtils::process_sample? How do I get samples.size()? Or do you mean strides.size() ?

aspeteRakete:
Ok i assume you meant ```strides.size() != 3```
implemented that.

arturoc:
yeah sorry i meant strides, looks great now thanks! i'll let the tests run and then merge it

arturoc:
thanks!

PR #4461, opened 2015-10-22
Fixes ofQuaternion product. closes #4458

### LINKED ISSUE #4458: Quaternion Product implementation is incorrect
In ofQuaternion.cpp [here](https://github.com/openframeworks/openFrameworks/blob/ac32f4d21ec54577298f5f9ad7636623ee3c441d/libs/openFrameworks/math/ofQuaternion.h), the second line of the multiplication operator is
    
     rhs._v.w*_v.y - rhs._v.x*_v.z + rhs._v.y*_v.w + rhs._v.z*_v.x,
    
but it should be
    
     rhs._v.w*_v.y + rhs._v.x*_v.z + rhs._v.y*_v.w - rhs._v.z*_v.x,
    
according to the definition. (See [this page](http://people.csail.mit.edu/bkph/articles/Quaternions.pdf) for the definition)

It's just one minus sign, but the net result is that multiplying two quaternions and then retrieving the resultant's angle-and-axis will always flip the sign on the y-axis.

I'm guessing that this will fix: https://github.com/openframeworks/openFrameworks/issues/785

Poofjunior:
correctly implements quaternion product. (sign error in the previous implementation.)

closes https://github.com/openframeworks/openFrameworks/issues/4458

### DIFF
--- a/libs/openFrameworks/math/ofQuaternion.h
+++ b/libs/openFrameworks/math/ofQuaternion.h
@@ -394,7 +394,7 @@ const ofQuaternion ofQuaternion::operator*(const ofQuaternion& rhs) const {
 //----------------------------------------
 ofQuaternion& ofQuaternion::operator*=(const ofQuaternion& rhs) {
     float x = rhs._v.w * _v.x + rhs._v.x * _v.w + rhs._v.y * _v.z - rhs._v.z * _v.y;
-    float y = rhs._v.w * _v.y - rhs._v.x * _v.z + rhs._v.y * _v.w + rhs._v.z * _v.x;
+    float y = rhs._v.w * _v.y + rhs._v.x * _v.z + rhs._v.y * _v.w - rhs._v.z * _v.x;
     float z = rhs._v.w * _v.z + rhs._v.x * _v.y - rhs._v.y * _v.x + rhs._v.z * _v.w;
     _v.w = rhs._v.w * _v.w - rhs._v.x * _v.x - rhs._v.y * _v.y - rhs._v.z * _v.z;
     
--- a/libs/openFrameworks/math/ofQuaternion.h
+++ b/libs/openFrameworks/math/ofQuaternion.h
@@ -178,7 +178,8 @@ public:
     inline const ofQuaternion operator -(const ofQuaternion& rhs) const;    ///< Binary subtraction
     inline const ofQuaternion operator -() const;                           ///< returns the negative of the quaternion. calls operator -() on the Vec4
     
-    inline ostream& operator<<(ostream& os);
+    friend inline ostream& operator<<(ostream& os, const ofQuaternion& q);
+
     inline istream& operator>>(istream& is);
     
     /// \}
@@ -199,8 +200,8 @@ ofQuaternion::ofQuaternion() {
 
 
 //----------------------------------------
-ostream& ofQuaternion::operator<<(ostream& os) {
-    os << _v.x << ", " << _v.y << ", " << _v.z << ", " << _v.w;
+ostream& operator<<(ostream& os, const ofQuaternion& q) {
+    os << q._v.x << ", " << q._v.y << ", " << q._v.z << ", " << q._v.w;
     return os;
 }
 
--- a/libs/openFrameworks/math/ofQuaternion.h
+++ b/libs/openFrameworks/math/ofQuaternion.h
@@ -385,7 +385,7 @@ ofQuaternion& ofQuaternion::operator *=(float rhs) {
 //----------------------------------------
 const ofQuaternion ofQuaternion::operator*(const ofQuaternion& rhs) const {
     return ofQuaternion(rhs._v.w*_v.x + rhs._v.x*_v.w + rhs._v.y*_v.z - rhs._v.z*_v.y,
-                        rhs._v.w*_v.y - rhs._v.x*_v.z + rhs._v.y*_v.w + rhs._v.z*_v.x,
+                        rhs._v.w*_v.y + rhs._v.x*_v.z + rhs._v.y*_v.w - rhs._v.z*_v.x,
                         rhs._v.w*_v.z + rhs._v.x*_v.y - rhs._v.y*_v.x + rhs._v.z*_v.w,
                         rhs._v.w*_v.w - rhs._v.x*_v.x - rhs._v.y*_v.y - rhs._v.z*_v.z);
 }


bilderbuchi:
I have to admit I only skim-read, but isn't `operator*=` also affected by the same problem?

arturoc:
it's kind of strange that we've been working with this for so much time without problem. I'm looking at other implementations and the signs seem to be the same as OF uses:

OF:
```
wx + xw + yz - zy 
wy - xz + yw + zx
wz + xy - yx + zw
ww - xx - yy - zz
```

nalgebra (https://github.com/sebcrozet/nalgebra/blob/master/src/structs/quat.rs#L143-L146)
The order of the rows is swapped but the signs are the same as in the OF impl
```
ww - xx - yy - zz   
wx + xw + yz - zy
wy - xz + yw + zx
wz + xy - yx + zw
```

Paul Bourke (http://paulbourke.net/fractals/quatjulia/)
This seems exactly the same as the nalgebra implementation
ww - xx - yy - zz   
rx + xw + yz - zy
wy + yw + zx - xz
wz + zw + xy - yx

http://de.mathworks.com/help/aeroblks/quaternionmultiplication.html
This one is completely different in the signs and order?
```
xx - yy - zz - ww   
xy + yx - zw + wz
xz + yw + zx - wy
xw - yz + zy + wx
```

memo:
Also OSG (which is where OF's quat implementation is from) is showing the same as OF
https://github.com/openscenegraph/osg/blob/master/include/osg/Quat#L194

bilderbuchi:
> it's kind of strange that we've been working with this for so much time without problem

yeah, that's also what I thought...

Poofjunior:
nalgebra's implementation is correct. Our's is different. Here's why:

3rd line of nalgebra's implementation (new y-term):
    
    self.w * right.j - self.i * right.k + self.j * right.w + self.k * right.i
    
That's the same as
    
    self.w * right.y - self.x * right.z + self.y * right.w + self.z * right.x
    
which is the same as
    
    _v.w * rhs._v.y - _v.x * rhs._v.z + _v.y * rhs._v.w + _v.z * rhs._v.x
    
which is the same as
    
    rhs_v.w*rhs._v.y  + rhs._v.x*_v.z + rhs._v.y*_v.w - rhs._v.z*_v.x
    
Our implementation in OF, however is (2nd line, new y-term)
    
    rhs._v.w*_v.y - rhs._v.x*_v.z + rhs._v.y*_v.w + rhs._v.z*_v.x
    
which is __not__ the same.

Paul Borke's implementation for the y-term looks like this:
    
    r1 b2 + b1 r2 + c1 a2 - a1 c2
    
which is the same as
    
    w * rhsY + y * rhsW + z * rhsX - z * rhsZ
    
which is the same as
    
    rhsW * y + rhsX * z + rhsY * w - rhsZ * z
    
which is the same as
    
    rhs._v.w*_v.y + rhs._v.x*_v.z + rhs._v.y*_v.w - rhs._v.z*_v.x
    
which is correct.
Our OF implementation, however (for the y-term) is
    
    rhs._v.w*_v.y - rhs._v.x*_v.z + rhs._v.y*_v.w + rhs._v.z*_v.x
    
which is, again, both different and incorrect.

(I was too lazy to do the last one from mathworks.)

I completely admit that with the different conventions and organization of the terms, it's really hard to compare them fairly, so I hope this clears it up.

> it's kind of strange that we've been working with this for so much time without problem.

I was a little weirded-out too, but I can take a guess as to why it's been lurking around for so long.
 * Quaternions are common among 3d animators, but they're less intuitive so many beginners and artists might seem less likely to use them because they aren't intuitive.
 * The only thing that's wrong is the product of two multiplied quaternions. Multiplying two quaternions together is handy for getting the net rotation, but I can see people mostly just using quaternions to encode a single rotation, not chaining them together.
 * The net result of this bug is that the y-axis of the represented vector has a sign error. If the results aren't what you'd expect, I can imagine people blaming themselves first and tacking on another negative sign.

Poofjunior:
@bilderbuchi, yes, *= needs to be fixed too. Adjusting my fork now...

arturoc:
I'm trying this:

With the current version and this code:

```cpp
ofQuaternion q1(30,ofVec3f(1,0,0));
ofQuaternion q2(20,ofVec3f(0,0,1));
auto q3 = q1 * q2;
for(auto & v: box.getMesh().getVertices()){
    v = v * q3;
}
```

screenshot from 2015-10-23 10-58-28

Only multiplying the vertices by q1:

screenshot from 2015-10-23 10-58-45

Only multiplying the vertices by q2:

screenshot from 2015-10-23 10-59-00

Multiplying the vertices by q1 and then by q2, as you can see you get the same as multiplying the vertices by q3:

screenshot from 2015-10-23 11-00-35


With your version and the original code, multiplying all the vertices by q3 the result is not the same, it seems to add an extra rotation in y:

screenshot from 2015-10-23 10-57-07


I don't think this fix is correct but can you put an small example in which you can show how it's behaving incorrectly?

bilderbuchi:
I guess doing a Quaternion multiplication and checking if it agrees with, e.g. [Wolfram Alpha](https://www.wolframalpha.com/input/?i=Quaternion%28%2B30%2C1%2C0%2C0%29+*+Quaternion%2820%2C0%2C0%2C1%29) could make sure we calculate correctly?

Poofjunior:
Aye, thanks for taking a closer look. My changes still hold after plugging it into WolframAlpha. 
Here's the breakdown:
I wrote some quick "box-rotating" code also and printed the angle and axis from getRotate. 
    
    int size = 100;    
    ofBoxPrimitive box(size, size, size);    
    ofQuaternion q1(30,ofVec3f(1,0,0));    
    ofQuaternion q2(20,ofVec3f(0,0,1));    
    auto q3 = q1 * q2;
    float angle;    
    ofVec3f axis;    
    q3.getRotate(angle, axis);    
    std::cout << "angle: " << angle << std::endl;    
    std::cout << "axis: " << axis << std::endl;      
                                                   
    for(auto & v: box.getMesh().getVertices()){    
        v = v * q3;                                
    }                  
         
    box.drawWireframe();
    
The result is the same that I noted on the issue: sign error on the y-axis.

original:
    
    angle: 35.9277
    axis: 0.826436, 0.145723, 0.543846
    
modified (my changes)
    
    angle: 35.9277
    axis: 0.826436, -0.145723, 0.543846
    
Doing the same math with [WolframAlpha](https://www.wolframalpha.com/input/?i=quaternion+0.965926+%2B+0.258819i+%2B+0j+%2B+0k+multiplied+by+0.984808+%2B+0i+%2B+0j+%2B+0.173648k) with 

q1 = (w = 0.965926, x = 0.258819, y = 0, z = 0) same as q1(30,ofVec3f(1,0,0));
q2 = (w = 0.984808, x = 0, y = 0, z = 0.173648) same as q1(20,ofVec3f(0,0,1));
yields the second of the two.
(Keep in mind that WolframAlpha doesn't drop the factor of sine(theta/2) on the i,j,k axis, so the vector is scaled.)

Does this hold as proof enough?

Poofjunior:
One last fix. the "print" function needs to be a friend function and outside the quaternion class; otherwise it prints junk. I've made the fix in my latest commit, but can anyone else confirm that the original is printing nonsense?

arturoc:
I still don't understand why the visual result seems correct with the current implementation, or at least the same as applying the rotations separately while it looks different with your changes. Can you explain why multiplying the box vertices by q3 yields a different result than multiplying the vertices by q1 and then by q2 when using your changes.

You can test it with:

```cpp
#include "ofMain.h"

class ofApp: ofBaseApp{
void setup(){
    ofQuaternion q1(30,ofVec3f(1,0,0));
    ofQuaternion q2(20,ofVec3f(0,0,1));
    auto q3 = q1 * q2;
    for(auto & v: box.getMesh().getVertices()){
        v = v * q3;
    }
    for(auto & v: box2.getMesh().getVertices()){
        v = v * q1;
        v = v * q2;
    }
    ofEnableLighting();
    light.enable();
    light.setPosition(-100,-100,100);
    ofEnableDepthTest();
    show1=true;
}

void draw(){
    cam.begin();
    ofSetColor(ofColor::red);
    if(show1){
        box.drawFaces();
    }else{
        box2.drawFaces();
    }
    cam.end();
    if(show1){
        ofDrawBitmapString("q3",20,20);
    }else{
        ofDrawBitmapString("q1 then q2",20,20);
    }
}

void keyPressed(int key){
    show1 = !show1;
}
    ofBoxPrimitive box;
    ofBoxPrimitive box2;
    ofEasyCam cam;
    ofLight light;
    bool show1;
};

int main( ){
    ofSetupOpenGL(1024,768,OF_WINDOW);			// <-------- setup the GL context

    // this kicks off the running of my app
    // can be OF_WINDOW or OF_FULLSCREEN
    // pass in width and height too:
    ofRunApp(new ofApp());

}
```

in main.cpp and no ofApp.cpp or .h

sorry to be so picky about this but it looks like a change that could break lots of code and we are about to do a release so don't want to merge it until i'm fully sure that is correct.

arturoc:
I've also added a new unit test to check this: https://github.com/openframeworks/openFrameworks/blob/master/tests/math/quaternionTests/src/main.cpp

it checks that the multiplication of a vector by the multiplication of 2 quaternions is the same as the multiplication of the vector by one quaternion and then by the other.

as you can see the test is passing in current master:
https://travis-ci.org/openframeworks/openFrameworks/builds/87173292

but if you merge your branch with current master, so it runs the tests, you'll see how it fails on any combination of rotations on x and z with:

```
[ error ] value1: 36.6025, 68.3013, 154.904
[ error ] value2: 68.3013, 81.6987, 136.603
[ error ] v*q12 == (v*q1)*q2 failed 
[ error ] /home/arturo/Code/openFrameworks/tests/math/quaternionTests/src/main.cpp: 28
[notice ] v*q13 == (v*q1)*q3 passed
[notice ] v*q23 == (v*q2)*q3 passed
[ error ] value1: 68.3013, 81.6987, 136.603
[ error ] value2: 36.6025, 68.3013, 154.904
[ error ] v*q21 == (v*q2)*q1 failed 
[ error ] /home/arturo/Code/openFrameworks/tests/math/quaternionTests/src/main.cpp: 31
[notice ] v*q31 == (v*q3)*q1 passed
[notice ] v*q32 == (v*q3)*q2 passed
[ error ] 2/6 tests failed
```

ofTheo:
@Poofjunior - totally puzzled with this PR :) 
Appreciate you bringing this up - we want to make sure there isn't a critical bug in something used as heavily as ofQuaternion in OF. 

What is also interesting is if you take @arturoc 's example and add this print out to the end of ofApp::setup():

    float angle = 0;
    ofPoint axis;
    q3.getRotate(angle, axis);
    std::cout << "q3 angle: " << angle << std::endl;
    std::cout << "q3 axis: " << axis << std::endl;

    std::cout << "q1 euler " << q1.getEuler() << std::endl;
    std::cout << "q2 euler " << q2.getEuler() << std::endl;
    std::cout << "q3 euler " << q3.getEuler() << std::endl;


 to the end of setup you get with your PR changes:

```
q3 angle: 35.9277
q3 axis: 0.826436, -0.145723, 0.543846
q1 euler 30, 0, 0
q2 euler 0, 0, 20
q3 euler 31.5667, -10.3141, 17.2294
```

and without the PR changes you get:

```
q3 angle: 35.9277
q3 axis: 0.826436, 0.145723, 0.543846
q1 euler 30, 0, 0
q2 euler 0, 0, 20
q3 euler 30, 0, 20
```

To me the `q3 euler 30, 0, 20` looks correct compared to `q3 euler 31.5667, -10.3141, 17.2294`

There is absolutely no rotation around Y when you multiply:
    ofQuaternion q1(30,ofVec3f(1,0,0));
    ofQuaternion q2(20,ofVec3f(0,0,1));
    ofQuaternion q3 = q1 * q2;

So seeing -10.3141 in the y value of the Euler result is troubling.

flvi0:
On 10/24/2015 09:57 AM, arturo wrote:
> Can you explain why multiplying the box vertices by q3 yields a
> different result than multiplying the vertices by q1 and then by
> q2 when using your changes.
> ...
>     for(auto & v: box.getMesh().getVertices()){
>         v = v * q3;
>     }

Hi, just want to note that rotating a vector by a quaternion is performed by multiplying the quaternion from the left and the complex conjugate from the right. Maybe this is done by the operator* implementation, have not looked into the code, but this could be a source of confusion.

Best

flvi0:
On 10/24/2015 09:57 AM, arturo wrote:
> Can you explain why multiplying the box vertices by q3 yields a
> different result than multiplying the vertices by q1 and then by
> q2 when using your changes.
> ...
>     for(auto & v: box.getMesh().getVertices()){
>         v = v * q3;
>     }

Hi, just want to note that rotating a vector by a quaternion is performed by multiplying the quaternion from the left and the complex conjugate from the right. Maybe this is done by the operator* implementation, have not looked into the code, but this could be a source of confusion.

Edit:

```
ofVec3f ofQuaternion::operator*(const ofVec3f& v) const {
    // nVidia SDK implementation
    ofVec3f uv, uuv;
    ofVec3f qvec(_v.x, _v.y, _v.z);
    uv = qvec.getCrossed(v);    //uv = qvec ^ v;
    uuv = qvec.getCrossed(uv);    //uuv = qvec ^ uv;
    uv *= (2.0f * _v.w);
    uuv *= 2.0f;
    return v + uv + uuv;
}
```

There are some things I do not understand, however. Thiis is actually inside class ofQuaternion, so I thought the parameter passed should be RHS of the operation. However in the code above, it is vector * quat, which I don't find an operator* implementation for. What am I missing here?

Also I'm trying to figure out what the above code actually does, did not know one could implement it like this. Can someone explain?

For reference, here is the implementation of Quaternion::Rotate and operator* (quat, vec) from the ViSTA VR toolkit:

```
inline VistaVector3D VistaQuaternion::Rotate( const VistaVector3D& v3Vector ) const
{
	VistaQuaternion qReturn = (*this) * v3Vector * GetComplexConjugated();
	return VistaVector3D( qReturn[0], qReturn[1], qReturn[2] );
}

VistaQuaternion operator*( const VistaVector3D& v3Vec, const VistaQuaternion& qQuat )
{
	// like quaternion rotation, with q2 = ( v3, 0 )
	return VistaQuaternion( v3Vec[0] * qQuat[3] + v3Vec[1] * qQuat[2] - v3Vec[2] * qQuat[1],
							v3Vec[1] * qQuat[3] + v3Vec[2] * qQuat[0] - v3Vec[0] * qQuat[2],
							v3Vec[2] * qQuat[3]	+ v3Vec[0] * qQuat[1] - v3Vec[1] * qQuat[0],
							-v3Vec[0] * qQuat[0] - v3Vec[1] * qQuat[1] - v3Vec[2] * qQuat[2]); 
}
```

[reposted since it seems github will not parse markdown when replying by mail?...]

Best

Poofjunior:
Thanks for the sample code! That definitely made replying a lot faster!

> it checks that the multiplication of a vector by the multiplication of 2 quaternions is the same as the multiplication of the vector by one quaternion and then by the other.

Ok, although that would __seem__ correct, it's actually the opposite. It turns out that a combined quaternion, q3 = q1 * q2 represents a rotation first by q2; then by q1. The details are on [Math Stack Exchange](http://math.stackexchange.com/questions/331539/combining-rotation-quaternions), but in a nutshell, let's say I've got two quaternions, __q__ and __p__. I first rotate my vector v by __p__ and then by __q__. This works out to 

a = __p__ v __p*__
b = __q__ a __q*__ =  __qp__ v __p* q*__ = ...proof... = __(qp)__ v __(qp)*__

(note: I didn't go through the full proof here, just the result)
If we wanted a single quaternion that represented a rotation first by __p__ and then by __q__, it'd be

__r__ = __qp__

With my implementation (the one in the PR), I get the image you'd expect from the combined rotations:

combined_rotation_operator


One thing to keep in mind is that multiplication with quaternions is __not__ commutative, just like rotating first about one axis; then about another axis.

Finally, @flvi0 is right. Rotating a vector by a quaternion is done by left and right multiplying that vector by the quaternion and it's conjugate respectively, like this:

__a__ = __q__ v __q*__

I'm assuming that this is the net result of the overloaded multiply operater in the code @flvi0 brought up:
    
    ofVec3f ofQuaternion::operator*(const ofVec3f& v) const {
        // nVidia SDK implementation
        ofVec3f uv, uuv;
        ofVec3f qvec(_v.x, _v.y, _v.z);
        uv = qvec.getCrossed(v);    //uv = qvec ^ v;
        uuv = qvec.getCrossed(uv);    //uuv = qvec ^ uv;
        uv *= (2.0f * _v.w);
        uuv *= 2.0f;
       return v + uv + uuv;
    }
    
At least, our output suggests that this code basically does __a__ = __q__ v __q*__, although I disagree with overloading the multiply operator and prefer a member function like ViSTA's:

    ofVec3f ofQuaternion::rotate(const ofVec3f& v) ...
    
I can live with it for now until we get the nuts-and-bolts of the class working correctly.

Yes, I realize that this goes against our initial intuition as to how we'd want these rotation operators to work. (This is my new hunch as to why the bug has been around for so long: because the corrected implementation, while mathematically correct, goes against our intuition about the order of combining rotation operators by multiplying them.)

Finally, yes; I know this is going to break some code. But in the end, I think it's worth using quaternions in a way that can be referenced against a textbook or someone's academic notes rather than propagate an incorrect understanding about how they fwork. I'll be honest and say that the only reason I discovered this bug is because I was writing my own [quaternion library](https://github.com/Poofjunior/QPose/blob/master/src/quaternion.hpp) and was using the OpenFrameworks implementation to check it when I discovered that the OF implementation had a flipped sign error.

arturoc:
mmh ok, then this might have to do with the order of multiplication being inverted in OF. it's the same for matrices, in glsl for example you would do:

```
vec4 v = matrix * v1;
```

but in OF (and OSG) where we took the implementation from the way to do that same multiplication is:

```
ofVec4f v = v1 * matrix;
```

i guess the same happens with quaternions, you should do the multiplication with the reverse order, that's in the same order as you want to apply the rotations for it to work the way you are expecting

we have plans to switch all our vector math to glm in the near future so i don't think we should fix it now since there's a lot of code out there using ofQuaternion and expecting it to work as it does now. once we switch to glm there will be different class names which would make it easier to detect this and fix it otherwise it would be a really hard bug to solve if we just swap the order in which you multiply quaternions to get a rotation

flvi0:
> At least, our output suggests that this code basically does a = q v q*, although I disagree with overloading the multiply operator and prefer a member function like ViSTA's

Full acknowledgement. We deliberately chose to use the rotate member to make that notion explicit for users of the library. The fact also being that both a vec3 and a quat are 4-element vectors, and in ViSTA they are actually derived from the same base class. There operator* is defined as the scalar product between the two vectors, which is intuitive. Operator overloading with "additional semantics" is something one should be very careful with, since it will suggest users a wrong idea of what is actually happening, which forces errors. Just see the post about GLM, that's a valid solution of course as well. If we want to provide abstractions for this functionality ourselves, we should take care that it contains no such bugs or misconceptions, since dealing with the transform math will produce enough head-scratching with people anyways, and discovering that you spent all those hours debugging because the math library has errors is really frustrating.

Poofjunior:
Ok, thanks a bunch for taking a look at this! I'm happy to write an updated version of this class when the time comes. At the end-of-the-day, OF has done some amazing things for me, and I'd be honored to help write improvements.

arturoc:
glm has it's own quaternion class so we'll probably just use those but we'll need some way to go from OF <-> glm at the beginning to provide an easy way to port projects, addons...

i'm going to close this then. thanks so much for looking into it and take the time to explain it!

Poofjunior:
Sounds good! Thanks again for taking a close look!

bakercp:
Hey @Poofjunior thanks for all of your analysis on this.  I learned a lot and really appreciate the time you put in on it.

Poofjunior:
@bakercp NP! Thanks for taking some time to look into it, and, heck, I get to learn a lot from this stuff too!

PR #4469, opened 2015-10-24
added .m and .mm files to project source discovery makefile

### LINKED ISSUE #4468: Makefile not detecting .mm files in src
I have a few .cpp and a few .mm files in my `src/` directory. While the .cpp files are being recognized and built, the .mm files are not. I'm kind of stumped here, as I've commented out the rule that builds .cpp files from `$OF_ROOT/libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk`, which does indeed stop building the .cpp files. But the rule, literally two or so rules down from it, that looks for .mm files doesn't seem to get hit.

Am I missing something here?

Qix-:
This fixes #4468.


arturoc:
thanks!

PR #4475, opened 2015-10-25
Issue 4470 msys2 documentation

oxillo:
Start documentation for MSYS2 based on 0.9.0RC2


bakercp:
I can confirm this setup guide worked for me.  Would be nice to figure out an easier way to include include the DLLs for distribution.  Perhaps copying them like we do for VS?  Then again, perhaps this isn't possible because we need the whole MSYS2 collection of dlls?

arturoc:
yes there's 2 weird things with this:

- there shouldn't be any need to recompile videoInput, i've tested in my install some days ago and it was working without problem but i'll try to recreate the whole thing from scratch just in case. if it doesn't work what we should do is just install the recompiled version in the repository not add instructions to recompile it.

- copying all the dlls is not an option because there's too many, to run an application compiled with this toolchain you need to first install the dependencies as we do in linux. but there should be an explanation about how to modify the PATH in the windows settings, even if a user doesn't have administrative rights they can change the user environment variables so i think it's ok to not include the second option with the bat file...

Apart from that it looks great. There's also no projectGenerator in this install since the only supported IDE is qtcreator which is already able to create new projects... through the wizard so you can remove that section.

Thanks!

oxillo:
@arturoc libs/videoInput/lib/win_cb/videoInputLib.a is not on github nor in the RC2 win_cb zip... So there was no other way but to recompile. Should we track the binary on gihub or just compile it when we create the zip package...

About the dlls, I agree that the bat file solution is not the best one. With no administrative rights, I get some 'TLS_V1_client' error if I set the user PATH and some classical "missing dll" errors if I don't set it. I think some dlls in the system PATH overrides those from /mingw32. I think it has something to do with openSSL but I need to have a closer look...

I also have to add how to use on a corporate network with a proxy....

arturoc:
ok that explains why it's not working :) i'll push it later

oxillo:
I figured out the "TLSv1_1_client_method missing" error. It was using ssleay32.dll and libeay32.dll from some Intel iCls program that use openSSL 1.0.0. TLS v1.1 was introduced after that in OpenSSL...
The solution is to copy (or better link)  these 2 libs in the bin folder. If someone has a better solution...

Issue #4451, opened 2015-10-18
Apothecary - Audit Build scripts for OSX / iOS - Xcode 7


danoli3:
Xcode 7 has some sneaky new rules regarding compile time arguments for OSX and iOS.

This is due to them opening up to multi-sub platforms such as tvOS, WatchOS.

The issue is all libs have to specify their intended target platform.
From Apple: 
> In Xcode 7, the linker enforces that all files linked together were built for the same platform. You control the platform via -mmacosx-version-min=XX or -miphoneos-version-min=XX.

What does this mean?
----
In all apothecary scripts we must make sure we specify target's for the libs via that command.

### For iOS: 
```-miphoneos-version-min=7.0``` (or proxy iOS simulator command) ```-mios-simulator-version-min=7.0```

### For OSX 

```-mmacosx-version-min=10.9 ```

### For tvOS

```-mtvos-version-min=9.0 ```

----

Symptoms of this issue being an error:
----
> building for iOS simulator, but linking in object file built for OSX, for architecture x86_64

# Audit / Edit Status:

| Lib                             | osx |  ios  | tvos  |
|---------------------------------|-----|-------|-------|
| FreeImage                       |  ✓  |   ✓   |    ✓  |
| FreeType                        |  ✓  |   ✓   |    ✓  |
| tess2                           |   ✓   |    ✓    |    ✓    |
| poco                            |  ?   |     ✓   |    ✓    |
| openssl                         |  ✓  |   ✓   |   ✓   | 
| ofxAssimpModelLoader -> assimp  |   ?   |    ✓    |    ✓    |
| ofxOpenCV -> opencv             |   ✓   |    ✓    |    ✓    |
| glew                            |  ?   |  N/A     |   N/A    | 
| glfw                            |   ?  |   N/A    |   N/A    |
| cairo                           |   ?  |  N/A  |  N/A  |
| kiss                            | N/A |  N/A  |  N/A  |
| portaudio                       |  ?   |  N/A  |  N/A  |
| fmodex                          | ✓?  |  N/A  |  N/A  |
| rtAudio                         | N/A |  N/A  |  N/A  |
| Boost 1.58.0 FileSystem/System  |  ✓  |   ✓   |   ✓   |
| URI  							   |  ?   |    ?   |    ?   |

---------------------------------------------------------

danoli3:
Yeah this was looking good after the last Apothecary Script updates merge. 
- iOS was all good and all libraries already merged with the command. 
[...]

PR #4477, opened 2015-10-26
tvOS - Apple TV OS Feature - Updated

danoli3:
Updated PR from https://github.com/openframeworks/openFrameworks/pull/4413

Working version of openframeWorks for AppleTVOS tvOS 

Requirements: 
------
- Xcode 7.1+ 

Changes:
---------

- Added Seperate **ofxtvOS.h**
- Added Seperate **ofxtvOSAppDelegate / ofxtvOSViewController**
- Added New Seperate **tvOS+OFLib** static project file for tvOS (as it is quite different from iOS)
- Added tvOS template project
- Recompiled all libraries with **embedded bitcode**
- Changes to **ofConstants.h** to add tvOS definitions
- Changed ```#ifdef TARGET_OS_IOS``` to ```#if TARGET_OS_IOS``
   - (as it is defined in tvOS sdk however defined as 0, so #if works / #ifdef doesn't not.)
- Using ```#if TARGET_OS_IOS || (TARGET_OS_IPHONE && !TARGET_OS_TV)``` 
- Changes to a lot of ofxiOS files

Testing:
--- 
- tvOS Empty Example (Working on **AppleTV 4** AppleTV Simulator ).
- tvOS Static Lib (Building successfully for **AppleTV 4** and AppleTV Simulator)
- iOS Examples (All working! *Regression Test*)

*Changes made include modifications to ofxiOS with some pre-processor defines to not expose certain functions that are not in the tvOS SDK (such as video grabber, mapkit, orientation code etc).*

----

Pre-Requisites (Apothecary library builds) (WORK IN PROGRESS) - BITCODE = YES.
----

- [x] ```./apothecary -t tvos update FreeImage```
- [x] ```./apothecary -t tvos update openssl```
- [x] ```./apothecary -t tvos update poco```
- [x] ```./apothecary -t tvos update freetype```
- [x] ```./apothecary -t tvos update tess2```
- [x] ```./apothecary -t tvos update ofxAssimpModelLoader```

~~- [x] ```./apothecary -t tvos update ofxOpenCV```~~ Removed.

Library Changes.
- Requires Poco **1.6.2** (tvOS support).
- OpenSSL **1.0.2d** (why are we not already here D:)
- Removed ofxOpenCV *(no vision).*


danoli3:
@openframeworks/ios  @julapy Updated this PR since Apple TV officially out today (in the US at least).

addons/ofxiOS/src/video/ofxiOSVideoPlayer.mm
```diff
@@ -14,7 +14,7 @@
     bResetPixels = false;
     bUpdatePixels = false;
     bUpdateTexture = false;
-    bTextureCacheSupported = (CVOpenGLESTextureCacheCreate != NULL);
+    bTextureCacheSupported = (&CVOpenGLESTextureCacheCreate != NULL);
```
danoli3: This comparison was flagged by Xcode as always being false. Have to compare the reference.

danoli3:
Yeah changed all those pre-processors to be < 9.0 SDK safe.

danoli3:
Rebased. Adding controller support

danoli3:
> For watchOS and tvOS apps, bitcode is required. If you provide bitcode, all apps and frameworks in the app bundle (all targets in the project) need to include bitcode.

Okay Apothecary hat again!
f924926c-f8d6-11e4-91cf-d401d0b69f2a

danoli3:
bitcode
Awesome!

ofTheo:
awesome work @danoli3

danoli3:
img_4120
Yay! (On device)

ofTheo:
nice!! can't wait to get this in.

arturoc:
@danoli3 can you perhaps divide this in several PRs? it seems like there's several things here that are not directly related to having TV support.  like there's a new event for controller pressed/released (is that controller as in xbox controller?) and also seems like poco has been updated  but there's only libraries for ios and tv while the headers have changed.

perhaps send the changes for poco to the apothecary repo since now it's already building the libraries automatically you will only need to send the changes to the formula not the headers or the binaries and it'll get built for all the platforms

danoli3:
@arturoc I believe all the Apothecary changes for this branch are already included in the main repo, so everything should be good to build the tvOS libs in the other repo. I'll have to configure some things travis to work with tvOS though.

All relate to the tvOS changes... the controller event changes were for the "apple tv controller" and external controllers. I'm not sure this is the right approach, however it is the only input methods given, so I think it has to be part of the core in someway, I followed what the touch events did. This could be used I guess as a "controller" support for xbox or other types. I tried to make it generic in ways it could be used in that way.

danoli3:
And yeah I'll split this one up!

arturoc:
ok, then perhaps better to remove the headers and libraries from this one, keep them by now if you want and we can remove them before merging once we have all the libraries compiling in the new repo.

about the controller, since it's somethign very specific for this platform perhaps it's easier to have it in the addon by now, with the touch events we did that, we had them in the ofxiPhone addon at the beggining but then once we did the android port and since windows and other platforms were adding touch capabilities we turned it into a core feature. that'll also help figuring out the best api and once it's more stable it can become a core feature if we thing it makes sense

also why do we need different libraries for tv and ios? isn't it the same architecture?

danoli3:
@arturoc same architecture, different settings and tvOS uses LLVM bitcode embedded.

Addon sounds correct I'll work on that change.

arturoc:
that's strange if it's the same OS (at least the same kernel) and same chip sounds like the same binaries should work, i don't think it's a big deal and much less if we get the auto apothecary thing working but would be nicer if the same binaries worked for iOS and tv

danoli3:
Technically they could but it would put iOS on iOS 9.0+ minimum and we would have to recompile all the libraries with bitcode enabled and embedded.

Aside from that Apple mandated that when linking libraries they are to be built with the target architecture set, see https://github.com/openframeworks/openFrameworks/issues/4451

arturoc:
can't you specify both flags?  having to recompile the libraries is not
a big deal now but having compatibility for iOS9 only is, i guess we can
merge both platforms binaries in a couple of years then :)

Issue #4478, opened 2015-10-26
openCv libs outdated in MSYS2


oxillo:
openCV libs in win_cb are version 2.3.1 and use a different exception scheme (SjLj). Examples using OpenCV cannot be linked.

kylemcdonald:
closed by #4480

PR #4480, opened 2015-10-26
Issue #4478 msys2 outdated opencv

oxillo:
update ofxOpenCv (and opencv libs) to work for MSYS2. Not tested on other platforms...


ofTheo:
Thanks for this PR - curious why you changed files in ofxOpenCv though?
On other platforms these changes aren't required, so it makes me think that you might have built a newer version of OpenCV than is being used on other platforms. Is this version 2.4.9 ?

oxillo:
@ofTheo MSYS2 pacman installs openCV 3.0.0 where some legacy code (I guess from OpenCv 1.x) has been dropped. 
Most changes were just #define in compat.hpp.
The harder one was the unDistortOnce where I just copied the 3 lines content of the function from legacy code...
The good thing is that switching to openCv3 looks pretty easy.

ofTheo:
thanks @oxillo - I figured as much. 
Is there anyway to get pacman to generate 2.4.9? 
I think we could switch to OpenCV 3.0 for OF 0.9.1 but we can't release 0.9.0 with different versions of OpenCv across platforms. 

Thanks!!

arturoc:
in msys2 we use the same system as in linux now, the opencv version as well as other libraries depend on the version installed in the system, in ubuntu right now is 2.4.9 but for next release, which happens in april, it'll be 3.0 surely, in archlinux is probably 3.0 already.

we should just test this changes with 2.4.9 and if they work too just merge them or add a couple of idefs in case they don't.

oxillo:
All the functions are part of 2.4.9 so it should work on platforms with 2.4.9....

arturoc:
oh, yes actually this is compiling in the tests for both linux and osx which compile the allAddonsExample which includes opencv so it seems to be correct. i'm going to merge it then, thanks!

PR #4483, opened 2015-10-27
remove ofx3DModelLoader

arturoc:
was failing to compile in some platforms cause it uses some really old GL calls


bilderbuchi:
should we already add a changelog entry about that?

arturoc:
done, we haven't updated the CHANGELOG in a long time though so we'll need to go through the changes before the release. perhaps we should add something so people get a notification that they should update the CHANGELOG when sending a PR that doesn't

bilderbuchi:
> we should add something so people get a notification that they should update the CHANGELOG when sending a PR that doesn't

I think that's too complex to work properly (i.e. detect when a changelog entry is needed), but if we would do it, it would probably be best to fail a Travis build to notify - we could add a new build which would also check style, etc.

>  we'll need to go through the changes before the release

yes, I know. Be aware that [this script](https://github.com/bilderbuchi/OF_repo_utilities/blob/master/get_merged_prs_since_tag.py) exists for that purpose (in case I'm not around at the time) - it will generate a markdown-formatted list you can easily put in a Gist to share.

Issue #4481, opened 2015-10-26
ofxOpenCv Error 


NickHardeman:
When compiling the opencvExample on OSX in 10.8 in 09RC2, the following error occurs
`dyld: Symbol not found: ___sincos_stret
  Referenced from: /Applications/OpenFrameworks/of_v0.9.0RC2_osx_release/examples/addons/opencvExample/bin/opencvExampleDebug.app/Contents/MacOS/opencvExampleDebug
  Expected in: /usr/lib/libSystem.B.dylib`

This error can be remedied by building the opencv lib against the 10.8 SDK by modifying the apothecary script and adding:
`-DCMAKE_OSX_SYSROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.8.sdk \`

NickHardeman:
Related to: https://github.com/openframeworks/openFrameworks/issues/4412

arturoc:
can you also send the recompiled libraries for opencv? we've tried to recompile them with this flags before in osx > 10.8 and it didn't work i think you really need to recompile it in a 10.8 machine which only a few people still have around

NickHardeman:
Yes. I have openCv and assimp recompiled, but can't seem to get the cairo apothecary script to work. Any suggestions? Keep getting this error:
`/Applications/OpenFrameworks/of_v0.9.0RC2_osx_release/scripts/apothecary/apothecary: line 68: /Applications/OpenFrameworks/of_v0.9.0RC2_osx_release/scripts/ostype.sh: No such file or directory
/Applications/OpenFrameworks/of_v0.9.0RC2_osx_release/scripts/apothecary/apothecary: line 934: XS: unbound variable`

ofTheo:
I think the issue might be here:
https://github.com/openframeworks/openFrameworks/blob/master/scripts/apothecary/apothecary#L901

When it runs the apothecary script again for each dependency its path for APOTHECARY_DIR is incorrect 

This is what the value is when it starts:
```
 START - APOTHECARY_DIR is now openFrameworks/scripts/apothecary 
```

This is called from when the dependency is run as a separate script 
```
 depend is pkg-config 
 START - APOTHECARY_DIR is now openFrameworks/scripts/apothecary/build 
```

ofTheo:
Okay this is definitely the issue:
https://github.com/openframeworks/openFrameworks/blob/master/scripts/apothecary/apothecary#L901

[...]

Issue #4412, opened 2015-10-06
ofxAssimpModelLoader linker errors on OSX 10.8 / master branch.


bakercp:
Compiling the ofxAssimpModelLoader example on 10.8 with the latest master branch produces this linking error:

![screenshot_10_5_15__9_18_pm](https://cloud.githubusercontent.com/assets/300484/10298617/a2fab2f2-6ba6-11e5-8888-3725adc94f9e.png)

Trying to recompile it on 10.8 now to see if that fixes the issue.

arturoc:
this seems like the same problem as with other libraries, can you recompile it under 10.8 as you did with the others?

danoli3:
Yeah 
```
extern double __exp10(double) __OSX_AVAILABLE_STARTING(__MAC_10_9, __IPHONE_7_0);
```

Note: ```__MAC_10_9```

NickHardeman:
I get this error when compiling the 09RC2 assimpExample in Xcode on OSX 10.8:
dyld: Symbol not found: ___sincos_stret
  Referenced from: /Applications/OpenFrameworks/of_v0.9.0RC2_osx_release/examples/addons/assimpExample/bin/assimpExampleDebug.app/Contents/MacOS/assimpExampleDebug
  Expected in: /usr/lib/libSystem.B.dylib

The error is fixed and the example compiles when rebuilding the Assimp lib using apothecary and specifying the 10.8 SDK. 
`cmake -G 'Unix Makefiles' -DCMAKE_OSX_DEPLOYMENT_TARGET=10.8 -DCMAKE_OSX_SYSROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.8.sdk $buildOpts -DCMAKE_C_FLAGS="-arch i386 -arch x86_64 -O3 -DNDEBUG -funroll-loops" -DCMAKE_CXX_FLAGS="-arch i386 -arch x86_64 -stdlib=libc++ -O3 -DNDEBUG -funroll-loops" .`

arturoc:
@NickHardeman can you send a PR with the recompiled libraries?

kylemcdonald:
closed by #4487

PR #4487, opened 2015-10-27
Bugfix libs for 10.8

NickHardeman:
Recompiled the libs against the 10.8 SDK. Cairo and dependencies, freetype, assimp and opencv. 
Resolves: https://github.com/openframeworks/openFrameworks/issues/4481#issuecomment-151406337 and https://github.com/openframeworks/openFrameworks/issues/4412


danoli3:
Nice! 0.9 close now!

ofTheo:
@danoli3 does iOS feel pretty solid at the moment? or does it need some heavy testing?
were you going to wait to 0.9.1 to do AppleTV or try and get it in?

danoli3:
All the examples work on the devices / simulators I've tested. 
I think it is good to go. Always could use more testing though :D! 

AppleTV and other future changes definitely 0.9.1 (we going to push iOS 7.0 minimum then as well). So be good to get 0.9.0 released, stable with support for iOS 5.1.1. 
Anyone wanting to keep up to date, like all the iOS openFrameworks users currently, will just keep up with master (post 0.9.0 release).

ofTheo:
@danoli3 - thats awesome!
yes - think thats a solid plan. 

Also want to work on cleaning up the iOS keyboard and image picker code for for 0.9.1 
But will worry about that after 0.9.0 is out! :)

Issue #4372, opened 2015-09-14
Multi window fullscreen issue


HalfdanJ:
I've been sitting with a lot of problems today with the multi window support. I'm not excatly sure what is going wrong, but i'm pretty sure there are some bugs / missing features. 

The case: I'm have an app (LightLeaks) that needs to run on a new Mac Pro (the bin). We need the installation to run on 3 projectors, so my thought was to change the app to initialize 3 windows, all sharing context, and making them fullscreen on each one display on startup. 

The projectors are running 1920x1200, and i had a primary display running 1920x1080. If i in main.cpp run

```
 window->setWindowPosition(1920, 0);
 window->setFullscreen(true);
``` 
on all 3 windows, they do fullscreen, but only 1080px high (not 1200px). If i toggle fullscreen twice they fullscreen correctly, but often they jump to main screen when I do it. I changed my main monitor to also run 1200px high, "fixing" this problem, but often they dont fullscreen correctly anyway. 

Second thing i tried was to run the windows without decoration, and setting the dimensions and position of the windows so they should fullscreen in setup. Initially this worked, but suddenly when one of the proejctors turned off and on again, the window was offset by half of the screen height, also after rebooting. So setting ofSetWIndowPosition(1920,0), did not set the window to the top left corner of the second screen, but instead above the screen. 

My last test was to run this code every frame in update. And that worked, now the windows are properly filling out the screens, but it seems like an odd way to do it. 

My suggestion is to add a function call that will tell a window to fullscreen on a monitor with a specific name or id. It seems like a very usefull feature of the multiwindows to be able consistently to fullscreen on the right monitors, even when the layout changes.

bilderbuchi:
It's slightly tangent to the problem at hand, but have you considered getting a Triplehead2Go (matrox device) - it's a pretty great line of devices, and saves all kinds of multi-desktop, multi-window, v-sync, maximization, etc. issues if you just run your app in _one_ window with dimensions 3*width x height. :-)

bilderbuchi:
also, are your monitor/projector ids constant over a power cycle of involved devices? what are the names/ids if you have a set of identical devices?

kylemcdonald:
we've used a th2g in the past, but it only works for low resolution projectors (up to 1360 pixels wide on each iirc).

ofTheo:
I've run into this issue quite a bit with my own GLFW multi window implementation. 
[...]

PR #4489, opened 2015-10-27
Bugfix glfw fullscreen multi osx

NickHardeman:
Related to https://github.com/openframeworks/openFrameworks/issues/4372. Adds a check to determine if the point lies on the monitor rectangle. Allows setting the window position to the exact monitor position and then setting to full screen to enter full screen on that monitor.


arturoc:
yeah, makes a lot of sense, thanks!

PR #4490, opened 2015-10-27
Bugfix glfw fullscreen multi - fix for setting monitor and fullscreen in glfw settings on OSX

NickHardeman:
`settings.monitor = 1;
settings.windowMode = OF_FULLSCREEN;`
will now work on OSX, as long as the position is not set in the settings.
Must set the window shape and then the position, as previously commented here
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L707


kylemcdonald:
this might need manual testing on other platforms to make sure it doesn't introduce any regressions, but looks good. would be a great feature to have cross-platform :)

PR #4496, opened 2015-10-28
fix area lights for programmable renderer

### LINKED ISSUE #4495: shader error with programmable renderer in AreaLightExample on Windows / VS
It seems like it doesn't like the:
#define MAX_LIGHTS 
#define USE_TEXTURE

The shader doesn't compile, but the app runs. 
Going to paste the print out.

tgfrerer:
Because the old implementation of `ofMaterial.cpp` mixed the `STRINGIFY`preprocessor macro with GLSL preprocessor directives, the behaviour of the c preprocessor for this file was undefined.

This lead to strange GLSL code in VS2015 builds, and `ofMaterial` may have been considered broken on this platform(s).

To close #4495, this PR suggests & implements the following:

* Using **Raw String Literals** for inline GLSL code instead of `STRINGIFY`. This allows us to
  concat different cpp constants based on the cpp build environment. Raw String Literals are a C++11 feature. See: http://en.cppreference.com/w/cpp/language/string_literal

* Anything within Raw String Literals is not interpreted by the preprocessor, also escapes are not scanned. It is literally a literal.

Tested on XCode 7.1 / clang 7.0.0 and VS2015; Raw string literals are advertised as supported by GCC from 4.5 onwards. https://gcc.gnu.org/projects/cxx0x.html

### DIFF
--- a/libs/openFrameworks/gl/ofMaterial.cpp
+++ b/libs/openFrameworks/gl/ofMaterial.cpp
@@ -308,13 +308,9 @@ static const string fragmentShader = R"(
 
 		ambient += light.ambient.rgb * attenuation;
 		diffuse += light.diffuse.rgb * nDotVP * attenuation;
-)"
 #ifndef TARGET_OPENGLES
-	R"(
 #define SPECULAR_REFLECTION
-	)"
 #endif
-R"(
 #ifndef SPECULAR_REFLECTION
 		// ha! no branching :)
 		pf = mix(0.0, pow(nDotHV, mat_shininess), step(0.0000001, nDotVP));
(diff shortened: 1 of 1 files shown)


ofTheo:
Nice fix! 
Just tested this on Windows 8 and it works perfectly. 
Also tested on OS X 10.9 with Xcode and it works fine too. 

@danoli3 / @julapy - could you check on iOS?

arturoc:
yes much better than the stringify macro. btw, TARGET_OPENGLES is also defined in shaders so there's no need to put that out of the shader string.

danoli3:
iOS Working with above. Tested iosES2ShaderExample

tgfrerer:
Thanks @danoli3! 
@arturoc - makes sense =) I've removed the redundant check.

arturoc:
thanks!

PR #4501, opened 2015-10-30
Android: use fixed data zip file name

HalfdanJ:
I had an issue having an android app with a app name including a space. The bin/data folder got saved into a .zip file using the name of the folder the app is in. When unzipped, it used the app name (so the label of the app), these where not the same in my case since i wanted a space in my app name. 

This PR is making a fixed name (`ofdataresources`) instead for the data zip folder. I cant think of issues having a fixed name instead of a dynamic names based on app name.


arturoc:
thanks!

PR #4506, opened 2015-10-31
win_cb removal. Replace by msys2

oxillo:
Remove win_cb directories, scripts and replace them by msys2


arturoc:
wow thanks!

i see there's a script to install the 64bits version. i didn't do it because fmod the library we use for sound doesn't have a 64bit version for mingw so not sure we should include this. otherwise people will probably install that version rather than 32bit and then won't get sound.

at least the script should advice that when installing 64bit version soundplyaer won't work.

another option, perhaps better, is to let the install_dependencies script receive a parameter that makes it install the 64bit version but only document that in the setup guide advising of the limitations of doing so.

arturoc:
also just realizing that the libraries that come with the download are compiled for 32 anyway so the 64bit version won't work at all unless people recompile the libraries so it's probably better to just leave it out by now.

for next release we can look into adding 64bit libraries and find a substitute for fmod

oxillo:
Forgot to delete the 64bits install. It's partially in the install_dependencies but the 'if' has been commented out.
I do some more cleaning to use system libs (installed with pacman) and then I'll take a look at the 64 bits problem. Msys2 has openAL so I may go with that.
I was thinking of having MSYS2 apothecary scripts to call pacman when applicable. Does it sound like a good way to do it ?

arturoc:
we are only using apothecary for the libraries that are not in pacman, same as in linux, so there's no need to adapt the apothecary scripts.

we could look into installing openAL but unless it's really straight forward i'd left it for next release.

arturoc:
just checked the combined install_dependencies.i think we should only support i686 by now, you can still use msys2 64bits but it'll install the 32bits tools which will be compatible with fmod.

oxillo:
Ok. I have a branch with openAL support
I still have some errors when running example 
[warning] ofSoundPlayer: ofSoundUpdate() not implemented on this platform
AL lib: (EE) alc_cleanup: 1 device not closed````

arturoc:
Hey, that's great, i think i get similar errors in linux sometimes, the openAL player really needs some refactoring.

I'm a little worried to introduce this changes so late in the release, we are planning to release 0.9 this week. do you mind separating this in 2? one PR for the naming changes with support only for 32bits and then another one for 64bit support and openAL?

i'll merge the first one right away then the second right after 0.9 is out

oxillo:
This current PR only has the naming changes (32 bits only). So you can merge but beware that you have to reflect the changes on the site (due to create_package.sh).
I'll submit a new PR for openAL support and use of pacman distributed libs (when possible). This PR will be 32 bits only.
A second PR will be for 64 bits support in MSYS2 (with openAL).
These two PR should go in 0.9.1 as I share your feelings of introducing these changes so late.

arturoc:
thanks so much! i'm going to merge this already since it doesn't make sense to release as win_cb if we don't have support for codeblocks any more

PR #4507, opened 2015-10-31
OFAndroid - try to load x86 lib before arm to resolve houdini crashes

tallavi:
1. Build x86 native lib to libOFAndroidApp_x86.so (instead of libOFAndroidApp.so)
2. try to load x86, and if failed, load the arm. This resolves crashes on arm emulated asus devices that report arm instead of x86.

See discussion here: http://forum.openframeworks.cc/t/ofandroid-doesnt-work-on-x86-devices/21164/25


arturoc:
thanks! i guess trying to load the x86 library in arm won't crash the app right?

tallavi:
yep, tested to work on several arm devices. It catches the exception and then loads the arm lib.

arturoc:
great, thanks!

PR #4509, opened 2015-10-31
Arm updates

jvcleave:
- removed conditional for old firmware (old firmware won't be possible to use as we are requiring Jessie sources anyway)

- added $(RPI_ROOT) for cross-compiling on arm7


arturoc:
thanks!

Issue #4508, opened 2015-10-31
Android: ofGetOrientation doesn't get updated when phone is rotated


HalfdanJ:
I have an app where the orientation is not specified (ofSetOrientation is never called) and the manifest has ```android:configChanges="orientation|screenSize"```. I can rotate the phone and the openframeworks window gets rotated, and the size gets updated. But the orientation keeps returning ```OF_ORIENTATION_DEFAULT```

Shouldn't the function return the actual orientation?

arturoc:
yes i guess it should return the correct orientation. it's not as bad cause the orientation in OF is default 90 -90 180 where default could be read as whatever the hardware is set to even automatic rotation, and the rest are offsets from that as opposed of having orientation settings as portrait, landscape which would be absolute.

i guess having it change wouldn't break anything and would be useful though. another option would be to add a specific android function that returned the absolute rotation, as portrait, landscape...

HalfdanJ:
How is iOS handling this?

HalfdanJ:
@danthemellowman, @arturoc, @julapy i'm trying to make orientation work the same way on android and iOS, i've implemented an event now that triggers on orientation change, but i'm a bit unsure if the way its implemented in iOS is the way we want to go. In [the iOS example](https://github.com/openframeworks/openFrameworks/blob/master/examples/ios/iosOrientationExample/src/ofApp.mm#L195), the event triggers ofSetOrientation(). In iOS, this doesn't force the device to a certain orientation, but it does on android. So if I copy the sample, the first time you rotate the device, you freeze the orientation. 

Wouldn't it be better if ```ofGetOrientation()``` always returns the actual orientation of the device, and not what ofSetOrientation has been called with?

danoli3:
It shouldn't force you to the device physical orientation on Android.

With iOS it doesn't force you because what if you wanted a landscape mode only App. 
Then just don't respond to the orientation commands associated with Portrait in then your own code.
I think the iOS orientation code really needs some work though.

Issue #4505, opened 2015-10-31
Android: Bump min sdk requirement


HalfdanJ:
Currently we are having a min sdk requirement of sdk 8. Thats Android 2.2 (Froyo). Can we bump this up a bit? Currently only [0.3% of android phones](https://developer.android.com/about/dashboards/index.html) are running Froyo :) 

I'm not sure what a reasonable SDK requirement would be, i would prefer 16 (Jelly Bean) or 19 (Kit Kat that has the largest market share). 

The reason for doing this is to be able to use some more modern apis, specifically i'm having trouble with the camera code right now.

ofZach:
+1 to this, on osx / ios we tend support 2-3 versions of the api, sometimes a bit more (ie, OS 10.11, 10.10, 10.9...) for exactly this reason, it would be good to do here as well.

HalfdanJ:
Then the question is how to count that in android.. http://developer.android.com/guide/topics/manifest/uses-sdk-element.html 
I would say 3 versions back is Kit Kat (Android 4.4, API 19)

arturoc:
i haven't incremented the version since the beginning cause there's been no need since then. sometimes i've added support for new apis but in java you can use introspection to know if an api is present or not and only call it conditionally and through introspection not direct calls.

that kind of code is really unreadable though so i've only used it for relatively small things like a feature in the camera where there's only 1 call that is done like that.

if we want to use a whole class like camera2 then it doesn't make sense to use introspection all over. 

i guess we can just use the api level that whatever feature we want to use needs.

i would support 4.0 if possible (if camera2 is already supported there)

arturoc:
actually i'm checking and camera2 is not supported until 5.0 which is pretty new. i would wait to implement it unless we can do it in a way were we can fallback to camera if camera2 is not supported

HalfdanJ:
Yes, camera2 is a bit to new, but there are other changes that would make life easier. For example is surface texture introduced in 11, right now there is a lot of hard to read code around it that would be great to remove.

PR #4510, opened 2015-11-01
Android camera refactor

### LINKED ISSUE #3752: Android video grabber broken
I've just updated my openframeworks version to latest git version, and a android app using ofVideoGrabber stopped working. It turns out its initGrabber function that is broken. 

@arturoc I can see that there where made changes in that code last time in https://github.com/openframeworks/openFrameworks/commit/29e7c6b6f708ab1ce13c640a31a53050622356bc . I can't see that it should change how the code is used, is there an error in this commit?

### LINKED ISSUE #4502: Android: Camera updates very slowly
Running the android camera example on a Nexus 6, gives me a really slow refresh rate of the camera. The app is running ~60fps, but the camera is only running around 10fps. 

I can see that the current camera implementation is using the deprecated `android.hardware.Camera` module instead of the new `android.hardware.camera2`. Would it make sense to update this? 

There is some example code of using it here made by google creative lab in their shadercam example: https://github.com/googlecreativelab/shadercam/blob/master/shadercam/src/main/java/com/androidexperiments/shadercam/fragments/CameraFragment.java

HalfdanJ:
This PR tries to improve the android camera. 

- By default it will now select the best frame rate available instead of first (that on nexus 6 is 15fps). Fixes #4502 
- ```getPixels()``` now works, and the pixel conversion and copying only happens if needed (drastically improves performance). Fixes #3752
- Running without calling ```getPixels()```allows the code to run 30fps in full HD on a Nexus 6
- Now the deviceID can be changed while the camera is running, so its now possbile to swap camera easily
- ```listDevices()``` is now implemented 
- Functions for getting the facing and orientation of cameras
- Removed legacy code regarding supporting very old API 8 that didn't support surface textures
- Updated android grabber sample to show more functionality 

There are some outstanding issues: 
- ~~This PR requires the API level to be raised to 11  from 8 #4505~~
- ```getPixels()``` and the texture are sometimes not flipped the same way on some devices. Need to look at the matrix transformation
- In the sample, the orientation of the image is not correct. Depends on #4508 

The code is tested on following devices: 
- Nexus 6 (Running Android 5.1) 
- Nexus 9 (Running Android 5.1) 
- Fairphone 1 (Running Android 4.4)

### DIFF
--- a/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidVideoGrabber.java
+++ b/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidVideoGrabber.java
@@ -41,7 +41,7 @@ public class OFAndroidVideoGrabber extends OFAndroidObject implements Runnable, 
         }
 	}
 
-    public int getNumCameras() {
+    public int getNumCameras(){
         return Camera.getNumberOfCameras();
     }
 
@@ -215,10 +215,14 @@ public class OFAndroidVideoGrabber extends OFAndroidObject implements Runnable, 
         return cameraOrientation;
     }
 
-    public boolean getIsCameraFacingFront(){
-        return cameraFacing == Camera.CameraInfo.CAMERA_FACING_FRONT;
+    public int getFacingOfCamera(int _deviceID){
+        if(_deviceID == -1) _deviceID = deviceID;
+
+        Camera.CameraInfo info = new Camera.CameraInfo();
+        Camera.getCameraInfo(_deviceID, info);
+        return info.facing;
     }
-	
+
 	public boolean setAutoFocus(boolean autofocus){
 		
 		if(initialized){
--- a/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidVideoGrabber.java
+++ b/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidVideoGrabber.java
@@ -35,74 +35,74 @@ public class OFAndroidVideoGrabber extends OFAndroidObject implements Runnable, 
 		}
 
 		try {
-            // Create surface texture instance
+			// Create surface texture instance
 			surfaceTexture = new SurfaceTexture(texID);
 
 			// set texture as preview surface for camera
-            camera.setPreviewTexture(surfaceTexture);
-        } catch (Exception e1) {
-            Log.e("OF","Error initializing gl surface",e1);
-        }
+			camera.setPreviewTexture(surfaceTexture);
+		} catch (Exception e1) {
+			Log.e("OF","Error initializing gl surface",e1);
+		}
 	}
 
-    public int getNumCameras(){
+	public int getNumCameras(){
 		if (Build.VERSION.SDK_INT < 9) {
 			return 1;
 		}
 
 		return Camera.getNumberOfCameras();
-    }
+	}
 
-    public int getCameraFacing(int facing){
+	public int getCameraFacing(int facing){
 		if (Build.VERSION.SDK_INT < 9) {
 			return 0;
 		}
 
 		int numCameras = Camera.getNumberOfCameras();
-        Camera.CameraInfo cameraInfo = new Camera.CameraInfo();
-        for(int i=0;i<numCameras;i++){
-            Camera.getCameraInfo(i, cameraInfo);
-            int _facing = cameraInfo.facing;
-            if(_facing == facing){
-                return i;
-            }
-        }
-        return -1;
-    }
+		Camera.CameraInfo cameraInfo = new Camera.CameraInfo();
+		for(int i=0;i<numCameras;i++){
+			Camera.getCameraInfo(i, cameraInfo);
+			int _facing = cameraInfo.facing;
+			if(_facing == facing){
+				return i;
+			}
+		}
+		return -1;
+	}
 
-    public void setDeviceID(int _deviceId){
-        deviceID = _deviceId;
+	public void setDeviceID(int _deviceId){
+		deviceID = _deviceId;
 
-        if(initialized){
-            stopGrabber();
+		if(initialized){
+			stopGrabber();
 
-            initGrabber(width, height, targetFps, texID);
-        }
-    }
+			initGrabber(width, height, targetFps, texID);
+		}
+	}
 
 	public void initGrabber(int w, int h, int _targetFps, int _texID){
-        if(camera != null){
-            camera.release();
-        }
+		if(camera != null){
+			camera.release();
+		}
 
 		if(deviceID==-1)
-            deviceID = getCameraFacing(0);
+			deviceID = getCameraFacing(0);
 
-        try {
+		try {
 			if (Build.VERSION.SDK_INT >= 9) {
 				camera = Camera.open(deviceID);
 			} else {
 				camera = Camera.open();
 			}
 		} catch (Exception e) {
-            Log.e("OF","Error trying to open specific camera, trying default",e);
-            camera = Camera.open();
-        }
+			Log.e("OF","Error trying to open specific camera, trying default",e);
+			camera = Camera.open();
+		}
 
-        if(_texID != -1) {
-            texID = _texID;
-            setTexture(texID);
-        }
+		if(_texID != -1) {
+			texID = _texID;
+			setTexture(texID);
+		}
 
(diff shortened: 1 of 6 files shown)


danthemellowman:
I ran into a bunch of issues today with the camera and started fixing things both in the java and cpp side also.  Let me check out your changes to see how they differ.

HalfdanJ:
doh.... Let me know how it goes.

arturoc:
hey jonas, i'm checking this and although it seems to fix some issues, it's bumping the version without really needing it? the functions to get the device and set the texture were already working without need to use higher APIs. 

I wouldn't mind bumping the API in a further refactoring but it seems slightly unnecessary here since the parts that require it were already working?

also the way you are getting the pixels now moves the colorspace conversion from the camera thread to the GL thread which doesn't make much sense given that most phones nowadays have 4 cores it seems logical to leave the GL thread as free as possible for other stuff and do the colorspace conversion in the camera thread.

@danthemellowman do you have your fix somewhere? otherwise i can take a look, i think this is a relatively straight forward fix that only requires to swap the front and back buffers.

We can look into refactoring this classes for next release along with @tallavi's changes but i think we should just introduce the minimum changes possible right now in order to fix this since we  are already in RC2

HalfdanJ:
Okay, its not to bad to roll that part back requiring the higher api. I did it because i had a very hard time understanding the flow of the code, so needed to understand it better. Can we at least remove the whole check if surface textures are supported, and instead just throw an error if the camera is used on api < 11? 

Regarding the pixel stuff, i know the conversion has moved core, my reasoning behind doing is the following: 
- If you dont call getPixels, no conversion/copying will happen at all. Allowing high res texture usage from the camera. I was not able to get a full-hd fluid image out of the other setup
-  Less copying is necessary since we dont need a back and front buffer
- I think a very typical use scenario of getPixels is with openCV, and in that case the grayscale image is probably often usable, and therefore there wont be any conversion, and no big benefit of doing it on the camera thread.

HalfdanJ:
I think i've made a better solution on SDK 8 support now.

arturoc:
ok great. regarding the pixels, it make sense to not doing it unless necesary but we could just add a flag on the first call to getPixels or similar which will trigger the colorspace conversion in the next frame

HalfdanJ:
I considered that solution as well. The cons are that first time you call getPixels, you wont get any (or we do the conversion og gl thread first time), and I think its a bit messy that getPixels call is modifying the performance of the camera in future calls, and not only the single thread. But it can be a solution until we add a ```setUsePixels ``` function to grabbers.

arturoc:
another posibility is to add setUsePixels for android alone by now and then make it mandatory through ofBaseVideo later. if someone wants to disable it right now the syntax would be kind of complicated:

```
((ofxAndroidVideoGrabber*)grabber.getGrabber().get())->setUsePixels(false);
```

but i think it's totally ok if we plan to refactor this for next release

HalfdanJ:
I like that more. Ill implement that. 
Also, i saw that ofPixels has support for NV21 format directly, but i tried to feed in the raw data from the camera in this format, trying to copy it into a ofImage and draw that gave grayscale noise. Should it work?

tallavi:
Hi, thanks for tagging me. I sifted through the changes and I believe there won't be any contridictions with my refactored branch. The sore point is mainly the file OFAndroid.java (which is untouched). Since some methods from the class moved to other files, I have to merge changes manually in order to keep my branch mergeable, like the zip file name change: 

https://github.com/openframeworks/openFrameworks/commit/3214e08014a9eed5b6f4c975b5b380c3bb360a12 (also by you, @HalfdanJ :-))

In regards to where android addons are going, I think it would be great if people could easily include jar files as part of their addon, and the java side of the addon will consume OFAndroid, instead of being a part of it. Just like the iOS folks can include .mm files that are not part of the core, only allot more complicated :-)

Further more, it would be very nice to include a general purpose conduit to allow data transfer from the c++ to java side without having to write all that jni boiler plate. 

Combining the two will create a platform for allot of quality Android addons that will benefit OF.

arturoc:
@HalfdanJ NV12 should work with the programmable renderer since that has the corresponding shaders to do the colorspace conversion in the GPU. i'm planning to implement it at some point but haven't got to it yet

HalfdanJ:
Okay now the conversion is moved back on camera thread, and the flip flopping of pixels has been implemented.

danthemellowman:
I have my stuff on a branch let me clean it up and do a PR.

I basically implemented closing the camera thread, setDeviceID,  swapping
the pixel buffers.

arturoc:
Jonas, i don't think this solves multithread access to the same pixels. the camera thread is changing the bufferFlipFlip variable without locking and both threads are accessing both buffers so it can happen that the main thread tries to access the pixels while the other is writting to that buffer.

also i don't think the SDK_INT>=... and then using the symbols that require a higher api, is a valid check, if you want to make the class supported under earlier versions all the symbols in it have to be supported on that version otherwise it will fail, that's why i'm using introspection.

why do you need to remove the way it was don before? texture loading was working for me and changing the camera to the front facing one was working just by calling setDeviceID(1) is that failing now?

arturoc:
by the way the way i'd implement the pixels thin is something like:

```
Data{
...
bool newBackFrame;
bool newFrontFrame;
ofPixels fontBuffer, middleBuffer, backBuffer;
std::mutex mtx;
}

void update(){
    if(newBackFrame){
        std::swap(middleBuffer, frontBuffer);
        newBackFrame = false;
        newFrontFrame = true;
    }else{
        newFrontFrame = false;
    }
}

void java_callback(buffer, ...){
    // colorspace convert passed buffer to backBuffer
    std::unique_lock lck(mtx);
    std::swap(backBuffer, middleBuffer);
    newBackFrame=true;
}
```

that way the locking time is minimal and there's never parallel access to the same pixels from both threads.

HalfdanJ:
You are right.. If the framerate of the camera is higher then the fps of the app, then there is a risk. Let code on it using your code

Regarding the sdk version, then i'm very sure its valid: http://developer.android.com/training/basics/supporting-devices/platforms.html, without the check android studio is giving me errors, and is suggesting this as the solution also (and removes the errors)

arturoc:
ok, that's great i did the introspection thing a while ago and it was the recommended way, i guess they've changed it later. it really makes a lot of sense, using introspection is really hard to understand

danthemellowman:
If we want faster/better performance from the camera we should probably
toggle on and off reading into pixel buffers via setUseTexture(bool)?  This
way if you want higher fps and only need the textures you can have them.
Otherwise if you need the pixel data to dump into openCV you get them by
default.

Figuring out how to properly handle flipping the front facing camera pixels
when needed would be nice.

Let's push some of this stuff to 0.9.1 and only fix the regressions for
reading back the pixels, the missing setDeviceID java call (or name
mismatch its implemented in the cpp but fails to find the java method), and
the missing implementation of ofVideoGrabber close().

Increasing the MIN_SDK should be OK. I mean 2.3 is reallllly old now. That
was wayyyy back when I was working on Android haha.. But we can't raise it
all the way to 5.0. Only like 8% of the market is at 4.0 or below but only
24% is 5.0+.

Also we should look into switching between the Camera Api and the newer
Camera2 Api in the JAVA layer so that if the phone supports the newer Api
it gets used otherwise it falls back to the older API.

Would be nice to wrap the MediaRecorder to toggle native android video
recording.  We could also add support for USB On-The-Go webcams
https://github.com/openxc/android-webcam

HalfdanJ:
@danthemellowman you can now in this pr toggle the use of pixel buffers with setUsePixels call. I totally agree on doing a second refactor, but thats probably 0.10.0, and raise the api version.

danthemellowman:
Is that a base ofVideoPlayer call? We should keep everything as close to
the underlying OF API as possible and not have extra calls for platform
specific items.

arturoc:
i've been going through this with jonas and it seems good to go now, going to merge it, there's still some issues with the front camera being mirrored on some devices.

@danthemellowman we've agreed on leaving this as android only for this release since there's already plans to have it as a core feature, we'll add it for 0.10 in ofBaseVideo so it forces everyone to implement it in their platforms

thanks!

danthemellowman:
Awesome!  I'll trash my branch and pull this in so I can test with my app.

@arturoc haha that makes sense! Should it default to using Pixels and make people disable it for faster frame rates?

arturoc:
yes that's the default already

PR #4513, opened 2015-11-02
Android: Raise min sdk version to 11 (Android 3.0)

HalfdanJ:
See more here: #4505
Android 3.0 is so old that i think its a very fair requirement, and simplifies code in the camera class.


HalfdanJ:
We'll wait with this to a later point with a larger bump.

PR #4516, opened 2015-11-02
Android orientation

HalfdanJ:
- Implements ```deviceOrientationChanged(ofOrientation newOrientation)``` in ofxAndroidApp like in iOS that gets called whenever the orientation on the app is changed
- Implements support for ```ofSetOrientation(OF_ORIENTATION_90_RIGHT)``` and ```ofSetOrientation(OF_ORIENTATION_180)``` (these were implemented as their flipped versions before)

PR #4517, opened 2015-11-03
Android camera: Orientation working

HalfdanJ:
I've gotten the orientation to work for the android camera. It's mostly done in  the example code to show how to correctly rotate the image based on the app and camera orientation. But i've also removed the matrix multiplication on the texture, since it only confused, and made the raw pixel data and texute be different, and now there is no mirroring of the textures on frontal cameras (easy for the user to implement if thats the use case) 

The example code only works correctly with PR #4516 merged in, since the app orientation change event 
otherwise doesn't work. 
 
Tested on Nexus6, Nexus9 (tablet) and Fairphone 1


arturoc:
mmh it seems weird that the user has to take care of fixing the rotation, we could bake the fix for the mirroring in the texture matrix or something like that?

HalfdanJ:
there is no fix for mirroring, the matrix that previously was multiplied to the texture mirrored the image. The user has to counter rotate the texture when the device is rotated, and take in account the orientation of the camera, but there are functions now to get that knowledge. I think its weird if we rotate the texture based on the device rotation without the users knowledge

HalfdanJ:
also, i think its important that the pixel data and texture are the same, otherwise it will be really hard to synchronize opencv overlay with a texture, so if we decide to bake rotation in, then it should also happen on the pixel data

arturoc:
this won't merge any more can you take a look?

HalfdanJ:
Fixed, note that the orientation in the example is not responding on rotated device until #4516  has been merged (that implements the app orientation listener)

arturoc:
mmh, ok then i'll wait until @tallavi refactoring is merged since that's been waiting for ever and merging anything else android related will make merging it more complicated

HalfdanJ:
Seems like a good idea

ons. 3. feb. 2016 kl. 09.05 skrev arturo <notifications@github.com>:

tallavi:
:-)

HalfdanJ:
@arturoc ok to merge?

PR #4520, opened 2015-11-03
OpenAL support in MSYS2 (32bits only)

oxillo:
Replace FMOD by openAL in MSYS2.


DomAmato:
I think my concern here is that it doesn't support both 32 and 64 bit architectures, I would be hesitant to add features to the core that dont work universally since people will undoubtedly raise issues here and in the forum that they are getting an error building in 64 bits.

oxillo:
At this time, MSYS2 only support 32 bits. One of the blocking point to go to 64 bits is the FMOD player as we talked with @arturoc. This PR is/was the first step towards the 64 bits support in MSYS2 (cf PR #4526).
The target of the PR is 0.9.1 as this PR is too close to 0.9.0 release. 
Should we keep this PR separated from #4526 or delete it to keep #4526 only ?

DomAmato:
I guess theres no harm in keeping them separate, certainly default to what the other maintainers suggest and whatever is easiest.

I think FMOD was talked about being removed like, back in 0.7, so if anything this might be a good push to start phasing it out

arturoc:
thanks! what version of poco is there in msys2? there's usually pretty old versions of poco in official repos, at least in debian

oxillo:
POCO is version 1.6.0-2 in MSYS2.

arturoc:
oh, great i guess it's better to leave it as it is for 0.9 but good idea to remove it for next release.

thanks!

libs/openFrameworks/sound/ofFmodSoundPlayer.cpp
```diff
@@ -1,3 +1,7 @@
+#include "ofConstants.h"
+
+#ifdef OF_SOUND_PLAYER_FMOD
```
arturoc: can you remove this? sometimes we don't have that define but still want to use it optionally like in linux and this would prevent from doing so. i guess if we have the headers even if the library is not linked it should compile. also we can remove this file from the makefiles by excluding it for mingw 32bit

oxillo:
This PR is messy. Removing poco, adding Kiss apothecary, ...
I close it and submit one for each subject

PR #4521, opened 2015-11-04
Changelog 0.9.0

tpltnt:
Hi oF people,

I created the changelog according to the meeting on Sunday. I went through the PRs and excluded only code formatting changes, typo corrections, warning fixes and bug fixes to features which were introduced since 0.8.4.

Cheers,
tpltnt


kylemcdonald:
This is great! Going through all this must have been a ton of work... thank you :)

A few thoughts:

- This might need one more pass for double-checking that things are in the right place. We don't usually have a "General" category at the top, and it's fairly long with this release. It might be possible to divide it up better. We can merge before that happens though.
- It looks like there are 168 lines added. But if I remember correctly, there were around 500 PRs. I know a lot of them were things to ignore (code formatting changes, typo corrections, etc.) but was it really only 2/5 PRs that actually need a note in the changelog? It would be great if you could comment on any trends you noticed, for example: were there many changes spread across multiple PRs that can be summarized succinctly?
- It looks like this PR itself introduces a bunch of commits (48) instead of just one for updating the changelog. It might have something to do with how you resolved the merge conflict. My suggestion is to make a manual copy of the changelog you have, then make a new branch from the freshest OF/master, and replace the old changelog with the new one, finally pushing this new branch and sending a PR for it. If you'd like any help with this, please let me know!

jvcleave:
+1 on the General section. I also felt it needed it as I was going through it

bilderbuchi:
>  was it really only 2/5 PRs that actually need a note in the changelog?

Just as a single datapoint (I know, bad statistics), adding Travis support was spread over ~5 PRs before it was in a shape I was satisfied with.

One thing I think would be **very** useful (from my own experience digesting changelogs), would be a "summary" section on top, describing the major, high-level changes in about <=7 bullet points. This is especially true with such a huge and detailed changelog.

tpltnt:
I take in your comments and do a new PR.

PR #4524, opened 2015-11-05
fix ofMouseEventArgs for scroll event

### LINKED ISSUE #4523: ofMouseEventArgs inconsistent for scrolling event
All mouse events appear to consistently set `x` and `y` of `ofMouseEventArgs` to the current mouse cursor position.

When capturing a scrolling event, however, `x` and `y` represent the scrolling offset. When using a unique responder to mouse events, similar to what #2176 is suggesting, this means two extra global calls to `ofGetMouse[XY]`, which seems far from elegant. 

Wouldn't it be nice (and less surprising) to have the scroll xy-offset (which appears to be a float rather than an int anyway) in its own member variable of `ofMouseEventArgs`?

With a slight change to `ofBaseApp.h#L107` (to forward the new scroll offset rather than the mouse pos) the current beginner-friendly API for mouse scroll events could be kept, while more advanced users might profit from being able to tell where on screen the scroll event has occured.

tgfrerer:
When scrolling, `ofMouseEventArgs` would report the scroll
offset in the same place where it usually would report the
current mouse coordinates: `ofMouseEventArgs.[xy]`

This meant that to find out where on screen a scroll event
occured, you had to call `ofGetMouse[XY]` when responding to a
scroll event.

All other mouse events, e.g. Press/Release pass on the
current mouse position as `ofMouseEventArgs.[xy]`.

This brings the scroll event into line, by:

+ adding `scrollX` and `scrollY` to `ofMouseEventArgs`
+ passing the current mouse position in `.x` and `.y`,
  in line with other mouse event responders

I've double-checked our legacy GLUT window, and
ofAppEGLWindow if they needed updating too, but these window
classes appear not to capture scrolling events anyway.

Closes #4523

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -963,7 +963,7 @@ void ofAppGLFWWindow::entry_cb(GLFWwindow *windowP_, int entered) {
 void ofAppGLFWWindow::scroll_cb(GLFWwindow* windowP_, double x, double y) {
 	ofAppGLFWWindow * instance = setCurrent(windowP_);
 	rotateMouseXY(instance->orientation, instance->getWidth(), instance->getHeight(), x, y);
-	instance->events().notifyMouseScrolled(x, y);
+	instance->events().notifyMouseScrolled(instance->events().getMouseX(), instance->events().getMouseY(), x, y);
 }
 
 //------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


arturoc:
this seems much more correct than using x,y for scrolling in the scroll event and position in the rest and we should merge it before 0.9 so we don't break the api later. if there's no problem with it i'll merge it later

bakercp:
+1 for this.  I think the reason it was this way previously is because it was mirroring glfw / glut, which just provided the scroll x/y in their callbacks.

arturoc:
thanks!

PR #4526, opened 2015-11-05
Msys2 64bits

oxillo:
64 bits support with MSYS2. Use OpenAL for sound support


oxillo:
integrated in master with #6413

PR #4527, opened 2015-11-05
Improve variable quoting in cleanAllExamples.sh and buildAllExamples.…

raphaelss:
…sh linux scripts and fix minor formatting/whitespace issues.

Most important contribution is the use is of quotes in the variables in the lines containing "rm -rf" in cleanAllExamples.sh as a undefined, empty or variable containing spaces could cause unrelated files to be deleted. I used the "${VAR:?}" idiom that terminates the script immediately if the variable is empty (whitespace only) or unset.
I could have put it in only one appearence of the variables, but it makes sense to put it  in every statement so that if more code is inserted in between them, an eventual accident is avoided.
Considering the context in which the variables are used, it's unlikely that we will ever have a problem, but in a script that is meant to be run as root, it's better to be safe when running 'rm -rf'

Added quotes to variables where necessary in other less critical places (where spaces could generate issues) and fixed formatting/whitespace/style issues on the way.

### DIFF
--- a/scripts/linux/buildAllExamples.sh
+++ b/scripts/linux/buildAllExamples.sh
@@ -5,34 +5,34 @@ export LC_ALL=C
 for category in $( find ../../examples/ -maxdepth 1 -type d )
 do
     if [ "$category" = "../../examples/android" -o "$category" = "../../examples/ios" -o "$category" = "../../examples/" ]; then
-            continue
+        continue
     fi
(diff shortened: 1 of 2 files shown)


bilderbuchi:
nice, thanks. ping @arturoc for review

arturoc:
thanks!

PR #4528, opened 2015-11-05
platform list

tpltnt:
Hi,

another attempt at addressing #3506.

Cheers,
tpltnt

### DIFF
--- a/docs/PLATFORMS.md
+++ b/docs/PLATFORMS.md
@@ -6,7 +6,7 @@ versions: 2.2+
 
 architectures:
   * armeabi-v7a
-  * x86 (i686)
+  * x86
 
 
 # Emscripten #
@@ -21,7 +21,7 @@ versions: iOS 5 - 8 (FAT Bundle with libc++ for all archs)
 architectures:
   * ARMv7
   * ARM64
-  * i386
+  * x86
   * x86_64
 
 notes:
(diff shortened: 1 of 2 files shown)


kylemcdonald:
this looks correct, and all the ?s have been replaced. :+1:

docs/PLATFORMS.md
```diff
@@ -0,0 +1,68 @@
...
+# Microsoft Windows #
+
+versions: Win7 - Win10
+
+architectures:
+  * x86
+  * x86_64
+
+notes:
+  * Visual Studio 2012 for x86
```
arturoc: this is not correct it's vs2015 for every platform

docs/PLATFORMS.md
```diff
@@ -0,0 +1,68 @@
...
+  * Visual Studio 2015
+  * MSYS2
+  * MinGW / Codeblocks 12.11 for x86
+
```
arturoc: we don't support codeblocks anymore, the msys2 download only has support for x86 by now

docs/PLATFORMS.md
```diff
@@ -0,0 +1,68 @@
+openFrameworks currently supports the following platforms:
+
+# Android #
+
+versions: 2.2+
+
+architectures:
+  * armeabi-v7a
+  * x86 (i686)
```
arturoc: i think it's better to specify x86 only, in all the platforms, i686 is correct in the case of android but i386 specified in other places is surely not compatible with the libraries we ship

arturoc:
thanks!

PR #4530, opened 2015-11-06
fix ofEasyCam scroll behaviour

tgfrerer:
Unfortunately #4524 introduced an issue with ofEasyCam, which was tracking the mouse argument's `.y` instead of the new `.scrollY`.

This PR fixes this.


arturoc:
we should grep all the core and addons for appeareances of the scroll event. ofxGui is also probably broken

tgfrerer:
I'll look into it.

CNek:
I'm not sure if it is related but i get strange scrolling behavior too with ofEasyCam :
Scrollwheel zoom's are fluid and smooth until I pan, it becomes stepped (OS X default system behavior) i have also some "jumps" when panning and scrolling as they were'nt aware of each others movements.

PR #4533, opened 2015-11-07
fixing closing and restarting ofxAndroidVideoGrabber

danthemellowman:
fixes closing the Android Video Grabber

### DIFF
--- a/addons/ofxAndroid/ofAndroidLib/project.properties
+++ b/addons/ofxAndroid/ofAndroidLib/project.properties
@@ -9,4 +9,4 @@
 
 android.library=true
 # Project target.
-target=android-22
+target=android-19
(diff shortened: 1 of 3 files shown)


HalfdanJ:
Uh, forgot that one! You removed close from setDeviceID, in that case the example app should be updated to call close when swapping camera.

arturoc:
i think that's how it works currently in other platforms so i'm going to merge it but it's probably a good idea to make it restart if you change the device ID we could add that as a feature in ofVideoGrabber

HalfdanJ:
Let me then update the example app so it works again. I will add a log notice if you change the device id while its running

arturoc:
ok, thanks!

PR #4536, opened 2015-11-08
Fix android camera example

HalfdanJ:
Adds in close an setup in swapping of camera. I dont like that you have to close and open it this way, but thats the only way to do it with the current api


arturoc:
going to close this since @HalfdanJ is going to revert to setDeviceID resetting the grabber

PR #4540, opened 2015-11-08
Android examples

HalfdanJ:
- Deleted 3dModelLoaderExample (since the addon does not exist) 
- Updated build target to sdk 22 on androidVibratorExample (like all other examples)


arturoc:
thanks!

PR #4541, opened 2015-11-08
Android script for building and running all android examples

HalfdanJ:
I've created two scripts that will build and run all android examples (on mac os at least)
It requires some enviroment variables set (see top of the scripts)


arturoc:
nice thanks!

PR #4555, opened 2015-11-10
Fix for copying of ofTrueTypeFonts

NickHardeman:
Changed the face variable to be a shared_ptr that only gets deleted when the use count is 1 or less.
Currently the face variable is not getting copied over and will crash on copy.
To replicate the issue that this fixes:
`ofTrueTypeFont tfont;
tfont.load("verdana.ttf", 14, true, true);

verdana14 = tfont;
verdana14.drawString("Font Example - use keyboard to type", 30, 35);`

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -375,8 +375,9 @@ ofTrueTypeFont::ofTrueTypeFont(){
 
 //------------------------------------------------------------------
 ofTrueTypeFont::~ofTrueTypeFont(){
-
-	if(face) FT_Done_Face(face);
+	if(face && face.use_count() <= 1) {
+		FT_Done_Face(*face.get());
+	}
 	#if defined(TARGET_ANDROID)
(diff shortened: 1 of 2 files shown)


ofTheo:
this PR looks good to me! 
note: that this issue didn't exist in 0.8.4 so not sure what changed for 0.9.0 to introduce the bug.

arturoc:
the problem is that the face was being destroyed after load finished in 0.8.4, we are now keeping it around to query it for some info that we didn't made available before.

there's a better way to do this which is  passing FT_Done_Face to the shared_ptr when creating it as it's second argument, that way the shared_ptr will call that function instead of delete when the last instance gets destroyed.

in any case there's some more issues with that i think, mostly related to the android calls, always there's a destructor there should be a copy constructor and equal operator that in this case are missing. i'm going to close this and fix everything in another PR

arturoc:
i've fixed this here:

https://github.com/openframeworks/openFrameworks/commit/df6be4dfd2d9372fe4464e12f6146ca68e6c94d4

along with other similar problems with copy/move constructors and assigment operators

there's still some more similar cases but i'll open a different issue

Issue #4559, opened 2015-11-11
Windows 10 0.9.0 Build Fail OF_TARGET_WINGCC


JordanParsons:
I found what seems to be a simple bug when building 0.9.0 under windows 10 with Msys. OF_TARGET_WINGCC is not defined and the build fails. 

If OF_TARGET_WINGCC is defined inside of ofConstants.h with the other platform defines oF builds fine. Below is the error and what I did to fix.
**Error**
```
D:/Code/openFrameworks/libs/openFrameworks/utils/ofUtils.cpp: In function 'ofTargetPlatform ofGetTargetPlatform()':
D:/Code/openFrameworks/libs/openFrameworks/utils/ofUtils.cpp:1016:16: error: 'OF_TARGET_WINGCC' was not declared in this scope
         return OF_TARGET_WINGCC;
```
**ofConstants.h**
```c++
enum ofTargetPlatform{
	OF_TARGET_OSX,
    OF_TARGET_MINGW,
    OF_TARGET_WINGCC,
	OF_TARGET_WINVS,
	OF_TARGET_IOS,
	OF_TARGET_ANDROID,
	OF_TARGET_LINUX,
	OF_TARGET_LINUX64,
	OF_TARGET_LINUXARMV6L, // arm v6 little endian
	OF_TARGET_LINUXARMV7L, // arm v7 little endian
	OF_TARGET_EMSCRIPTEN
};
```

I cannot find any other references to OF_TARGET_WINGCC so maybe this isn't the best fix, but it seems accurate. I can submit a pull request for what I changed if people think its the right way to go.

arturoc:
hey thanks. the correct fix is actually to change that line in ofUtils to return OF_TARGET_MINGW instead of adding a new platform in ofConstants.

JordanParsons:
Ok sorry about that. Should I submit the correct fix?

arturoc:
yes please

JordanParsons:
Ok #4561 is updated to fix this. Sorry for the confusion, still working out the details on contributing.

oxillo:
@JordanParsons 
Just to make sure. Your are building with MSYS not MSYS2. Is that correct ?

arturoc:
@oxillo no, this is a problem i introduced some days ago, i changed the platform enums from win_cb to mingw which was still missing in the changes you did but missed that one.

@JordanParsons thanks!

JordanParsons:
[...]

PR #4561, opened 2015-11-11
Updated OF_TARGET_WINGCC changed to OF_TARGET_MINGW 

JordanParsons:
Fix for #4559 adding OF_TARGET_WINGCC to the ofTargetPlatform enum in ofConstants.h. This fixes build issues on Windows 10 when using GCC with Msys2.


JordanParsons:
Closing this, its the wrong fix.

JordanParsons:
Updated to have ofUtils return OF_TARGET_MINGW instead of adding a a new platform.

arturoc:
Theo, i was waiting until the tests were done, the change was pretty clear but better to let the tests finish, that way we know everything is correct for sure.

PR #4565, opened 2015-11-12
added note about sharing context between windows

genekogan:
added a commented line about sharing window contexts, which is necessary to share pixels between windows. this might be helpful to include in the example if it's not clear from the docs.

### DIFF
--- a/examples/events/multiWindowOneAppExample/src/main.cpp
+++ b/examples/events/multiWindowOneAppExample/src/main.cpp
@@ -15,6 +15,8 @@ int main( ){
 	settings.height = 300;
 	settings.setPosition(ofVec2f(0,0));
 	settings.resizable = false;
+	// uncomment next line to share main's assets with gui
+	//settings.shareContextWith = mainWindow;	
 	shared_ptr<ofAppBaseWindow> guiWindow = ofCreateWindow(settings);
 	guiWindow->setVerticalSync(false);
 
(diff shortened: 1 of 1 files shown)


arturoc:
can you set assets with opengl resources? it would be also useful to comment out that some resources like shaders or fbos can't be shared. in general everything that is memory is usually shareable but everything that is just representing state or logic can't be shared

genekogan:
why can't fbos or shaders be shared? 
maybe it makes sense to put this into a tutorial or into the docs? rather than overloading the example with all the details? i thought for the example should cover big use cases, like sharing images. 
but i am also not clear on what can and can't be shared.

tgfrerer:
Because openFrameworks objects like ofFbo or ofVbo encapsulate OpenGL objects, which are, really, just a way for OpenGL to capture opengl (driver) state. (what is plugged into what and how internal variables are set, and which OpenGL objects correspond to what OpenGL ids etc.)

If you were to render such an object in another rendering thread there is no guarantee (it is in fact most unlikely) that this rendering thread is in the same state as the current render thread, which would lead to race conditions and (outrageously) undefined behaviour.

The same does not apply to things which are "pure memory," that is, effectively just an address in VRAM, these might be shared in a shared context, but you might have to (manually, using GL sync objects) sync them. Doing shared contexts right can be real hard!

arturoc:
yes perhaps it makes sense to leave the details to the docs. you can copy/paste what tim just posted :)

but in this PR can you change assets with opengl resources?

thanks!

arturoc:
thanks!

PR #4570, opened 2015-11-12
bugfix for drawing wireframe and verts on an assimp model. closes #4553

### LINKED ISSUE #4553: ofxAssimpModelLoader drawVertices drawWireframe doesn't work.
Trying to draw a wireframe or point version of a model draws the model with faces. 
However getting the meshes from the model and calling drawVertices does work. 

It looks like it might be this line:
https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp#L749

The #ifndef should be #ifdef ? 

Changing that fixes the issue on OS X.

ofTheo:
see #4553 
added checks to avoid calling glPolygonMode too many times.

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
@@ -722,6 +722,8 @@ void ofxAssimpModelLoader::draw(ofPolyRenderMode renderType) {
     ofPushMatrix();
     ofMultMatrix(modelMatrix);
     
+    int curRenderType = -1;
+    
     for(unsigned int i=0; i<modelMeshes.size(); i++) {
         ofxAssimpMeshHelper & mesh = modelMeshes[i];
         
@@ -747,6 +749,13 @@ void ofxAssimpModelLoader::draw(ofPolyRenderMode renderType) {
         
         ofEnableBlendMode(mesh.blendMode);
 #ifndef TARGET_OPENGLES
+        
+        //avoid setting it over and over again for the same type.
+        if( curRenderType != renderType ){
+            glPolygonMode(GL_FRONT_AND_BACK, ofGetGLPolyMode(renderType));
+            curRenderType = renderType;
+        }
+        
         mesh.vbo.drawElements(GL_TRIANGLES,mesh.indices.size());
 #else
         switch(renderType){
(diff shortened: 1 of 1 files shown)


arturoc:
mmh, doesn't seem correct, if the passed type is wireframe for example, it'll be set to faces at the end but then never set back to wirframe later no?

also i'm just checking and the renderer has a way to do this so just call:

```
ofGetCurrentGLRenderer()->draw(mesh.vbo, renderType);
```

btw, the travis servers we've setup check the PRs in almost every platform so it's good practice to leave it run until it tests every platform to check that nothing is wrong before merging

ofTheo:
@arturoc  thats what I intended. 

I'm assuming here that before the draw call glPolygonMode is set to Faces. 
So the function always restores the state to faces after it finishes drawing all the meshes.  

If you call model.drawWireframe() it should end up restoring it to faces after the end of the drawWireframe() call. 

Thinking about it though - I don't need to be doing the glPolygonMode in the loop. 
All meshes get drawn with the same mode so it could just be set once. 

Happy to push the push/pop attributes back if you think thats a better approach.

arturoc:
i think the best would be to call the renderer which already does the right thing

ofTheo:
@arturoc - sorry somehow missed the ofGetCurrentGLRenderer() part of your reply. 
That makes total sense. Better to have less GL calls in the code too.

ofTheo:
@arturoc - so the issue with `ofGetGLRenderer()->draw(mesh.vbo, renderType, 0, mesh.vbo.getNumVertices());` 
is that you don't get outlines of the triangles as you do with glPolygonMode. 

It looks like this: 
screen shot 2015-11-16 at 12 05 38 pm

vs: 
screen shot 2015-11-16 at 12 06 49 pm

Without using glPolygonMode I think we would need to create a new mesh to draw triangles as outlines. 
As GL_TRIANGLES can't easily be converted to GL_LINE_LOOP ( needs multiple begin / end ) and GL_LINES expects a pair of points per line.

julapy:
in the past ive written helper functions which convert a GL_TRIANGLE_STRIP mesh to a GL_TRIANGLES mesh, which can then be drawn as a wireframe.

i don't think this should be done internally or automatically.
but maybe these ofMesh helper functions should be included inside ofMesh as static functions?

PR #4577, opened 2015-11-14
Improve logging while building opencv in osx

r9y9:
The fixes are summarized as follows:

- add `-a` option to `tee` to append outputs to the log file
- redirect stderr to stdout and save both to log file

It could be useful for debugging.

Please see the second commit for the actual changes. First one tweaks trailing white-spaces (my editor automatically does).


arturoc:
thanks!

Issue #4566, opened 2015-11-12
iOS Template Project needs LaunchScreen and Image Assets to support new devices


danoli3:
Need to review and add the LaunchScreen.xib and change over to Assets. to support native resolutions of iPhone 6/6s / iPhone 6/6s Plus, iPad Pro

This will then be used and built into projects built with the project generator and prevent this issue some users are experiencing.

ofZach:
if you alter the xcproj file, be sure to use plutil to convert it back to xml -- the project generator parses the xml file (it doesn't read it as a plist).  

http://hints.macworld.com/article.php?story=20050430105126392

on first open an xml based xcproj file is turned into a plist again....

danoli3:
Thanks @ofZach I knew I had to do this, but was going to have to look into how to do it. Cheers! :+1:

julapy:
@danoli3 i was thinking it might be time to introduce a `media assets` folder to the ios xcode template project... seems to be a much better way of managing icons and splash screens assets.
what do you recon?

danoli3:
@julapy Yeah sounds like a great idea

danoli3:
Okay since we need new resolutions / images... any ideas for new image style? 

Current from 0.7:
Icon:
![icon-72 2x](https://cloud.githubusercontent.com/assets/830748/11178285/fc50c034-8c9f-11e5-8299-b380e16addeb.png)

Splash:
![default](https://cloud.githubusercontent.com/assets/830748/11178359/71da153a-8ca0-11e5-9642-0a4738ca86e3.png)

--

OSX has a white oF logo and black textured background. I'll have a play around with the AppIcon template and the logo

danoli3:
Hows this look for AppIcon: (Going for the Black oF on white/silver gradient).
<img width="558" alt="screen shot 2015-11-17 at 4 38 10 am" src="https://cloud.githubusercontent.com/assets/830748/11189648/429d338e-8ce5-11e5-967b-0f03ae487316.png">

danoli3:
Splash:
![default-portrait 2x iphone4](https://cloud.githubusercontent.com/assets/830748/11190360/260d7eb4-8ce9-11e5-91b3-c54e8fb8658f.png)

ofTheo:
awesome! :)

danoli3:
Hmm Travis is having an issue with the storyboard... looking into this.

julapy:
icons and splash screens are looking very slick!
will review the PR now.

[...]

PR #4591, opened 2015-11-16
iOS Template Project Update - Icon / Splash / iPhone6 / iPad Pro

danoli3:
Fixes issues related to iPhone6 and greater not being supported by template project out of the box.
Fixes: https://github.com/openframeworks/openFrameworks/issues/4566

- Adds Launch Screen
- Adds ImageAssets for Icons / SplashScreen
- New and updated Icon and SlashScreen for iOS
- iOS 7 Minimum for Project.
screen shot 2015-11-17 at 4 38 10 am


julapy:
hey @danoli3,
one small quirk ive noticed with media assets,
sometimes it unnecessarily makes a copy of files, even though they are the same resolution.
so you may notice that you have some files ending with `-1`, like `Icon-Small@2x-1.png`
its basically duplicated `Icon-Small@2x.png` twice

so what i usually do to fix this is go into the media assets json file,
and change `Icon-Small@2x-1.png` => `Icon-Small@2x.png` (for example) where ever there is a duplicate. and then delete `Icon-Small@2x-1.png` (the duplicate file)

can see that it has happened a few time in this PR.

danoli3:
@julapy Ah, Good call! I noticed the icons were way too big as well. Fixed that too!

danoli3:
Yeah to support iOS 9 devices, iPad Pro, iPhone 6s, iPhone 6s Plus and all devices running iOS 9 we need to set Minimum Xcode version to 7.0 (which is fine as you need it to submit to the AppStore, and it is free upgrade for all platforms that Xcode 6 ran on).

HalfdanJ:
Seems like two of the images are changed, but the resolution is the same (`scripts/templates/ios/mediaAssets/Images.xcassets/AppIcon.appiconset/Icon-76.png` and `scripts/templates/ios/mediaAssets/Images.xcassets/AppIcon.appiconset/Icon-76x2.png`) 
They are lower quality now (look more fuzzy). Is that on purpose?

danoli3:
Probably a bug with my exporter I'll have a look thanks

danoli3:
Going to re-pr this one. Squash commits. Ah yes @HalfdanJ the quality changed on those icons as those original images were like 400kb for a 76x76... compressed versions still looking great on devices with current.

danoli3:
Closing in favour of new PR.

Issue #4556, opened 2015-11-11
ofxTCPManager broken on Windows


traumedy:
ofxTCPManager::Receive() calls recv() which returns -1, WSAGetLastError() returns WSAEWOULDBLOCK which is correct because the socket is non-blocking, but it is still returned as an error so for instance when called from ofxTCPClient::receiveRawBytes():
```
	messageSize = TCPClient.Receive(receiveBuffer, numBytes);
	//	0 is not an error... -1 is
	if ( messageSize == SOCKET_ERROR ){
		close();
```
There are related issues from other places it is called, for instance ofxTCPClient::receiveRaw():
```
	messageSize = TCPClient.Receive(tmpBuff, TCP_MAX_MSG_SIZE);
	if(messageSize==0){
		close();
	}else if(messageSize<TCP_MAX_MSG_SIZE) {
        // null terminate!!
        tmpBuff[messageSize] = 0;
```
tmpBuff[-1] would be overwritten in this case.

Also by the time ofxNetworkCheckError() is entered WSAGetLastError() returns 0, I think the error value has been reset by some successful call to a Windows API in between (I read in some places that WSAGetLastError() just wraps GetLastError()).

ofTheo:
thanks for the bug report @traumedy !
we're aiming to do a 0.9.1 release soon to address small bugs/regressions in 0.9.0 

Could you submit a PR with fixes for the issues you mentioned?
It would be great to double check any changes before you submit it. 

thanks!!

traumedy:
I'm guessing this is broken for Linux and OSX too from the looks of it.

I can do the leg work, but it might be more than a simple fix or two.  Some
minor design stuff might should be rearranged regarding error handling.

On Tuesday, November 10, 2015, Theodore Watson <notifications@github.com>
wrote:

> thanks for the bug report @traumedy <https://github.com/traumedy> !
> we're aiming to do a 0.9.1 release soon to address small bugs/regressions
> in 0.9.0
>
> Could you submit a PR with fixes for the issues you mentioned?
> It would be great to double check any changes before you submit it.
>
> thanks!!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openframeworks/openFrameworks/issues/4556#issuecomment-155631395>
> .
>


-- 
*JOSH BUCHBINDER* | Interactive Developer | +1 925 351 8049 | *OBSCURA*
<http://www.obscuradigital.com/> | SF
[...]

PR #4592, opened 2015-11-17
This at least gets basic TCP networking working again by fixing three…

traumedy:
… issues:

Addressing issue #4556

The macro ofxNetworkCheckError() and the function it calls ofxNetworkCheckErrno() make no attempt to convert parameters to strings avoiding calling any function that might reset the errno or (WSA)GetLastError() value.
ofxNetworkCheckErrno() will not generate a log entry if (WSA)EWOULDBLOCK is the error.
ofxTCPManager::Receive() checks the error returned from ofxNetworkCheckError() and sets the return value to 0 (instead of -1) if (WSA)EWOULDBLOCK is returned, indicating no data available and not an error.

### DIFF
--- a/addons/ofxNetwork/src/ofxNetworkUtils.h
+++ b/addons/ofxNetwork/src/ofxNetworkUtils.h
@@ -19,9 +19,9 @@
 #endif
 
 
-#define ofxNetworkCheckError() ofxNetworkCheckErrno(__FILE__,ofToString(__LINE__))
+#define ofxNetworkCheckError() ofxNetworkCheckErrno(__FILE__, __LINE__)
 
-inline int ofxNetworkCheckErrno(const string & file, const string & line){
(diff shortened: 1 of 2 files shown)


traumedy:
I see from https://travis-ci.org/openframeworks/openFrameworks/builds/91533458 that my commit broke the build on some other platform because:
#if !defined(TARGET_WIN32)
	case OFXNETWORK_ERROR(AGAIN):
...

is the same value as case OFXNETWORK_ERROR(WOULDBLOCK).
This seems to be a somewhat involved Unix/Linux issue:
http://stackoverflow.com/questions/7003234/which-systems-define-eagain-and-ewouldblock-as-different-values
EAGAIN might be defined as the same or different value as EWOULDBLOCK depending on the OS.

This could be fixed with some clever pre-processor instructions, like:
#ifdef EAGAIN
and
#if EAGAIN == EWOULDBLOCK
...

PR #4603, opened 2015-11-19
Fix UDP timeouts not working

### LINKED ISSUE #4597: ofxUDPManager timeouts do nothing in blocking mode
When ofxUDPManager is set to block on IO, SetTimeoutReceive and SetTimeoutSend do nothing. 

udpManager receive will block forever even if SetTimeoutReceive is set if there is no data.

SetTimeoutReceive and SetTimeoutSend both simply set their respective member variables that are never used.

zminjie:
fixes #4597 

Tested on OSX 10.11.1
Not tested on windows

### DIFF
--- a/addons/ofxNetwork/src/ofxUDPManager.cpp
+++ b/addons/ofxNetwork/src/ofxUDPManager.cpp
@@ -345,10 +345,28 @@ int	ofxUDPManager::Receive(char* pBuff, const int iSize)
 void ofxUDPManager::SetTimeoutSend(int	timeoutInSeconds)
 {
 	m_dwTimeoutSend= timeoutInSeconds;
+
+	if (m_hSocket != INVALID_SOCKET) {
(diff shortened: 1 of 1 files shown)


arturoc:
looks good to me but i'm pretty sure that timeval is posix only so it won't compile in windows. we need some implementation for win too and the corresponding ifdefs

ofTheo:
going to ping @traumedy who is looking at ofxNetwork on Windows. 
maybe @traumedy you could do a PR against @zminjie 's branch?

arturoc:
i'm going to close this since this way of setting the timeout wont' work on non-blocking sockets. 

i've implemented it using select to wait until there's data to receive or the socket is available to send so it works both with blocking and non blocking sockets here:

https://github.com/openframeworks/openFrameworks/commit/405fa8bd86505b1c256e52d743aa7bb208c81848

PR #4610, opened 2015-11-23
ofFIleUtils tidying up

2bbb:
- add `std::`
- add iterators `cbegin`, `cend`, `crbegin`, `crend`
- return values change from copy to const reference if that was forgiven.


tpltnt:
looks good to me.

2bbb:
i think this PR is not needed already.

PR #4613, opened 2015-11-23
better fix for drawing wireframes on non GLES addresses #4553

ofTheo:
fixes regressions in 0.9.0 
doesn't solve the ability to do proper wireframe outlines on ES - but that was the case for 0.8.4 as well.


arturoc:
looks better. at some point we should add a function to the renderers to draw vbos with polygon mode as we have for meshes. right now the renderer is letting the polygon mode to lines or fill depending if ofFill or ofNoFill was called so this is still not 100% correct but let's leave it like this by now and i'll refactor the renderer for 0.10

Issue #3851, opened 2015-05-19
Using ofSoundStream and ofSoundPlayer together on OSX creates terrible noise


dimitre:
here is the minimum code to reproduce (on git master)
in ofApp.h
```c++
    ofSoundStream soundStream;
    ofSoundPlayer soundPlayer;
```
in ofApp.cpp
```c++
void ofApp::setup(){
    soundStream.setup(this, 2, 0, 44100, 128, 4);
    soundPlayer.load("iuea-violao2.wav");
    soundPlayer.stop();
}
```

This sequence generates an audio glitch (it was not present until the great update some days ago)
if I comment soundStream.setup line it doesn't happen. 
If I comment soundPlayer.stop() line it doesn't happen too
tested with wav and aif files.
Let me know if you can reproduce it.

arturoc:
what platform are you in?

dimitre:
OS X 10.10 XCode 6.3

admsyn:
I can reproduce this, sounds like an uncleared buffer being looped (different every time). Implementing a buffer-clearing ofApp::audioOut() doesn't fix it.

bilderbuchi:
@openframeworks/sound could any of you take a stab at this?

admsyn:
> any of you

:p

I've been too busy to get involved the last couple months, but should be able to get back into things in ~a week. I won't be able to tackle this immediately though.

The bug seems very similar to #1694, i.e. a conflict between rtAudio and FMOD. I'd guess one of the libs is configuring some shared resource in the low level OSX audio system in a way that's incompatible with the other. Sounds like a hair-puller, but I'll get to it ASAP (though if anyone else would like to jump in, by all means do! :)

bilderbuchi:
> :p

sorry, I only realized after posting. :-P

arturoc:
it seems this happened already sometime ago:

https://github.com/openframeworks/openFrameworks/issues/1694

we solved it by rolling back to the previous version of FMOD but i'm not sure which version that was.

arturoc:
also changing the buffer size to 1024 in the sound stream fixes the problem as it happened with #1694 so it's surely something that could be fixed by rolling back to the old version of FMOD. FMOD ex is anyway deprecated in all it's versions so it shouldn't be a problem to go back a couple of versions

arturoc:
[...]

PR #4615, opened 2015-11-23
set path id to libfmodex.dylib to avoid having to call install_name_tool

arturoc:
calling:

```
install_name_tool -id "@executable_name/libfmodex.dylib" libfmodex.dylib
```

on the library itself rather than on the executable avoids having to call install_name_tool at all later.

this only removes that step from the qtcreator project but could be removed from xcode too.

this could also be a solution for addons we just need the PG to call this on dylibs in the addons folder and then copy those to the bundle and the exe will have the right path


ofZach:
this is interesting --- I wonder if it will be hard to know what's going on if dylibs get recompiled / fmod gets switched for another version, etc...   Ie, if this isn't part of a build process, it would be easy to miss this step.

arturoc:
yes i guess it should be a step in the apothecary scripts. also fmod is deprecated and the next thing will do will be probably remove it.

in any case we can still leave the xcode projects as they are by now, this mostly fixes the problem for qtcreator where it was kind of complicated to add this kind of post step.

but for addons it would be a way to have a universal solution for any dylib

arturoc:
also it can still be a step on the xcode project but run it on the library instead of the executable that way it's easier to make it work for any library

danoli3:
Just a note: This / the old way actually breaks the OS X AppSandbox as it ends up not signed.
Is there a way to package this into the oF lib?

Example Build / Archive Failure:
```
exampleApp.app: code object is not signed at all
In subcomponent: exampleApp.app/Contents/MacOS/libfmodex.dylib
Command /usr/bin/codesign failed with exit code 1
```

danoli3:
@arturoc It seems for OSX fmod is still the defined sound player by default. Not deprecated yet.

arturoc:
i mean that fmodex itself, the library, is deprecated. they released something else sometime ago, fmod studio but the api is completely different

danoli3:
Hmmm this code signing issue is really making it impossible to use for actual production environment (albeit also the GLUT, but that can be easily removed). I'll see if we can compile this into a standard archive.

danoli3:
Ah it seems the dylib itself is broken. 
http://www.fmod.org/questions/question/error-submitting-app-with-fmod-designer-4-44-to-the-mac-app-store/

danoli3:
API is the same @arturoc for this, want me to PR a FMOD Ex Update to the latest on all platforms? / dylibs?

arturoc:
i think we had to stay on the current version because later ones had a problem when using the sound player with the soundstream simultaneously but if you can test if this version fixes that? i think you just need to test an example with both soundplayer and soundstream and one of them where noisy.

danoli3:
I'll try it out now

danoli3:
@arturoc Yeah I tried a merge example of soundBufferExample (sound stream) and soundPlayerExample and it works great!! Actually really cool hahaha

arturoc:
this is the original issue:

https://github.com/openframeworks/openFrameworks/issues/3851

just so you can double check since i remember we tried a lot of newer versions and the one we currently use is the last one that didn't had this problem

danoli3:
@arturoc yeah that specific shrink the buffer size down issue does occur still on latest fmod. However by synchronising the sizes with ```ofFmodSetBuffersize(bufferSize);``` it is fixed.

danoli3:
@arturoc try it out here: https://github.com/openframeworks/openFrameworks/pull/4620

danoli3:
Okay then only issue with this PR is the install location. 
https://github.com/openframeworks/openFrameworks/issues/4619
> Using the executable path for the distribution location (this causes issues in validation).

I changed it in this PR:

Like so:

Copy: 
```
rsync -aved ../../../libs/fmodex/lib/osx/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Frameworks/";
```

Install_name_tool:
```
install_name_tool -change ./libfmodex.dylib @executable_path/../Frameworks/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME";
```

arturoc:
can you then run install_name_tool -id @executable_path/../Frameworks/libfmodex.dylib libfmodex.dylib 

on the libraries both in libs/fmod/lib/osx and in export/osx/libs?

i need this in order for binaries created by qtcreator to work when run from the finder

danoli3:
@arturoc so the command to run would be then:

```
install_name_tool -id "@executable_name/../Frameworks/libfmodex.dylib" libfmodex.dylib
```

arturoc:
yes, that has the same effect as running with -change in the final binary so we wouldn't even need it in xcode either

arturoc:
also if you could add that in the apothecary script that would solve this problem forever :)

danoli3:
@arturoc Yeah I opted for the Apothecary Script rather than  do it manually :D!

1 thing about the fmodex apothecary script, is you have to manually download the dmg and deploy it into the build directory for apothecary... (site requires a login). Ah well it still works.

danoli3:
@arturoc Okay that's done. Want to merge this first then the Fmod version one.

arturoc:
no, just merge yours if i merge this it'll give you conflicts in the binary. i'll close this one and redo the changes in the project files which are pretty minimal

danoli3:
I'm not sure why... but Xcode didn't seem to like the change and I had to manually it regardless.

```
dyld: Library not loaded: @executable_name/../Frameworks/libfmodex.dylib
  Referenced from: emptyExample.app/Contents/MacOS/emptyExample
  Reason: no suitable image found.  Did find:
	/usr/local/lib/libfmodex.dylib: stat() failed with errno=13
```

```@executable_name/../Frameworks/libfmodex.dylib``` is correct.

When I add this to the run script again in Xcode: (however modify the ./libfmodex.dylib to the new changed location... It works.

```
install_name_tool -change @executable_name/../Frameworks/libfmodex.dylib @executable_path/../Frameworks/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME";
``` = Works

**Comment by danoli3** [comment]
Could be just an Xcode thing... but yeah think have to leave that in there.

**Comment by arturoc** [comment]
are you copying the library to the correct folder? something that i don't like much, thinking of it, is that your change will break old projects which shouldn't happen in a minor release

**Comment by danoli3** [comment]
Yeah copied to the correct folder. I think it was more the ""$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME" part that fixed it. Where $PRODUCT_NAME is the App exe.

**Comment by danoli3** [comment]
@arturoc I think your right! I'm going to change the install location back to standard and handle it in Xcode differently if we need.

**Comment by danoli3** [comment]
@arturoc Hmmm... Okay so I ran the apothecary script with what you had above:
```		install_name_tool -id "@executable_name/libfmodex.dylib" libfmodex.dylib```

And the following happens when running the App:
```
dyld: Library not loaded: @executable_name/libfmodex.dylib
  Referenced from: /Users/danielrosser/Documents/openFrameworks/scripts/templates/osx/bin/emptyExampleDebug.app/Contents/MacOS/emptyExampleDebug
```

danoli3:
So I don't think this method is going to work ;( 
I'm going to revert those changes

arturoc:
this was working for me fine yesterday, be sure to do it for the library in export too since it's possible that's the one we are copying, i'll take a look tomorrow too

arturoc:
@danoli3 the problem is that in the apothecary script, here: https://github.com/openframeworks/openFrameworks/pull/4620/files#diff-747f89f0298dc636c3d643f8b105a42dR65 you where calling:

`install_name_tool -id "@executable_name/../Frameworks/libfmodex.dylib" libfmodex.dylib`

cause i told you so :) but it's actually

`install_name_tool -id "@executable_path/../Frameworks/libfmodex.dylib" libfmodex.dylib`

so not executable_name but executable_path

in any case i think the best would be to call 

`install_name_tool -id "@executable_path/libfmodex.dylib" libfmodex.dylib`

on the libraries

and then on the xcode project do

`install_name_tool -change @executable_path/libfmodex.dylib @executable_path/../Frameworks/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME"`

since that will solve the problem in qtcreator but also keep working with the new required path in osx and keep working for old projects.

as far as i understand we don't need the new version of fmod anymore if we merge this: https://github.com/openframeworks/openFrameworks/pull/4622/files

no?

danoli3:
Ah damn!! Nice catch! My Brain just melted I guess. I'll fix that in a few hours and apply it where we were before.

Still need the new fmod version as only the latest dylib works for AppStore submissions now, they changed something that requires a flag that is only set in the latest build of fmod.(Some sort of dylib info thing).

Issue #4619, opened 2015-11-24
OS X AppStore - FMOD issues and problems


danoli3:
As discussed in this PR . https://github.com/openframeworks/openFrameworks/pull/4615
### FMOD:

There are 4 current problems with FMOD when building for OS X (App Store).
- [x] 1. Using the **executable path** for the distribution location of fmod.dylib (this causes issues in validation).
- [x] 2. You must **code sign** all packaged dylibs / frameworks else failure (need to run a script on install to fix this).
- [x] 3. **Outdated fmod library** makes it impossible to validate app build post Xcode 7.0 (Fix only found in  4.44.59+ http://www.fmod.org/questions/question/error-submitting-app-with-fmod-designer-4-44-to-the-mac-app-store)
### Other Issues:
- [x] - You must **package the data folder** into the App Package (like iOS).
- [x] - Your application bundle must install to "**/Applications**"'. (currently set to $HOME/Applications)
- [x] - Must **remove all references to GLUT** and remove copying GLUT into application. (Using prohibited internal framework). 
- [x] - Must **remove references to QuickTime / QTKit**. (as it is deprecated and only 32bit)
- [x] - openFrameworks Lib must be set to **"Skip Install : YES"** (Else every archive of the project will be categorised into "Other Items" and not OSX Apps" in Organiser)

danoli3:
Solutions:
1. Change the path to frameworks
```
rsync -aved ../../../libs/fmodex/lib/osx/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Frameworks/"; 
install_name_tool -change ./libfmodex.dylib @executable_path/../Frameworks/libfmodex.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME";
```
2. Add a new script to automatically code sign the dylib on install
- Currently doing this:
```
codesign --entitlements exampleApp/exampleApp.entitlements -s "Mac Developer: Name (THEID493)" "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/libfmodex.dylib"
```
However I have found a script someone wrote that does this automatically.
3. Update FMOD Ex to 4.44.59 for all platforms.
> Changelog 12/11/15 4.44.59 - Stable branch update (build 69790)
[...]

PR #4620, opened 2015-11-24
Fmod Update to 4.44.59

danoli3:
FMOD Update 

- Fixes part 3 of https://github.com/openframeworks/openFrameworks/issues/4619


arturoc:
moving the discussion here since the other PR is not related. 

The problem with updating fmod is that it produces this problem:

https://github.com/openframeworks/openFrameworks/issues/3851

there's some workarounds but the ideal would be to not have the problem at all.

if i understand it correctly this would fix some signature problem with fmod in osx, isn't there a different fix for this that would still use the old version of fmod?

danoli3:
No fix other than the new FMOD i'm afraid. 
Can we report this issue to FMOD? FMOD Ex is still maintained and active.

arturoc:
another possible fix by now would be to have soundstream call ofFmodSetRate() in osx whenever a new soundstream is created

danoli3:
That's probably the perfect fix yeah

danoli3:
Okay made that change. Tested it. All good no horrible noise with latest. Tried an assertion of different buffer sizes too. All good.

danoli3:
Okay I update the Apothecary script, yeah we totally forgot the license file before.
Also Added the export directory now.

PR #4621, opened 2015-11-24
ofutils: get the clock instance on init instead of everytime.

arturoc:
Haven't tested thread safety but according to @ofTheo it should be ok


ofTheo:
@arturoc haha - was putting together an almost identical PR. 
I'll test this one instead :)

ofTheo:
just checked it with multiple threads and the timing code. 
looks good to me!

arturoc:
ok great!

PR #4622, opened 2015-11-24
OSX AppStore Fixes

danoli3:
Fixes a lot of the issues in: https://github.com/openframeworks/openFrameworks/issues/4619

#### OSX Template Updates:
1. Fixes FMOD using the **executable path**. Moves to the **framework directory**.
2. Adds a **new Run Script** post build to **Code sign frameworks or dylibs** found in the App Package. (Automatic Script, Only runs if code signing set and entitlements).
3. Adds **Copy Resources into App** Package (bin/data) like iOS
4. Fixes Installation Directory bug (Needed to be **/Applications**) instead of $HOME/Applications

#### openFrameworksLib Updates: 
5. The **Skip Install** flag is set to YES. (For archiving/deployment), fixes Organizer issues.

### DIFF
--- a/libs/openFrameworksCompiled/project/osx/openFrameworksLib.xcodeproj/project.pbxproj
+++ b/libs/openFrameworksCompiled/project/osx/openFrameworksLib.xcodeproj/project.pbxproj
@@ -867,6 +867,7 @@
 				OBJROOT = "$(SRCROOT)/../../lib/osx/build/debug/";
 				PRODUCT_NAME = openFrameworksDebug;
 				SDKROOT = "";
+				SKIP_INSTALL = YES;
 			};
 			name = Debug;
 		};
@@ -895,6 +896,7 @@
 				OBJROOT = "$(SRCROOT)/../../lib/osx/build/release/";
 				PRODUCT_NAME = openFrameworks;
 				SDKROOT = "";
+				SKIP_INSTALL = YES;
 			};
 			name = Release;
 		};
--- a/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj
+++ b/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj
@@ -10,17 +10,6 @@
 	<string>46</string>
 	<key>objects</key>
 	<dict>
-		<key>BB4B014C10F69532006C3DED</key>
-		<dict>
-			<key>children</key>
-			<array/>
-			<key>isa</key>
-			<string>PBXGroup</string>
-			<key>name</key>
-			<string>addons</string>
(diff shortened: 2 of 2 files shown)


danoli3:
I would also like to remove references to QTKit, QuickTime and remove GLUT references completely from the template as these are all outdated, the wrong architecture or illegal to package :)

danoli3:
@ofZach would having 2 run scripts be handled in the Project Generator?

ofTheo:
+1 for removing QTKit / GLUT and QT - though this might make our OS min be 10.9 no? 
Would this maybe be for OF 0.10.0 ?

danoli3:
We can do that part for 0.10.0 this fix is for 0.9.1 though.

ofTheo:
@danoli3 with the copying of the files into the resources folder, does the ofApp code need to then change the default data path from ../../../data to ../Resources/data ?

danoli3:
@ofTheo I'm not sure if that should be the default, It really depends how you are using it I guess. 

I think this method of copying the bin/data into the App/resources works for the "read only" / AppStore uses.

However I do know a lot of people make OS X apps that never get "deployed" to any App Store so, you can use the bin/data/ folder as a export directory for results / output etc.

By copying the resources we cover both, and it just relies on the user to set the default if they want to deploy via ```ofSetDataPathRoot("../Resources/");

The current default is:
```../../../data/```

ofZach:
@danoli3 @ofTheo I'm late to the party here but I wonder if we want to do this copying of files into the package by default?  This seems like a big change compared to how things have worked in the past and one that can be really expensive if you are working with big files.  perhaps it would make sense to have a specific template for "app store" apps and something more general (and more like 0.9.0 in terms of using bin/data) for most all / other apps...    cc @kylemcdonald

danoli3:
A specific Template is actually sounding like a good idea for the use case of the old workflow, it may be a better idea to just create one called legacy that doesn't do the packaging.

ofZach:
@ofTheo what do you think?  I think in 99% of the cases, we don't need to do packaging / code signing / moving content into the app package.  the edge case here seems to be preparing an app for app store...   

My suggestion is: 

- osx template (same as 0.9.0)
- osxAppStore (@danoli3's changes) 

if there's important (and more general) changes to the template from this PR we can also copy them over to the 0.9 template

ofTheo:
@ofZach @danoli3 - yeah just replied in #3563 
I think a new template for distributing to the App Store would be the best way to go. 
With the default being the same as in 0.9.0 

Maybe we can have the template set a #define so you can query the if its for the App Store or not in code ( for setting the data path ).

PR #4623, opened 2015-11-24
Add .gitignore template for project generator. Closes #2791

### LINKED ISSUE #2791: RFC: .gitignore template file for project generator

bilderbuchi:
ping @arturoc 
In the template.config, I also added linuxarm entries, but did not find any android or ios entries already existing, so I guess those aren't affected yet? This template is in principle useful for all our platforms/OSes.

### DIFF
--- a/scripts/templates/gitignore/.gitignore
+++ b/scripts/templates/gitignore/.gitignore
@@ -1,0 +1,83 @@
+#.gitignore is a file which makes git ignore files which should
+# not go into version control in the first place
+# Questions? See
+# http://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository#Ignoring-Files
+# http://git-scm.com/docs/gitignore
+
+###########################
+# ignore generated binaries
+# but not the data folder
+###########################
+
+/bin/*
+!/bin/data/
+
+#########
+# general
+#########
+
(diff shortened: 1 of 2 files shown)


HalfdanJ:
Android Studio should also be added. Something like 

```
.gradle
/local.properties
/.idea/workspace.xml
/.idea/libraries
/build
```
(http://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project)

LeoColomb:
:+1:

arturoc:
thanks!!

PR #4625, opened 2015-11-25
ofNode: better default for up vector on lookAt. Closes #4594

### LINKED ISSUE #4594: ofNode::lookAt gives incorrect results
When an `ofCamera` (inheriting from `ofNode`) is situated in a position with the `x` and `z` components at 0, the result of `lookAt` is inconsistent.

To reproduce the bug:
```
camera.setPosition(0, 600, max(0, ofGetMouseX()-100));
camera.lookAt(ofVec3f());
ofSetColor(255, 0, 0);
camera.begin();
ofDrawBox(100);
camera.end();
```

arturoc:
Instead of using {0,1,0} get's the vector of the camera position to
the lookAt point. then figures out the rotation that transforms {0,0,1}
into that vector and rotates {0,1,0} by that rotation to get the up vector.

### DIFF
--- a/libs/openFrameworks/3d/ofNode.cpp
+++ b/libs/openFrameworks/3d/ofNode.cpp
@@ -231,6 +231,14 @@ void ofNode::rotateAround(float degrees, const ofVec3f& axis, const ofVec3f& poi
 }
 
 //----------------------------------------
+void ofNode::lookAt(const ofVec3f& lookAtPosition){
+    ofQuaternion q;
+    q.makeRotate({0.f,0.f,1.f}, getPosition() - lookAtPosition);
+    auto up = q * ofVec3f(0.f,1.f,0.f);
+    lookAt(lookAtPosition, up);
+}
+
+//----------------------------------------
 void ofNode::lookAt(const ofVec3f& lookAtPosition, ofVec3f upVector) {
 	if(parent) upVector = upVector * ofMatrix4x4::getInverseOf(parent->getGlobalTransformMatrix());	
 	ofVec3f zaxis = (getGlobalPosition() - lookAtPosition).getNormalized();
@@ -245,6 +253,14 @@ void ofNode::lookAt(const ofVec3f& lookAtPosition, ofVec3f upVector) {
 		
 		setGlobalOrientation(m.getRotate());
 	}
+}
+
+//----------------------------------------
+void ofNode::lookAt(const ofNode& lookAtNode){
+    ofQuaternion q;
+    q.makeRotate({0.f,0.f,1.f}, getPosition() - lookAtNode.getGlobalPosition());
+    auto up = q * ofVec3f(0.f,1.f,0.f);
+    lookAt(lookAtNode, up);
 }
 
 //----------------------------------------
@@ -351,7 +367,7 @@ void ofNode::orbit(float longitude, float latitude, float radius, ofNode& center
 void ofNode::resetTransform() {
 	setPosition(ofVec3f());
 	setOrientation(ofVec3f());
-    setScale({1.f,1.f,1.f});
+    	setScale({1.f,1.f,1.f});
 }
 
 //----------------------------------------
--- a/libs/openFrameworks/3d/ofNode.cpp
+++ b/libs/openFrameworks/3d/ofNode.cpp
@@ -232,10 +232,15 @@ void ofNode::rotateAround(float degrees, const ofVec3f& axis, const ofVec3f& poi
 
 //----------------------------------------
 void ofNode::lookAt(const ofVec3f& lookAtPosition){
-    ofQuaternion q;
-    q.makeRotate({0.f,0.f,1.f}, getPosition() - lookAtPosition);
-    auto up = q * ofVec3f(0.f,1.f,0.f);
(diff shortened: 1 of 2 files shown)


tgfrerer:
Hold your fire =) I'm testing this currently and it "wobbles" quite a bit, let me rule out it's my fault first.

arturoc:
ok perfect

arturoc:
how are you setting the position?

i'm testing with:

```cpp
#include "ofMain.h"

class ofApp: public ofBaseApp
{
    ofNode target;
    ofNode nodeNew, nodeOld;
    ofEasyCam cam;
public:
    void setup(){ // root
        ofSetBackgroundColor(0);
    }

    void update(){
        auto now = mouseX/float(ofGetWidth())*360.f;
        nodeNew.orbit(now,-30,100,target);
        nodeOld.orbit(now+30,-30,100,target);
        nodeNew.lookAt(target);
        nodeNew.lookAt(target, {0.f, 1.f, 0.f});
    }

    void draw(){
        cam.setPosition(0,0,200);
        cam.begin();
        target.draw();
        ofSetColor(ofColor::red);
        nodeNew.draw();
        ofSetColor(ofColor::green);
        nodeOld.draw();
        cam.end();
    }
};

int main()
{
    settings.setGLVersion(3,2);
    ofCreateWindow(settings);
    ofRunApp(new ofApp);
}
```

and it works the same in both cases: http://rin.rs/nodes.mp4

tgfrerer:
I believe your update function sets`nodeNew` to the old way:
`
void update(){
        auto now = mouseX/float(ofGetWidth())*360.f;
        nodeNew.orbit(now,-30,100,target);
        nodeOld.orbit(now+30,-30,100,target);
        nodeNew.lookAt(target);
        nodeNew.lookAt(target, {0.f, 1.f, 0.f}); // should be nodeOld(...)
    }
`

arturoc:
mmh, yeah i'm getting the same problem if i fix that.

arturoc:
i think that should work in all cases, i'm figuring out the longitude and latitude and from that creating the same quaternion we are using to orbit, that way lookAt doesn't need an up vector. it seems to work for all cases, it rotates around the poles in the opposite way than the old version but the effect is pretty much the same and it solves the degenerate case in #4594 and i guess anything else where the lookAt direction is parallel to {0,1,0}

this is the code i'm using for testing:

```
#include "ofMain.h"

class ofApp: public ofBaseApp
{
    ofNode target;
    ofNode nodeNew, nodeOld;
    ofEasyCam cam;
public:
    void setup(){
        ofSetBackgroundColor(0);
        ofEnableDepthTest();
        target.setPosition(10,10,10);
    }

    void update(){
        auto lon = mouseX/float(ofGetWidth())*360.f;
        auto lat = mouseY/float(ofGetHeight())*360.f;
        nodeNew.orbit(lon,lat,100,target);
        nodeOld.orbit(lon+30,lat,100,target);
        nodeNew.lookAt(target);
        nodeOld.lookAt(target, {0.f, 1.f, 0.f});
    }

    void draw(){
        cam.setPosition(0,0,200);
        cam.begin();
        ofSetColor(255);
        target.draw();
        ofSetColor(ofColor::red);
        nodeNew.draw();
        ofSetColor(ofColor::green);
        nodeOld.draw();
        cam.end();
    }
};

int main()
{
    settings.setGLVersion(3,2);
    ofCreateWindow(settings);
    ofRunApp(new ofApp);
}
```

libs/openFrameworks/3d/ofNode.cpp
```diff
@@ -231,6 +231,19 @@ void ofNode::rotateAround(float degrees, const ofVec3f& axis, const ofVec3f& poi
 }
 
 //----------------------------------------
+void ofNode::lookAt(const ofVec3f& lookAtPosition){
+    auto relPosition = (getGlobalPosition() - lookAtPosition);
+    auto radius = relPosition.length();
+    if(radius>0){
+        auto latitude = ofRadToDeg(acos(relPosition.y / radius)) - 90;
+        auto longitude = ofRadToDeg(atan2(relPosition.x , relPosition.z));
```
tgfrerer: The only way I could see this blow up is if `relPosition.x == 0 && relPosition.y==0` - for which `atan2` is undefined. Could this ever be the case?

arturoc: it seems to behave properly even then, in the original issue the degenerate case would have x and z as 0 but it still looks ok visually and the rotation values seem correct. atan2 is returning 0 in that case.

tgfrerer: Interesting! 

One c++ reference says that in that case a `domain error` (<http://www.cplusplus.com/reference/cmath/atan2/>) will be raised, but this reference (http://en.cppreference.com/w/cpp/numeric/math/atan2) says a domain error will not occur "If the implementation supports IEEE floating-point arithmetic," so we should be safe for floats, doubles, and long doubles =) 

And when i try atan2(0.f,0.f) myself i get 0, at least in clang, as you reported. `0`, anyway, is what we might want for longitude in that case. 

+1

arturoc:
thanks!

PR #4640, opened 2015-12-02
fix an issue with arrow heads not drawing centered

tgfrerer:
The cones used to draw arrow heads were drawing at the wrong position, offset slightly (0.5 * headSize) from where they should be. This was especially apparent when arrowheads were larger.

This commit fixes this.

Before (was): 
testarrows_debug 2015-12-02 15-28-02-912

After this fix:
testarrows_debug 2015-12-02 15-28-23-213


arturoc:
i saw this in your examples the other day but though it was a perspective thing. thanks!

tgfrerer:
=)
cheers

PR #4641, opened 2015-12-03
ofxiOSCoreMotion

julapy:
this PR introduces ofxiOSCoreMotion into the ofxiOS core,
taken from @trentbrooks's solid ofxCoreMotion addon >> https://github.com/trentbrooks/ofxCoreMotion

it also removes ofxiOSAccelerometer which needed to be removed due to being a deprecated API.

TouchAndAccelExample has been renamed to iosCoreMotionLegacyExample,
and is now using ofxiOSCoreMotion for retrieving the accelerometer values.
i figured it would be good to leave this example in as it shows people how to migrate to the new ofxiOSCoreMotion API.

two new examples were added,
`iosCoreMotionExample`
`iosCoreMotionCameraExample`
which demo all the new ofxiOSCoreMotion functionality.

ping @ofTheo @danoli3 for a review.


arturoc:
i'm milestoning this for 0.10

danoli3:
Awesome!! Just tested examples! This fixes so many of those old deprecated warnings.

Only issue is the iOS template needs CoreMotion Framework added by default.

Also for old projects that upgrade to 0.10 oF and get the following error:
```
  "_OBJC_CLASS_$_CMMotionManager", referenced from:
      objc-class-ref in libofxiOS_iphoneos_Debug.a(ofxiOSCoreMotion.o)
```
- Fixed by Adding CoreMotion Framework to project.

julapy:
ah yeah, good catch @danoli3.
just added CoreMotion.framework to the ios xcode template project.

@arturoc, is 0.10.0 the current master?

julapy:
will merge this in a couple days if theres no objections.
cheers.

gameoverhack:
Good to see an update to CoreMotion! If it's of any help there is this: https://github.com/gameoverhack/ofxIOSDeviceMotion ...maybe the checks for availability of each DeviceMotion functionality are useful? For instance the magnetometer isn't available on iPods...also there's code in there for backwards compatibility with ofxAccelerometer etc - such as setting a filtering factor, as well as some hi and low pass filtering stuff. I wrote it before I totally got const correctness, so I'm returning by reference when I should be returning by const&, but for apps that hammer CoreMotion it's probably a good idea to return the values as const& getSomething() to save on the copy.

arturoc:
hey, no current master is 0.9.1 let's finish that first before merging any new features or api changes, there's really few things to fix:

https://github.com/openframeworks/openFrameworks/milestones/0.9.1

julapy:
ah ok, sounds good @arturoc

julapy:
hey @gameoverhack, yeah nice... 
do you perhaps want to give a crack at merge your addon with ofxiOSCoreMotion in this PR?
i can also take a look and see what we can merge.

gameoverhack:
@julapy yep happy to have a look ....one initial question I have is a style thing - are we preferring setupSomeOption() or setUseSomeOption(bool b)?

arturoc:
i think enableAccelerometer/disableAccelerometer is what we are using in other places

danoli3:
@arturoc should we just make a branch for 0.10 and merge these hanging PR's there. As we are already getting some rot with a lot to merge post 0.9.1. 

Then later merge that branch into master?

arturoc:
i think if we start a branch for 0.10 everybody is going to concentrate on that rather than fix the remaining issues :)  so let's try to finish 0.9.1 asap (is almost there) and then we can merge things into master directly

PR #4644, opened 2015-12-03
Bug fix - Set max to max().

bakercp:
ofParameter seems to have its internal value max() set incorrectly.  Both min and max are set to `of::priv::TypeInfo<ParameterType>::min()`.  This fixes that.


arturoc:
thanks!

bakercp:
:)

Issue #4600, opened 2015-11-18
Rename ofGetLastFrameTime


HalfdanJ:
I had for some time wondered why there was no `ofGetDeltaTime()` that returns the delta time from last from to current frame, until I now realized that `ofGetLastFrameTime() ` does exactly this (after looking in its source). 
In my view it has a very bad name, i expected to get the absolute time of the last frame (I was about to calculate the delta time my self using `ofGetElapsedTimef() - ofGetLastFrameTime()` when i realized my mistake). 

Am I the only misunderstanding its name?

ofZach:
+1, this also seems unintuitive to me -- some alternative names that might be worth considering (since ofGetDeltaTime() doesn't really specify what the delta is): 

ofGetFrameDeltaTime() 
ofGetTimeSinceLastFrame()
ofGetElapsedFrameTime()

or something similar ?

HalfdanJ:
Yeah one of those is better. 
Also i just looked a bit at the reason for the current naming. It seems to me that the name means "The time last frame took", but thats not correct, its the time between the last `newFrame` and the call before it (the time i'm interested in also). If the framerate is locked to something slower then the framerate the system can run, this wont be the time the last frame took.

opheliagame:
@HalfdanJ came across this issue while trying to figure out what this function does myself. 

@ofZach 
I am wondering if there is interest and has been any attempt at writing the documentation to make it easier to understand for beginners. If documentation PRs are being reviewed, would love to write some documentation myself.

clankill3r:
> [@HalfdanJ](https://github.com/HalfdanJ) came across this issue while trying to figure out what this function does myself.

same

GitBruno:
It would be nice to make clear what unit the delta time is returned:
I assumed milliseconds but I think it's seconds?

    ofGetLastFrameSeconds()
    ofGetLastFrameMillis()

This can be added as is.

Adding Delta gets a bit long though ...

    ofGetLastFrameDeltaSeconds()
    ofGetLastFrameDeltaMillis()

When we call it frame delta we can remove last? So maybe:

    ofGetFrameDeltaSeconds()
    ofGetFrameDeltaMillis()

clankill3r:
`ofGetDeltaTime()`

GitBruno:
ofxGetDeltaMS()

PR #4646, opened 2015-12-04
Feature issue 4600

tpltnt:
Hey,

i renamed ofGetLastFrameTime() to ofGetTimeSinceLastFrame() as suggested in #4600.
Cheers.
tpltnt

### DIFF
--- a/examples/3d/advanced3dExample/src/Swarm.cpp
+++ b/examples/3d/advanced3dExample/src/Swarm.cpp
@@ -91,7 +91,7 @@ void swarm::customDraw(){
 void swarm::update(){
 
 	// Calculate time past per frame
-	float dt = ofGetLastFrameTime();
+	float dt = ofGetTimeSinceLastFrame();
 
 	// Update positions, velocities
 	for(unsigned int i = 0; i < particles.size(); i++){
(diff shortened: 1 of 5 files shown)


tpltnt:
ping

ofTheo:
looks good to me - thanks!

tpltnt:
Thank you for the fast response @ofTheo

arturoc:
mmh we never agreed on this, the original issue says that the current function is confusingly named and some suggestions but i don't think this one is much better either, i would argue that is even worse.

get time since last frame, strictly speaking would return different values if you called it several times during one frame.

what's really going on is that we meassure the time since we call update until swap buffers returns and then pass that time to the next frame, so really get last frame time is pretty accurate describing what's going on

perhaps we can add a time delta parameter to the update callback in ofApp and update event but i don't think this naming is very correct and i think it's even too long

arturoc:
also this is changing the global c function but not the equivalent method in ofCoreEvents and ofFpsCounter, making the api less unifom.

i'll revert it later unless someone thinks it should stay like this

tpltnt:
@arturoc why haven't you spoken up earlier when you added it to the 0.10.0 milestone on 2015-12-05?  I only mapped the consensus reached in #4600 into code. Since that ticket points out the "original issue (which) says that the current function is confusingly named", wouldn't that be the place to have a meaningful discussion? Others might also attempt at fixing it and probably won't check all discussions in all related PRs. As far as I am aware I also was not "changing the global c function", but only added deprecation warnings. The original code is still there, but the compiler will point out function calls to investigate for compatibility.

ofTheo:
I think the way the PR handles changing the API is totally fine, as it simply makes the old function deprecated. 

However if the naming is not really describing what is going on correctly then that is obviously a problem. If the original describes the value correctly then we should revert it and keep the original.

arturoc:
it's not, because the function calls other methods with the same name in ofCoreEvents and ofFpsCounter which are not renamed so now there's the global c function with one name and the methods in those classes with the old name.

in any case i think the naming is not very accurate anyway.

Issue #4639, opened 2015-12-02
ofToggleFullscreen doesn't record window size & position


danomatika:
OSX 10.11, OF 0.9.0 & using ofSetupGL(w, h, OF_WINDOW)

ofToggleFullscreen enters fullscreen fine but doesn't seem to record the original position and size of the window and remains fullscreen sized when transitioning to windowed mode. Looking at ofGLFWWindow.cpp, exiting fullscreen simply sets the window size using getWindowSize() which does nothing since it returns the *current* size of the window aka the fullscreen size. Same with the position. An empty default position opens the window centered in the screen but this reverts to 0,0 when exiting fullscreen.

Looking further, I'm able to get the original window size by using the ofWindowSettings object size [here](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L741):

    int nonFullScreenW = settings.width;
    int nonFullScreenH = settings.height;

_This is not a real fix, just checking._

It seems to me, the window should record it's rect before going fullscreen and reapply those settings when coming back, unless there has been an explicit change.

This is a rather fundamental function and I'm a bit surprised it was allowed to break this way and make it into the release.

danomatika:
Followup:

Saving the current size & pos to an ofRectangle before going to OF_FULLSCREEN and reapplying these when going back OF_WINDOW work in my limited testing. This preserves manual window movements and resizing as well.

arturoc:
this is working properly in linux and the windows implementation seems correct, can you send a PR?

PR #4655, opened 2015-12-08
save window position on OS X before going fullscreen

danomatika:
This is a small fix for #4639 using an ofRectangle.

For some reason, there was an update to ofAppGLFWWindow.cpp line 156 that wasn't merged when I fetched for upstream, so that is reflected in the small whitespace change there.


arturoc:
thanks!!

PR #4664, opened 2015-12-11
implement more robust loading of ply meshes

tgfrerer:
ofMesh has the ability to store and retrieve meshes as text files, using the fairly common ply format. Although it is almost ubiquitous, the format is not all too well documented, and many exporters from 3D modelling software do order (and name) attributes in slightly different ways than oF does.

This fix makes it possible to parse attributes based on the order of how these are defined in the file header (as the file format suggests), and cleans up the code slightly.

It also allows parsing of texture coordinates should they be named 's/t', as some 3d export plugins prefer over openFrameworks' export naming: 'u/v'.

### DIFF
--- a/libs/openFrameworks/3d/ofMesh.cpp
+++ b/libs/openFrameworks/3d/ofMesh.cpp
@@ -803,6 +803,16 @@ void ofMesh::load(string path){
 		Faces
 	};
 
+	
+	enum Attribute {
+		Position,
+		Color,
+		Normal,
+		TexCoord,
+	};
+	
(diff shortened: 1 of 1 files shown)


arturoc:
thanks! i'm milestoning for 0.10 and will merge it right away as soon as 0.9.1 is done

micuat:
Tested on 0.9.0 release / VS2015 with `x y z nx ny nz` and `x y z nx ny nz r g b a` and both worked. Great job!

arturoc:
thanks!!

PR #4675, opened 2015-12-15
Clean up main.cpp/mm templates.

bakercp:
- Remove erroneous comments.
- Clarify.
- Format.
- Use newer std::make_shared<ofApp>() syntax.
- Remove ofMain.h where appropriate since it is included in ofApp.h templates.

### DIFF
--- a/scripts/templates/gl3.1/src/main.cpp
+++ b/scripts/templates/gl3.1/src/main.cpp
@@ -1,15 +1,8 @@
-#include "ofMain.h"
 #include "ofApp.h"
 
-//========================================================================
-int main( ){
+int main(){
 	ofGLWindowSettings settings;
 	settings.setGLVersion(3,1);
 	ofCreateWindow(settings);
-
-	// this kicks off the running of my app
-	// can be OF_WINDOW or OF_FULLSCREEN
-	// pass in width and height too:
-	ofRunApp(new ofApp());
-
+	ofRunApp(std::make_shared<ofApp>());
 }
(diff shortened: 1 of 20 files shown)


arturoc:
would be better to have something like:

```
ofGLWindowSetings settings;
settings.setGLVersion(3,2);

auto app = std::make_shared<ofApp>();

ofRunApp(window, app);
return ofRunMainLoop();
```

that's the most complete syntax and the required one if you want to use multiwindow. it also returns a value when main is done which will never done. that way if you do ofExit(1) the app will return 1 which you can check from the OS to for example see if the app finished correctly

arturoc:
also i don't think we should rely on including something through other header files, in any case we could remove ofMain.h and use ofAppRunner.h which is the only thing needed by default. but as it is right now if someone wants to modify their app to use forward declarations main will fail which it shouldn't

bakercp:
Hey @arturoc I think I got this cleaned up now.  Please check it out.

scripts/templates/android/src/main.cpp
```diff
@@ -1,21 +1,19 @@
...
 
 int main(){
-	ofSetupOpenGL(1024,768, OF_WINDOW);			// <-------- setup the GL context
-
-	// this kicks off the running of my app
-	// can be OF_WINDOW or OF_FULLSCREEN
-	// pass in width and height too:
-	ofRunApp( new ofApp() );
-	return 0;
+	// ofSetupOpenGL sets up a default OpenGL window for ofApp. The initial
```
arturoc: this doesn't make much sense in android since you can't change the window shape like that, better to leave it like it was.

bakercp: So should we be using the ofSetupOpenGL function here or should we be setting up the windows?  Or are you just suggesting removing the comment?

arturoc: yes just the comment, ofSetWindowShape doesn't make sense on mobile

PR #4669, opened 2015-12-14
ofAVFoundationVideoPlayer: simplified setVolume

i-n-g-o:
a redo of #3803 for #3763 due to changes in ofAVFoundationVideoPlayer

PR #4681, opened 2015-12-17
simplified setVolume

i-n-g-o:
same as:
https://github.com/openframeworks/openFrameworks/pull/4669/files
but for iOS


arturoc:
@danoli3 if you want to double check? otherwise i'll merge it tomorrow. thanks @i-n-g-o !

danoli3:
Tested all good!

danoli3:
Although I think it was the idea to have the iOS/OSX video player using the same one, it's currently not hooked up that way. @julapy is that right?

Issue #4689, opened 2015-12-23
 setVolume() of ofVideoPlayer overwriten in update() 


dmelladom:
Hi,
v0.8.4 allowed to set the volume during setup() ant its value was kept during all the execution.
In v0.9.0, under mac os x 10.11.2, volume value is reset in update(), this meaning volume is set to 1 every cyce unless it is explicitly set to other value.

i-n-g-o:
Hello.
This should be fixed in the current master by PR #4669 and #4681
related ticket: https://github.com/openframeworks/openFrameworks/issues/3763

i-n-g-o:
you are right, there was an issue, the internal volume should be set even if the player is not ready.
here the PR: https://github.com/openframeworks/openFrameworks/pull/4690

kflak:
Seems this problem is still happening in Linux.

This works for me: 
```void ofApp::update(){

    if ((ofGetElapsedTimef() - startTime) > videoDuration ){
        file = ofApp::getFile();
        videoDuration = ofApp::getVideoDuration();
        startTime = ofGetElapsedTimef();
        movie.load(file);
        ofLog()<<volume<<endl;
        movie.play();
    };

    movie.setVolume(volume);
    movie.update(); 
}
```
This doesn't:
```
void ofApp::update(){

    if ((ofGetElapsedTimef() - startTime) > videoDuration ){
        file = ofApp::getFile();
        videoDuration = ofApp::getVideoDuration();
        startTime = ofGetElapsedTimef();
        movie.load(file);
        movie.setVolume(volume);
        ofLog()<<volume<<endl;
        movie.play();
    };

    movie.update(); 
}
```

Platform: Arch Linux, 
OpenFrameworks version: of_v20200912_linux64gcc6_release

Solcany:
The same issue is still on Mac Os 10.14.6 Mojave and of_v0.11.0_osx_release

PR #4690, opened 2015-12-24
set volume even if we are not ready

i-n-g-o:
should fix: https://github.com/openframeworks/openFrameworks/issues/4689


arturoc:
thanks!

dmelladom:
Thanks!!

PR #4697, opened 2015-12-28
bugfix for 90 degree orientation not working for fullscreen

### LINKED ISSUE #4682: ofSetOrientation OF_ORIENTATION_90_RIGHT/LEFT won't go fullscreen
It looks like its trying to set the Window Width to the Window Height and vice versa. 

This is because ofAppGLFWWindow::getScreenSize();  swaps the width and height based on the orientation, but the fullscreen code doesn't account for orientation when setting fullscreen. 

see getScreenSize(): https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L477 

and setFullscreen():
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L693

This looks like it could be an issue for all platforms. 
Also potentially an issue when restoring the window shape from fullscreen too.

### LINKED ISSUE #4696: toggling fullscreen on HiDpi screen causes window to shrink in half each time.
This is because this line doesn't multiply the windowW / windowH variables by the pixelScreenCoordScale variable: 

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L1126

multiplying w and h by pixelScreenCoordScale stops the window shrinking on fullscreen toggle. 

PR coming shortly.

ofTheo:
and small bugfix for window shrinking on hidpi screens. 
closes #4696 closes #4682

Needs testing on Windows. 
Will merge once confirmed fix on Windows.

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -780,10 +780,12 @@ void ofAppGLFWWindow::setFullscreen(bool fullscreen){
 	}
 #elif defined(TARGET_WIN32)
     if( windowMode == OF_FULLSCREEN){
-        int nonFullScreenX = getWindowPosition().x;
-        int nonFullScreenY = getWindowPosition().y;
-		int nonFullScreenW = getWindowSize().x;
-		int nonFullScreenH = getWindowSize().y;
+		// save window shape before going fullscreen
+		ofPoint pos = getWindowPosition();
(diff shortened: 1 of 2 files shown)


ofTheo:
okay this works for windows as well as OSX. 
it also fixes an issue where coming back from fullscreen on windows didn't restore the window to the correct size.

PR #4702, opened 2015-12-28
bugfix for being able to set the multidisplay fullscreen with the legacy window approach. 

### LINKED ISSUE #4698: Multi window fullscreen broken on current master.
If you enable setMultiDisplayFullscreen on OS X via the window pointer it doesn't end up being set. 

If you do ofCreateWindow with settings and then cast the returned window to a ofAppGLFWWindows shared_ptr it works . 

Also this seems like an incorrect deprecation as there isn't : 
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.h#L140

```
	OF_DEPRECATED_MSG("use ofGLFWWindowSettings to create the window instead", void setMultiDisplayFullscreen(bool bMultiFullscreen)); //note this just enables the mode, you have to toggle fullscreen to activate it.
```

Looking into this more. 
Its also confounded for me by issues with retina displays mixed with non retina displays.

ofTheo:
closes #4698

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.h
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.h
@@ -94,6 +94,7 @@ public:
     
     GLFWwindow* getGLFWWindow();
     void * getWindowContext(){return getGLFWWindow();}
+	ofGLFWWindowSettings getSettings(){ return settings; }
 
 	ofVec3f		getWindowSize();
 	ofVec3f		getScreenSize();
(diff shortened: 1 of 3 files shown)


arturoc:
this is failing cause some platforms don't include that class the best is probably to ifdef it by now. can you also add a TODO or FIXME to remove that method once the setter in the window class is removed? the solution is kind of hacky

ofTheo:
doh - of course :) 
I added an #ifdef for the include but not around the function.

PR #4705, opened 2015-12-29
support for jetson tk1

arturoc:
this adds a check in ofConstants for TARGET_OPENGL in which case it doesn't assume that LINUX ARM == OPENGLES

it also adds a specific makefile config for the jetson tk1.

it still needs a recompiled tess2 but i need a way to detect this board in order to change the apothecary formula and the install_dependencies script. anyone with one of this boards can check if there's anyway to detect it?

running:

```
  gcc -dM -E - < /dev/null
```

will dump a list of defines which can be useful for the apothecary formula. for the install_dependencies script any file that is specific to this platform will do


jvcleave:
to detect you may be able to use this

```
$ cat /etc/nv_tegra_release
# R19 (release), REVISION: 3.0, GCID: 4143443, BOARD: ardbeg, EABI: hard, DATE: Wed Jun 18 18:34:47 UTC 2014

```

Issue #4679, opened 2015-12-17
ofDirectShowPlayer can't load a movie if the width/height of the movie are not divisible by 4.


motoishmz:
Discussion on forum.openframeworks.cc:
https://forum.openframeworks.cc/t/ofdirectshowplayer-cant-load-uncommon-width-height-movies/21740

Solution for this problem:
https://gist.github.com/yoggy/df598c31bd3a74df10a0/revisions

[ofDirectShowPlayer::setPixelFormat()](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/video/ofDirectShowPlayer.cpp#L1181) looks strange because it sets anything. `ofDirectShowPlayer::pixelFormat` should be used as a hook for switching RGB24/RGB32 in `DirectShowVideo::loadMovie()` and `processPixels::processPixels()` ?

Thanks,
Motoi

ofTheo:
Thanks @motoishmz  

I guess the only issue would be the change from pixels being in a RGB format to an RGBA format. 
Would we manually strip the alpha out so as not to break existing code? 

Or just allow it do be 4 bytes per channel and let users deal with it via the higher level existing ofPixels queries.

PR #4707, opened 2015-12-30
DS VideoPlayer: add support for all different RGB formats

arturoc:
This adds support for setPixelFormat() for the directshow player which allows also to play videos with widths that are not multiple of 4. there's probably someway to have the library pad the video if it's not a standard size but by now this at least allows to play those videos by adding:

```cpp
player.setPixelFormat(OF_PIXELS_BGRA);
player.load("weird_size_video.mp4")
player.play();
```

Closes: #4679


arturoc:
this also removes a couple of unecessary copies of the pixels which should make the player slightly faster for big videos

PR #4714, opened 2016-01-05
optimise ofPixels::setImageType() on Windows VS

whg:
Using Visual Studio, calling getNumChannels() inside the for each pixel
loop has a huge effect on performance. By simply caching this result we
can get a 20x speed up. 180 -> 9ms on an 1920x1080 RGB to grayscale.

The difference with Xcode is minimal.


arturoc:
thanks!

PR #4723, opened 2016-01-12
Fixed video rendering into non-ARB textures on OS X.

### LINKED ISSUE #4722: ofVideoPlayer doesn't work with ofDisableArbTex() set OS X
From the forum: 
https://forum.openframeworks.cc/t/ofvideoplayer-and-ofdisablearbtex/22033

On OS X I could see how it might break with the new AVFoundation player using its own texture cache. 
Might mark this for 0.9.2 or 0.10.0 - but if we can easily fix it before then thats fine. 

Edit: misread the post - this isn't an issue on Windows.

prisonerjohn:
Doing this by drawing the ARB texture into a RECTANGLE_2D FBO when `ofGetUsingArbTex()` is `false`. Should fix #4722

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -11,6 +11,11 @@
 #include "ofTexture.h"
 #include "ofThread.h"
 
+#ifdef TARGET_OSX
+#include "ofFbo.h"
+#include "ofGraphics.h"
+#endif
+
 #ifdef __OBJC__
 #import "ofAVFoundationVideoPlayer.h"
 #endif
@@ -118,6 +123,8 @@ protected:
 #ifdef TARGET_OSX
 	CVOpenGLTextureCacheRef _videoTextureCache = nullptr;
 	CVOpenGLTextureRef _videoTextureRef = nullptr;
+
+	ofFbo videoFboRectangle2D;
 #endif
 };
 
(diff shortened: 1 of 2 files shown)


ofTheo:
not sure if this is the right fix. 
as ofGetUsingArbTex() can change during the course of an app - IE: it can be false in setup but true later on. Also I think using an FBO might not be the best approach. 

Going to see if there is a slightly simpler solution, but I'll keep this PR open incase this ends up being th best approach :)

arturoc:
we should probably deprecate get/set ARB texture. it's naming is incorrect and this kind of global state makes it really hard to maintain things. not saying to remove it right away but i would be against introducing new code that uses it and instead would add a property to the video grabber in initialization to set the texture target.

i would like to add more settings struct to objects that need some kind of complex initialization for 0.10 so for example to setup this you would call:

```
ofVideoPlayerSettings settings;
settings.file = path;
settings.textureTarget = GL_TEXTURE_RECT;
grabber.setup(settings);
```

ofTheo:
I wonder at least if the state should be held by the renderer instead? 
It can be quite useful to disable non power of 2 textures for a whole project, when you need to work with a codebase or shaders that assume POT / 0-1 tex coords.

tgfrerer:
+1 to deprecating ARB texture. 

`ArbTex` is actually not power of two related, but in oF terms means whether the texture target is `GL_TEXTURE_2D` or `GL_TEXTURE_RECTANGLE` (both of which may or may not be POT dimensions). 

`ARB` is a really generic term, it just refers to "Architecture Review Board", and is used as a prefix for any GL extensions which have been proposed to the API, but not yet entered core.

PR #4731, opened 2016-01-14
return correct bounding boxes for empty strings

### LINKED ISSUE #4730: ofTrueTypeFont/ofBitmapFont: bounding box width of empty strings should be 0
when calculating bounding boxes of an empty strings for ofTrueTypeFont or ofBitmapFont the width and height is negative ( = std::numeric_limits<float>::min() ). 

the width should be 0 and the height probably should also be 0 too (or maybe equal to the actual line height?).

fx-lange:
fixes #4730


arturoc:
thanks! :)

PR #4733, opened 2016-01-14
ofTexture hasMipmap() - fixes #4732

### LINKED ISSUE #4732: ofTexture - can't find out if mipmaps already exist or not
I see enableMipmap(), disableMipmap() and generateMipmap(). None of them provide info regarding the existence of the mipmap.

getTextureData() could work, but "hasMipmap" is defined as private... Maybe its just a matter of making this public? Otherwise adding a method hasMipmap() to ofTexture could work too.

armadillu:
see #4732

### DIFF
--- a/libs/openFrameworks/gl/ofTexture.cpp
+++ b/libs/openFrameworks/gl/ofTexture.cpp
@@ -960,6 +960,11 @@ void ofTexture::disableMipmap(){
 }
 
 //------------------------------------
+bool ofTexture::hasMipmap() const{
+	return texData.hasMipmap;
+}
+
+//------------------------------------
 void ofTexture::draw(float x, float y) const{
 	draw(x,y,0,getWidth(),getHeight());
 }
--- a/libs/openFrameworks/gl/ofTexture.h
+++ b/libs/openFrameworks/gl/ofTexture.h
@@ -927,6 +927,12 @@ class ofTexture : public ofBaseDraws {
 	/// \sa ofEnableArbTex()
 	/// \sa ofDisableArbTex()
 	void generateMipmap();
+
+	/// \brief Find out if a mipmap has been generated for the current texture.
+	///
+	/// \sa generateMipmap()
+	/// \sa enableMipmap()
+	bool hasMipmap() const;
 	
 	/// \}
 


bakercp:
Seems ok to me.  There seem to be lots of convenience functions that just forward to `texData`.

armadillu:
Yep I could see hasMipmap() being a method in texData instead; what does everyone prefer?

bakercp:
Hmm ... probably both since it's a private variable in the `texData`.  That said, you'll probably have to change the private variable name if to something else so it doesn't conflict with a function of the same name.  It's odd, (I prefer an `_` prefix for this reason), but the convention may be `bHasMipmap` ... ?

bakercp:
Also, the fact that `bAllocated` and other members of `texData` are inconsistently private / protected is odd.  The user could call

```c++
ofTexture tex;

ofTextureData& data = tex.getTextureData();
// or even just data = tex.texData
data.bAllocated = true;
```

... and create trouble.  Perhaps `ofTexture` and `ofTextureData` need a more general oop cleanup.

arturoc:
yes let's add this only to ofTexture by now, ofTextureData should be similar to other settings methods in other parts, where you just use it to allocate a texture with more complex settings but cannot access it later

tgfrerer:
@armadillu the PR looks good, the CI fail on OS X appears to be related to a networking unit test failing. +1 for merging.

@bakercp agreed - the interface is rather messy permissions-wise. Ideally, data fields for ofTexture would be only accessible through getters/ and setters which help enforce correct use. What kept us back from cleaning this up in the past has been, i believe, that it might break legacy code which depends on direct access to these elements. 

But I'd be all in favour of deprecating, and simplifying the API by narrowing it. 

@arturoc the crux with textureData is that it's something of a "setup/defaults object" on the one hand side - but appears also to transform into a store for object state during ofTexture's lifetime. Might be worth thinking of which fields should rather belong into `ofTexture` proper and which ones we might want to migrate into something like a more transient `ofTexture::Settings`.

armadillu:
+1 for ofTexture::Settings

arturoc:
i think it's ok if ofTextureData or Settings is later used internally by ofTexture to store it's parameters but it shouldn't be accesible from outside, if you want to query the id, or the width you just do tex.getId() or tex.getWidth() not tex.getTexData().id and much less tex.texData.id

PR #4736, opened 2016-01-15
Support darwin platform in apothecary android builds

HalfdanJ:
The current build script only supports linux, this fixes this problem


arturoc:
thanks!

PR #4737, opened 2016-01-15
bugfix PR now correctly shows when a video is loaded for AVF based players. 

### LINKED ISSUE #4629: ofAVFoundationPlayer: isLoaded returns false after a movie loaded for a short time
Even if a movie is successfully loaded synchronously, the `isLoaded` function returns false for a short time, and then return true.  I wonder it should always return true after a movie is loaded, but is this normal behavior or a bug?

## Code

In ofApp.h: 
```cpp
 ofVideoPlayer videoPlayer;
```

In ofApp.cpp:
```cpp
void ofApp::setup() {
    // sample video from:
    // http://www.sample-videos.com/video/mp4/720/big_buck_bunny_720p_1mb.mp4
    bool success = videoPlayer.load("SampleVideo_1080x720_1mb.mp4");
    ofLogNotice() << ofGetTimestampString() << ", load: " << success;
    ofLogNotice() << ofGetTimestampString()
                  << ", isLoaded: " << videoPlayer.isLoaded();

    videoPlayer.play();
}

void ofApp::update() {
    ofLogNotice() << ofGetTimestampString()
                  << ", isLoaded: " << videoPlayer.isLoaded();
    videoPlayer.update();
}

void ofApp::draw() { videoPlayer.draw(0, 0); }
```

## Results

```
2015-11-27 17:49:30.619 videoPlayerExampleDebug[53520:3446574] video loaded at 1280 x 720 @ 25.000000 fps
[notice ] 2015-11-27-17-49-30-626, load: 1
[notice ] 2015-11-27-17-49-30-626, isLoaded: 0
[notice ] 2015-11-27-17-49-30-626, isLoaded: 0
[notice ] 2015-11-27-17-49-30-811, isLoaded: 0
[notice ] 2015-11-27-17-49-30-816, isLoaded: 0
[notice ] 2015-11-27-17-49-30-819, isLoaded: 0
[notice ] 2015-11-27-17-49-30-917, isLoaded: 0
[notice ] 2015-11-27-17-49-30-934, isLoaded: 0
[notice ] 2015-11-27-17-49-30-935, isLoaded: 0
[notice ] 2015-11-27-17-49-31-022, isLoaded: 1
[notice ] 2015-11-27-17-49-31-031, isLoaded: 1
[notice ] 2015-11-27-17-49-31-067, isLoaded: 1
[notice ] 2015-11-27-17-49-31-068, isLoaded: 1
[notice ] 2015-11-27-17-49-31-164, isLoaded: 1
[notice ] 2015-11-27-17-49-31-194, isLoaded: 1
[notice ] 2015-11-27-17-49-31-207, isLoaded: 1
[notice ] 2015-11-27-17-49-31-220, isLoaded: 1
[notice ] 2015-11-27-17-49-31-259, isLoaded: 1
[notice ] 2015-11-27-17-49-31-277, isLoaded: 1
[notice ] 2015-11-27-17-49-31-281, isLoaded: 1
[notice ] 2015-11-27-17-49-31-286, isLoaded: 1
...
```

As you can see, there are a few zeros after a movie loaded.

## Environment

- Mac OS X 10.10.4
- oF current master f2149a97c85154a531c1395f21e29f257d67901b
- debug app for 64bit

ofTheo:
- Also adds a couple of extra safety checks for null ptr textures. 
 - Removes unneeded error messages when texture hasn't yet been loaded. 
closes #4629

The main change here is that ofAVFoundationPlayer.mm now uses the Objective C **bLoaded** state for ofAVFoundationPlayer::isLoaded instead of the **bReady** flag. bReady isn't true immediately but only when the first frame can be delivered and that can take a few mills to happen and it doesn't seem possible or practical to block until thats true as well. 

This means in synchronous loading mode after the load function is called isLoaded will now return true. 
The player still takes time to have frames ready, but as long as you wait for the is FrameNew() you will have good pixels ( which is what you should wait for anyway ). 

For anything internally needing the pixels, we still use isReady() so we don't return an empty ofPixels or try and draw a texture with no data loaded. 

This should be a good fix to #4629

### DIFF
--- a/libs/openFrameworks/video/ofVideoPlayer.cpp
+++ b/libs/openFrameworks/video/ofVideoPlayer.cpp
@@ -403,7 +403,9 @@ void ofVideoPlayer::resetAnchor(){
 
 //------------------------------------
 void ofVideoPlayer::draw(float _x, float _y, float _w, float _h) const{
-	ofGetCurrentRenderer()->draw(*this,_x,_y,_w,_h);
+	if( isUsingTexture() && getTexture().isAllocated() ){
+		ofGetCurrentRenderer()->draw(*this,_x,_y,_w,_h);
+	}
 }
 
 //------------------------------------
--- a/libs/openFrameworks/video/ofVideoPlayer.cpp
+++ b/libs/openFrameworks/video/ofVideoPlayer.cpp
@@ -403,9 +403,7 @@ void ofVideoPlayer::resetAnchor(){
 
 //------------------------------------
 void ofVideoPlayer::draw(float _x, float _y, float _w, float _h) const{
-	if( isUsingTexture() && getTexture().isAllocated() ){
-		ofGetCurrentRenderer()->draw(*this,_x,_y,_w,_h);
-	}
+	ofGetCurrentRenderer()->draw(*this,_x,_y,_w,_h);
 }
 
 //------------------------------------
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -94,6 +94,7 @@ protected:
 	
     bool loadPlayer(string name, bool bAsync);
 	void disposePlayer();
+    bool isReady() const;
 
 #ifdef __OBJC__
     ofAVFoundationVideoPlayer * videoPlayer;
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -221,7 +221,7 @@ void ofAVFoundationPlayer::update() {
     
     bFrameNew = false; // default.
     
-    if(!isLoaded()) {
+    if(!isLoaded() || !isReady()) {
         return;
     }
     
@@ -254,8 +254,14 @@ void ofAVFoundationPlayer::draw(const ofRectangle & rect) {
 }
 
 void ofAVFoundationPlayer::draw(float x, float y, float w, float h) {
-    if(isLoaded()) {
-        getTexturePtr()->draw(x, y, w, h);
+    if(isLoaded() && isReady()) {
+	
+		ofTexture * texturePtr = getTexturePtr();
+		if( texturePtr != NULL ){
+			if( texturePtr->isAllocated() ){
+				texturePtr->draw(x, y, w, h);
+			}
+		}
     }
 }
 
@@ -292,7 +298,7 @@ const ofPixels & ofAVFoundationPlayer::getPixels() const {
 }
 
 ofPixels & ofAVFoundationPlayer::getPixels() {
-    if(isLoaded() == false) {
+    if(isLoaded() == false || isReady() == false) {
         ofLogError("ofAVFoundationPlayer") << "getPixels(): Returning pixels that may be unallocated. Make sure to initialize the video player before calling getPixels.";
         return pixels;
     }
@@ -377,7 +383,7 @@ ofTexture * ofAVFoundationPlayer::getTexturePtr() {
 		return NULL;
 	}
 	
-    if(isLoaded() == false) {		
+    if(isLoaded() == false || isReady() == false) {
         return &videoTexture;
     }
     
@@ -567,6 +573,15 @@ bool ofAVFoundationPlayer::isPaused() const {
 
 //--------------------------------------------------------------
 bool ofAVFoundationPlayer::isLoaded() const {
+    if(videoPlayer == nullptr) {
+        return false;
+    }
+    
+    return [videoPlayer isLoaded];
+}
+
+//--------------------------------------------------------------
+bool ofAVFoundationPlayer::isReady() const {
     if(videoPlayer == nullptr) {
         return false;
     }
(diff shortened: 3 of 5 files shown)


ofTheo:
Waiting for the CI tests to complete. 
Might be a good idea for @danoli3 @julapy to take a quick look. 

I think this will close the last 0.9.1 issue!

arturoc:
the pixels should be allocated right after load returns if not using loadAsync, (not have a new frame but allocated) if that's not the case we can use an  std::condition to wait on in load and signal it from the avf callback whenever everything is ready

danoli3:
Looks good!

ofTheo:
@r9y9 - thats strange I don't see that issue with your example code and movie.
this is the code I am using - it draws and plays just fine for me. 

could you double check that you started on the master branch, switched to a new clean branch and then pulled in this PR? 

```
ofVideoPlayer videoPlayer;
void ofApp::setup(){
    videoPlayer.load("big_buck_bunny_720p_1mb.mp4");
    ofLogNotice() << ofGetTimestampString()
                  << ", isLoaded: " << videoPlayer.isLoaded();

    videoPlayer.play();

}

void ofApp::update(){
    ofLogNotice() << ofGetTimestampString()
                  << ", isLoaded: " << videoPlayer.isLoaded();
    videoPlayer.update();
    if( videoPlayer.isFrameNew() ){
        cout << " now we have a video frame " << endl;
    }else{
        cout << " no frame " << endl;
    }
}

void ofApp::draw(){
    if( videoPlayer.isFrameNew() ){
        videoPlayer.draw(0, 0);
    }
}
```

r9y9:
@ofTheo umm, I will give it a try again tomorrow.

ofTheo:
@arturoc - not sure if that will be simple to do. 

After the video is loaded we have the width and height available from the internal player, but the pixel format is not available until we receive the first frame. 

We could allocated with an assumption of RGB pixels or RGBA pixels, but I don't see a simple way to get back the pixel format at the end of the load call.

arturoc:
it would be more or less something like:

in the .h:

```
std::condition readyCondition;
std::mutex mtx;
```

at the end of the load method 

```
//start pipeline
readyCondition.wait();
```

in the callback:

```
if(!pixels.isAllocated){
    pixels.allocate(...)
    readyCondition.notify_all();
}
....
```

we've always had the pixels ready right after load so you can query width, height... and if we break that some applications might crash. i do more or less the same in gstreamer to ensure that everything is allocated when load finishes.

ofTheo:
@arturoc I can give that a try, but any type of waiting in the main thread for the callback to complete seems to block the callback - that was the issue I've been running into. Haven't found a way round this yet. 

EDIT: just tried it and the callback never gets fired if we lock at the end of the load call.

i-n-g-o:
@ofTheo with the work i was doing to optimize playback speed we would know the pixelformat before the first frame arrives based on the movies codec (using CMFormatDescription)
maybe this helps?
see this branch https://github.com/i-n-g-o/openFrameworks/tree/ofAVFoundationPlayer_set_pixelformats

i did not come around to finish it and make a PR out of it.
maybe i can do so in the next couple of days.

r9y9:
@ofTheo Sorry, I forgot to play a video `videoPlayer.play()`.... Your code works perfect.

r9y9:
@ofTheo After looking more carefully, something might be going wrong at the loop boundary? I'm seeing brinking. See https://www.youtube.com/watch?v=GB8qy35-Ro8&feature=youtu.be

ofTheo:
@r9y9 if you remove the isFrameNew check around vidPlayer.draw you won't get that issue. 
I don't think thats a regression - more that there is a check in draw which you wouldn't normally do.

r9y9:
@ofTheo Understood, thank you for the explanation!

ofTheo:
okay - it was actually pretty simple to solve. 
right now the pixels are either going to be RGB or RGBA and if you want a RGBA video to play with alpha you have to set the player to use the RGBA format before load. 

I tested with videos with and without the alpha channel, and the pixels are correct immediately after load now. 

So this should be good to merge.

libs/openFrameworks/video/ofVideoPlayer.cpp
```diff
@@ -403,7 +403,9 @@ void ofVideoPlayer::resetAnchor(){
 
 //------------------------------------
 void ofVideoPlayer::draw(float _x, float _y, float _w, float _h) const{
-	ofGetCurrentRenderer()->draw(*this,_x,_y,_w,_h);
+	if( isUsingTexture() && getTexture().isAllocated() ){
+		ofGetCurrentRenderer()->draw(*this,_x,_y,_w,_h);
+	}
```
arturoc: this check belongs in the renderer, if we use a renderer that doesn't use the texture it won't work like this. you could draw a video with cairo, doesn't make much sense :) but you can do it, and this would break it without producing any error either

arturoc:
other than that it looks good to me

ofTheo:
@arturoc - that is a strange edge case - but probably worth accounting for :) 
I'll pull that check out as its really a separate issue and open a new issue for it - I think its a fairly easy fix.

PR #4739, opened 2016-01-18
[AVFoundationPlayer] Fix crashes when unloading a video

### LINKED ISSUE #4738: Several crashes when unloading a video with ofAVFoundationVideoPlayer
Stack traces:

crash at https://github.com/openframeworks/openFrameworks/blob/90f3582a1b54b94506b038cf79fc36be1d486780/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m#L451

```
(lldb) bt
* thread #4: tid = 0x236316, 0x00007fff8f83d44f libobjc.A.dylib`objc_release + 15, queue = 'com.apple.root.default-qos', stop reason = EXC_BAD_ACCESS (code=1, address=0x20)
  * frame #0: 0x00007fff8f83d44f libobjc.A.dylib`objc_release + 15
    frame #1: 0x00000001003efc97 AVFoundationPlayerCrashDebug`__destroy_helper_block_((null)=0x00006000000e9d80) + 23 at ofAVFoundationVideoPlayer.m:0
    frame #2: 0x00007fff8d1ea6e5 libsystem_blocks.dylib`_Block_release + 196
    frame #3: 0x0000000100d8ad43 libdispatch.dylib`_dispatch_client_callout + 8
    frame #4: 0x0000000100d8e283 libdispatch.dylib`_dispatch_root_queue_drain + 1471
    frame #5: 0x0000000100d9fcd0 libdispatch.dylib`_dispatch_worker_thread3 + 106
    frame #6: 0x00007fff87c12637 libsystem_pthread.dylib`_pthread_wqthread + 729
    frame #7: 0x00007fff87c1040d libsystem_pthread.dylib`start_wqthread + 13
```

crash at https://github.com/openframeworks/openFrameworks/blob/90f3582a1b54b94506b038cf79fc36be1d486780/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m#L455

```
(lldb) bt
* thread #19: tid = 0x237dbb, 0x00007fff8f8350dd libobjc.A.dylib`objc_msgSend + 29, queue = 'com.apple.root.default-qos', stop reason = EXC_BAD_ACCESS (code=1, address=0x40dedeadbec0)
    frame #0: 0x00007fff8f8350dd libobjc.A.dylib`objc_msgSend + 29
  * frame #1: 0x00000001003ef815 AVFoundationPlayerCrashDebug`__45-[ofAVFoundationVideoPlayer unloadVideoAsync]_block_invoke(.block_descriptor=0x00006080000eb980) + 50 at ofAVFoundationVideoPlayer.m:455
    frame #2: 0x0000000100d902bb libdispatch.dylib`_dispatch_call_block_and_release + 12
    frame #3: 0x0000000100d8ad43 libdispatch.dylib`_dispatch_client_callout + 8
    frame #4: 0x0000000100d8e283 libdispatch.dylib`_dispatch_root_queue_drain + 1471
    frame #5: 0x0000000100d9fcd0 libdispatch.dylib`_dispatch_worker_thread3 + 106
    frame #6: 0x00007fff87c12637 libsystem_pthread.dylib`_pthread_wqthread + 729
    frame #7: 0x00007fff87c1040d libsystem_pthread.dylib`start_wqthread + 13
```

I wonder there might be a threading issue. 

The code below should reproduce the issue easily:

```cpp
void ofApp::setup() {}

void ofApp::update() {
    ofVideoPlayer videoPlayer;
    bool success = videoPlayer.load("SampleVideo_1080x720_1mb.mp4");
    ofLogNotice() << ofGetTimestampString() << ", load: " << success;
    ofLogNotice() << "will crash!";
}

void ofApp::draw() {}
```

though it's not practical, just to see if any issue in load/unload of a video.

Tested with Xcode 6.4 on osx 10.10.4 with oF current master.

cc: @openframeworks/video

r9y9:
Updated: Ok, now I think I finally fixed a real issue by https://github.com/openframeworks/openFrameworks/commit/fdea3689050d39b133ade2dac2293d2a975ce981. The reason of crashes is that AVPlayerItemVideoOutput delegate wasn't released property. See http://stackoverflow.com/questions/29103254/copy-helper-block-crash-in-avfoundation for explanation.

Also, I added a few cleanup that I noticed while debugging. 

### Previous description (please ignore below. just keep this old description for someone read discussion on the thread)

There are two commits in this PR:

- [ofAVFoundationPlayer] Internal AVFoundationVideoPlayer will be dealloced synchronously instead of asynchronously. This ensures that dealloc runs before the ofAVFoundationPlayer's destructor finished.
- [AVFoundationVideoPlayer] Avoid unloading a video twice. This should prevent unexpected future or current bugs.

Let me illustrate the timings of the dealloc of AVFoundationVideoPlayer and destructor of ofAVFoundationVideoPlayer runs w/o this patch.

Before:
```
[notice ] start: Destructor of ofAVFoundationPlayer
[notice ] videoPlayer is not nullptr! needs dealloc property
2016-01-18 19:47:38.246 AVFoundationPlayerCrashDebug[53601:2963124] start: unloadVideoAsync
2016-01-18 19:47:38.247 AVFoundationPlayerCrashDebug[53601:2963124] pass: unloadVideoAsync
[notice ] end: Destructor of ofAVFoundationPlayer
2016-01-18 19:47:38.263 AVFoundationPlayerCrashDebug[53601:2963123] dealloc cond signal!
2016-01-18 19:47:38.264 AVFoundationPlayerCrashDebug[53601:2963124] dealocCond waited. must have finished unloadVideoAsync
2016-01-18 19:47:38.264 AVFoundationPlayerCrashDebug[53601:2963124] AVFoundationVideoPlayer.m: start dealloc
2016-01-18 19:47:38.265 AVFoundationPlayerCrashDebug[53601:2963124] start: unloadVideoAsync
2016-01-18 19:47:38.265 AVFoundationPlayerCrashDebug[53601:2963124] pass: unloadVideoAsync
2016-01-18 19:47:38.266 AVFoundationPlayerCrashDebug[53601:2963222] dealloc cond signal!
2016-01-18 19:47:38.266 AVFoundationPlayerCrashDebug[53601:2963124] dealocCond waited. must have finished unloadVideoAsync
2016-01-18 19:47:38.267 AVFoundationPlayerCrashDebug[53601:2963124] AVFoundationVideoPlayer.m: end dealloc
```

After:
```
[notice ] start: Destructor of ofAVFoundationPlayer
[notice ] videoPlayer is not nullptr! needs dealloc property
2016-01-18 19:46:04.691 AVFoundationPlayerCrashDebug[47725:2666285] start: unloadVideoAsync
2016-01-18 19:46:04.691 AVFoundationPlayerCrashDebug[47725:2666285] pass: unloadVideoAsync
2016-01-18 19:46:04.705 AVFoundationPlayerCrashDebug[47725:2947859] dealloc cond signal!
2016-01-18 19:46:05.252 AVFoundationPlayerCrashDebug[47725:2666285] dealocCond waited. must have finished unloadVideoAsync
2016-01-18 19:46:05.252 AVFoundationPlayerCrashDebug[47725:2947859] AVFoundationVideoPlayer.m: start dealloc
2016-01-18 19:46:05.252 AVFoundationPlayerCrashDebug[47725:2947859] _player is nil: already unloaded.
2016-01-18 19:46:05.253 AVFoundationPlayerCrashDebug[47725:2947859] AVFoundationVideoPlayer.m: end dealloc
[notice ] end: Destructor of ofAVFoundationPlayer
```

 
With these changes, I'm not having any crash so far using the code as I noted in #4738. I'm running a sample app over 30 mins and getting no crash.

fixes #4738 

~~My commits include debug logging that would be helpful for someone who can test this branch. I will remove it and rebase this branch if this PR is acceptable.~~ done

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.h
@@ -31,7 +31,7 @@ typedef enum _playerLoopType{
 
 
 //---------------------------------------------------------- video player.
-@interface ofAVFoundationVideoPlayer : NSObject <AVPlayerItemOutputPullDelegate> {
+@interface ofAVFoundationVideoPlayer : NSObject {
 	
     AVPlayer * _player;
 	AVAsset * _asset;
@@ -44,7 +44,6 @@ typedef enum _playerLoopType{
 	
 #if USE_VIDEO_OUTPUT
 	CMVideoFormatDescriptionRef _videoInfo;
-	dispatch_queue_t _myVideoOutputQueue;
 	AVPlayerItemVideoOutput * _videoOutput;
 #endif
 	
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -501,6 +501,7 @@ static const void *PlayerRateContext = &ItemStatusContext;
 				
 #if USE_VIDEO_OUTPUT
 				// remove output
+				[currentVideoOutput setDelegate:nil queue:nil];
 				[currentItem removeOutput:currentVideoOutput];
 				
 				// release videouOutput
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -44,9 +44,6 @@ static const void *PlayerRateContext = &ItemStatusContext;
 		deallocCond = nil;
 		
 #if USE_VIDEO_OUTPUT
-		// create videooutput queue
-		_myVideoOutputQueue = dispatch_queue_create(NULL, NULL);
-		
 		// create videooutput
 		_videoOutput = nil;
 		_videoInfo = nil;
@@ -107,7 +104,6 @@ static const void *PlayerRateContext = &ItemStatusContext;
 	}
 	
 	self.videoOutput.suppressesPlayerRendering = YES;
-	[self.videoOutput setDelegate:self queue:_myVideoOutputQueue];
 }
 #endif
 
@@ -120,10 +116,6 @@ static const void *PlayerRateContext = &ItemStatusContext;
 	}
 	
 	[asyncLock lock];
-	
-#if USE_VIDEO_OUTPUT
-	dispatch_release(_myVideoOutputQueue);
-#endif
 	
 	[asyncLock unlock];
 	
@@ -501,7 +493,6 @@ static const void *PlayerRateContext = &ItemStatusContext;
 				
 #if USE_VIDEO_OUTPUT
 				// remove output
-				[currentVideoOutput setDelegate:nil queue:nil];
 				[currentItem removeOutput:currentVideoOutput];
 				
 				// release videouOutput
@@ -581,14 +572,6 @@ static const void *PlayerRateContext = &ItemStatusContext;
 	[asyncLock lock];
 	[self unloadVideoAsync];
 	[asyncLock unlock];
-}
-
-
-#pragma mark - AVPlayerItemOutputPullDelegate
-
-- (void)outputMediaDataWillChange:(AVPlayerItemOutput *)sender
-{
-	NSLog(@"outputMediaDataWillChange");
 }
 
 
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -115,7 +115,9 @@ static const void *PlayerRateContext = &ItemStatusContext;
 //---------------------------------------------------------- cleanup / dispose.
 - (void)dealloc
 {
-	[self unloadVideo];
+	if (_player != nil){
+		[self unloadVideo];
+	}
 	
 	[asyncLock lock];
 	
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -460,6 +460,7 @@ static const void *PlayerRateContext = &ItemStatusContext;
 			if (currentReader != nil) {
 				[currentReader cancelReading];
 				[currentReader autorelease];
+				currentReader = nil;
 				
 				if (currentVideoTrack != nil) {
 					[currentVideoTrack autorelease];
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -527,7 +527,7 @@ static const void *PlayerRateContext = &ItemStatusContext;
 
 				if (currentTimeObserver != nil) {
 					[currentPlayer removeTimeObserver:currentTimeObserver];
-					[currentTimeObserver release];
+					[currentTimeObserver autorelease];
 					currentTimeObserver = nil;
 				}
 				


arturoc:
thanks! ping @i-n-g-o who has been doing most of the dev on this classes lately

i-n-g-o:
hi.

as far as i can see (in `ofAVFoundationPlayer.mm`) by removing the `dispatch_async` from:
```
@autoreleasepool {
    [currentPlayer unloadVideo]; // synchronious call to unload video
    [currentPlayer autorelease]; // release
}
```
you make the cleanup-call synchron.

the most expensive part when cleaning up the videoPlayer is disposing the assetReader.
how long does a `ofAVFoundationPlayer::disposePlayer()` take for you with those changes?

r9y9:
@i-n-g-o  I think it's tolerable. I tried to load/unload 1MB ~ 2.5GB movies many times and got about 15 ~ 40msec in `ofAVFoundationPlayer::disposePlayer()` regardless of the size of movie.

Code for testing:

```cpp
void ofApp::setup() { }

void ofApp::update() {
    ofSleepMillis(1);
    ofVideoPlayer videoPlayer;
    assert(videoPlayer.load("test.mov"));
}

void ofApp::draw() {}
```

Code for time measuring:

```cpp
ofAVFoundationPlayer::~ofAVFoundationPlayer() {
	std::chrono::system_clock::time_point start, end;
	start = std::chrono::system_clock::now();
    disposePlayer();
	end = std::chrono::system_clock::now();
	ofLogNotice() << "Elapsed time in disposePlayer: "
		<< std::chrono::duration_cast<std::chrono::milliseconds>(end-start).count() << " msec";
}
```

i-n-g-o:
@r9y9 the reason for the threaded disposal of the movie was to get rid of this 15 ~ 40 ms. Think of a realtime application, which runs with 60fps (each frame has ~16ms to render) in which you want to load and unload movies. it would mean you skip 1 - 3 frames everytime when unloading a movie.

as it turns out, disposing a AVAssetReader on a background thread in tricky and leads to crashes and/or leaks, when done under certain circumstances.
the current implementations of ofAVFoundationPlayer/ofAVFoundationVideoPlayer is not perfect at all, it actually needs to be reworked: the first step we should do is to figure out how to dispose AVAssetReader on a background thread so it does not leak or crash. around that part we should build our player...
this is quite a bit of work (and a lot of testing!) and __might__ be feasible for 0.10

r9y9:
I understand what you mean, but I do think that the first step is to make it stable (i.e. no crash), and then optimize performance. Needless to say, stable and high performance video player is cool though.

i-n-g-o:
jep. i aggree. it should be stable when disposed on a background thread.

i think it would make sense to collect tests for the player. like yours which produces a crash.

for now i think the way to go is to use the player in a way, so it does not crash (e.g.: don't unload a movie right after you loaded it)

r9y9:
I'm not sure how to automate tests for oF apps, but having tests would be nice.

r9y9:
I think merging this is fine for now, but to be make sure, don't you think it's a good solution even if as a tentative fix for crashes? In my option, crash is serious than slowness.

i-n-g-o:
i disagree. pulling this in renders it unusable for certain applications.
to avoid a crash, don't use the videoplayer in that particular way.

i know this is far from optimal and it needs to be addressed. as soon as possible.

r9y9:
uum, I understood. 

JYI: I had crashs in my real application, not in the testing code above. I didn't load and unload repeatedly such as my testing code.

ofTheo:
In general I think we should be targeting stability over performance. 

I wonder though if we could do something slightly different in the destructor so it does sync unloading when the destructor is called, but async otherwise? 

@i-n-g-o this would allow close / open calls from the same player to avoid the unload slowdown 
@r9y9 this would solve the issue you pointed out and solves potential crashes on exit when video players are being destroyed.

arturoc:
yes i'd say we should try to find a solution that keeps things fast and it doesn't crash but we should do 0.9.1 asap. if we can't find a fix for this that solves both problems then i'd say we should go with whatever is safer even if it's slower.

it's kind of tricky to delay destruction on a different thread, usually detaching the thread would allow for the thread to keep running even if the owner class is already destroyed but transferring ownership of the data to that thread can be tricky, i think what @ofTheo proposes is what the gstreamer implementation does currently, calling close or reloading async happens completely asynchornously and doesn't block the main thread at all but destruction is synchronous to be safe.

even if you managed to let the destruction happen in a detached thread, doing so during the app destruction would be really tricky

i-n-g-o:
@ofTheo: hm jep. this could be a strategy.
@arturoc: in case of async disposal we would not detach the destruction of ofAVFoundationPlayer, but the destruction of it's internal player object. as we currently do it. ofAVFoundationPlayer would get destructed as usual, the internal player continues living inside a async block. since the internal player does not have a reference to the player this should be fine.
@r9y9: can you isolate the crash in your application?

generally i think we should isolate the source of all the trouble with that player: for what i know this is disposing the AVAssetReader async.

i-n-g-o:
... i mean not only isolating, but fixing ;)

r9y9:
@i-n-g-o Thanks for your explanation. I'm convinced that detaching destruction of the internal player should be fine. Also, I have a good news; I think that I finally fixed the issue #4738 without making the ofAVFoundationPlayer destruction synchronous! I have updated the issue description, rebased and pushed a few  commits. Please see the new issue description and commit messages.

r9y9:
Running a test app over an hour. Verified no crash on both Debug and Release build.

i-n-g-o:
that's great!

we don't need that delegate in the first place.
removing `[self.videoOutput setDelegate:self queue:_myVideoOutputQueue];` in `createVideoOutput` has the same effect.

thank you so much for looking at it!

r9y9:
Ok, indeed we didn't implement delegate methods for `AVPlayerItemOutputPullDelegate`. I pushed a further cleanup. 

I think other cleanups can be address as a separate PR. I would like to focus on "fixing crashes" in this PR.

arturoc:
can we merge this then?

i-n-g-o:
looks good to me

r9y9:
Ah, I forgot to remove an unused delegate method. Will push with the following change:

```diff
diff --git a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
index feb1cea..1c4dc1f 100644
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -575,14 +575,6 @@ static const void *PlayerRateContext = &ItemStatusContext;
 }


-#pragma mark - AVPlayerItemOutputPullDelegate
-
-- (void)outputMediaDataWillChange:(AVPlayerItemOutput *)sender
-{
-       NSLog(@"outputMediaDataWillChange");
-}
-
-
```

r9y9:
Done. I think it's ready to merge.

arturoc:
thanks!

r9y9:
You are welcome :)

PR #4744, opened 2016-01-21
small fix - checks if texture is allocated before binding. closes #4743

### LINKED ISSUE #4743: error message for all AVF video players on draw when the first frame is not ready.
Right now all videos will give you this error messages on OS X when you try and draw them until the video has the first frame come in which usually takes a few frames after load. 

`
[ error ] ofTexture: getTextureData(): texture has not been allocated
`

So even the videoPlayer example dumps out a bunch of these errors on launch. 
A simple fix would be to add something in the line below here:
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofGLRenderer.cpp#L362

Where it checks if the texture has been allocated before trying to draw. 
We would need to add it for the programmable renderer too - but it would be the same.

ofTheo:
Also improves isInitialized() check to account for videoplayers which have textures internally. 
The changes to GLRenderer and GLProgRenderer make it consistent with the other calls. 

The change with isInitialized() fixes the issue that is noticeable with the AVF player.


ofTheo:
only needed the fix to isInitialized as that is used in both renderers.

PR #4745, opened 2016-01-22
Update changelog for 0.9.1

jedahan:
This is not ready! I just wanted to start a discussion - if I should continue summarizing the commits, and putting the issue numbers up, or if I should just copy-paste the commit log in question. I am trying to summarize to make the changelog more human readable, but maybe instead of linking to the github issue number, I should link to the git log commit hashes?


danzeeeman:
don't forget the ASCII ART!

2bbb:
nice ASCII ART likes esher! :+1:

arturoc:
is this ready to be merged?

jedahan:
Yep, everything listed in https://github.com/openframeworks/openFrameworks/issues?utf8=%E2%9C%93&q=milestone%3A0.9.1+is%3Aclosed+ has some sort of entry

r9y9:
It would be great to add a mention to #4739.

arturoc:
mmh, yes there's some issues that weren't milestoned as 0.9.1 so they won't appear on that query. a more accurate one would be https://github.com/openframeworks/openFrameworks/issues?page=1&q=is%3Aclosed+closed%3A>%3D2015-11-08&utf8=✓

arturoc:
i wonder if we could automate this with something like: https://github.com/skywinder/Github-Changelog-Generator

i've give it a quick look and it's not generating everything but it's probably worth looking into it instead of having to do this manually for every release

Issue #4720, opened 2016-01-08
makefiles for msys2 not copying export libraries


kritzikratzi:
the current msys2 version doesn't copy the export libraries, so `make run` fails. 
not sure what's the best way to go about this. locally i fixed it by adding a line to the `afterplatform` task. 
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/msys2/config.msys2.default.mk#L292

```
cp -r $(OF_EXPORT_PATH)/$(ABI_LIB_SUBPATH)/* bin/
```

oxillo:
@arturoc @bakercp : This issue is solved by the copy_dlls task. 
It may be closed.

Issue #4748, opened 2016-01-23
MSYS2: compilation error on DirectShowPlayer


oxillo:
Compiling DirectShowPlayer.cpp crashes g++ (v5.2.0)

```
In file included from C:/msys64/mingw32/include/c++/5.2.0/memory:79:0,
                 from Z:/SMU_GenTool/of/libs/openFrameworks/utils/ofConstants.h:394,
                 from Z:/SMU_GenTool/of/libs/openFrameworks/ofMain.h:5,
                 from Z:/SMU_GenTool/of/libs/openFrameworks/video/ofDirectShowPlayer.h:6,
                 from Z:/SMU_GenTool/of/libs/openFrameworks/video/ofDirectShowPlayer.cpp:1:
C:/msys64/mingw32/include/c++/5.2.0/functional: In member function 'decltype (*(forward<_Tp>)(*std::_Mem_fn_base<#'offset_type' not supported by simple_type_specifier#)#'offset_type' not supported by direct_abstract_declarator#, false>::_M_call::__object).*(declval<#'offset_type' not supported by simple_type_specifier#&#'offset_type' not supported by direct_abstract_declarator#>)()) std::_Mem_fn_base<_Res _Class::*, false>::_M_call(_Tp&&, _Up* const*) const [with _Tp = IMediaSample*&; _Up = IMediaSample; _Res = long unsigned int(); _Class = IUnknown; decltype (*(forward<_Tp>)(*std::_Mem_fn_base<#'offset_type' not supported by simple_type_specifier#)#'offset_type' not supported by direct_abstract_declarator#, false>::_M_call::__object).*(declval<#'offset_type' not supported by simple_type_specifier#&#'offset_type' not supported by direct_abstract_declarator#>)()) = long unsigned int (IUnknown::)(); std::_Mem_fn_base<_Res _Class::*, false>::__pm_type = long unsigned int (IUnknown::*)()]':
C:/msys64/mingw32/include/c++/5.2.0/functional:658:43: internal compiler error: in gimplify_expr, at gimplify.c:8786
  { return (*std::forward<_Tp>(__object)).*_M_pm; }
```

oxillo:
Track down the error to the line in function bool ofDirectShowPlayer::load(string path)
    
    player.reset(new DirectShowVideo());

oxillo:
Going deeper...
In `STDMETHODIMP SampleCB(double Time, IMediaSample *pSample)`

[...]

PR #4749, opened 2016-01-24
MSYS2 compilation enhancements

oxillo:
Remove a 'file not found error' due Raspbian checks.
Copy required dlls to the executable directory so there is no more need to change the PATH
Updated MSYS2 documentation.


arturoc:
thanks! for the docs, can you just remove that paragraph? also if you can send a similar PR to the web that'd be great we'll merge it there after 0.9.1 is released

oxillo:
Forgot : close #4720
I've removed the paragraph. 
However I think we should keep it for the web version as we may have users still running/installing 0.9.0.

arturoc:
perfect, one thing, i think we use opencv from the system too so we would need to copy the opencv libraries too. we can just copy for every example by now and check if we can copy them only if you are using opencv in the future so the builds are not so big

about the web docs, i think it's ok as long as we merge it after 0.9.0. 0.9.1 is fully compatible with 0.9.0 so anyone who might want to install 0.9.x should go for 0.9.1 which fixes several things

arturoc:
i'm going to milestone this for 0.10 and merge it right away then. i'm a little concerned that we are forgetting some library and it'll break 0.9.1 since we are already in RC1 and there'll be little time to test.

also can you remove the change to ofDirectShowPlayer? it's a different problem and in any case the way you've fixed it is removing the call to the destructor and introducing a memory leak. i'll check what's the best way to fix this

thanks!

oxillo:
Oops... DirectShowPlayer changes shouldn't have been committed. I needed this temporary fix to be able to compile as OF on MSYS2 is broken (see #4748) .

oxillo:
Introduced a COPY_DLL make option. 
If not set, we keep the 0.9.0 behavior (minus #4720).
If set, we copy the mingw32 dlls to the bin directory.
Let's put it in 0.9.1, so we can get feedback for 0.10.0 and decide to make it default then.

arturoc:
this looks good, it seems like a good solution even on the long term since this way you can debug without copying the dlls which would take a lot of space and then generate a dll included version if you want to distribute the program. i wonder if we could use the target rule instead of a variable so you could do:

```
make DebugDlls
```

or something like that and it'll compile in debug and also copy the dlls. we could just need to add a dependency to the rule. we probably just need to add:

```
DebugDlls: Debug copydll
```

and remove the ifdef
 
that way you can also call

```
make Debug
make copydll
```

arturoc:
I've merged this with some slight modifications, instead of always copying the dlls you can optionally copy them with

```
make copy_dlls
```

also the install_dependencies.sh script will automatically if it needs to add the path and add it permanently so the extra step to add the path is gone now without needing to copy all the dlls which where too big.

This way the applications will run right away in the computer where you usually work and if you need to move it somewhere else you can run the copy_dlls target to get a copy that is self contained

in the future perhaps the best option would be to use mingw directly instead of msys2 and create apothecary scripts to compile the needed libraries as static but lets see

oxillo:
Thanks for the modification. I wanted to do it after I get the computer back from my children....
Abouts the libs, I think MSYS2 has the both the shared and static versions. Just have to figure out the correct compilation flags.
The way to do it should be discussed. I've opened #4759 for this purpose.

Issue #4571, opened 2015-11-12
ofxOsc examples are outdated


thomasgeissl:
- oscSenderExample misses bin/data/sendImageTest.jpg
- examples do not show how to use new osc types 

Is there common data, which can be used across all examples? A couple of images, soundfiles, videos?
How would one use the timetag argument correctly?

PR #4751, opened 2016-01-24
added missing image

thomasgeissl:
The oscSender example missed an image. I added the of logo.
This bug was reported in #4571.


arturoc:
thanks!

PR #4754, opened 2016-01-27
bugfix for OS X. stops window moving on ofSetWindowShape closes #4753

### LINKED ISSUE #4753: On OS X if you call ofSetWindowShape the position of the window changes.
The window resizes from the bottom left corner, with the window top left corner moving to accommodate the size change.

A simple test to reproduce is below
```
    if(key =='1'){
        ofSetWindowShape(1024, 768);
    }
    if(key =='2'){
        ofSetWindowShape(800, 200);
    }
}
```

Working on a fix. 

I wonder if this is a bug with GLFW?
I feel like we ran into this issue before.

EDIT: an old one: #2183

ofTheo:
Its not ideal that we have to do this, but I think this is due to OS X having lower left as the origin for cocoa windows. ( http://stackoverflow.com/questions/12408334/cocoa-nsview-origin-x-at-the-bottom ) 

If we want a quick fix that works this should cover it, but it might be a good idea to file an issue with GLFW and get it fixed there.

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -540,7 +540,15 @@ void ofAppGLFWWindow::setWindowPosition(int x, int y){
 
 //------------------------------------------------------------
 void ofAppGLFWWindow::setWindowShape(int w, int h){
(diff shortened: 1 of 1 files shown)


ofTheo:
Okay I tested this with fullscreen on multiple monitors and across multiple monitors and everything works as expected.

PR #4764, opened 2016-02-01
add any folder with name prefix apps to gitignore

fred-dev:
From forum post https://forum.openframeworks.cc/t/branches-and-my-of-folder/22277, to allow for folders named apps* in the of directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,5 +126,5 @@ Desktop.ini
 #########################
 
 .mailmap
-/apps*
-!/docs/apps/*
+/apps*/
+
--- a/.gitignore
+++ b/.gitignore
@@ -127,4 +127,4 @@ Desktop.ini
 
 .mailmap
 /apps*
-!/docs/apps*
+!/docs/apps/*
--- a/.gitignore
+++ b/.gitignore
@@ -126,4 +126,4 @@ Desktop.ini
 #########################
 
 .mailmap
-/apps*
+/apps**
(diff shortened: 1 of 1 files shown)


arturoc:
have you tested if this really works and doesn't ignore other apps* folders in other levels than root? i posted that syntax in the forum but not 100% sure is completely correct

fred-dev:
I tested but not completely- you are correct, it ignores any folder named apps in all folders. The correct functioning (tested by adding folders with various names beginning with apps to **all** subfolders from the root of the repo is actually  "/apps*/" it is in my last commit 

https://github.com/openframeworks/openFrameworks/commit/a8fa38546b4436144f7753358ef7b1263f477133

Now I am unfamiliar with protocol, should I make another PR?

arturoc:
no that's fine any new commit gets added to the current PR

arturoc:
going to merge this then. thanks!

fred-dev:
Cheers

arturoc:
also something that would be really useful if you want to look into it is to update any guides we have about workflow with this kind of info. i'm not sure where they live right now. perhaps @bilderbuchi knows?

bilderbuchi:
I'm not aware of any docs like this. Most stuff that I wrote was concerned with the developer-side of things, and not the user-side, and lives in the Wiki on Github.

Btw, it would probably have made more sense to add that line farther up, where the `apps/*` entry is, and/or add a little explanation what the intent of the change is.

PR #4767, opened 2016-02-02
Added a method to copy a fbo to a pbo using glReadPixels.

smallfly:
In order to get better performances when copying a FBO to a PBO, I have implemented the following method:

```
void ofFbo::copyTo(ofBufferObject & buffer) const{
	if(!bIsAllocated) return;
	bind();
	buffer.bind(GL_PIXEL_PACK_BUFFER);
	glReadPixels(0, 0, settings.width, settings.height, ofGetGLFormatFromInternal(settings.internalformat), ofGetGlTypeFromInternal(settings.internalformat), NULL);
	buffer.unbind(GL_PIXEL_PACK_BUFFER);
	unbind();
}
```
When copying big FBOs, using ```glReadPixels()``` instead of ```glGetTexImage()``` (via the ofTexture of the ofFbo) improves the performances quite substantially. (~20ms instead of ~80ms on a Mac Book Pro w/ a NVIDIA GeForce GT 650M 1024 MB)

Here a simple test app to see the differences in performance:
[https://gist.github.com/smallfly/75761a9178a917d13bc5](https://gist.github.com/smallfly/75761a9178a917d13bc5)

### DIFF
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -1008,6 +1008,18 @@ void ofFbo::readToPixels(ofFloatPixels & pixels, int attachmentPoint) const{
 #endif
 }
 
+#ifndef TARGET_OPENGLES
+//----------------------------------------------------------
+void ofFbo::copyTo(ofBufferObject & buffer) const{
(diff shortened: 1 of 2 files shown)


arturoc:
that's great, thanks!

PR #4768, opened 2016-02-03
WIP: harfbuzz shaping support

bakercp:
Apothecary Formulas
- [x] osx 
- [ ] ios 

Note: it would be nice to avoid the zlib.pc but since we are currently using system zlib it's a hack.  One workaround would be to use set up our local pkg-config to search system pkg-config paths e.g.

PKG_CONFIG="$BUILD_ROOT_DIR/bin/pkg-config:`pkg-config --variable pc_path pkg-config`"

but that introduces some links to system libraries that we don't want.  Anyway ....

### DIFF
--- a/scripts/apothecary/formulas/harfbuzz/harfbuzz.sh
+++ b/scripts/apothecary/formulas/harfbuzz/harfbuzz.sh
@@ -1,0 +1,124 @@
+#! /bin/bash
+#
+# Harfbuzz
+# HarfBuzz OpenType text shaping engine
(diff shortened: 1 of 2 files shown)


bakercp:
p.s. I'm int the progress of moving my code from ofFont into the updated ofTrueTypeFont that was merged earlier today.

arturoc:
nice, how are you doing the shaping? i'm looking in your repo but can't see anything related to harfbuzz

PR #4771, opened 2016-02-03
add https://github.com/nlohmann/json as default OF json

arturoc:
this adds https://github.com/nlohmann/json which is a 1 header json parser with a really nice syntax as our default json implementation. ofConstants re-exports the json class to the global namespace as ofJson to avoid name clashes.

this allows to follow up on the work on ofxGui. ping @frauzufall @fx-lange


fx-lange:
great, when do you plan to merge this into master?  after 0.9.2?

ofZach:
I have to say really love this library and have been using it for some projects recently -- excited to have this in the core.

frauzufall:
Thanks!

arturoc:
@fx-lange i'll merge it in master as soon as it's ready, 0.9.2 is in a different branch since we had already started to merge breaking changes into master. i thought it was going to be pretty straight forward but android won't compile it right away since the c++11 implementation in libstdc++ is not complete so i've had to change the whole toolchain to use clang, that in turn won't compile the version of opencv we are currently using so i've had to update opencv to 3.1 

all of that is done already but we need to compile opencv 3.1 for some platforms still, once that's done i'll merge the clang toolchain for android and this PR

2bbb:
:+100:

arturoc:
i've added an example too in utils/jsonExample

PR #4776, opened 2016-02-04
fix warning in ofTrueTypeFont

tgfrerer:
since glyphPositions are `emplaced` back, visual studio (and possibly
other compilers) warn bitterly about implicit type conversion.

this pr makes the conversion explicit to end the whining.

Since we're storing the glyph position in an ofVec2f and we don't have
an integer vector type there's no way not to have a type conversion at
some point.


arturoc:
thanks!

PR #4780, opened 2016-02-05
Add function for createWindow and ofCreateWindow for EGL

cgiles:
I noticed than when you tried to create an egl window, you lose some data, like if the ofAppEgl::Setting was casted into a simpler kind of setting.

Ass requested by arturo, my simple fix, I guess it can be highly improved.

### DIFF
--- a/libs/openFrameworks/app/ofAppRunner.cpp
+++ b/libs/openFrameworks/app/ofAppRunner.cpp
@@ -195,6 +195,11 @@ shared_ptr<ofAppBaseWindow> ofCreateWindow(const ofWindowSettings & settings){
 	ofInit();
 	return mainLoop()->createWindow(settings);
 }
+// special function for ofAppEGLWindowSetting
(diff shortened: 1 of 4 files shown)


arturoc:
mmh, this is not very correct. instead of creating a new method, the EGL window should check that the windows settings being passed are an EGLWindow settings by using a dynamic cast, this is how GLFW does it:

```
void ofAppGLFWWindow::setup(const ofGLESWindowSettings & settings){
	const ofGLFWWindowSettings * glSettings = dynamic_cast<const ofGLFWWindowSettings*>(&settings);
	if(glSettings){
		setup(*glSettings);
	}else{
		setup(ofGLFWWindowSettings(settings));
	}
}
```

this is kind of convoluted though and kind of error prone for people implementing new windows classes so there's porbably a way to do it using tempaltes that solves the problem for new windows. i'll take a look

PR #4785, opened 2016-02-05
Added numCoords parameter when setting color and texcoord data in ofVbo.

prisonerjohn:
I added an extra parameter for the number of coords when setting color and texture coordinates to an `ofVbo`. This allows binding 3D textures, amongst other things.

The change is transparent in most cases, i.e. when using `ofVec2f` for tex coords and `ofFloatColor` for colors. However, it is likely to break code using `float` arrays directly. I'm not sure how to keep backwards compatibility, but I'm open to suggestions!

### DIFF
--- a/libs/openFrameworks/gl/ofVbo.cpp
+++ b/libs/openFrameworks/gl/ofVbo.cpp
@@ -329,12 +329,12 @@ void ofVbo::setVertexData(const float * vert0x, int numCoords, int total, int us
 
 //--------------------------------------------------------------
 void ofVbo::setColorData(const ofFloatColor * colors, int total, int usage) {
-	setColorData(&colors[0].r,total,usage,sizeof(ofFloatColor));
+	setColorData(&colors[0].r, 4, total,usage,sizeof(ofFloatColor));
 }
 
 //--------------------------------------------------------------
-void ofVbo::setColorData(const float * color0r, int total, int usage, int stride) {
-	colorAttribute.setData(color0r, 4, total, usage, stride);
+void ofVbo::setColorData(const float * color0r, int numCoords, int total, int usage, int stride) {
+	colorAttribute.setData(color0r, numCoords, total, usage, stride);
 	enableColors();
 }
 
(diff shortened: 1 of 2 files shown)


tgfrerer:
Hey Elie, 

if you want to add attributes with more than the default number of fields, you can already do something like: 

`testvbo.setAttributeData(ofShader::TEXCOORD_ATTRIBUTE,  &addr, numCoords, numAttributes, GL_STATIC_DRAW);`

This allows you to use 3D textures, etc.

`setAttributeData()` should map the attributes correctly in the background, and will automatically detect texcoord/color/normals default attribute positions, if you use the ones from `ofShader`. 

In a way `setTexCoordData()` `setColorData()` are convenience methods for the most commonly used `setAttributeData()` data settings - it could make sense to add some more convenience methods, but I'm not sure if documenting the behaviour of the more universal `setAttribute` method wouldn't do the same trick...

prisonerjohn:
Ah ok, that makes sense!

But would that also work with the fixed pipeline? I'm asking cause I see [this line](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofVbo.cpp#L420) in `ofVbo`.

tgfrerer:
It will probably not work with the fixed function pipeline. 

That said, if you wanted to use 3D texture coordinates and more specialised vertex attribute settings, there is a good argument right there to use the programmable pipeline.

arturoc:
i'm ok merging something along this lines but as it is right now i think it might introduce some really hard to debug bugs. because of the default argument something that before was:

```
setColorData(colors.data(), colors.size(), usage, sizeof(ofFloatColor));
```

now will still compile but the usage will be passed as the number of cordinates and the sizeof as the usage with default stride 0 which will be super hard to debug.

perhaps we can add new methods with different names?

prisonerjohn:
@tgfrerer Cool, thanks for the info. I'm on GL3 but was just thinking out loud...

@arturoc Right, that was my concern too. I opted for this as the params match [setVertexData](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofVbo.h#L32), so it's somewhat consistent. Maybe it would be best to wait until the next major release 0.10.0 where things are expected to break a little ;)

arturoc:
current master is already 0.10 but still i don't think we should break things in a way that is not really obvious, if something stops compiling and you have to change it or you get a deprecation message saying to use something else but the old functionality works the same that's fine but in this case you would get no error or deprecation message, anyone using ofVbo and passing a stride to colors or normals would just get an application that won't show anything but will still compile.

we could leave a new method like setCustomColorData(...) but i'm not sure is really necessary to extend the api only to add functionality to the fixed pipeline, mostly if the new api would be kind of strange or break in a really nasty way like in this case

prisonerjohn:
Right. `setAttributeData()` is working fine, so let's just keep it at that for anything custom.

Thanks for taking the time to go over this so quickly!

PR #4794, opened 2016-02-08
Const corrections in ofImage.  Fixes #4791.

### LINKED ISSUE #4791: ofImage::saveImage() should not modify the passed ofPixels.
If the pixel format passed to `ofSaveImage` (aka `ofImage::saveImage()`) is `OF_PIXELS_BGRA` (for example color bitmaps in FreeType2 are stored this way), ofSaveImage will modify the ofPixels object to swap the B/R pixels.

Besides leading to a very long debugging process, on principle, I'd argue that an image saving function should not modify the pixels to be saved.

Thus, I'd suggest that `ofSaveImage(ofPixels & pix, std::string fileName, ofImageQualityType qualityLevel)` and similar be `ofSaveImage(const ofPixels& pix, std::string fileName, ofImageQualityType qualityLevel)` and likewise `ofImage::saveImage(const ofPixels& pix ...);`.

Any modifications to the pixels that need to be done to make things easier for freeimage should be done on a local copy (saving is a slow operation, so memory manipulations probably won't be the bottleneck).

bakercp:


arturoc:
thanks!

PR #4801, opened 2016-02-09
iPad Pro Pencil and generic 3D touch inputs

borg:
Here is one possible implementation of adding support for device specific touch input devices without having to create a separate event system.

The Apple Pencil can still be used as a normal finger touch input device, and the normal touch methods called in ofApp. However if the app needs access to the additional Pencil specific properties such as pen azimuth and force the native UITouch can be accessible through the nativeTouchData property.

https://developer.apple.com/library/ios/documentation/UIKit/Reference/UITouch_Class/#//apple_ref/c/tdef/UITouchType

eg.
```
void ofApp::touchMoved(ofTouchEventArgs &touch){
    if(touch.nativeTouchData){
        UITouch * pencil = (UITouch *)((ofxiOS3dTouch*)touch.nativeTouchData)->data;
        CGFloat azimuth = [pencil azimuthAngleInView:ofxiOSGetGLView()];
       
    //A value of 0 radians indicates that the stylus is parallel to the surface; when the stylus is perpendicular to the surface, altitudeAngle is Pi/2.
    CGFloat altitudeAngle = pencil.altitudeAngle;
    }
}
```

Since new devices keep coming out this would be a flexible way of passing on native data to devices that can handle it without upsetting other. In this case the ofxiOSEAGLView added support for new iOS features, but other platforms could do the same. It is using the same paradigm used for passing custom data in bullet physics. The shortcoming is that it does not include strongly typed properties in the normal oF way, and in this case it is passing an ObjC pointer.

Unless someone wants the job to map all properties on all devices moving forward this is one way of handling it. This was put together quickly and works, but hasn't been road tested. Would welcome other design ideas for this.

### DIFF
--- a/addons/ofxiOS/src/core/ofxiOSEAGLView.mm
+++ b/addons/ofxiOS/src/core/ofxiOSEAGLView.mm
@@ -10,6 +10,7 @@
 #include "ofAppiOSWindow.h"
 #include "ofGLRenderer.h"
 #include "ofGLProgrammableRenderer.h"
+#include "ofxiOS3dTouch.h"
 
 static ofxiOSEAGLView * _instanceRef = nil;
 
@@ -267,6 +268,15 @@ static ofxiOSEAGLView * _instanceRef = nil;
 			ofNotifyEvent(window->events().touchDoubleTap,touchArgs);	// send doubletap
         }
 		touchArgs.type = ofTouchEventArgs::down;
+
+		//add stylus specific data if available
+		//can be extended as new types become available
+		if(touch.type == UITouchTypeStylus){
+			static ofxiOS3dTouch stylusTouch;
+			stylusTouch.data = touch;
+			touchArgs.nativeTouchData = &stylusTouch;
+		}
+        
 		ofNotifyEvent(window->events().touchDown,touchArgs);	// but also send tap (upto app programmer to ignore this if doubletap came that frame)
 	}	
 }
@@ -299,6 +309,15 @@ static ofxiOSEAGLView * _instanceRef = nil;
 		touchArgs.y = touchPoint.y;
 		touchArgs.id = touchIndex;
 		touchArgs.type = ofTouchEventArgs::move;
+        
+		//add stylus specific data if available
+		//can be extended as new types become available
+		if(touch.type == UITouchTypeStylus){
+			static ofxiOS3dTouch stylusTouch;
+			stylusTouch.data = touch;
+			touchArgs.nativeTouchData = &stylusTouch;
+		}
+        
 		ofNotifyEvent(window->events().touchMoved, touchArgs);
 	}
 }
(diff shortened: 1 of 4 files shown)


bakercp:
On a related note, I'd suggest that we consider pointer-based system -- ofxPointer demonstrates a drop-in, well-documented backward compatible system (as described in the road map).

https://github.com/bakercp/ofxPointer

PointerEventArgs capture pressure, depth, multitouch, pen input, pointer shape, rotation, etc.

https://github.com/bakercp/ofxPointer/blob/master/libs/ofxPointer/include/ofx/PointerEvents.h

It would make integrating pen events like this very easy.

danoli3:
Yeah if we could integrate ofxPointer and all this new code! Be so awesome!

Exciting improvements!!

borg:
That's great @bakercp. 
There certainly cannot be any harm in mapping the most common properties using the W3C standard, and it seems you have done most of that work already. How are you suggesting handling properties not mappable to current standards though? I can imagine new devices introducing unique properties (eg. muscle tension). Would you propose always trying to map those properties for devices that support them, or allow for a native data pointer as in my suggestion?

I imagine the ofxPointer and ofxTouchPad make working with devices like [sensel]( http://www.sensel.com/)  and [tactonic](http://tactonic.com/) simple?

How about 3D inputs though? Would it be a smart thing to try to integrate things like Leap Motion or even Kinect 3D pointers into the same framework?

bakercp:
I think I'd probably try to limit the pointer events to things that specifically engage user interface / interaction in the traditional sense.  

So, if a Leap Motion or Kinect or tactonic could be used to generate pointer events, then they could inject them into the event pipeline easily.  Any data that would be considered experimental or truly hardware specific and not widespread and not cross platform (this is a moving target of course) would still come via custom events.  I'm not completely convinced that the native data pointer approach mixed with a standards based approach is the best fit for openFrameworks (i.e. with the desire to avoid pointers and associated memory management confusion).  I'd say that those highly customized or native / platform-specific behaviors are in contrast to the spirit of the PointerEvent spec which attempts to unify hardware platforms through a common interface.

Regarding your specific 3d pointer data -- I'm pretty sure PointerEvents can capture all of that data -- there is a lot there already about pressure, 3d points, pointer, shape, etc.  You can even define the device type via a string so that it's easier to interpret.

borg:
If PointerEvents already allow for a free form device string, a simple way around pointers & memory issues without losing a possibility to pass custom data would be to simply add string, float and int vectors where device specific data can be added and parsed as needed. (eg. how OSC is commonly used). New devices could define those properties with their own typedef enums to enforce some gentle strong type.

Maybe it's just a matter of taste, but it appears handling the two situations a) multiple concurrent input devices to one app and b) running the same app on different devices with different input device support would be handled gracefully this way by piping all input events through the same handlers and simple check for custom features as need be. I realize it means opening the door for some hacking, but the alternative would be to have to write custom events and event handlers for every device and/or having to hack the core to support new device features (which will always be the case, but not having to time core releases with device releases would be nice). 

Also piping all input events through the same handler is nice when using other touch managers eg. Gal's [scenegraph](https://github.com/galsasson/ofxInterface/blob/master/example-basic/src/ofApp.cpp).

After all avoiding separate event handlers for mouse, touch and stylus input seems to be the purpose of the standard. I'm just trying to stretch the implementation a bit further to arrive at the same goal. I'm sure some clever templating could solve it nicer than I propose here. 

Thanks for your thoughts. Integrating your system into the core would be fab.

bakercp:
In the last week, I've significantly updated ofxPointer to fully support all of this stuff (including Apple Pencil features, etc).

It now supports a host of `UITouch` feature on iOS including coalesced and predicted touches, pressure, tilt, estimated property updates, etc.

https://github.com/bakercp/ofxPointer

Please check it out if you're still looking for solutions here.

danoli3:
Oh awesome!!!!

PR #4806, opened 2016-02-10
Msys2 freeimage fixes

oxillo:
FreeImage package name update in MSYS2 package repository


arturoc:
is this broken in 0.9.2 then?

oxillo:
I haven't checked but, unfortunaltely,  I would say Yes if you have a clean MSYS2 install or upgraded it duraing the last days. The FreeImage package requires the libraw-10.dll that has been upgraded to libraw-15.dll.

arturoc:
so both things are because of something that has changed in msys it self not a bug on our side?

oxillo:
Not an OF bug !

arturoc:
i've opened a 0.9.3 milestone and created a branch of the same name since there's a couple of critical and regressions that we can fix quickly, can you send this PR to that branch instead? thanks!

arturoc:
this is the same we merged yesterday into 0.9.3 right? in that case we can close it and i'll merge 0.9.3 into 0.10

oxillo:
Yes, they are the same. I close it.

arturoc:
great, thanks!

PR #4807, opened 2016-02-10
MSYS2 : replace Fmod by OpenAL

oxillo:


oxillo:
integrated on MSYS2 64 bits #6413 - MSYS2 32 bits stays on FMOD

PR #4808, opened 2016-02-10
Update Poco and openssl formulas.

bakercp:
- Poco - whitespace (made tabs / spaces consistent).
- Poco - include and link to our openssl build (not easily available on newer OS X as a system lib).
- OpenSSL - update library to latest release.
- OpenSSL - Copy opensslconf.h to build directory so Poco on OS X can use it as a dependency.


arturoc:
can you send this PR to https://github.com/openframeworks/apothecary

i'm in the process of moving apothecary back to it's own repo where the binary libraries will be built by the test servers. that way we will get rid of the binaries in the main repo

arturoc:
this is already in the apothecary repo going to close this one

PR #4810, opened 2016-02-11
Adds tests for raspberry pi

arturoc:
Creates a raspbian image using multistrap and the raspberry tools
to be able to cross compile from linux64 to linuxarmv6l


tpltnt:
looks good to me :+1:

PR #4818, opened 2016-02-14
Make Mouse Events to Floats

DomAmato:


arturoc:
we should also change the app callbacks to receive floats then and have the old method with ints being called by that in case the new version is not overriden so old apps still work

PR #4822, opened 2016-02-15
OF_USE_GST_GL fixes

gohai:


arturoc:
thanks, this is breaking the tests in linux though. i think the problem is that gstreamer gl is not installed in the test server. i think the best solution would be to put the corresponding pkg-config package in an ifdef in the makefile that way you can call

make USE_GL_GST 

and it'll use that package but won't use it by default or in the test server.

once this api is more mature and we add proper support for it then we'll remove that ifdef and add the package to install_depdencies

gohai:
@arturoc Something like https://github.com/gohai/openFrameworks/commit/4489b91199521d720f26e109489f17ec4749ed11?

arturoc:
yeah looks good perhaps can you change the constants so both have GL GST in the same order? right now the one in the code is GST GL and the one in the makefiles GL GST.

gohai:
(still untested btw)

arturoc:
seems good to me and the tests are passing so i'm going to merge it. thanks!

gohai:
Thanks for hacking this in the first place. I'll see if I can get it to work on the Pi with Gstreamer 1.4.4 and/or 1.6.

gohai:
Not quite, but I was at least able to test the earlier commits in this PR.

arturoc:
thanks!

PR #4824, opened 2016-02-16
Add trySend to ofThreadChannel

elliotwoods:
Also:

* `TrySendStrategy` enum
* `TrySendresult` struct (set of bools)
* `size()` is now a public method (i didn't do any locking, since `empty()` didn't do any locking either
* `private` -> `protected` (in case anybody else needs to add their own functionality)

### DIFF
--- a/libs/openFrameworks/utils/ofThreadChannel.h
+++ b/libs/openFrameworks/utils/ofThreadChannel.h
@@ -219,7 +219,7 @@ public:
 	///
 	/// This method moves the contents of the sent value using `std::move`
 	/// semantics. This avoids copying the data, but the original data data will
-	/// be invalidated. Note that the original data will be invalideated even if
+	/// be invalidated. Note that the original data will be invalidated even if
 	/// the send fails because the channel is already closed.
 	///
 	/// ~~~~{.cpp}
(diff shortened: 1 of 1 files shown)


elliotwoods:
oops
i made an error
will edit again...

libs/openFrameworks/utils/ofThreadChannel.h
```diff
@@ -256,6 +256,153 @@ class ofThreadChannel{
...
+					true,
+					true
+				});
+			}
+			case DiscardStrategy::WaitForSpace:
+			{
+				do {
+					lock.unlock();
+					lock.lock();
+				} while (queue.size() >= maxSize)
```
elliotwoods: This do while is totally evil
but what is the best strategy?
Ideal is wait until next release of lock (e.g. keep lock unlocked, and wait until next unlock).

We could add another conditional value which `receive` but that seems 'selfish' for this function to add weight to receive. Or we could add an optional sleep time argument, but sleep is always at least 50ms AFAIK.

elliotwoods:
also we need to use `queue(move(value))` in our `(T && value` functions
otherwise we get:
image

elliotwoods:
p.s. sorry for doing this on openFrameworks directly
i didn't realise that GitHub would make the branch for the PR within the oF repo itself

arturoc:
Some comments:

- size should probably be something like sizeHint since the number you get can have changed before you even read it
- I don't like the struct much, it seems very error prone, in the sense that it has several states that are sometime muttually exclusive so it doesn't make much sense to have them all i think. an enum would probably be clearer and if we use ChannelClosed=0 then the old syntax will still work
- Something that other languages do is mix the error code and the ok result in one struct that contains an enum too so for example we could have something like

```
template<typename T>
class ofThreadChannel{
public:
    enum ErrorCode{
        ChannelClosed,
        Ok,
        ChannelFull,     
    }
    struct RecvResult{
        ErrorCode code;
        T value;
    }

   ErrorCode send(T&);
   RecvResult receive();
....
```

But not very sure about this, is not very idiomatic in c++ and not sure if it will incur on extra copies.

- In your implementation you have both WaitForSpace and trySend which seems redundant no?

elliotwoods:
ok at dinner now
if i:
* Change the return from struct to enum (as you outlined)
* change `size` to `sizeHint`
Then can we pull?
changing the return type of `receive` would be a breaking change for anything using ofThreadChannel so maybe comes later

do you agree with the `&&`s?
* I changed the code for the existing one to use `move` (it just calls the non-`&&` method on `queue.push` otherwise)
* The way I implemented my 2 functions to use the code in the `&&` version

in a bit of a rush to get this fixed so be good to know what's needed to get this moving

arturoc:
yeah let's keep receive as it is, we can see later if it's something we want.

move is ok, visual studio seems to be much less "smart" about guessing when it needs to move things and you explicitly have to use move much more than with clang or gcc.

apart from the struct and sizeHint i don't understand very well why you need both trySend and the wait strategy.

btw are you testing this? seems like you are hacking it directly in the github page :)

i would like to be really careful with this class it has to be really sound in terms of thread safety... and i would like it to be the main class for thread sync in OF so the API has to be really consistent too, if you are in a hurry perhaps it's better if you pull it out as an external class and experiment with it a little to see how it feels.

in any case we are at the beginning of the release cycle so we can always change things later but there's still some things i'm not very convinced about

elliotwoods:
My local oF git has a corrupt submodule so i'm copy/pasting from my working repo into the GitHub editor. Testing here with about 8 thread channels (using different parts of the API) in 2 applications running in sync.

On my XCode, it seems `send(T && value)` is still calling `void push(const value_type& __v) {c.push_back(__v);}` in `<stack>`, but it won't let me put a breakpoint in there (and XCode intellisense is pants) so it's difficult to tell. I presume we need to stay explicit to ensure we're getting what we pay for.

Yes totally, we can eliminate `WaitForSpace`, and just use `DiscardNew` (the user then looks at what is returned and decides whether to wait).

if we should wait for the `trySend` feature, then i suggest we do 2 things now:

* add `size` or `sizeHint` (since we have `empty`, i would either stay without the `...Hint` or change both
* change `private` to `protected`

without any type of `size`, this class is unsafe to use in the wild. so i'd consider that a bug.

changing `private` to `protected` just makes it less aggressive to anybody who wants to use this class who needs patterns which it doesn't cater for.


then for later, `trySend` can be considered a feature which enables a specific use-case (not actually the one i need).

arturoc:
yes, adding move is not a problem at all and better to be really explicit about it, it just casts whatever var to an rvalue so it can be used without problem in most cases.

you are right about size/empty sizeHint/emptyHint but don't see how not having size is a bug? all the implementations i've used in other languages don't have size and i don't remember even needing it. if we add something like drop and block on add i don't see the need for it but i'm not opposed to add it, really just curious why you think it's a bug or unsafe not to have it.

changing vars to protected is not a good idea, i wouldn't even inherit from a class like this but we can create a protected api through methods that allow to do certain operations that we don't want to be public.

having protected vars means that if for example we later change the current queue to something like: https://github.com/cameron314/concurrentqueue or even our own internal impl it will break the code for anyone extending the class.

elliotwoods:
if we can set a maxsize then that's great, but i'd still use size also

but without either, there's no way for the user to assure that they're not sending faster than the receiver can handle (and therefore causing an allocation crash at max user memory).

in my specific case, one `ofThreadChannel packetizerToSocket`. On the receiving side is something like:

```c++
while(threadRunning) {
    DataGram dataGram;
    if(packetizerToSocket->receive(dataGram)) {
        socket->send(dataGram);
    }
}
```

On my system (and on my network) this never backs up, but we've found that on some systems the packetizerToSocket just gets really really full.
my solution is to (3 threads earlier) not send any frames into the thread chain whilst there's too many packets to send (i.e. need `size` not `maxSize` for this solution).


the implementation by cameron314 has `size_approx`, but also is extremely 'generous' in terms of methods exposed.

arturoc:
I usually solve this by using tryReceive until there's no more frames in the side that runs slower, that way you consume the whole queue. the problem with that is that you may be doing more work than really needed in the first place but that's solved in a more efficient way if we have a limit that blocks the sender

anyway as i said no problem adding sizeHint just curious why you needed it.

also that implementation i posted is not a channel is just a thread safe, lock free queue that we could use internally in ofThreadChannel to substitute the current std::queue. That would break any code using protected variables in several ways: because of the different methods names but for example also because of not having to use the mutex to access the queue which is much more difficult to fix

arturoc:
btw, there's some other pattern/library that we should look into cause it's something similar to channels but it aims to solve the exact same problem you were discribing above. it's called reactive streams and it's mostly about sending messages from a producer to a consumer that potentially work at very different rates in the most efficient way, without locking and without doing any extra work on any of the sides.

i've been wanting to look into it for a while but haven't looked at the precise implementation or api details yet but i'm pretty sure there's some ideas we can adapt

PR #4830, opened 2016-02-17
OS X Xcode Template project / Legacy and AppStore Fixes

danoli3:
Fixes issues brought up by @kylemcdonald in https://github.com/openframeworks/openFrameworks/pull/3563

Also introduces a new deployment target of AppStore which provides the functionality required by https://github.com/openframeworks/openFrameworks/issues/4619

---

### What has changed?
- Added new deployment target (AppStore) which defines a GCC Pre-processor defined as APPSTORE that allows the build/run script to only code sign and package up an app when building for AppStore (distribution) target.
appstore 2x
- Run Script Example:
runscript 2x

---

@ofZach @ofTheo This should be good with PG, just verify not going to be an issue.


ofTheo:
Thanks @danoli3
This looks good on my end!

danoli3:
Probably merge this post ```0.9.3``` is merged back into master. @arturoc (since that branch modified the same project file).

ofZach:
this looks good to me!    

as a side note, I wonder if once the PG gui supports templates (which is planned, templates just came kind of late in the dev process for 0.9) we could split this into two and remove the app store things from the main template (and folks can use the app store template if they want this functionality).  I think it's good to avoid making the main template more complicated with multiple targets (just advocating here for beginners seeing maybe too many choices for targets, etc).   I think it's also because I haven't seen that many OF app store apps (see for example, that apple rejected this: http://nickhardeman.com/603/bass-shapes/) I feel like this doesn't necessarily belong in every project and is kind of specialized, similar to how the no-fmod template is specialized, etc.    If you want this feature, you can find it, but most people won't need to use this or see this. 

at any rate, I think this is a good solution for 0.9.3 and happy to revisit this down the line!  it's def. great that we have this as a possibility...  thanks for doing this!

danoli3:
This change here is for 0.10.x
Currently 0.9.3 has just the old functionality (pre 0.9.1 no copy process) ... should I PR this into 0.9.3

ofZach:
oh sorry I misunderstood --  I think moving back to 0.9 templates is good for now.

danoli3:
All good! I'll keep this for 0.10.0

danoli3:
This one is good to merge now once travis is green.

danoli3:
:+1:

PR #4832, opened 2016-02-17
Fix for FreeImage BOOL issues on OSX

danoli3:

(2009) https://github.com/openframeworks/openFrameworks/issues/76
(2013) https://github.com/openframeworks/openFrameworks/pull/1889


arturoc:
mmh, better send this to 0.9.3 only i'll merge all of it into master later

danoli3:
Yeah makes sense!

PR #4833, opened 2016-02-17
0.9.3 - FreeImage BOOL fix for OS X 10.8

danoli3:
Fixes issues with OS X and BOOL issues on some compilers.

Only seems to be an issue for older compilers / systems.

Older issues that solved this: 
(2009) https://github.com/openframeworks/openFrameworks/issues/76
(2013) https://github.com/openframeworks/openFrameworks/pull/1889

This has cropped up again in 2015 due to updating to 3.17.0 and this no longer being an issue on Modern OS X Xcode compilers.

PR #4834, opened 2016-02-17
0.9.3 - Fix for FreeImage 3.17.0 OS X and Windows.

danoli3:

(2009) #76
(2013) #1889

--

Previous PR:
https://github.com/openframeworks/openFrameworks/pull/4833

### DIFF
--- a/libs/FreeImage/include/FreeImage.h
+++ b/libs/FreeImage/include/FreeImage.h
@@ -150,22 +150,22 @@ FI_STRUCT (FIMULTIBITMAP) { void *data; };
 #ifndef _MSC_VER
 // define portable types for 32-bit / 64-bit OS
 #include <inttypes.h>
-typedef int32_t BOOL;
-typedef uint8_t BYTE;
-typedef uint16_t WORD;
-typedef uint32_t DWORD;
-typedef int32_t LONG;
-typedef int64_t INT64;
-typedef uint64_t UINT64;
+#define BOOL int32_t;
+#define BYTE uint8_t;
+#define WORD uint16_t;
+#define DWORD uint32_t;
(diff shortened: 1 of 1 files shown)


danoli3:
Okay this should work now!

danoli3:
!!!!!!!!!!!!!!!!!!!!!

ofTheo:
its the semi colons! 
just remove them and you should be good :)

libs/FreeImage/include/FreeImage.h
```diff
@@ -158,14 +158,25 @@ typedef int32_t LONG;
...
-typedef long LONG;
-typedef signed __int64 INT64;
-typedef unsigned __int64 UINT64;
+#define BOOL long;
+#define BYTE unsigned char;
+#define WORD unsigned short;
+#define DWORD unsigned long;
+#define LONG long;
+#define INT64 signed __int64;
+#define UINT64 unsigned __int64;
```
ofTheo: anything with #define shouldn't have the ; at the end 
also from travis looks like the the apple check isn't working as its using the #define BYTE and not the typedef

ofTheo:
@danoli3 also curious why there is both #define and typedef definitions - I think in the past we just did #define and then undef'd them at the end of the header, so it doesn't pollute other classes.

danoli3:
I went crazy with trying to fix it! I'll remove semi-colons! Seems more legit!

danoli3:
Lets just see if this builds regardless :3

libs/FreeImage/include/FreeImage.h
danoli3: Reverting!

PR #4838, opened 2016-02-17
FreeImage FIX!

danoli3:

(2009) #76
(2013) #1889

Previous PR:
#4833

## This is the final PR OMG


danoli3:
OGM!

PR #4841, opened 2016-02-17
VBO fixes for custom attributes on fixed pipeline.

prisonerjohn:
Hello, I'm bringing this up again because I noticed that the current ofVbo still doesn't behave as expected. The main issue is that `setAttributeData` is going through a different code path with the fixed pipeline, and not using the existing `VertexAttribute` objects for vertices, normals, colours, tex coords, etc.

With this change, something like the following will work the same under both fixed and programmable pipeline.
```
volVbo.setVertexData(volVerts, 24, GL_STATIC_DRAW);
volVbo.setNormalData(volNormals, 24, GL_STATIC_DRAW);
volVbo.setAttributeData(ofShader::COLOR_ATTRIBUTE, (float *)volVerts, 3, 24, GL_STATIC_DRAW);
volVbo.setAttributeData(ofShader::TEXCOORD_ATTRIBUTE, (float *)volVerts, 3, 24, GL_STATIC_DRAW);
volVbo.setIndexData(volIndices, 36, GL_STATIC_DRAW);
```

Sorry for not using the previous PR, but it seems to have completely vanished, probably because I got rid of the branch associated with it. For reference, this all started because I'm trying to use 3D textures with ofVbo.

cc @arturoc @tgfrerer

### DIFF
--- a/libs/openFrameworks/gl/ofVbo.cpp
+++ b/libs/openFrameworks/gl/ofVbo.cpp
@@ -382,42 +382,36 @@ void ofVbo::setIndexData(const ofIndexType * indices, int total, int usage){
 //--------------------------------------------------------------
 ofVbo::VertexAttribute & ofVbo::getOrCreateAttr(int location){
 	VertexAttribute * attr = nullptr;
-	if (ofIsGLProgrammableRenderer()) {
-		switch (location){
-			case ofShader::POSITION_ATTRIBUTE:
-				attr = &positionAttribute;
-				break;
-			case ofShader::COLOR_ATTRIBUTE:
-				attr = &colorAttribute;
-				break;
-			case ofShader::NORMAL_ATTRIBUTE:
-				attr = &normalAttribute;
-				break;
-			case ofShader::TEXCOORD_ATTRIBUTE:
-				attr = &texCoordAttribute;
-				break;
-			default:
-				customAttributes[location].location = location;
-				attr = &customAttributes[location];
-				vaoChanged = true;
-				break;
-		}
-	}else{
-		customAttributes[location].location = location;
-		attr = &customAttributes[location];
-		vaoChanged = true;
+	switch (location){
+		case ofShader::POSITION_ATTRIBUTE:
+			attr = &positionAttribute;
+			break;
+		case ofShader::COLOR_ATTRIBUTE:
+			attr = &colorAttribute;
+			break;
+		case ofShader::NORMAL_ATTRIBUTE:
+			attr = &normalAttribute;
+			break;
+		case ofShader::TEXCOORD_ATTRIBUTE:
+			attr = &texCoordAttribute;
+			break;
+		default:
+			customAttributes[location].location = location;
+			attr = &customAttributes[location];
+			vaoChanged = true;
+			break;
 	}
 	return *attr;
 }
 
 //--------------------------------------------------------------
 void ofVbo::setAttributeData(int location, const float * attrib0x, int numCoords, int total, int usage, int stride){
-	if(ofIsGLProgrammableRenderer() && location==ofShader::POSITION_ATTRIBUTE){
+	if(location==ofShader::POSITION_ATTRIBUTE){
 		totalVerts = total;
 	}
 
 	bool normalize = false;
-	if(ofIsGLProgrammableRenderer() && !hasAttribute(location)){
+	if(!hasAttribute(location)){
 		vaoChanged = true;
 		bUsingVerts |= (location == ofShader::POSITION_ATTRIBUTE);
 		bUsingColors |= (location == ofShader::COLOR_ATTRIBUTE);


arturoc:
i remember we did this for some reason, in GL 2 there's standard "attributes" for vetices, colors, textures... that you access with gl_Position... in the vertex shader. I think if you do it the way it's done in this PR then some custom attribute locations become unusable, for example vertex is location 1 in OF so if you have for example a point size attribute in your shader it'll be given location 1 by default which will become inaccessible with this change

arturoc:
btw i'm not 100% sure on the details right now but i remember it's related to that

tgfrerer:
I've been looking into this too, and i think this PR should do the trick.

Probably also a good idea to also check the `clear()` and `updateAttributeData()` methods are in line with this PR to make sure we would not be leaking attributes or somehow make them inaccessible.

The difference between GL 1.2 and modern programmable GL was that the default attributes needed specialised methods to upload to the GPU, like `glTexCoordPointer` to push texture coordinates. 

`ofVbo` deals with this by having a set of `VertexAttributes` (1) defined by default that match the fixed pipeline, and a map for any custom attributes that might be set. Both, fixed and programmable pipeline are routing the default attributes (Position/TextureCoordinate/Normal/VertexColor) through the set described in (1), with the fixed pipeline using the specialised methods to submit attribute buffer data.

The way it's currently implemented, the fixed function implementation will correctly submit 2D *and* 3D texture coordinates, so i would expect this PR to work in these respects. But you won't be able to, for example, submit 4D normals, since that call in fixed OpenGL will get routed through `glNormalPointer` which is rather opinionated about what it will accept. In programmable OpenGL, that's no problem.

As for how custom attributes will interfere with this setup in OpenGL fixed, I would have to do some more research, but I can't immediately see the issue @arturoc describes.

That said, may i suggest we eventually deprecate the fixed function renderer? I feel like maintaining backwards compatibility with an old and complex api is costing us a lot of time and code simplicity.

arturoc:
@tgfrerer not sure as i said but if we have a shader like:

```glsl
attribute float pointSize
void main(){
...
}
```

and you do:

```
auto loc = shader.getLocation("pointSize"); // this would usually be 0?
volVbo.setAttributeData(loc, ...)
```

that will end up in the position or at least falsely enabling bUsingVerts instead of the pointSize since 0 == ofShader::POSITION_ATTRIBUTE.

tgfrerer:
I think the problem is that in fixed function an `attribute` was everything that was attached to a vertex that was extra, therefore not: `Position`, `Texture`, `Normal`, `Color`, which was covered by the default use case. These extra attributes in fixed GL were addressed by index, and the first one would be placed at index 0.

In programmable GL, everything apart from indices is an attribute, and attributes are indexed, starting at `0=position`,`1=color`, `2=normal`,... in the most common use case. 

Therefore additional attributes in fixed GL would most likely start with an index higher than 0, so yes, in the case outlined above, when querying a shader to get the attribute mapping, the shader would return wrong indices. I wonder if this isn't an issue with `ofShader` rather than `ofVbo`, at least in the fixed function code path?

arturoc:
well you could bind those extra attributes starting from 5 lets say to skip the ones we use as default ones but we've seen in the past that that incurred in a huge performance hit in some platforms so it's probably not an option. 

we could also do some offseting inside ofShader to return +5 but then use -5 for attribute locations so we could skip the locations asigned to the "default" attributes but that sounds really hacky.

i'd say that unless we found a really elegant way of doing this it should stay as it is and we just support this feature in the programmable renderer

prisonerjohn:
Hmm I had a feeling this would be more complicated than I originally thought... 

How come offsetting everything by 5 causes a performance hit? Is it an OF issue or is it a GL thing where it doesn't like skipping over empty attribute locations?

If it's along the lines of the latter, would it make sense to have this "5" offset be variable? So if you're using verts and normals but no texcoords or colors the offset would be 2.

tgfrerer:
The offset performance problem is most likely a driver issue. We noticed it with NVidia cards, and it was related to the first attribute not being 0-indexed. If you do a search over the oF code base for "performance hog" you will find the line in question, and nearby an url to some more docs as a memento =)

I fear any attempt at backporting this to fixed function GL is going to be rather complex, also because there are no good tools (debuggers/inspectors like NSight, RenderDoc) that support inspecting fixed function GL anymore that I know of- it appears that all have moved on to programmable GL.

Are you locked in to fixed function with your project because of dependencies/limited hardware or is there a chance you could use programmable GL? I'm asking to understand why fixed function GL still appears to popular...

prisonerjohn:
Not at all, I've been on programmable pretty much exclusively! I just noticed that it behaved differently on fixed (and I'm updating the ofxVolumetrics add-on), so I figured I'd try to fix it, or at least bring up the issue. 

But if it's a complex problem that might break in a lot of unexpected places, I understand we might just want to leave as-is :)

arturoc:
going to close this since it's a known issue but doesn't have an easy solution

Issue #4842, opened 2016-02-17
Resizing the window breaks ofGetWidth & ofGetHeight


braitsch:
Hi, it looks like ofGetWidth is acting up again. If I enable hi-res ofGetWidth reports 2x the window width on window resize but only reports 1x everywhere else. To see what I mean create an empty project, enable hi-res and print out the width within setup, a keypress and resize handler.

````
void ofApp::setup(){
    cout << "ofGetWidth() " << endl; // 1024
}
void ofApp::keyReleased(int key){
    cout << "ofGetWidth() " << endl; // 1024
}
void ofApp::windowResized(int w, int h){
    cout << "ofGetWidth() " << endl; // 2048
}
````
OF v0.9.2 
OSX v10.10.5

braitsch:
I found the problem. The pixelScreenCoordScale multiplication was happening twice, once in the resize callback and again in the getWidth call. 
````
void ofAppGLFWWindow::resize_cb(GLFWwindow* windowP_,int w, int h) {
	ofAppGLFWWindow * instance = setCurrent(windowP_);
	cout << "w = " << w << " scale = " << instance->pixelScreenCoordScale << endl;
// don't modify instance dimensions here since queries (e.g. getWidth & getHeight)  
// will return the window values adjusted by the pixelScreenCoordScale
	instance->windowW = w;// * instance->pixelScreenCoordScale;
	instance->windowH = h;// * instance->pixelScreenCoordScale;
	cout << "sending " << w*instance->pixelScreenCoordScale << endl;
	instance->events().notifyWindowResized(w*instance->pixelScreenCoordScale, h*instance->pixelScreenCoordScale);
	instance->nFramesSinceWindowResized = 0;
}

int ofAppGLFWWindow::getWidth(){
	if(windowMode == OF_GAME_MODE)
	{
		return getScreenSize().x;
	}
	else {
		if( orientation == OF_ORIENTATION_DEFAULT || orientation == OF_ORIENTATION_180 ){
			return windowW * pixelScreenCoordScale;
		}else{
			return windowH * pixelScreenCoordScale;
		}
	}

}
````

I'll submit a pull request in a few minutes.

braitsch:
Noting that this bug was introduced in v0.9.1 and breaks any app running in hi-res fullscreen mode that relies on querying the window dimensions to lay itself out.

arturoc:
[...]

Issue #4746, opened 2016-01-22
ofGetWidth() returning incorrect value when High Resolution Capable is enabled in 0.9.1


braitsch:
Seems like something's changed between 0.9.0 & 0.9.1. If I create an empty project in each and enable "high resolution capable" in the plist settings, ofGetWidth() returns different values.

v0.9.1

    ofGetWidth() -> 1024
    ofGetScreenWidth() -> 2880

v0.9.1

    ofGetWidth() -> 2048
    ofGetScreenWidth() -> 2880

One problem this presents is that if I position something based on what ofGetWidth() returns it will potentially be drawn off screen.

Xcode 7.1.1 OSX 10.10.5

braitsch:
Hi Theo,
I just pulled down 0.9.1 and ofGetWidth() is still returning 2x the actual window width when retina is enabled. For example if I run an app fullscreen on my rMBP I get the following:

    ofSetFullscreen(true);
    ofGetWidth() >> 5760
    ofGetScreenWidth() >> 2880

Is this the desired behavior?
If so I think this is kind of confusing as the window isn't actually 5780 pixels wide, it's 2880.

ofTheo:
@braitsch - thats strange on my rMBP gives:

 ofGetWidth() 3840
 ofGetScreenWidth() 3840

My laptop is set to do 1920x1200 so the 3840 is 2x that. 
That would be expected behavior - I don't know why your seeing something different.

ofTheo:
I should note, I am about to start working on a proper retina support for the 0.10.0 release, so things might change a bit.  

The goal being that 1024x768 in retina would give you a window that is the same size as a window that is 1024x768 in a non retina app, but with double the pixels and OF internally handling the pixel to point ratio.

arturoc:
not sure what's the way to work with "retina" but isn't this confusing? if you set the pixel size you probably want the exact size in pixels? there could be other units in which to set the screen size that are resolution independent but setting the size in pixels and getting something else sounds kind of weird. btw i've created a 0.9.2 branch since there was a couple of small bugs that need to be fixed so if you need to fix this please send the PR there instead of master

ofTheo:
@arturoc @braitsch - I've set another issue to discuss the retina stuff. #4789

PR #4844, opened 2016-02-17
fix for #4842 removing pixelScreenCoordScale multiplier from window dimensions in resize callback

braitsch:


braitsch:
This resolves #4842 & #4746

ofTheo:
Doesn't the OF app resize callback then get the wrong values? 

Thanks!!

braitsch:
I don't believe so. It appears that the callback receives the dimensions at 1x and dispatches them out at 2x if high-res is enabled which I think is fine since ``getWidth()`` & ``getHeight()`` are also checking if high-res is enabled before returning a value.

PR #4851, opened 2016-02-18
ignore addons.make in examples folders

grayfuse:
Ignoring addons.make in examples folders, so that project generator doesn't create a great deal of new untracked files to the local git projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ examples/**/[Dd]ebug*/
 examples/**/[Rr]elease*/
 examples/**/gcc-debug/
 examples/**/gcc-release/
+examples/**/addons.make
 tests/**/[Dd]ebug*/
 tests/**/[Rr]elease*/
 tests/**/gcc-debug/
@@ -127,4 +128,3 @@ Desktop.ini
 
 .mailmap
 /apps*/
-


ofZach:
Not sure I understand the problem you are trying to fix here, can you elaborate ?

arturoc:
@ofZach when you run the PG on an example in the github version it always creates an addons.make which is not ignored so if you do git status it always appears as unstaged

the problem with this solution is that some examples need addons so this files can't be ignored. a posibility would be to add an addons.make file to every example, empty for those that don't use any

bakercp:
I think we should have addons.make and other makefiles in each project by default.  They are the only "project" files that don't need to be updated.  Minimally, addons.make should be git add -f to those directories.

arturoc:
this is wrong @bakercp it's ignoring all the addons.make so from now on examples with addons.make won't get their addons.make file into the repo. can you revert it?

bakercp:
Will fix.

grayfuse:
Sorry guys, didn't mean to screw with examples in this way, just didn't want to unnecessarily commit a bunch of addons.make files by accident during the doc sprint. Makes sense to me to have addons.make be part of every example (even if empty). Could help to provide consistency that will help explain the build process using make, etc...

PR #4853, opened 2016-02-19
Inline documentation for ofBaseVideoPlayer

brannondorsey:


bakercp:
I'm reviewing now.

bakercp:
Done reviewing.  PR sent against your branch.

brannondorsey:
Merged PR, thanks for the edits!

arturoc:
thanks!

bakercp:
@arturoc just moving forward -- how was that PR -- is this style (we tried to keep it as clean as possible and limited to one small portion) -- seem workable?  Any changes you'd propose?  Or should we modify our strategy?

arturoc:
no that's great.

bakercp:
Suuuper.  Thanks.  We'll keep them flowing today.

arturoc:
Actually one thing: Something that i'd like to change is to not use the \brief tag and instead just use a first line with a double break line which doxygen and clang (so i guess xcode too) parse without problem and it's much nicer to read when you are looking directly at the headers:

```cpp
	/// \brief Set the anchor point the item is drawn around in pixels.
	///
	/// This can be useful if you want to rotate an image around a particular
	/// point.
	///
	/// \param x Horizontal texture position in pixels.
	/// \param y Vertical texture position in pixels.
```

would be

```cpp
	/// Set the anchor point the item is drawn around in pixels.
	///
	/// This can be useful if you want to rotate an image around a particular
	/// point.
	///
	/// \param x Horizontal texture position in pixels.
	/// \param y Vertical texture position in pixels.
```

we use \brief everywhere else but perhaps we can start using this notation?

bakercp:
Yes -- I've been thinking about that as well.  I'll do a double check on doxygen's ability to infer this and modify our style guide.

bakercp:
For the moment the low level working group is going to keep including `\brief` for consistency's sake (we are in c.  If, in the future we want to remove these, it should be straight forward.

arturoc:
don't mind much but this is the kind of thing we'll never change because it touches so many places at once, so even if it's not very consistent i would just start using the new style

Issue #4846, opened 2016-02-18
ofEasyCam : separated viewport and mouse inputs area


BaptisteTheoriz:
Hi guys,
Should it be possible to add a function to separate camera's viewport and mouse input area ? It could be useful when rendering easycam inside a fbo which can be larger than app window, and then drawn scaled or translated in window.

arturoc:
I don't think we want to have this as part of ofEasyCam, it would probably complicate the internal logic quite a bit to cover a case that seems pretty rare.

Something that could work is:

- making the event callbacks public, that way you can unregister the camera from the core events and pass your own mouse coordinates that are pre transformed to fit your needs
- ofEasyCam has a setEvents function that allows to use ofEasyCam in different windows, we could make that function more flexible so people could create their own "core events" class that transforms the coordinates and ofEasyCam would register to those
- Add some kind of transformation matrix to ofCoreEvents so you can just plug a custom transformation into the events

or something along those lines, i guess the idea is transform the events not add more logic to ofEasyCam.

The first one of them is the easiest to implement but is probably more open to confusion, if someone for example uses ofEasyCam normally and tries to call the mouse events too but seems really unlikely

BaptisteTheoriz:
I didn't considered the problem from events point of view. 

But, for me, it sounds more complicated and confusing to create some custom events with remapped mouse positions than simply telling to the camera which screen area can control it. 

I tried to simply add a new ofRectangle inside easycam, and tell mousePressed ... functions to check mouse inside this rectangle instead of viewport. In my case, it does perfectly the job, allowing me to allocate a part of my screen to the control of the camera (and so, I can render my easycam inside a FBO of any size, then draw this FBO anywhere, scaled or not, and tell my easycam to take mouse inputs exactly where and how my fbo's texture is drawn).

In my opinion it's easier and clearer than creating custom events, but I could be wrong.

tgfrerer:
I agree with @BaptisteTheoriz here -- i've actually run into this situation a couple of times -- the classic scenario is to draw into an ofFbo and then render the fbo at half scale - if the mouse coordinates are not modulated by the render rectangle, the mouse input is super inaccurate. 

i believe it could be solved by passing a rect for the mouse rect as an optional parameter to ofEasyCam.begin() - with the default being the viewport rect...

arturoc:
[...]

PR #4854, opened 2016-02-19
Custom input area for mouse events in ofEasyCam

BaptisteTheoriz:
This proposal refers to #4846 : it allows user to set a custom area to control an ofEasyCam. 

It is especially useful when you want to render an ofEasyCam in an ofFbo : if you draw the fbo's texture scaled or translated on the screen (so texture drawn doesn't match camera's viewport), mouse inputs becomes inaccurate. This feature allows user to allocate a specific part of screen to camera's control, and in my example, you can set the input area so controls match where fbo is drawn.

To do that, I added an ofRectangle representing the area where mouse events are tested, and in mousePressed etc functions, I check if mouse is inside this area instead of the whole camera's viewport.
This area can be set optionally in begin() function (as viewport), and is set to viewport by default.

### DIFF
--- a/libs/openFrameworks/3d/ofEasyCam.cpp
+++ b/libs/openFrameworks/3d/ofEasyCam.cpp
@@ -64,11 +64,18 @@ void ofEasyCam::update(ofEventArgs & args){
 }
 
 //----------------------------------------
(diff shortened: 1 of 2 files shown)


arturoc:
this has been solved already with a different solution, a method to specify the active area

PR #4858, opened 2016-02-19
fix for #4842 removing pixelScreenCoordScale multiplier from window resize callback

braitsch:
Fix for #4842, #4746 & [datgui#39](https://github.com/braitsch/ofxDatGui/issues/39) where the window instance's dimensions were being multiplied by the pixelScreenCoordScale within the internal resize callback. The problem this created is that the dimensions are also multiplied by pixelScreenCoordScale within any query against the window e.g. getWidth, getHeight, getWindowSize etc so the multiplication was happening twice resulting in those calls returning a 4x value instead of a 2x value when pixelScreenCoordScale is greater than 1.


ofTheo:
Thanks @braitsch this works for me. 
Apologies for not getting this right for 0.9.1/0.9.2 

The travis error looks like it is unrelated and with android - so going to merge this in.

braitsch:
No worries, thanks for helping me resolve this.

PR #4860, opened 2016-02-19
Fix documentation for ofBaseVideoPlayer::getDuration()

brannondorsey:
Previously stated the value was in milliseconds but it is actually representing seconds.


bakercp:
@arturoc for small typos like this, I'll merge, unless you would like to all core merges.  Let me know.

arturoc:
no that's fine, i merged this class cause it's highly unlikely that any previous PR is modifying it

bakercp:
Got it. Thx.

PR #4861, opened 2016-02-19
Brings ofxCv into the core

arturoc:
Still needs some work. 

- openCv libraries should go out once we remove the binaries from the core and are built through apothecary
- ofxOpenCv addon_config.mk should be now ofxCv's and ofxOpenCv will depend on ofxCv

Something to discuss:

- ofxCv uses namespaces but nothing in the core is using namespaces yet. i think that's fine and we should probably move to namespaces at some point. What i find strange is the way files are laid out:

libs/ofxCv

has addon source which usually only contains external libraries. i would move the source inside src even if it's

src/ofxCv

or just rename the files so 

libs/ofxCv/Calibration.h -> src/ofxCvCalibration.h

ping @kylemcdonald


ofZach:
sweet !  -- another discussion point: 

 do you think it's worthwhile to move CLD out of ofxCv  (cc @kylemcdonald ) ?  it has always felt like it should be it's own addon.  (I'm not sure if there's anything else like that in ofxCv but that has always felt like it would benefit from being on it's own since it's not really opencv and it's kind of a specific algorithm).  I like the idea of core addons being more like "building blocks"

kylemcdonald:
not sure if there's still interest in this -- but yes, i'd move CLD out of ofxCv if it were to be added to the core.

bakercp:
Is this sync-d with the latest ofxCv master branch (which was overhauled recently for 0.10.0 support?).  I can't tell from the commit history ...

bakercp:
Opps, sorry, just realized this was a thread from 2016 that Kyle just woke up :) So, no, it probably isn't sync'd with the latest ofxCv.

ofZach:
I'm still supportive of bringing ofxCv into the core and sunsetting ofxOpenCv.....

PR #4864, opened 2016-02-19
Doc sprint examples

grayfuse:
This is the primary branch we were all branching off of. If these initial changes look alright, feel free to fold in, and then future PRs will be much smaller bite-sized commits from the other individuals at the Doc-Sprint.


bakercp:
Niiiiice!

arturoc:
this PR instroduces a new category in examples that shouldn't be there with a new example. @bakercp can you remove it?

PR #4867, opened 2016-02-19
systemSpeakExample: linux support, converting spaces into tabs

frauzufall:
adds support for linux by using this [espeak](https://en.wikipedia.org/wiki/ESpeak):
 
    echo "some words"|espeak

and converts some spaces into tabs.


bakercp:
Thanks!

PR #4880, opened 2016-02-20
docs_low_level: ofTimer.h

workergnome:
@bakercp, please review?


bakercp:
Reviewing ...

bakercp:
```c++
	/// \warning This will crash your application if you haven't called setPeriodicEvent().
```

true fact?  seems like this should be reported as a bug?

bakercp:
Related https://github.com/openframeworks/openFrameworks/issues/4883

bakercp:
Would like to see https://github.com/openframeworks/openFrameworks/issues/4883 fixed and the \warning removed from these docs :)

arturoc:
yes @workergnome can you remove the warning and instead open an issue? thanks!

workergnome:
It looks like issue #4883 already exists, and it references this pull request, so should we leave the warning in until that issue's resolved?

arturoc:
ah, ok, no better remove the warning.

workergnome:
Fair enough.  It's removed.

arturoc:
thanks!

PR #4882, opened 2016-02-20
Deprecate and replace ofRectangle::getPositionRef().

bakercp:


arturoc:
i'd say getPosition should be const only it can lead to surprises in some edge cases and i don't think we are returning basic types like this by non const reference anywhere else

bakercp:
I agree and I fixed this @arturoc Let me know what you think.

PR #4872, opened 2016-02-19
ofRectangle::intersects(const ofPoint& p0, const ofPoint& p1) fails in one case. Fixes #4871

bakercp:
See #4871 for discussion.

Issue #4871, opened 2016-02-19
ofRectangle::intersects(const ofPoint& p0, const ofPoint& p1) fails in one case. 


bakercp:
This should return true:

`ofRectangle(0, 0, 100, 100).intersects(ofPoint(0, 50), ofPoint(50, 0))`  It does not.

Thanks @anikahhh!

bakercp:
![blank](https://cloud.githubusercontent.com/assets/300484/13192211/bfa12e5e-d726-11e5-84d6-0426ab6e577b.png)

PR #4888, opened 2016-02-20
Add ofIsFloatEqual() for better floating point equality comparisons.

bakercp:
This is in response to a discussion in #4871 #4872 but also can be applied more consistently in other places in the core.

pinging various people that have been a part of the discussion here docsprint
@danomatika @arturoc @ofZach @golanlevin @anikahhh @workergnome


arturoc:
that seems overly complicated, i would just add something like ofIsPartiallyEqual or ofIsApproximatelyEqual to work with floats. not sure about the name but i think it should make clear that it's not exactly equal and for integers you will probably always use == anyway

bakercp:
So to be clear, what part is overly complicated?  I've attempted to clearly document the functionality and references and basically copied it from the C++FAQ.  It seems that we'd want to support both float and double (particularly as GLM etc) comes online.  The fact that in the C++FAQ they use the same name (`isEqual`) seems to be a good case for calling it isEqual (and the documentation makes it clear what it is actually doing and why).  The float/integral templating ensures that the the more complicated epsilon approach won't be used for integral types.  Most advance users would know to use `==` for integral types, but new users would benefit from having it "do the right thing" for integral types.

bakercp:
Also, if it matters, I chose return type specialization (instead of function parameter specialization) because I saw that it was done this way in ofEvent of::priv, etc.

arturoc:
what i was trying to say is that i don't think we need a function to compare every possible type. just something to compare floats because doing == with floats can fail.

I would prefer something like ofIsFloatEqual or isAproxEqual or something like that. and that perhaps it's easier to just have explicit versions for float and double than the current template that will fail with an undecipherable error message if you try to do do something like:

```cpp
ofVec2f p1, p2;
ofIsEqual(p1, p2)
```

danomatika:
I agree with Arturo in that the generic naming leads people to expect an implementation for other datatypes as well which is not needed. I think ofIsFloatEqual() would work fine with commenting along the lines of 

> In general, using == is fine when comparing floats without too many decimal places aka image widths, mouse positions, etc. However, when the number of decimal places increases, the floating point precision will go down and it becomes harder to compare things. Also, certain mathematical computations with result in infinitely large, infinitely small, or even irrational "Not A Number" values. In those cases it's recommended to use ofIsFloatEquals() which filters out such precision issues."

danomatika:
Also, maybe I'm a bad programmer but I have almost never needed to do extended checks when comparing floats. I guess I don't need much precision and when I do, I use doubles anyway.

bakercp:
OK.  I understand the concerns better now.  Am modifying it ...

bakercp:
@arturoc @danomatika I believe this resolves the concerns.

As soon as this is merged, I can update the ofRectangle bugs and take a pass to make sure it's in use in other appropriate places in the core.

bakercp:
@arturoc ping

arturoc:
sorry to be picky but isn't it easier to just have a float and a double version? if you try to call it with any other type it'll throw a really complicated error, if there were more types it could make sense to have a worse error message in exchange for code re usability but since it's only 2 functions i think it's worth

bakercp:
OK.  Makes sense.  I'll update it.

arturoc:
thanks, sfinae is such a terrible hack to solve that c++ doesn't have a way to specify the type of the template parameters. it's useful in some cases but is really ugly in terms of syntax, the error messages it produces... there's a couple of places in the core where it's used but there were no other option

bakercp:
OK I think this is cleaned up and ready to go as soon as CI finishes.  Once this is in I'll send fixes for the ofRectangle, etc.

bakercp:
@arturoc The updates failed hard on appveyor.  Probably because it gets ambiguous with float,float double,double.  How should we solve this?

arturoc:
mmh, yeah makes sense then i guess we need to use enable_if after all

bakercp:
Closing this to remove all the in-progress commits.  Check out the final version here #4958.

PR #4903, opened 2016-02-20
 Example for the point cloud example including readme and screenshot

bitchwhocodes:
did the readme and screenshot


bakercp:
Thanks!

PR #4907, opened 2016-02-20
Added the Quaternion spelled the incorrectly Arc Ball Example

bitchwhocodes:
did the readme and screenshot for the point picker which loads in a .ply file into a mesh, and allows mouse interaction to determine the index of the point.


bakercp:
Thanks!

PR #4917, opened 2016-02-21
Necessary files for 3DModelLoaderExample

crecord:
These files are necessary for the 3DModelLoaderExample, but weren't included in my original pull request.


bakercp:
Thanks!! (I moved this from addons to 3d).

PR #4920, opened 2016-02-21
Graphics example comments

grayfuse:
Example of adding in README stub in comments of example, to use as guide for future examples.


bakercp:
Add periods. :)

bakercp:
I'd propose changing it to :

`// Refer to the README.md in this example's root folder for more information.`

PR #4925, opened 2016-02-21
fixes #4923 allAddonsExample path in test

### LINKED ISSUE #4923: allAddonsExample gone
We were using it for testing that all the addons are still compiling so right now all the tests are failing. The PR that introduced this in the first place and broke the tests shouldn't have been merged until the tests were correct.

Anyway, can anyone fix it?

frauzufall:
This should fix https://github.com/openframeworks/openFrameworks/issues/4923.


arturoc:
thanks!

PR #4932, opened 2016-02-23
clean up network examples

firmread:
Improve code readability and small visual adjustments for
* oscReceiveExample
* oscSenderExample
* networkTcpServerExample
* networkTcpClientExample
* networkUdpReceiverExample
* networkUdpSenderExample


arturoc:
there's several things in this changes that go against the style guides for the code, for example the headers have been changed so the functions are in the same indentation level as public. there's extra tabs everywhere in the examples that weren't there before. there's some places where ofSetBackgroundColor is changed for ofBackground, we've been talking lately about which one is preferred and i think ofBackground will probably be deprecated. 

There's also a change in one of the osc examples that removes one of the posible messages, the one in mouseMoved

firmread:
Thank you for lightning fast feedback @arturoc! 

I think I've already fixed the indentation to follow the style guide but maybe something messed up in the pushing process. I copy and pasted the code from different location for this one. will try to rearrange the code from the original one instead and send a new commit in a bit!

I can switch back to ofSetBackgroundColor for those examples. but I also put ofBackgroundGradient in some of the example though. (I love how they looks! eye candies!) And since they won't work if it doesn't stay in draw function I think if we are to depreciate ofBackground maybe we have to implement ofSetBackgroundGradient too? 

I moved stuffs inside mouseMoved to update instead because having it there cause the mouseX and Y to be blocked when mousePressed event is called. (eg. dragging mouse in original oscSender and mouseX and Y won't update in the receiver end until the mouseReleased is called)

arturoc:
ok, perfect, yes if you can fix the indentation? also leave the main files as they were: we'll refactor all the main files soon to use the new ofCreateWindow function and it they all have the same format it will be easy to automate the process.

about background gradient, it's kind of an strange function, in my opinion it should just be ofDrawGradient since it's really not the same as clearing the background and it makes assumptions about depth testing... that probably shouldn't be making.

> I moved stuffs inside mouseMoved to update instead because

cool didn't noticed that but sounds good then

if you can revert the indentation, main and ofSetBackgroundColor bits everything else looks good

firmread:
This should be good now. 
I'm having strange stuffs going on when I push code to github though. Indentation is looking fine locally on my computer both on Xcode and Sublime Text but when I push it, it's all messed up on github. So I had to hand fixed on github a bit :\

arturoc:
that's because xcode uses spaces instead of tabs, we should change to spaces at some point. i

t's still incorrect, the main problem is that this will look wrong in some platforms, if it was core code i or anyone else could fix it at some point but in the examples it will look strange.

if you set sublime (not sure how to do it in xcode) to use tabs instead of 4 spaces you should be able to fix it easily. i think sublime even has an option to turn spaces to tabs and viceversa

otherwise a tool like: http://tabstospaces.com/ should do it automatically

firmread:
Sure, I get the point. Will try to see if I can correct this in Sublime Text. 
That site looks cool but seems like it only do the inverse of what we want at this point :)

firmread:
This commit should fix it. I do this in Sublime Text and reindented everything, they should now be all tab indented. 

I'm wondering if this tab vs. space thing should also be included in the style guide? I think it's actually straightforward if we say that everyone on Xcode should set it this way. 
screen shot 2016-02-23 at 12 20 15 pm

and if we think it's better to switch to spaces at some point we then can change the style guide accordingly.

arturoc:
great, seems it's even solved some previous indentation problems :)  thanks!

PR #4941, opened 2016-02-24
Make ofTexture::loadData with different glFormat/glType public

gameoverhack:
Currently this is a protected member - by making it public it allows users to allocate and loadData without having to modify glUtils/ofTexture for formats such as GL_BGRA/GL_UNSIGNED_INT_8_8_8_8_REV and GL_RGB_422_APPLE/GL_UNSIGNED_SHORT_8_8_APPLE which both require a mismatch between internal type and glFormat to work correctly.


arturoc:
makes sense, thanks!

PR #4944, opened 2016-02-24
Fix for ply mesh data loader with uchar color type

### LINKED ISSUE #4935: PLY vertex color data not working 9.2
Hi there,

It looks like ply vertex colors have stopped working between 8.4 and 9.2. [Here](https://forum.openframeworks.cc/t/ply-color-data-not-working-properly/21408
) is a forum post describing the issue. 

here is a reduced size sample ply with vertex colors:
[sampleVertexColor.ply.zip](https://github.com/openframeworks/openFrameworks/files/143190/sampleVertexColor.ply.zip)

Here is what is look  like loaded into OF 9.2 using mesh.load(); and mesh.drawFaces(); 
screen shot 2016-02-23 at 1 59 50 pm

Here is what is look like loaded into OF 8.4 
screen shot 2016-02-20 at 8 25 18 am

I am on xcode 6.2 on a 10.9 mac with a newly downloaded copy of 9.2 incase any of that is relevant.

tgfrerer:
When loading meshes from .ply files, color vertex attributes may be specified as `uchar`, not just float. 

The existing implementation was not allowing for this, and `uchars` were righteously (though not rightfully) interpreted as `float`s.

This PR suggests adding a flag for the vertex color type derived whilst reading the file's manifest section.

It then makes interpretation of color data type dependent on this flag, and scales `uchar` color data to normalised float range, so that both types, either `uchar` or `float` may be used to the same effect.

Fixes #4935

### DIFF
--- a/libs/openFrameworks/3d/ofMesh.cpp
+++ b/libs/openFrameworks/3d/ofMesh.cpp
@@ -792,6 +792,8 @@ void ofMesh::load(string path){
 	ofIndexType currentVertex = 0;
 	ofIndexType currentFace = 0;
 
+	bool colorTypeIsUChar = false; /// flag to distinguish between uchar (more common) and float (less common) color format in ply file
+	
 	enum State{
 		Header,
 		VertexDef,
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

PR #4947, opened 2016-02-25
update to close Android video player properly

noyanc:
The ofxAndroidVideoPlayer doesn't close properly.

I Play a video file, I stop it, I pause-resume the app and the video restarts again himself without calling him. Another issue, not closed mediaplayer causes ioConfigChanged() error on some devices when trying to start a soundplayer after the app resume.

I see that the inside of the close() method is empty. I added the necessary code lines inside this method.


arturoc:
thanks!

PR #4949, opened 2016-02-25
Bugfix correctly load non square pixels for AVF player. Closes #4937

### LINKED ISSUE #4937: AVFoundation player doesn't load HDV ( non-square pixels ) correctly.
https://forum.openframeworks.cc/t/issues-with-pixel-ratios-in-ofvideoplayer/22120/10

Test video is here:
https://www.dropbox.com/sh/t2rezaeeghyj581/AAAAkGZvI6uJmG6KYOKEFCNUa?dl=0

On initial research the way to get the true size of the frame is: 

    CVImageBufferRef imageBuffer = [videoPlayer getCurrentFrame];
      int realW =  CVPixelBufferGetWidth(imageBuffer);
      int realH = CVPixelBufferGetHeight(imageBuffer);

The problem is that this info comes in later ( once the first frame is delivered ). 
So for texture playback we need to change the texture coords based on the difference between the reported width ( 1920 ) and the pixel width ( 1440 ) 

For getting the pixels out we would need to figure out whether we should return the pixels as they are ( 1440 x 1080 ), or scale the pixels up which would add overhead. 

ping @i-n-g-o @openframeworks/macos

ofTheo:
Before the AVF player was loading a non square video as ( for example ) 1920x1080 but both the pixel data and the texture coming from AVFoundation were actually 1440x1080. 

This caused a crash when accessing pixels and resulted in the texture being drawn in a strange way. 
see : https://forum.openframeworks.cc/t/issues-with-pixel-ratios-in-ofvideoplayer/22120/10?u=theo

This PR loads the video at the size that represent the real pixel data and adds the ability to query what the intended stretched display size is meant to be. 

For now the new calls to getDisplayWidth() / getDisplayHeight are isolated to the ofAVFoundationPlayer class - if this is something we might want to support across platforms on linux and windows we could add it to the ofVideoPlayer and potentially even make the draw calls aware of the scaling needed. 

For now this seems like a conservative fix which leaves the display scaling up to the user, but provides them with the information they need to account for the non square pixels. 

An example of how you would do that is:

```
ofAVFoundationPlayer player;

void ofApp::setup(){
    player.load("HDV_test.mov");
    player.play();
    ofSetWindowPosition(0, 0);
}

void ofApp::update(){
    player.update();
}

void ofApp::draw(){
    if( player.isAnamorphic() ){
        player.draw(0, 0, player.getDisplayWidth(), player.getDisplayHeight());
    }else{
        player.draw(0, 0);
    }
}
```

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.h
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.h
@@ -52,10 +52,13 @@ public:
 	
     float getWidth() const;
     float getHeight() const;
-	
+    float getDisplayWidth() const;
+    float getDisplayHeight() const;
+
     bool isPaused() const;
     bool isLoaded() const;
     bool isPlaying() const;
+	bool isAnamorphic() const;
 	
     float getPosition() const;
     float getSpeed() const;
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -567,6 +567,24 @@ float ofAVFoundationPlayer::getHeight() const {
 }
 
 //--------------------------------------------------------------
+float ofAVFoundationPlayer::getDisplayWidth() const {
+    if(videoPlayer == nullptr) {
+        return 0;
+    }
+    
+    return [videoPlayer getDisplayWidth];
+}
+
+//--------------------------------------------------------------
+float ofAVFoundationPlayer::getDisplayHeight() const {
+    if(videoPlayer == nullptr) {
+        return 0;
+    }
+    
+    return [videoPlayer getDisplayHeight];
+}
+
+//--------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


ofTheo:
okay - finally not erring on travis ( must have tougher warning/error compile settings ). 
the travis errors I see now seem unrelated. 

should be good to go.

i-n-g-o:
hi theo.
missed that ping for a while. (changed my mail-filters, let's see if i get pings earlier now)

once you got a `CMFormatDescriptionRef`  already you could get the trackDimensionts like this:

```
CGSize trackDimensions = CMVideoFormatDescriptionGetPresentationDimensions(formatDescription, false, false);
				
videoWidth = trackDimensions.width;
videoHeight = trackDimensions.height;
```

with the test-movie this gives: 1440 x 1080 pixels.
maybe better to use the CMs description than calculating with pixel-aspect?

PR #4955, opened 2016-02-29
Doxygen sections ofRectangle & ofVec2f

HalfdanJ:
Changed sections to  `\section` syntax and internal functions to `\internal` syntax

@bakercp

PR #4956, opened 2016-02-29
Doxygen sections ofTexture & ofUtils

HalfdanJ:
just like #4955 changed sections syntax.

### DIFF
--- a/libs/openFrameworks/gl/ofTexture.h
+++ b/libs/openFrameworks/gl/ofTexture.h
@@ -21,9 +21,7 @@
 
 
 
-/// \name Global texture settings
-/// \{
-
+/// \section Global texture settings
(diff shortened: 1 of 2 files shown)


bakercp:
Cool.  The only thing I'd add is that I think we should consistently apply "Title Case" to the section names (e.g. "Size and Coordinates").  Currently it is not consistent.  I'll update the doxygen style guide.

HalfdanJ:
Good point. Like this?

bakercp:
@HalfdanJ looks perfect to me!

HalfdanJ:
Ok to merge?

bakercp:
Looks good to me.  @arturoc can we merge this?

arturoc:
let me check all of these against the glm branch before merging, don't want to have to merge them all at once later

bakercp:
+1

PR #4958, opened 2016-02-29
Add ofIsFloatEqual function.  Replaces #4888

bakercp:
Replaces #4888


bakercp:
@arturoc looks like this passed.

arturoc:
thanks!

Issue #3082, opened 2014-07-11
ofSetupOpenGL no longer takes a raw pointer to ofAppNoWindow 


bakercp:
`ofSetupOpenGL` now accepts classes that extends `ofAppBaseGLWindow`.  `ofAppNoWindow` only extends `ofAppBaseWindow`, leaving it unusable.  

Making it extend `ofAppBaseGLWindow` should work ... but leads to bad access exceptions on OSX.

Demo:

``` c++
#include "ofApp.h"
#include "ofAppNoWindow.h"


int main()
{
    ofAppNoWindow window;
    ofSetupOpenGL(&window, 0, 0, OF_WINDOW);
    ofRunApp(new ofApp());
}
```

PR #4960, opened 2016-03-01
noWindowExample. needs input, please review

wasawi:
maybe some more explanation about 
things that will not happen in headless mode?
keystrokes? events? sound? rendering?

### DIFF
--- a/examples/windowing/noWindowExample/README.md
+++ b/examples/windowing/noWindowExample/README.md
@@ -1,0 +1,39 @@
+#About noWindowExample
+
+
+###Learning Objectives
+
+This example demonstrates how to use openframeworks to run processes without a window. 
+
+In the code, pay attention to: 
+
+* Use of ````	ofAppNoWindow window```` in the main.cpp.
+* The fact that no draw commands will be displayed. See ````draw()````
+
+
+### Expected Behavior
+
+When launching this application you will see the following:
+
+* Openframeworks has no window.
+* CommandLine is running as expected.
+
+
+
+When the key ````r```` is pressed:
+
+* Attention keystrokes will not have effect!
+
+* the counter is NOT set to 0.
+See ````keyPressed()````
+
+
+Instructions for use:
+
+* Running the application will result in a commandline displaying a constantly increasing number.
+* Close the program by pressing ````ESC````
+
+
+###Other classes used in this file
+
+ofAppNoWindow.
--- a/examples/windowing/noWindowExample/src/main.cpp
+++ b/examples/windowing/noWindowExample/src/main.cpp
@@ -1,0 +1,14 @@
+#include "ofMain.h"
+#include "ofApp.h"
+#include "ofAppNoWindow.h"
+
+int main() {
+	// if you want to see the window	
+	// comment these two lines 
+	ofAppNoWindow window;
+	ofSetupOpenGL(&window, 1024, 768, OF_WINDOW);
+
+	// and uncomment this line
+	// ofSetupOpenGL(300, 300, OF_WINDOW);
+	ofRunApp(new ofApp());
+}
--- a/examples/windowing/noWindowExample/src/ofApp.cpp
+++ b/examples/windowing/noWindowExample/src/ofApp.cpp
@@ -1,0 +1,91 @@
+#include "ofApp.h"
+
+//--------------------------------------------------------------
+void ofApp::setup(){	 
+
+	number = 0;
+}
+
+//--------------------------------------------------------------
+void ofApp::update(){
+
+	// maybe some more explanation about 
+	// things that will not happen in headless mode?
+	// keystrokes? events? sound? rendering?
+	
+	// everything that does not run on screen runs as expected
+	number++;
+	ofLogNotice() << number;
+
+}
+
+//--------------------------------------------------------------
+void ofApp::draw(){
+	
+	// these circles will never be drawn
+	for (int i = 0; i < number; i++) {
+		ofDrawCircle(ofRandom(ofGetWidth()), ofRandom(ofGetHeight()), 15);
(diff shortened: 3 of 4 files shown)


brannondorsey:
Super rad to see this as an example. I think this will be very helpful to folks who are looking for this functionality but aren't familiar with `ofAppNoWindow()`. Is the code in `ofApp::update()` limited to the framerate set with `ofSetFramerate()`? Either way it might be helpful to mention the relationship to the framerate/loop when using `ofAppNoWindow()`. Also I would point out that even though things get kicked off with `ofSetupOpenGL()` there is no instance of OpenGL when using `ofAppNoWindow()` so any calls like `ofDraw*` will cause the app to crash. I ran into [this PR](https://github.com/openframeworks/openFrameworks/issues/3082) the other day which suggests an alternative method `ofSetWindow()` but I believe that method no longer works. @arturoc, can you confirm that `ofSetupOpenGL()` is in fact the correct method for launching an `ofAppNoWindow` instance?

brannondorsey:
Also, it might be helpful to include how to access/parse command line arguments w/ a headless instance of OF. Something like this: https://gist.github.com/rc1/5876289

wasawi:
Thank you very much @brannondorsey . 
Yes i think this example must explain how `ofAppNoWindow` works internally. As you say, would be good to explain if framerate can be detached from `ofSetFramerate()` and also explain why some calls like `ofDraw` could result in a crash (I didn't konw about it). Also I found it a bit interesting that `keyPressed()` didn't work but the `ESC` key to quit OF actually worked.
+1 for command line arguments! 
I will wait for more comments.

brannondorsey:
Glad it was a help. I've been using `ofAppNoWindow()` recently the first thing I did was search for an example like this, so I think this will be very helpful. Let me know if I can help in any way.

HalfdanJ:
Great with this example! I guess that many of the empty functions (for example mouseEnter), aren't doing anything in noWindow? I think the functions that don't add anything should be stripped so the example is as simple as possible.

examples/windowing/noWindowExample/README.md
```diff
@@ -0,0 +1,39 @@
+#About noWindowExample
+
+
+###Learning Objectives
+
+This example demonstrates how to use openframeworks to run processes without a window. 
+
+In the code, pay attention to: 
+
+* Use of ````	ofAppNoWindow window```` in the main.cpp.
```
bakercp: Just three backticks ````` are needed.

examples/windowing/noWindowExample/README.md
```diff
@@ -0,0 +1,39 @@
...
+### Expected Behavior
+
+When launching this application you will see the following:
+
+* Openframeworks has no window.
+* CommandLine is running as expected.
+
+
+
+When the key ````r```` is pressed:
```
bakercp: Instead of giving examples that won't produce a result, just that there is no (current) way to accept command line or mouse input with no-window mode.

examples/windowing/noWindowExample/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+* Attention keystrokes will not have effect!
+
+* the counter is NOT set to 0.
+See ````keyPressed()````
```
bakercp: Having ofLog output might be the best thing to watch for.

examples/windowing/noWindowExample/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+Instructions for use:
+
+* Running the application will result in a commandline displaying a constantly increasing number.
+* Close the program by pressing ````ESC````
```
bakercp: Does ESC work?  Or does it need to be ctrl+\ or ctrl+c etc?

examples/windowing/noWindowExample/src/main.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+int main() {
+	// if you want to see the window	
+	// comment these two lines 
+	ofAppNoWindow window;
+	ofSetupOpenGL(&window, 1024, 768, OF_WINDOW);
+
+	// and uncomment this line
+	// ofSetupOpenGL(300, 300, OF_WINDOW);
+	ofRunApp(new ofApp());
```
bakercp: This will be changing soon, but leave it this way for now :)

examples/windowing/noWindowExample/src/ofApp.cpp
```diff
@@ -0,0 +1,91 @@
...
+	// maybe some more explanation about 
+	// things that will not happen in headless mode?
+	// keystrokes? events? sound? rendering?
+	
+	// everything that does not run on screen runs as expected
+	number++;
+	ofLogNotice() << number;
+
+}
+
```
bakercp: Get rid of everything below here.

examples/windowing/noWindowExample/src/ofApp.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "ofMain.h"
+
+class ofApp : public ofBaseApp{
+	
+	public:
+		
+		void setup();
+		void update();
+		void draw();
```
bakercp: Get rid of everything after `void update()` since it won't have any effect.  Not having it there underlines that fact (and also reminds users that ofBaseApp doesn't require overrides :)

bakercp:
Nice example!  One addition that might be nice is command parsing in the main file.  Not sure if we want to include command line ... it's used in the projectGenerator

https://github.com/openframeworks/projectGenerator/blob/master/commandLine/src/optionparser.h

bakercp:
Might be nice to link that.

bakercp:
Hi @arturoc Anything holding this up?

PR #4975, opened 2016-03-07
add ofNode regression tests

tgfrerer:
+ these tests to help transition to glm by testing current ofNode
  behaviour against glm calculated ofNode behaviour

+ the tests currently focus on orbit, roateAround, and some
  basic parenting

+ these can in the future be extended and used to test ofNode more
  thoroughly.


arturoc:
thanks!!

Issue #4812, opened 2016-02-12
SSL-related Segmentation faults when using Poco SSL manager with custom settings on OSX, traced to . 


bakercp:
After tracking down the source of this issue https://github.com/bakercp/ofxSMTP/issues/12 and this issue https://forum.openframeworks.cc/t/ofxhttp-segfault/22391/7 (both of which use ofxSSLManager to require valid SSL certs via a ca cert bundle ... which the core currently does not do) I traced it to this commit https://github.com/openframeworks/openFrameworks/pull/4390 that recompiled freeimage on 10.8.  There appears to be some mysterious issue that is fixed for users by recompiling freeimage on the local system (e.g. `./apothecary update freeimage`).  I'm honestly not sure what the solution is because if we compile FreeImage on a 10.9+ machine, it seems to break 10.8 machines (https://github.com/openframeworks/openFrameworks/issues/4385).  My initial suspicion is that maybe the faulty 10.8 compiled freeimage is not compatible with Poco because it linked to a different non ABI compatible system lib or something.  Any thoughts are appreciated.

danoli3:
This is a strange one. We can assume majority developers are above board
with 10.9+ so we should compile with that for distribution. For those
under, they can just use Apothecary to recompile it.

Next major is 10.9+ anyway so this is really a legacy <0.9.2 issue.

bakercp:
I'm all for 10.9+.  Supporting 10.11, 10.10, 10.9 seems great.  I'm doing a little testing now with 10.8 to see if the recommendation to custom compile via apothecary for 10.8 actually solves the problem.  Funny thing is I just tried the stable branch on 10.8 and got the same segfaults which leads me to believe that it may be an incompatibility between our precompiled static libs (Poco / FreeImage), rather than a specific 10.8 issue.  I'm wondering if we make sure that _everything_ is compiled on 10.8 (and make sure we aren't linking into _any_ system libs ... which I believe we may accidentally be doing now in some cases ... and just getting lucky) -- perhaps it will just work.  Anyway, firing up the 10.8 virtual machine and will report back sometime soon.

danoli3:
Yeah awesome! Nice work!

bakercp:
[...]

PR #4977, opened 2016-03-07
0.9.3 osx freeimage fix

bakercp:
Fixes https://github.com/openframeworks/openFrameworks/issues/4812


arturoc:
is it ok to remove the undefs? i think they were there to solve some other problem but the FT headers are not exposed globally any more, only included in the cpp, so it might not be necessary any more

bakercp:
@danoli3? 

I just ran the apothecary scripts in the 0.9.3 branch.

danoli3:
@arturoc removing the undefs, I think it should be fine, I originally put them in to be sure it wasn't an issue in the old code, especially when they were in the headers, however it should be totally fine it is not exposed anymore. +1 merge

arturoc:
ok, great!

ofTheo:
I think this PR introduced a bug with OS X 32bit in the 0.9.3 release. 
Opening an issue for this but I was able to reproduce with the current download.

PR #4978, opened 2016-03-08
Async overloading

grayfuse:
In leiu of updates in 0.9 to allow for setting headers, I needed a method for doing async calls with an ofHttpRequest, instead of with just a url and optional name. This pull request adds a new method (getAsyncWithRequest) to ofURLFileLoader. 

A somewhat silly example below, but setting headers can be extremely useful for doing certain types of simple authentication. 

Happy to tweak naming scheme if needed, but thought this could be very useful in the core for others, and further expand options for teaching introductory topics around HTTP requests with the core library.

```
void ofApp::setup(){

    ofRegisterURLNotification(this);
    
    resID = loader.getAsync("http://localhost:1313", "config");
    
    ofHttpRequest req("http://localhost:1313/images/thumbs/01.jpg", "img");
    req.headers["Connection"] = "keep-alive";
    loader.getAsyncWithRequest(req);
 }

void ofApp::urlResponse(ofHttpResponse & response){
    if(response.status == 200 && response.request.name == "config"){
        cout << response.status << " " << endl;
        
    }else if(response.status == 200 && response.request.name == "img"){
        cout << response.status << " " << response.error << endl;
        networkImage.load(response.data);
    }
}
```


arturoc:
this is useful but we should find a different name, the idea behind get/getAsync is that internally it uses the GET method. until recently that was the only option but you can now use POST too so it would be good to not mix those.

the equivalent to the method you are adding is handleRequest. we could add a handleRequestAsync which would then be used internally by getAsync.

grayfuse:
Makes sense, renamed to keep it consistent with handleRequest.

```
void ofApp::setup(){

    ofRegisterURLNotification(this);
    
    resID = loader.getAsync("http://localhost:1313", "config");
    
    ofHttpRequest req("http://localhost:1313/images/thumbs/01.jpg", "img");
    req.headers["Connection"] = "keep-alive";
    loader.handleRequestAsync(req);
 }

void ofApp::urlResponse(ofHttpResponse & response){
    if(response.status == 200 && response.request.name == "config"){
        cout << response.status << " " << endl;
        
    }else if(response.status == 200 && response.request.name == "img"){
        cout << response.status << " " << response.error << endl;
        networkImage.load(response.data);
    }
}
```

Issue #4951, opened 2016-02-25
documentation fixes: ofFileUtils.h


bakercp:
Hey all, a few of these low level doc PRs got merged before I got to take a look at them -- for ofFileUtils.h a couple things -- 

To conform to doxygen style guide:
- Mixed use of \return \returns - both are valid, but style guide suggests \returns
- Need complete sentences with capitalization / punctuation, it's -> its where appropriate.
- 80 char hard breaks unless impossible.
- Need \brief line with a blank line after (for parsers -- \brief tag will be removed later) and order should 
be:

- brief
- blank
- description
- notes, warnings, throws, etc
- params
- returns

I'm happy to take a pass unless someone else wants to.

Thanks @danomatika for all of your work on these!

bakercp:
@danomatika Do you want to take this on?  Otherwise, I'm happy to do it.

danomatika:
What about \brief after \class? Whitespace there too?

bakercp:
Yes, whitespace there too.

danomatika:
So much for "minimal, 1 or 2 lines" :)

bakercp:
Yes :), with that said, after a discussion w/ @workergnome and @HalfdanJ earlier today, we're planning to get rid of the \class tag, but I'd say for now leave it in and we'll remove it all at once when we run through a conformance check (@HalfdanJ is researching the possibility of a documentation linter).  Fun times are ahead ...

danomatika:
I hope y'all aren't trying to outsmart doxygen too much ... sounds like heading down the road of "let's write out own documentation parser because the existing one isn't super pretty". :)

danomatika:
https://github.com/danomatika/openFrameworks/commit/64301c30072605d6fe5070fec5bb739fb61c7658

danomatika:
Closed with #4979

Issue #4931, opened 2016-02-22
ofFile api inconsistencies


danomatika:
Noticed a few things when documenting ofFileUtils.h:

ofFile(path, mode, binary) constructor opens w/ default binary=true while:
* ofFile::open() opens w/ binary=false
* ofFile::changeMode() uses binary=false
* ofDirectory::getFile() uses binary=false

ofFile::create(), how to set path before creating? Both the constructor and open() set the path but also try opening the file...

ofFile::setReadOnly() & ofDirectory::setReadOnly() should be setReadable() to match setWriteable(). I feel this naming also conflicts with the ReadOnly access mode.

arturoc:
oh, i thought i had changed everything to binary=true. that should be the default, it doesn't mean binary file really but only that it'll do no break lines transformations.

and yes it would be useful to have a version of create that accepts a path. you can do it with:

```
ofFile("file.txt", ofFile::Reference).create();
```

or any combination of open/create using Reference as open mode. or even:

ofFile("file.txt", ofFile::WriteOnly);

 but it would be easier to have a create with a parameter.

Also setReadOnly right now is not setReadable, it just removes the write permissions from the file, which is really strange cause you can just do setWritable(false) so yes having setReadable instead would be much better

do you want to send a PR?

arturoc:
we might even have an static create that accepts a parameter and returns an ofFile

danomatika:
Sure. Chris and I thought the binary flag might have been an oversight and we hope nobody has been *relying* on that so far.

arturoc:
it's really not so important, it only affects files that have windows breaklines in unix and unix breaklines in windows so it's hard to have any problem with it. the main use is to set it to true when you want to open a binary file so it doesn't change anything.

danomatika:
Right. So the doxy lines could basically be something like:

    opens as a text file by default: split individual lines & strip endline characters
    \param binary set to false when working with binary files so data will not be interpreted as lines of text

arturoc:
yep

bakercp:
+1 to this effort to clarify this api :)

danomatika:
Closed via #4979

PR #4979, opened 2016-03-08
ofFileUtils: low_level_docs & api updates

danomatika:
PR for fixes related to #4951 & #4931

I would have done this sooner but we decided to head to the mountains last weekend. I think everyone should :)


arturoc:
can you deprecate setReadOnly with a message asking to use setWritable instead of completely remove it?

also if you can change the test that is currently failing to use the new method instead?

danomatika:
sure thing

danomatika:
Updated.

arturoc:
thanks!

PR #4988, opened 2016-03-13
json serializer:

arturoc:
- removes ofBaseSerializer and instead uses a function based api to (de)serialize
```cpp
ofSerialize(Serializer &, ofAbstractParameter&)
ofDeserialize(const Serializer &, ofAbstrictParameter&)
```

which allows to implement serializers for external classes without modifying them

- implements the (de)serializing functions for ofJson

- changes ofXml and ofxXmlSettings to the new API

- changes ofxGui to the new api

- moves ofJson to it's own header and adds ofLoadJson/ofSaveJson to make it easier to load/save json from the filesystem


bakercp:
This looks like a stereo in the right direction. Currently there is no exception handling around parse errors or file not found exceptions that could be happen with ofJson/ofFile.  This should be documented with a \throws doxygen tag.  In general, it would be great to get this documented before merging so we can get in the habit.  

Looks cool!

bakercp:
Also, how does this "allows to implement serializers for external classes without modifying them"? Send like it all still must inherit abstract parameter?

PR #4992, opened 2016-03-14
Std regex

edap:
as continuation of this issue https://github.com/openframeworks/openFrameworks/pull/4422
(it was easier to open a new pr with the code needed for the new example, than continuing with the old one)


arturoc:
great, thanks!

PR #4998, opened 2016-03-16
[proposal] make ofxOsc strong

2bbb:
for example, we wrote like this:

```cpp
    ofxOscReceiver receiver;
    void setup() { receiver.setup(9005); }
```

and we do in terminal.app (or similar)

```
$ nc -4u localhost 9005
> a
```

maybe your app got crush.
so, this process is too extreme.
but, recently, some app got bad packet and crushed 10 machines at same time. (maybe bad packet is broadcasted)
i think this is very critical.

and, more critical point is "we cannot catch this exception from the outside now."

now, this PR is first aid.
(but, i hope to be accepted this PR as soon as possible.)

i think good solution is:
doing propagate exception with `std::promise`, `std::future`, or `std::async`.
or, simply, replace ofLog (is implemented now) with send error message with `ofThreadChannel` when catch exception. and notify error.
or, is this OK?

i don't know policy of exception handling in oF.
and i want to discuss for bright future we can use osc in peace.

best regards.

### DIFF
--- a/addons/ofxOsc/src/ofxOscReceiver.cpp
+++ b/addons/ofxOsc/src/ofxOscReceiver.cpp
@@ -59,7 +59,7 @@ ofxOscReceiver & ofxOscReceiver::operator=(const ofxOscReceiver & mom){
 	return *this;
 }
 
-void ofxOscReceiver::setup( int listen_port )
+bool ofxOscReceiver::setup( int listen_port )
 {
     if( osc::UdpSocket::GetUdpBufferSize() == 0 ){
     	osc::UdpSocket::SetUdpBufferSize(65535);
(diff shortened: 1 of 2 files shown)


arturoc:
i want to look into adding more expresive error handling, we don't throw exceptions but something that returns an error object with more information about the error would be really useful. 

by now, can you just return a boolean from setup to indicate that it was correctly setup or not

2bbb:
@arturoc 
thanks for comment.

i have 1 idea as following:

```cpp
void setup() {
    receiver.setup(8888);
    ofAddListener(receiver.gotOscError, this, &ofApp::oscError);
}

void oscError(std::string &what) {
    ofLogWarning() << what;
}
```

if we got error then this example print like:

> [warning] element size must be multiple of four

and this is already implement in [this repo (is first aid for worrier)](https://github.com/2bbb/ofxInsensitiveOsc).

maybe this implementation can make error handling more expressive.

"return a boolean from setup" is mean like above commit?
(this commit doesn't include imeplements of notify error message. if it seems to ok, i can merge soon.)

arturoc:
Thanks!

i don't think having to register an event to receive the possible errors. as i said i want to look into a generic solution for returning errors conditions so let's leave it as it is by now and we can com back to it once we figure it out

2bbb:
:+1:

valillon:
It seems related to the issue #5514. It would be nice to have some feedback here : )
Another recent handle exception in ofxOsc here #5467.

Issue #4993, opened 2016-03-14
Freeimage missing symbols for 32bit for 0.9.3 on OS X 


ofTheo:
From the forum: https://forum.openframeworks.cc/t/of-0-9-3-not-building-32-bit-on-osx/22777

Doing a lipo -i it shows a FAT lib. 
But it looks like its missing 32bit for TIFF/PNG etc 
I am able to reproduce with the current 0.9.3 download building an example with 32bit target. 

ping @bakercp 

```
Ld bin/videoPlayerExampleDebug.app/Contents/MacOS/videoPlayerExampleDebug normal i386
    cd /Users/theo/Downloads/of_v0.9.3_osx_release/examples/video/videoPlayerExample
    export MACOSX_DEPLOYMENT_TARGET=10.7
[...]

PR #5001, opened 2016-03-17
Fixes for 32bit OS X Freeimage. 

ofTheo:
Fixes the 0.9.3 bug detailed here: #4993


arturoc:
hey better to send this to this repo https://github.com/openframeworks/apothecary  which is already building the library for all platforms. that way we also know if this would introduce any errors in other platforms.

also 0.9.4 will end up in master once we merge the whole branch anyway

PR #5011, opened 2016-03-22
Add SO_REUSEADDR option to ofxTCPManager and ofxTCPServer.

rafirstfruit:
When you don't close your socket cleanly (e.g. your program crashes), it can linger on your OS for quite some time before you're allowed to reuse the same port.

In this state, calls to `bind` will return `EADDRINUSE` error.

Most software deals with this by setting the `SO_REUSEADDR` socket option before calling `bind`. However, this functionality is missing from ofxTCPManager.

It looks like it confuses some people too:
https://forum.openframeworks.cc/t/tcp-server-socket-not-closing-already-in-use/22793

I'm setting it to `false` by default, to maintain old behaviour, but openFrameworks should seriously consider enabling it by default (that is, if this patch is accepted, of course).

NOTE: not tested on Windows, but should work there AFAIK.

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPManager.cpp
+++ b/addons/ofxNetwork/src/ofxTCPManager.cpp
@@ -144,7 +144,7 @@ bool ofxTCPManager::Listen(int iMaxConnections)
 	return ret;
 }
 
-bool ofxTCPManager::Bind(unsigned short usPort, bool bReuse = false)
+bool ofxTCPManager::Bind(unsigned short usPort, bool bReuse)
 {
 	struct sockaddr_in local;
 	memset(&local, 0, sizeof(sockaddr_in));
--- a/addons/ofxNetwork/src/ofxTCPManager.cpp
+++ b/addons/ofxNetwork/src/ofxTCPManager.cpp
@@ -144,7 +144,7 @@ bool ofxTCPManager::Listen(int iMaxConnections)
 	return ret;
 }
 
-bool ofxTCPManager::Bind(unsigned short usPort)
+bool ofxTCPManager::Bind(unsigned short usPort, bool bReuse = false)
 {
 	struct sockaddr_in local;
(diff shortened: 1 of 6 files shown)


arturoc:
having this functionality would be great but we should come up with a different api. boolean parameters for functions like this are really confusing:

```
setup(ip, true, false)
```

is really hard to understand what's going on. even the parameter we have right now for blocking should go away in my opinion.

i would suggest to deprecate the current calls, add one with only the address that blocks by default and then some settings classes that allow to setup the sockets more explicitly, something like:

```
ofxTCPSettings settings;
settings.address = "localhost";
settings.blocking = false;
settings.reuseAddress = true;
...
``

and something similar for ofxUDPSettings. that would allow to add even more options that can be useful to have. do you want to modify this PR to add this settings classes?

**Comment by rafirstfruit** [comment]
Yes, I might try it, but given the patch will be more complex it'll take me some time.

One question: how do you envision those structs be used?
```
ofxTCPSettings settings;
settings.blocking = false;
...
TCPServer.setup(settings); // right?
settings.blocking = true; // should this work?
```

**Comment by arturoc** [comment]
```
TCPServer.setup(settings); // right?
```

yeah that's it

```
settings.blocking = true; // should this work?
```

No the struct is just data that it's passed to the object if you change it afterwards it'll have no effect. That also makes it clearer than having a setBlocking (or any other setting)  method that sometimes has to be called before calling setup or it won't work.

rafirstfruit:
I've created a separate PR #5013.

I feel it's a separate, larger issue. After we're done we can return to `reuse` issue. Sorry if I'm breaking your workflow.

rafirstfruit:
Now that #5013 is merged, here's another stab at address reuse issue.

ofTheo:
This looks good to me. It basically mirrors all the functionality of the UDP Manager class regarding reuse.

PR #5013, opened 2016-03-23
Add ofxTCPSettings and ofxUDPSettings structs to ofxNetwork.

rafirstfruit:
Dealing with transition to ofx[TCP/UDP]Settings as proposed by @arturoc at #5011.

I suspect this would require a lot of back and forth, so am using a separate branch/PR.

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPClient.cpp
+++ b/addons/ofxNetwork/src/ofxTCPClient.cpp
@@ -31,13 +31,9 @@ void ofxTCPClient::setVerbose(bool _verbose){
 //--------------------------
 bool ofxTCPClient::setup(string ip, int _port, bool blocking){
 
-	ofxTCPSettings settings;
+	ofxTCPSettings settings(ip, _port);
 
-	settings.address = ip;
-	settings.port = _port;
 	settings.blocking = blocking;
-
-	setMessageDelimiter(settings.messageDelimiter);
 
 	return setup(settings);
 }
@@ -55,6 +51,8 @@ bool ofxTCPClient::setup(const ofxTCPSettings & settings){
 	}
 
 	TCPClient.SetNonBlocking(!settings.blocking);
+
+	setMessageDelimiter(settings.messageDelimiter);
 
 	port		= settings.port;
 	ipAddr		= settings.address;
--- a/addons/ofxNetwork/src/ofxTCPClient.cpp
+++ b/addons/ofxNetwork/src/ofxTCPClient.cpp
@@ -31,19 +31,33 @@ void ofxTCPClient::setVerbose(bool _verbose){
 //--------------------------
 bool ofxTCPClient::setup(string ip, int _port, bool blocking){
 
+	ofxTCPSettings settings;
+
+	settings.address = ip;
+	settings.port = _port;
+	settings.blocking = blocking;
+
+	setMessageDelimiter(settings.messageDelimiter);
+
+	return setup(settings);
+}
+
+//--------------------------
+bool ofxTCPClient::setup(const ofxTCPSettings & settings){
+
 	if( !TCPClient.Create() ){
 		ofLogError("ofxTCPClient") << "setup(): couldn't create client";
 		return false;
-	}else if( !TCPClient.Connect((char *)ip.c_str(), _port) ){
-		ofLogError("ofxTCPClient") << "setup(): couldn't connect to " << ip << " " << _port;
+	}else if( !TCPClient.Connect((char *)settings.address.c_str(), settings.port) ){
+		ofLogError("ofxTCPClient") << "setup(): couldn't connect to " << settings.address << " " << settings.port;
 		TCPClient.Close(); //we free the connection
 		return false;
 	}
 
-	TCPClient.SetNonBlocking(!blocking);
+	TCPClient.SetNonBlocking(!settings.blocking);
 
-	port		= _port;
-	ipAddr		= ip;
+	port		= settings.port;
+	ipAddr		= settings.address;
 	connected	= true;
 	return true;
 }
--- a/addons/ofxNetwork/src/ofxTCPClient.h
+++ b/addons/ofxNetwork/src/ofxTCPClient.h
@@ -2,6 +2,7 @@
 
 #include "ofConstants.h"
 #include "ofxTCPManager.h"
+#include "ofxTCPSettings.h"
 #include "ofFileUtils.h"
 #include "ofTypes.h"
 
@@ -22,6 +23,7 @@ class ofxTCPClient{
 
 		void setVerbose(bool _verbose);
 		bool setup(string ip, int _port, bool blocking = false);
+		bool setup(const ofxTCPSettings & settings);
 		void setMessageDelimiter(string delim);
 		bool close();
 
--- a/addons/ofxNetwork/src/ofxTCPServer.cpp
+++ b/addons/ofxNetwork/src/ofxTCPServer.cpp
@@ -24,9 +24,8 @@ void ofxTCPServer::setVerbose(bool _verbose){
 
 //--------------------------
 bool ofxTCPServer::setup(int _port, bool blocking){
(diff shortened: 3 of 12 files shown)


rafirstfruit:
Could you please give me a quick code review, before I delve further?

In particular, 
- do you want to have the structs as members of ofxTCPClient / ofxTCPServer classes?
- should they go even deeper into ofxTCPManager / ofxUDPManager?
- should I mark setBlocking, etc with OF_DEPRECATED?

arturoc:
> do you want to have the structs as members of ofxTCPClient / ofxTCPServer classes?

not sure, that depends on how many settings we have perhaps leave it as it is by now if it makes it easier and we can change it later if we add more parameter to the settings object that need to be persisted during the live of the socket rather than only used on setup

> should they go even deeper into ofxTCPManager / ofxUDPManager?

i guess that's where the settings for blocking, reuse... are really set so yes it makes sense. no need to have 2 different structs, one for ofxTCPClient and another for ofxTCPManager. Not sure if we need different settings objects for client/server though.

> should I mark setBlocking, etc with OF_DEPRECATED?

let's leave them by now and we'll see later, my feeling is that most people will just setup everything at the beginning and that changing the state while the socket is running is more complex than doing it only in setup because this classes might use threads internally... so deprecating them would be a good idea but it might be that it's a feature that people use in some cases

in general it looks good to me.

rafirstfruit:
Thank you, that makes perfect sense. I'll go on then.

(In my experience, people change socket options very very rarely, most protocols/implementations require some very specific options to be set. I *could* come up with some contrived example of, say, blocking/non-blocking change during some handshake or something, but I really don't think that's how people use sockets, in general. However, I welcome other opinions regarding this).

ofTheo:
@arturoc @rafirstfruit - I was tempted to suggest making the Settings part of the class.  As in https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofFbo.h#L135 

The advantage would be that we would have only settings that relate to the server and the settings that only relate to the client in each classes Settings struct.  

Having a common Settings class means that settings for both server and client classes have to be in there.  Just a small downside to consider. :)

rafirstfruit:
@ofTheo I like this. But please do consider that 99% of the settings would be the same. Even the "address" setting, which is currently unused in ofxTCPServer is _potentially_ useful for binding to specific network interfaces (i.e. "127.0.0.1" to make it invisible to outside world or vice-versa or a variation of this).

I humbly request you guys to reach some kind of consensus before I go on.

rafirstfruit:
Oh, and one more approach to consider:
```
struct Settings {
   int port;
   ...
   union {
       struct Server {
           int maxConnections;
       };
       struct Client {
           int reconnectTimeout; // or whatever
           ...
       };
   };
}
```

Don't know if unions are frowned upon in C++/oF style guide.

arturoc:
@ofTheo yes i thought about that but as @rafirstfruit says i think this settings would be the same for all TCP or all UDP no matter if client or server.

perhaps the best is to try with one settings class only and then if we find out that some settings are not compatible we can separate them in to

@rafirstfruit unions make it really complicated let's keep it as simple as possible by now and then we can see if we need to separate it in more than one class in which case it won't matter much if it's slightly redundant. better that than having a complicated syntax

Also, i kind of like the inner classes for the settings structs but in some cases it's not possible because of different reasons so i'm thinking lately that it's better to keep them as separated classes for uniformity. perhaps whenever/if we start using namespaces this will have a better solution but right now inner classes is problematic in some cases

rafirstfruit:
Whelp, ofxUDPManager is rather different beast, but here we go.

Things I'm unsure about:
- `send`/`receive` bools, ability to have both or neither
- potentially requires 2 sets of ports and hosts (for listening and for sending)
- people might forget to set port or address, unlike the method call

Please advice further :)

arturoc:
one thing we can do for mandatory arguments is having a constructor with those so you can only create the settings object like:

```
ofxTCPSettings settings("127.0.0.1", 8080);
```

or have the setup method accept the address, port and settings object like:

```
client.setup("127.0.0.1", 8080, settings);
```

i like the first better

arturoc:
For udp i guess something like:

```
struct ofxUDPSettings{
    void sendTo(string address, int port);
    void receiveOn(int port);
    void receiveOn(string address, int port);
    int bindPort;
    bool blocking = true;
    bool reuse = true;
    bool broadcast = true;
    string sendAddress;
    int sendPort;
    string bindAddress = INADDR_ANY;
};
```

so you can do:

```
ofxUDPSettings settings;
settings.sendTo(address, port);
settings.receiveOn(port);
upd.setup(settings);
```

which in turn will call the corresponding Connect and/or Bind

rafirstfruit:
>client.setup("127.0.0.1", 8080, settings);

I like this one a little bit more, but I don't have any compelling argument for it, just preference. So I went with your version.

One thing that still bothers me is
```
ofxUDPSettings settings;
//no calls to receiveOn or sendTo
ofxUDPManager.setup(settings);
```

do you think `setup` should return an error in such case?

rafirstfruit:
Hmm, and after giving it more though, I think there are potential issues in several cases:

Let's say we decide to refactor `ofxTCPManager` and use `ofxTCPSettings` internally to keep all our values:
```
class ofxTCPManager {
    ofxTCPSettins settings; // all our settings
    ...
}
```
-- we can't!

Similar problems might occur in user-code.

What do you think?

Oh, and on a related note, are `ofxTCPSettings& ofxTCPClient::getSettings()` kinds of methods desired?

arturoc:
> One thing that still bothers me is
>```
ofxUDPSettings settings;
//no calls to receiveOn or sendTo
ofxUDPManager.setup(settings);
```
>do you think setup should return an error in such case?

yes, totally, another posibility so it's impossible to create a wrong settings object would be to have different settings classes like:

```
ofxUDPSendSettings(ip,port)
ofxUDPReceiveSettings(ip,port)
ofxUDPSendReceiveSettings(ipS,portS, ipR,portR)
```

or perhaps skip the last one and have:

```
udp.setup(ofxUDPSendSettings);
udp.setup(ofxUDPReceiveSettings);
udp.setup(ofxUDPSendSettings, ofxUDPReceiveSettings);
```

**Comment by arturoc** [comment]
about:

>Hmm, and after giving it more though, I think there are potential issues in several cases:
>
>Let's say we decide to refactor ofxTCPManager and use ofxTCPSettings internally to keep all our values:
>```
class ofxTCPManager {
    ofxTCPSettings settings; // all our settings
    ...
}
```
>
>-- we can't!

i think it's ok if people can't store this settings easily but you atually can store them anyway:

```
class ofApp{
   ofxTCPSettings settings{"127.0.0.1", 8080};
}
``

or

```
class ofApp{
   ofApp()
   :settings("127.0.0.1", 8080){}
}
```

the first is quite nice cause it allows you to check all your basic parameters from a quick look in the header

rafirstfruit:
Hmm, interesting. I think `ofxUDPSendReceiveSettings` is an overkill, but your shorter version is rather nice. Good point on reading the `ofApp.h` file, that indeed looks like a feature rather than a problem.

I'll tinker with it some more.

What about
> Oh, and on a related note, are ofxTCPSettings& ofxTCPClient::getSettings() kinds of methods desired?

arturoc:
it's probably better to leave it in simple getters for the individual properties instead of the full settings object. so instead of getSettings(), getAddress, getPort...

rafirstfruit:
I see. Simple getters already exist, so nothing to add here.

OK, I think that's it. If you have any further comments/criticisms/desires/etc, feel free to add them.

arturoc:
this looks really good, one thing i'm not 100% convinced is if leaving the things that can be setup through methods as public variables but i'm going to merge it as it is and will take a look at some point.

thank!

rafirstfruit:
Oh, good point. Could be useful for << cout'ing, though.

Thanks for you guidance and merging!

Issue #5019, opened 2016-03-28
ofEasyCam Release Inertia/Velocity


sieren:
ofEasyCam on iOS reacts very sensitive when you release the finger (aka touch-up), essentially spinning the camera wildly around at times.

The easiest way to handle this would be to add extra functionality to ofEasyCam along the lines of `disableReleaseInertia() / enableReleaseInertia()` with the default behaviour to what it is now.

I'd file a PR for this, but wanted to open this up for discussion first.

bakercp:
@roymacdonald do you have any thoughts on this as you've been working with ofEasyCam lately ...

roymacdonald:
@bakercp the enable/disable Inertia methods are already implemented. Inertia is always added on mouse/touch release. Also there is `setDrag(float)` which lets you set how much drag or friction the inertia has. 1 = never stop; 0= no inertia. by default it is 0.9
Maybe renaming this function so it reflects that it is related to inertia.

roymacdonald:
BTW, I think it is safe to close this issue

bakercp:
Thanks!

dimitre:
It seems dragging only a few pixels can set the camera spinning "Salyut 7 style". 
Maybe it considers absolute pixels and retina screens have more pixel density.
It would be great to be able to configure camera initial velocity per pixels dragged.

roymacdonald:
Hi @dimitre what happens if you lower the drag, using` setDrag(...)`? also, you can lower the rotation sensitivity using `setRotationSensitivity(...)`.
can you paste a video about this behavior?
Are you seeing this behavior on iOS or on a computer?

dimitre:
Yes, video is here, I'm seeing this on a computer
http://dmtr.org/easycam/
setDrag seems to have no difference, even setDrag(0)
things are better with setRotationSensitivity, thanks.

roymacdonald:
thanks @dimitre 
that's really odd. I have not seen that behavior. The fast rotation starts when you release the mouse button or when you stop dragging the mouse?

roymacdonald:
@dimitre I double checked and now I see it. Inertia is disabled by default and I never activate it.
I'll see whats going on.
thanks for pointing this out

dimitre:
mouse release. it seems to have a 10x impulse no matter which setRotationSensitivity is set.
not such a problem if I lower it but feels unnatural.
thanks @roymacdonald

roymacdonald:
[...]

PR #5023, opened 2016-03-29
ofEasyCam: Add setters for mouse release inertia

sieren:
Add setters to disable/enable mouse release inertia.
This can avoid unintended camera spinning on touch sensitive
input devices like iOS where the delta between different touch
points will cause the camera to spin rapidly.

See https://github.com/openframeworks/openFrameworks/issues/5019


roymacdonald:
I think that this approach is better
https://github.com/openframeworks/openFrameworks/pull/5142
Simply ditching inertia.
best!

arturoc:
going to close this since it was already solved with another PR

PR #5024, opened 2016-03-30
add keyboard hot plug support

DeqingSun:
I used Arduino as Keyboard on RPi with magsafe connection.
So I added multiple keyboard support and hotplug support. Based on V0.9.2

Basically I changed keyboard_fd to an array. And another array keeps path of each device.
setupNativeKeyboard() will first check if any device disappears, and then add new device.
readNativeUDevEvents() will call setupNativeKeyboard to config new keyboard settings.

### DIFF
--- a/libs/openFrameworks/app/ofAppEGLWindow.cpp
+++ b/libs/openFrameworks/app/ofAppEGLWindow.cpp
@@ -14,7 +14,10 @@ struct udev_device* dev;
 struct udev_monitor* mon;
 static int udev_fd = -1;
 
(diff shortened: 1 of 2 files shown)


bakercp:
Cool.  I've been meaning to do this for a while.  Would also be nice to do w/ Mice as well.  I'll give this a test in the next few days.

DeqingSun:
I'm not expert in linux and I don't think I'm doing it in the best way.
I don't know how to use "dev" in readNativeUDevEvents(). If "dev" can indicates which keyboard is changed, the code can be more efficient.

PR #5025, opened 2016-03-31
Feature uniformblock

jacres:
Uniform buffer object support added to ofShader

### DIFF
--- a/libs/openFrameworks/gl/ofGLUtils.cpp
+++ b/libs/openFrameworks/gl/ofGLUtils.cpp
@@ -712,7 +712,7 @@ bool ofGLCheckExtension(string searchName){
 	extensionsSet.insert(extensionsList.begin(),extensionsList.end());
 	return extensionsSet.find(searchName)!=extensionsSet.end();
 #else
-	return glewGetExtension(searchName.c_str());
+	return glewIsSupported(searchName.c_str());
 #endif
 }
 
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -100,7 +100,8 @@ program(mom.program),
 bLoaded(mom.bLoaded),
 shaders(mom.shaders),
 uniformsCache(mom.uniformsCache),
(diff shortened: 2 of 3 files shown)


arturoc:
looks great, i'll let the tests pass before merging. one thing. when we implemented ofBufferObject we found out that just checking the ifdef is not enough to see if an extension is present. there's a function to check if an extension is present. i guess checking it everytime might be slow so we check on the first buffer creation and cache that value:

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofBufferObject.cpp#L21-L30

prisonerjohn:
Added those extra extension checks and merged latest master.

jacres:
Thanks Elie!

@arturoc one thing I noticed with ofBufferObject is that the checking for the define for DSA support succeeds, but the glew call to check for the extension fails on my computer, so it falls back to non DSA/named buffer calls. This is on Windows 10, maybe a few week old nvidia drivers and a 4.5 context.
I'll try updating drivers to see if this fixes it.

I think glewGetExtension() is throwing GL errors with each call to ofGLCheckExtension(). This is because glGetString(GL_EXTENSIONS) is deprecated on core GL contexts. I believe it's fixed in newer glew versions, but not sure if there would be any breaking changes. It would be nice to fix to make it easier to debug GL errors in an app. What do you think?

prisonerjohn:
Ah @jacres you're right, the check is also failing for `GL_ARB_uniform_buffer_object`

prisonerjohn:
Looks like just switching the following in [ofGLCheckExtension()](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofGLUtils.cpp#L715) fixes the check.
```cpp
	return glewGetExtension(searchName.c_str());
```
to
```cpp
	return glewIsSupported(searchName.c_str());
```

arturoc:
oh, that's great to know, there's always been some problems with that functions, can you add that change in this PR?

prisonerjohn:
Done!

arturoc:
thanks!

PR #5026, opened 2016-03-31
fixes for ofxOsc

tpltnt:
Hi there,

i took a look at ofxOsc and fixed a few issues.
Cheers,
tpltnt


underdoeg:
What does this fix? It only failes because of the casting on 64bit systems 
I think...

tpltnt:
@underdoeg this PR was supposed to fix a few (implicit) problems like missing initializations etc. Yes, it is currently failing since I did not had the time to continue working on it. It was working on my machine at some point, but was (rightfully) stopped by the integration tests.

addons/ofxOsc/src/ofxOscMessage.cpp
```diff
@@ -93,7 +93,7 @@ std::int32_t ofxOscMessage::getArgAsInt32( int index ) const
 	    if ( getArgType( index ) == OFXOSC_TYPE_FLOAT )
         {
 	    	ofLogWarning("ofxOscMessage") << "getArgAsInt32(): converting float to int32 for argument " << index;
-            return (std::int32_t)((ofxOscArgFloat*)args[index])->get();
+            return reinterpret_cast<std::int32_t> (dynamic_cast<ofxOscArgFloat*> args[index])->get();
```
2bbb: maybe:
```
return reinterpret_cast<std::int32_t> ((dynamic_cast<ofxOscArgFloat*> args[index])->get());
```

tpltnt:
thanks for pointing it out @2bbb, but it doesn't seem to be the only issue.

Issue #5009, opened 2016-03-21
Fullscreen framerate drop on linux 64


frauzufall:
Hey,

the framerate of any program drops down to 4 fps when going into fullscreen mode on linux 64bit. I tried the graphics/colorExample and an empty application with nothing but a background drawing. 

I tried this `main.cpp`
```
int main(){

	ofGLFWWindowSettings settings;
	settings.multiMonitorFullScreen = true;
	settings.windowMode = OF_FULLSCREEN;
	auto window = ofCreateWindow(settings);

	ofRunApp(window, std::make_shared<ofApp>());

	ofRunMainLoop();

}
```
as well as the following with calling `ofToggleFullscreen()` on key press.
```
int main(){

	ofSetupOpenGL(1024, 768, OF_WINDOW); 

	ofRunApp(new ofApp());

}
```
I saw different issues related to fullscreen that seemed to be fixed but I found nothing about framerates. Any ideas?

arturoc:
tis that with vertical sync enabled or without? if it's without and depending on the hardware even with it enabled it's perfectly normal, it's just the fill rate limit of the hardware.

try using OF_GAME_MODE instead of OF_FULLSCREEN which might be faster in certain cases.

there's also some settings in compiz (if you are on ubuntu) that might help. on the compiz config settings manager (doesn't come installed by default) in the composite section. disable unredirect fullscreen windows

you can also try if glxgears has the same effect, run glxgears to get the fps in a window and then glxgears -fullscreen

frauzufall:
Thank you for your reply! The differences appear way too high to me.
```
OF_FULLSCREEN: ~3 FPS
OF_WINDOW: ~2100 FPS
OF_GAME_MODE: ~2100 FPS
```
(vertical sync on/off changes are minimal)
```
glxgears: ~12000 FPS
glxgears -fullscreen: ~3900 FPS
```
I used fullscreen before a while ago without noticeable problems. I run Xubuntu 15.10, no Compiz. Possibly still system related if I'm the only one with that problem, just wanted to ask. It bugs me because I need multiple monitor fullscreen and OF_GAME_MODE is not doing that. (Btw. there are also only minimal changes in FPS when unplugging the second screen) We can close this if no one else is affected.

PetrosKataras:
[...]

PR #5028, opened 2016-04-04
several fixes for glfw and main loop

arturoc:
Trying to fix GLFW slowness on some linux: #5009
Fixes window size values in some cases
Fixes main loop not finishing last window which in turned caused
game mode to let the screen in a wrong resolution when the app finished

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -48,6 +48,8 @@ ofAppGLFWWindow::ofAppGLFWWindow(){
 	windowP = nullptr;
 	windowW = 0;
 	windowH = 0;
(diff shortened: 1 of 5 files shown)


frauzufall:
Hi @arturoc, since this PR I get segmentation faults when exiting an application that uses events like `examples/events/simpleEventsExample`. In order to fix that, I have to revert `currentWindow` in `ofMainLoop` back from a weak pointer to a shared pointer. Then it works. Should I create a PR?

arturoc:
the problem with that is that then the window wasn't being destroyed anymore at the end, i found out because when using game mode the screen stayed in the resolution you set instead of going back to the original. anyway, it should stay as a weak pointer but we should avoid generating more events after the windows have been destroyed. i'll take a look

PR #5030, opened 2016-04-05
ofxCvImage pts array array delete fix

danomatika:
Fix for the following clang warning:

~~~
openFrameworks/addons/ofxOpenCv/src/ofxCvImage.cpp:394:5:
'delete' applied to a pointer that was allocated with 'new[]'; did you mean 'delete[]'?
~~~


arturoc:
i think you changed it to: `delete ptr[]` but it should be `delete[] ptr` most tests are failing because of that

danomatika:
Whoops. Fixed.

arturoc:
thanks!

PR #5036, opened 2016-04-08
ofxGui: fix possible crash during construction

### LINKED ISSUE #4769: ofxGui crashes on construction (if the ofMainLoop is not setup)
The constructor of any ofxBaseGui crashes if called before the ofMainLoop is up and running.

For example if I define an ofxPanel globally (in ofApp.cpp) the app will crash because of the line 69 in the ofxBaseGui constructor:

`currentFrame = ofGetFrameNum();`

Of course this can be easily avoided and is bad practice anyway to define it globally but also an constructor shouldn't call any oF internals right? That should happen in some setup function? Or is it often the case?

Thanks,
Felix

fx-lange:
fix for #4769 

Calling ofGetFrameNum() in the constructor can cause the application to crash if ofMainLoop is not up and running. 

Setting the current frame to zero should actually be sufficient. isGuiDrawing() will still return false until you are drawing it the first time and after that the value set during construction doesn't matter anymore.

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -66,7 +66,7 @@ ofBitmapFont ofxBaseGui::bitmapFont;
 
 ofxBaseGui::ofxBaseGui(){
 	parent = nullptr;
-	currentFrame = ofGetFrameNum();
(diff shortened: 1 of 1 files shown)


fx-lange:
already fixed via #5042

PR #5037, opened 2016-04-08
ofxNetwork - settings "cache" is broken

rafirstfruit:
`ofxTCPManager` stores various socket options to avoid doing extra work and to return them without querying the socket interface.

For example,

```
//----------------------------------------------------------------------------$
bool ofxTCPManager::SetNonBlocking(bool useNonBlocking)
{
        if(useNonBlocking==nonBlocking){
                return true;
        }
        ....
```

This avoids calling all the setsockopt/ioctl code.

However, the socket underneath might be destroyed and re-created via, for example, `Connect` and `Close`. After that, those "cached" values do not represent the actual state and are actually in the way.

Consider:

```
ofxTCPManager tcp;
tcp.Create();
tcp.SetNonBlocking(true);
tcp.Close();
tcp.Create(); // now blocking again, yet the "nonBlocking" bool is TRUE
tcp.SetNonBlocking(true); // does nothing !!!
```

I do not know the full extent of this problem, if the server is affected similarly, if the UDP code has similar issues, etc. But likely answers are: extent is wide, other options are affected too, server is broken too, UDP is broken too :)

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPManager.cpp
+++ b/addons/ofxNetwork/src/ofxTCPManager.cpp
@@ -130,6 +130,8 @@ bool ofxTCPManager::Create()
 
 	if(!ret) ofxNetworkCheckError();
 
+	nonBlocking = false;
+
 	return ret;
 }
 


rafirstfruit:
This patch fixes the blocking option from my example, but there are probably many more things to patch.

arturoc:
if there is a way of querying the state of the socket we could just remove this cached values and use that instead. keeping internal mutable state like this is always problematic. 

i think this settings are mostly set at the very beginning when you create/setup the socket and never touched again so even if it's slow i think it's fine. 

another option is to have all this settings in the settings object and have default values for all of them directly in the header that are the default ones in the system then when the socket is created just overwrite the settings object with a new one or the passed one.

rafirstfruit:
There *are* ways to query sockets for their current options, and yes, it would be a bit slower. But like you say, keeping track of the state is problematic.

One interesting example (to your second point) is buffer size. Unless we actively set it, all OSes have different default values. So if we do add "defaults in the header", we'll have to set *all* the values we care about.

So yeah, the performance loss would be negligible, but it certainly would occur.

Personally, I like both your approaches. Not sure which one is better in the long term, but I'll try to work on a patch.

arturoc:
that would be great thanks! i think querying every time is perfectly fine, it's not the kind of thing that you are going to query or change a lot anyway and in most cases it won't ever be changed or queried after setup

Issue #4802, opened 2016-02-10
Integrate Pointer event based system into the core.


bakercp:
At the developer conference at YCAM a few years back we proposed that one way to improve the cross platform experience would be to move to a unified touch/mouse/pen/pointer input event system.  An implementation of a W3C Pointer-events like UI input was suggested (See "Cross Platform Experience" https://docs.google.com/document/d/1GjgjQZCHTcOD3cT9OFSl0AZ1nhfYwwvxTl5L4OJE6qA/edit).

In the few years since, the W3C specification has lost and then picked up steam again and now seems to be the way forward on the web.  It's exciting!

So here's the system in a single image:

https://github.com/w3c/pointerevents && https://w3c.github.io/pointerevents/

![image](https://cloud.githubusercontent.com/assets/300484/12935279/4a01da64-cf5a-11e5-98ac-b1d25e8e7e74.png)

_Fig. 1 A pointer is a hardware agnostic representation of input devices that can target a specific coordinate (or set of coordinates) on a screen._

In the years since, I have continued to update, refine and test (ofxPointer)[https://github.com/bakercp/ofxPointer] an implementation (as close as possible) to the specification.  It is a drop-in replacement for the existing mouse/touch system.  In its current form, it simply repackages the existing of mouse / touch events and makes it much easier to write applications that work across platforms.  For an example of its use in a desktop multitouch context, see https://github.com/bakercp/ofxTouchPad.  To try it out on a mobile platform see the ios example.  To try it on desktop with a mouse see the desktop example.

Integrating this into the core would be fairly straight forward.  The only real modification needed would be to decide if we want to enable it at first by default (which would just require us to add event listeners).

If people are into this (I know a few are) I am happy to put together a PR.

bakercp:
See also #4802 #4801 #3332 #3481 #2176 @danoli3

danoli3:
Yeah this sounds like a great plan!
Anything that is going to make the cross platform developer experience easier is a winning change in my book!

frauzufall:
I like it. Seems like a thoughtful way to improve input event handling and cross platform usage.

bakercp:
In the last week, I've significantly updated ofxPointer to bring everything up to spec and integrating advanced iOS features (including Apple Pencil features, etc).

on iOS, it now supports a host of `UITouch` feature on iOS including coalesced and predicted touches, pressure, tilt, estimated property updates, etc.

https://github.com/bakercp/ofxPointer

PR #4306, opened 2015-08-27
Refactoring ofxGui

frauzufall:
Refactored ofxGui version with new controls, layouts, options..

PR #5041, opened 2016-04-09
ofxGui refactoring #2

frauzufall:
This is a new PR regarding ofxGui. After working on the first approach (https://github.com/openframeworks/openFrameworks/pull/4306), talking to Arturo and people from the doc sprint in Denver I decided to redo this one more time (sorry for all the commits, can merge them all into one, if that's bugging someone, I personally like the history ;) ). I only post this now because I had to have a proof of concept first.

These are the general goals:
1. make ofxGui extendable
2. simple usage for beginners
3. emphasize ofParameter
4. introduce ofJson for configuring style and layout in order to prospectively allow CSS-like styling in a separate file without recompiling

The first approach solved big parts of it but there were short-cuts in accessibility and the way the layout was done. That's why I wanted to combine ofxGui with ofxDOM while keeping ofxGui as much the same as possible. @bakercp did a great job implementing the DOM standard, it just fits together really nicely and aligns with the ofJson concept. ofxDOM natively works with ofxPointer events for mouse and touch interactions. I removed this dependency and used the existing mouse events in order to keep it minimal and focus on the layout right now, but I strongly hope ofxPointer will be integrated into the core in the future to provide an interface for all devices. This would give us the full power of ofxDOM in ofxGui. See https://github.com/openframeworks/openFrameworks/issues/4802 to discuss this topic.

This approach allows to use pointers to gui objects while avoiding ownership problems. The layout supports static and absolute positioning, floating, percental width and margin. The configurations can be done with ofJson or the respective functions. This is an example:

```c++
// ofApp.h
ofxGui gui;
ofxPanel* panel;
ofxGuiGroup* group;

ofParameter<float> value1;
ofParameter<bool> value2;
```

```c++
// ofApp.cpp
void setup(){

  value1.set("value 1", 0, 0, 1)
  value2.set("value 2", false)

  panel = gui.addPanel();
  panel->add(value1);               // adds slider
  panel->add(value2);               // adds button

  group = panel->addGroup();
  ...

}
```
There is one root ofxGui object from which you create your panels and groups. You can add parameters and new groups, panels etc. to them.

Have a look at the examples in the branch. The main examples are `guiExample`, `advancedGuiExample` and `layoutExample`. 

About the layout: In order to keep it simple for beginners but also extendable, all gui groups added to ofxGui directly are absolutely positioned objects with a default width. All objects and groups added to these groups have a static positioning. By default these objects are non floating, 100% percental width. That means that they change their width according to their parent. You can also configure the objects..

.. with the respective functions:
```c++

panel->setBackgroundColor(ofColor::red);
panel->setPercentalWidth(true, 0.3); // set to 30% of the parents width

// set to fixed width
panel->setPercentalWidth(false);
panel->setWidth(300);  
```

.. and with ofJson:
```c++
ofJson config = {
  {"background-color", "#ff0000"},
  {"border-color", "rgba(0,0,255,0.5)"},
  {"width", "30%"},
  {"float", "right"}
};
panel = gui.addPanel("panelname", config);
```
It still needs some work, but I wanted to have your opinion on the whole concept as early in the process as possible. I am testing it with some more complex projects right now and I like it. I will also start adding documentation to the code as soon as I have a general approval for the concept. 

Note that I spent no thoughts on the design, I hope to get some help there.

Tell me what you think!

(Since the topic is quite complex, I would love to discuss this in person with anyone interested, perhaps we can have another Berlin meet-up about this, @arturoc, @fxlange ?)


arturoc:
Hey thanks so much for looking into this.

Have to look into it in more depth but a couple of things:

- I've always tried to avoid returning pointers to the internal objects since they can become invalid and produce crashes, we can look into returning references to do some simple changes to the gui elements after they are added but i think having the json config and ofParameter for the data should be enough to have enough flexibility so you can setup everything from the beginning and not have to modify the specific controls later on which will also simplify the internal logic by not having to keep mutable state.

- Not sure what ofxDOM does but it seems a little bit too much. the idea behind having json config was mostly to be able to change the behaviour and the style in a simple way but having some kind of DOM emulation seems a little overkill. also the way the addon is integrated as a library is kind of strange and it doesn't seem an addon common enough to have it as a core addon.

Also even in web the float attributes are becoming deprecated by the new flexbox model so not sure it's the best idea to use it ourselves. i've personally always find it confusing and something with vertical and horizontal panels should be enough for something like ofxGui which i think should aim to be simple and easily extendible rather than super complete, much less in terms of layout.

Let's meet anyway and look into it. i'll mail you and felix.

frauzufall:
@arturoc I was expecting your response and I was really trying to do it differently in order to keep the methods we agreed on, but I like this approach too much. 

- ofxDOM is separating the layout functionality from ofxGui following a standard. The included minimal version provides a root object, lets elements own other elements, and arranges them according to a layout, it's not an overload although it being a different addon is challenging for sure. I was able to remove the hierarchy of elements and the layouting from ofxGui, which was one of the weak points in my view, because it was hard to maintain and to extend. Now the gui elements and the layout functionality are separated in a clean way. ofxGui is tidied up. Users don't own any gui objects, they can't initialize their own controls, add them to the gui and delete them too early.

- The float layout made me get rid of quite some options (and ofxGuiMatrix). I can take a look at the flexbox layout, in ofxDOM it is pretty easy to add another layout and just use the new one.

- Including ofxDOM could be handled differently, I agree. I didn't know what would be the best way.

Is the guiExample too complicated for beginners, what do the people in teaching think?

In my view it is about whether we want this to be truly extendable or not. If not, we can keep the existing super basic ofxGui and let everyone with higher demands use external addons. Then I would just make this an external addon as well. I really tried to implement something that is easy to use without any options and still provides a ground on which other people can extend it however they want. Should the gui in the core be extendable in controls but not in layout? Doesn't it make sense to use standards like DOM or existing layout models in this scenario? 

Don't get me wrong, this is not perfect or done or anything, it's more a discussion about the general direction we are heading to with ofxGui. And I have no hard feelings no matter how this turns out :) I need all this functionality so I had to give it a shot and I will surely use it for my projects.

arturoc:
i see. i think it's overly complicated for a core addon and would probably be better to have it as an external addition, something like ofxDOMPanel or something like that?

As i see it ofxGui should be a simple and flexible way to control the parameters in you application so you don't have to recompile every time you want to change a value but the layout is totally secondary so adding so much code, dependency with another addon... for that doesn't seem a good idea.

I think the priorities should be:
- be able to represent all kinds of data:
  right now we have: int/float range (sliders), bool (toggle), void/trigger (button)
  it would be really useful to have: 
    * text boxes
    * multi option with and without multi choice
    * read only int/float, to represent things that you can't set in the gui like fps of the app or similar
    * images
    * graphs for signals
  (i know you already did most of them)

- flexibility to use any external control without renouncing to use ofParameter/Group: This way we can have something really simple in the core but people can use more complex stuff if they want.

- simplicity: the core elements should be as simple as possible but allow others to add more complex controls as external additions, if we find that some addition is really stable and use we can then bring it into  the official addon

- performance, the gui should be really fast, you should be able to have tons of options in the screen and that should consume almost no cpu/gpu so testing apps is not slowed down by the gui. this could be hard if we have a really complex layout system


In terms of layout my idea was to have vertical and horizontal panels. i think that should be enough for what ofxGui is meant to be and if we have a nice syntax to create the controls from parameters it can be really powerful.

Something like:

```cpp
ofParameter<bool> onoff{"on/off", true};
ofParameter<bool> enablealpha{"enable alpha", true};
ofParameter<bool> enablecolor{"enable color", true};
ofParameter<float> amplitude{"amplitude", 1.f, 0.f, 1.f};
ofParameter<float> lowpass_cutoff{"cutoff", 200.f, 20.f, 20000.f};

ofxVPanel panel{
    ofxHPanel{
        onoff, enablealpha, enablecolor,
    },
    amplitude,
    lowpass_cutoff,
};
```

Allows you to specify a full gui with layout in the header without having to specify any extra layout instructions which could be really nice.

I think when working with ofxGui you mostly want something that is really quick to setup and modify and allows to represent any parameter that you might have in your app, it's mostly a tool to allow to debug and develop not so much a "production" gui but perhpas not get in the way for there's external controls that allow for more complex layouts.

frauzufall:
Thanks for explaining. I understand that you want to keep it simple and fast. Did you run the examples? On my machines, this code is as fast as the first attempt. The layout is, as before, only calculated on setup, child adding and window resize. It is also super solid in comparison to the first version where no one seems to be able to help with this segfault on exit.

Where exactly is this overly complicated? It does not add a lot of code, it just separates layout and hierarchy (which was also a part of ofxGui before, just not extendable) from the controls. Modularity is a good thing. It also makes it possible to use the layout for other things than gui, which is pretty nice. There are a lot of people who want to use the interface not just for debugging. That's why there are so many gui addons out there that are not compatible with ofxGui and our first attempt is probably not going to improve that.

Here is the current way to implement your example with the current version:
```c++
//.h
ofParameter<bool> onoff{"on/off", true};
ofParameter<bool> enablealpha{"enable alpha", true};
ofParameter<bool> enablecolor{"enable color", true};
ofParameter<float> amplitude{"amplitude", 1.f, 0.f, 1.f};
ofParameter<float> lowpass_cutoff{"cutoff", 200.f, 20.f, 20000.f};

ofxGui gui;
```

```c++
//.cpp with float layout
void setup() {
	ofxPanel* panel = gui.addPanel();
	ofJson config = {{"width", "33%"}, {"float", "left"}};
	panel->add(onoff, config);
	panel->add(enablealpha, config);
	panel->add(enablecolor, config);
	panel->add(amplitude);
	panel->add(lowpass_cutoff);
}

```

That's how it would look like with the flexbox layout implemented:

```c++
//.cpp with flexbox layout
void setup() {
	ofJson config = {{"display", "flex"}, {"flex-direction", "row"}};
	ofxPanel* panel = gui.addPanel();
	ofxPanel* vpanel = panel->addPanel("", config);
	vpanel->add(onoff);
	vpanel->add(enablealpha);
	vpanel->add(enablecolor);
	panel->add(amplitude);
	panel->add(lowpass_cutoff);
}
```

The json parts could also be loaded from an external file. I think setting up the interface in the setup function is more understandable than in the header. We could also add an `addHPanel()` function that adds a panel with the config shown in the last example.

I know this can still be simplified, I am open to changes in details as well as in the general concept and I'd love to get critics on the actual code. I just can't afford to work on something that I won't be able to use, I hope you understand.

frauzufall:
Hey,

I integrated a simple method to add controls to a panel without even having to create it yourself:
```c++
//.h
ofxGui gui;

//.cpp setup
gui.add(param1, param2, param3, ..);
```
I also wrote a simpler layout that aligns items horizontally or vertically and is used by default but you can tell the GUI to use another layout if you want to. Also added a bunch of documentation and examples but I am no longer working on this branch. Since I got no feedback other than Arturos (thank you btw!) everything is part of this standalone addon now: [ofxGuiExtended](https://github.com/frauzufall/ofxGuiExtended). Now I can use it with the master OF branch and I can also let people use and test it easier. I guess we can close this PR now if no one has any other opinion.

Please be sure that for me negative critique is still better than no feedback. I am here - on this planet - to learn :)

transat:
For what it's worth, I've been using ofxGuiExtended and find it to be fast, very flexible in terms of layout and ofParameter friendly. I'd love to see it integrated into the core, especially as the more complex aspects (like flexbox) are entirely optional.

martialgallorini:
hi,

thanks a lot for this gui improvement.
I don't understand every technical details you are debating here, so maybe my opinion is not that relevant, but i must say arturo's approch seems quite interesting. Adding more common controls to the core GUI while keeping it fast and simple is the priority. We need to be able to control a variety of common values and position them easily on a layout. 

The idea of integrating them into the actual core addon is ok. Extending it with more complex or specific controls can be really useful. And as arturo said, if those extensions are proven to be useful and stable enough, one could merge them into the core gui addon later.

Just my two cents.

Really great job by the way. I am eager to see all those new controls added to the core

PR #5051, opened 2016-04-11
Fixes missing returns from #5042.

bakercp:


bakercp:
Merging this, as it's a critical bug.  The appveyor build on x86 was cancelled for some reason, but the x64 and linux/osx, etc passed.

arturoc:
thanks!

PR #5054, opened 2016-04-12
Add support for OF_PIXELS_NV21 in android video grabber

HalfdanJ:
This adds support for the native pixel format NV21 of the android video grabber in the returned ofPixels. 

ofPixels does not support OF_PIXELS_NV21 very well, but it's useful for advanced pixel processing in openframeworks where you need the native pixels for better performance just using ofPixels for passing the data on.


arturoc:
thanks, why do you say that ofPixels doesn't support NV21 very well?

HalfdanJ:
Looked a bit more afterwards, and you are right, there are lots of functionality. Its only the getting and setting color that is missing and that i remembered.

arturoc:
ah yeah, that's missing, i added the formats mostly to avoid colorspace conversion in the cpu but would be nice to have it anyway

PR #5055, opened 2016-04-13
overflow in 32bits systems with bigger images

anagma:
in bytesFromPixelFormat function, the calculation can overflow in 32bits systems with bigger images.


arturoc:
This is breaking the tests cause some YUV formats don't have a bpp that's a exact multiple of 8. A better solution would be to change the parameters from int to size_t but it's probably something that needs to be done across all the methods in ofPixels.

arturoc:
going to close this since there's already a PR that changes everything in ofPixels to size_t. thanks!

PR #5057, opened 2016-04-15
Fix missing return.

bakercp:
Fixes missing return introduced by d751a54e75f48556c56d90fa68d0f845b5bde2fd


bakercp:
the x86 test keeps getting cancelled. This passed all other tests and may crash w/o the return.  So, merging.

arturoc:
yeah appveyor seems to fail sometimes because the repository is too big, i guess it'll be fixed once we remove the binaries from it but by now if one of the platforms passes the tests it should be ok too merge

Issue #4885, opened 2016-02-20
Android Studio build forever


whatisor:
Android studio never finish to build.
it keeps on step:
Gradle: Configure project: openFrameworksProject.

Stefan0:
I had the same problem using Win 10 (64 bit), OF release 0.9.1, Android Studio 1.5.1, jdk1.8.0_71 and android-ndk-r10e.
In my path.make I've changed NDK_ROOT=E:/Android/android-ndk-r10e with NDK_ROOT=/Android/android-ndk-r10e. Now it stops building but with this error:

Error:(157, 1) A problem occurred evaluating project ':openFrameworksProject'.
> make failed with exit status 2

In logs I have this exception: org.gradle.tooling.BuildException: Could not run build action using Gradle distribution 'https://services.gradle.org/distributions/gradle-2.8-all.zip'.

Stefan0:
Using Windows 7 it builds successfully.

whatisor:
I have to use Eclipse on Win 8 (64bit) instead of.

Issue #4549, opened 2015-11-09
Keep supporting android Eclipse?


HalfdanJ:
I know several of you are still using eclipse for android development. But i would like starting a talk about for how long we should support eclipse. It seems like there are several hacks both in the eclipse and android studio workflow to support each other. And to me it seems like Android Studio is more then ready to take over from Eclipse, especially now where all googles focus is on improving android studio, including the NDK tools. 

The eclipse official support for android is [ending by the end of this year](http://android-developers.blogspot.dk/2015/06/an-update-on-eclipse-android-developer.html)

danoli3:
+1 burn eclipse with fire.

ofZach:
I used eclipse really heavily this summer and I would very much in favor of a move to android studio.   the tooling, info and overall experience of eclipse at ths point is really sub-par.   As android studio supports NDK more and more this would be a really useful move.

arturoc:
does android studio already have auto completion in c++ or any basic support at all other than syntax highlighting?

HalfdanJ:
It has c++ auto completion, but its not autocompleteting openframeworks. But i don't know if its a setting we are missing, or its a current limitation. But the NDK support is currently being improved, so it's not able to it right now, then my guess is it will be able to very soon. Might be something worth asking android studio developers directly

HalfdanJ:
Here is what i've currently seen of official docs about NDK support: http://tools.android.com/tech-docs/android-ndk-preview

HalfdanJ:
Looked a bit more into it. It seems like the thing we wait for the most is the release of gradle-experimental plugin. The current 0.4.0-alpha version supports native NDK modules, when that is released it seems like android studio will support proper external c++ libraries. But it does not make sense yet to implement it since the know issues are the following:

> Known Issues
> - Editing support for the standalone plugin is not yet implemented in Android Studio
> - Modifying a source file in the library does not automatically cause the application to re-link the new library when building an application.

See more here: 
http://tools.android.com/tech-docs/new-build-system/gradle-experimental

HalfdanJ:
Would like to wake up this conversation again. 

Android Studio has evolved a lot lately, and the c++ support is becoming good. I have a working version of openframeworks in #5059 with auto complete working, and I've now used it for a large project for Google I/O with big success. At I/O a new version of android studio got released supporting CMake as well. 

I'm currently doing a lot of Android development for the google creative lab, and would love to use this opportunity to push the android studio support even further. 

[...]

Issue #4503, opened 2015-10-30
Android Studio: Run compiles both release and debug


HalfdanJ:
I'm not able to completely understand the flow through the make and gradle files. But i can see that when i click run in android studio, gradle is compiling both release and debug mode. Also in both compile steps the resource zip is being produced. 

I can disable this behavior by commenting out [this line](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/android/build.gradle#L166), but thats definitely not the solution :)  

I did some benchmarking on a small example where i didn't make any source changes. In the current mode with both release and debug mode, the build took 5.3 sec. Disabling release mode brings down compile time to 1.8 sec.

Any ways to fix this? 

Here is my gradle log output: 

```
Executing tasks: [:generateDebugSources, :generateDebugAndroidTestSources, :assembleDebug]

Configuration on demand is an incubating feature.
ofNdkSetup
HOST_OS=Darwin
HOST_ARCH=x86_64
make: Entering directory `/Users/jonas/Development/openframeworks/openFrameworks/examples/android/androidCameraExample'
Compiling OF library for Debug
/Users/jonas/Development/Android/android-ndk-r10e/prebuilt/darwin-x86_64/bin/make -C /Users/jonas/Development/openframeworks/openFrameworks/libs/openFrameworksCompiled/project/ Debug PLATFORM_OS=Android ABIS_TO_COMPILE_DEBUG="armv7"
HOST_OS=Darwin
HOST_ARCH=x86_64
make[1]: Entering directory `/Users/jonas/Development/openframeworks/openFrameworks/libs/openFrameworksCompiled/project'
HOST_OS=Darwin
HOST_ARCH=x86_64

HOST_OS=Darwin
HOST_ARCH=x86_64
Done!
make[1]: Leaving directory `/Users/jonas/Development/openframeworks/openFrameworks/libs/openFrameworksCompiled/project'


Compiling androidCameraExample for Debug
HOST_OS=Darwin
HOST_ARCH=x86_64
make[1]: Entering directory `/Users/jonas/Development/openframeworks/openFrameworks/examples/android/androidCameraExample'
make[1]: Leaving directory `/Users/jonas/Development/openframeworks/openFrameworks/examples/android/androidCameraExample'

HOST_OS=Darwin
HOST_ARCH=x86_64
make[1]: Entering directory `/Users/jonas/Development/openframeworks/openFrameworks/examples/android/androidCameraExample'
[...]

Issue #4168, opened 2015-07-28
Support new Android Studio NDK


nneonneo:
Android Studio 1.3 will finally have more official support for the NDK (though it will still be a preview). This is a placeholder issue to remind me/us to look at updating the Android Studio support for the new Android Studio.

1.3 isn't officially released yet, and they have explicitly indicated the new NDK support isn't final. So, right now we shouldn't do anything, but in a few months when the NDK support is finalized we will want to update the project.

mruegenberg:
Right now, the androidEmptyExample works with Android Studio 1.4 beta when adding the file `/libs/openFrameworksCompiled/project/android/gradle.properties` containing `android.useDeprecatedNdk=true`. 
(In my case only on the device, not in the emulator, though.)

The alternative is to switch to the experimental Gradle plugin (see http://tools.android.com/tech-docs/new-build-system/gradle-experimental). I tried that, and got relatively far, but I'm stuck right now with some pretty unhelpful Gradle errors. (If there is any interest, I can post the edited versions of the files I changed here.)

nneonneo:
Right now I have no interest in chasing a half-finished experimental build
system. We should just add the useDeprecatedNdk flag (and check if that
works with older Android Studio versions). We will migrate to the new
system once it matures.

Robert

HalfdanJ:
Related to #4549, it seems like gradle 0.4.0 experimental is the version that might do the difference. Its supporting native libraries without wrappers, so it should allow to include openframeworks in a "correct" way through gradle (without any shell make commands in gradle). Its currently in alpha, so its not ready yet to be used.

PR #5059, opened 2016-04-16
Android Studio Native building (Gradle-experimental) - Work in progress

HalfdanJ:
@openframeworks/android 

Here is a work in progress branch where I built openFrameworks using the new [gradle-experimental](http://tools.android.com/tech-docs/new-build-system/gradle-experimental). This enables proper c++ editing in Android Studio, including autocomplete. And it enables also breakpoints in c++ code. It also gives proper error messages when build fails, instead of the current problem where it can silently hang (#4885)

One of the big benefits of using gradle-experimental over the current make system is that it takes care of all the platform specific tweaks that the make systems needs to do, simplifying the configuration drastic. Also, now it only compiles the ABI's and debug/release mode needed, making it much faster to build and run. 

This will only work in Android Studio, so no Eclipse! Until now it has made sense to use both, but in my opinion this changes it, since all features i can think of are in Android Studio, and eclipse support is discontinued. This could work in parallel with the current build system for eclipse support if needed.

All examples have been tested. 

**Changes** 
- Modified gradle scripts to support Android Studio 2.0 
- Removed legacy Eclipse files 
- Made NEON support default (very few devices now don't support NEON, and Unity made it mandatory already in 2015). 
- Updated libraries to work with latest NDK r12
- Changed OFActivity to non abstract, this enables [direct inheritance in the manifest](https://github.com/HalfdanJ/openFrameworks/blob/gradle-experimental/examples/android/androidEmptyExample/AndroidManifest.xml#L21), and then there is no java activity needed in the app, but can optionally be added if you need access to Java activity stuff (i think its rare). 
- Added a checkPermissions function to support Android N and its new permissions model for runtime permission requests, used by videograbber

**Known issues**
- [x] ~~Addons are not implemented yet. Need to do a gradle script for that.~~ Basic addon support implemented
- [x] ~~When opening the project, it will ask you if you want to download NDK automatically. Unfortunatly the latest NDK does not work with openFrameworks, so you need to manually give it the path to you local `ndk-r10e` folder. (but no need for paths.make)~~ Updated to support latest ndk-r12b
- [x] Due to a [known bug in the NDK](https://code.google.com/p/android/issues/detail?id=204143), the llvm toolchain is missing `ar` binary for building static libraries. You can copy it from one of the other toolchains. Something like

```
$ cd android-ndk-r12b/toolchains
$ cp ./arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/arm-linux-androideabi/bin/ar ./llvm/prebuilt/darwin-x86_64/bin/ar
```

- [ ] There are [two core changes](https://github.com/HalfdanJ/openFrameworks/commit/f6ffecd677fbaafa4099c957b1fd96fcf8323483) i needed to make for it to work, not sure what right solution is.
- [x] NEON support. I'm not really sure how this works, and understand how this is supported in this setup yet.
- [ ] Tests scripts have not been updated

This PR should close following issues: #4885, #4549, #4503 and #4168

Give it a try, and let me know!

Tested on OSX with Android Studio 2.1.1


HalfdanJ:
I've now changed the way assets are handled. Instead of custom code for zipping the data, and unzipping inside the ofapp, instead gradle is just pointed to bin/data in `assets` setting, and will take care of copying the files. This removes all platform specific code.

danoli3:
Wow this is amazing! Nice changes @HalfdanJ !! Yeah that Zip process was always messy.

antoniohof:
@HalfdanJ incredible work! built fine here after I resolved this issue within the ndk:  (windows 10 only) https://code.google.com/p/android/issues/detail?id=195801

-
another issue is that its not copying the data folder over, I think it doesnt have anything to do with your last commit, since I have tried with old and new OFAndroidLifeCycleHelper.java

files folder is empty on sd card and and with 3 folders on internal storage: images, sounds and webkit... no sign of the my application .xmls

this message appears a lot, dont know what it is:
04-24 23:56:14.573 9240-9240/? W/ResourceType: For resource 0x0103009a, entry index(154) is beyond type entryCount(1)
04-24 23:56:14.573 9240-9240/? W/ResourceType: For resource 0x0103011c, entry index(284) is beyond type entryCount(1)

HalfdanJ:
@antoniohof If you unzip the APK files produces, is the content of bin/data then in there? I'm wondering if its the packaging or unpackaging that doesnt work. 

I'm also unsure on that error message.

antoniohof:
@HalfdanJ Just unzipped, its not there

HalfdanJ:
@antoniohof Just realized i hadn't update the empty example with correct path (since it didnt have assets in it). Its now updated for demonstration. (https://github.com/openframeworks/openFrameworks/pull/5059/commits/7c049afc2c6058f87030577cf16f5ec6cc817950#diff-eb0c6d254a18b44efac2206b01ebd114L75)

scripts/apothecary/apothecary
```diff
@@ -88,7 +88,8 @@ VS_64_BIT_ENV='VC\bin\x86_amd64\vcvarsx86_amd64.bat'
 # paths to android SDK, etc
```
arturoc: the latest apothecary is in it's own repository now, in https://github.com/openframeworks/apothecary  can you send this changes there?

HalfdanJ: Ohhh, thanks for noticing it!

arturoc: np, also the travis config in that repo aready auto builds the latest
libraries but you'll need to change it to download v12 of the NDK i guess

HalfdanJ: Is the idea to remove the formulas from main repo, but keep updating the .a files with new builds? Or is it to remove the .a files entirely from git soon, and depend on external download or manual build?

danoli3: Eventually to remove all the binaries from the repo and history.

antoniohof:
Nice progress!
I'm getting new errors here after moving to the new NDK and updating with your newer commits, they have something to do with exception handling, i'm using windows 10:

~~~~ android-ndk-r12b\sources\cxx-stl\llvm-libc++\libcxx\include\ostream
Error:(401) undefined reference to '__cxa_begin_catch'
---
\libs\openFrameworks\events\ofEvent.h
Error:(579) undefined reference to '__dynamic_cast'
Error:(91) undefined reference to '__cxa_begin_catch'
Error:(421) undefined reference to '__dynamic_cast'
Error:(421) undefined reference to '__dynamic_cast'
Error:(421) undefined reference to '__dynamic_cast'

android-ndk-r12b\sources\cxx-stl\llvm-libc++\libcxx\include\fstream
Error:(789) undefined reference to 'std::bad_cast::bad_cast()'
Error:(821) undefined reference to 'typeinfo for std::bad_cast'

\libs\openFrameworks\types\ofParameterGroup.cpp
Error:(151) undefined reference to 'typeinfo for std::runtime_error'
Error:(151) undefined reference to 'std::runtime_error::~runtime_error()'
Error:(151) undefined reference to 'vtable for std::out_of_range'
Error:(151) undefined reference to 'typeinfo for std::out_of_range'

C:\Android\android-ndk-r12b\toolchains\arm-linux-androideabi-4.9\prebuilt\windows-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin\ld: the vtable symbol may be undefined because the class is missing its key function (see go/missingkeymethod)
Error:error: undefined reference to 'typeinfo for std::overflow_error'
Error:error: undefined reference to 'std::overflow_error::~overflow_error()'
~~~~

tallavi:
There's a yearly survey coming soon, isn't it?

Perhaps we should ask Android OF users about this, see how many of them prefer Eclipse vs Android Studio? If it's split exactly in the middle, perhaps it would make sense to put the effort in supporting both..

HalfdanJ:
@antoniohof what project are you trying to build? One of the examples? It looks to me like its trying to compile against a binary library build with the wrong STL (I have rebuild all the binary libraries in openFrameworks). 

@tallavi not sure the survey is the best way of doing it. Its a very small percentage to my knowledge that develop for android, and most people are developing in some other IDE and just compiling using eclipse (based on looking a previous surveys). And also not sure how many of these are aware of how much better Android Studio is for Android since Eclipse has been deprecated by Google. All OSS projects i see today for Android are focusing on Gradle and Android Studio. 

The reason I'm hesitant just implementing Eclipse is that I know I wont have time to do it. I know very very little about the build system for Android Eclipse, and it would require bunch of updates since I modernized the NDK and structure. I also know it shouldn't be the reason for releasing something (not having time), but its just the reality for me, and would love to see this update to get into a oF release since I believe its a much needed update drastically improving the experience developing for Android.

tallavi:
@HalfdanJ, I'm all for it!

I believe that if the migration instructions will be clear and simple and the process will be straight forward, it shouldn't be a problem to break backwards compatibility on a major release and force everyone to download android studio..

I promise to try out your branch on my project as soon as possible and report back.

antoniohof:
I was developing a project with the official android studio support for openframeworks. No auto-complete, no debugging and slow compile times. Almost every new addon had a problem compiling and it was impossible to find out what went wrong. When I found this branch it was awesome, it unleashed the full potential of Android Studio. It is incredible, thank you. I think the c++ support will only get better and everyone would benefit from this.

tallavi:
Hmm, is there a specific branch of projectGenerator I should take that will generate the new project structure? 

Running the normal one with -p"android" returns "platform not supported". Running without -p builds the project alright but the build.gradle file is empty. 

Help.

cerupcat:
Any updates on the status of this? Will be this be merged with core?

HalfdanJ:
This PR is replaced with #5461

PR #5060, opened 2016-04-17
Update ofxOpenCv addon_config for android

HalfdanJ:
the addon_config.mk reffered to armeabi libraries, but they dont exist. Instead the x86 libs should be pointed to.


arturoc:
thanks!

PR #5070, opened 2016-04-25
use pkg-config for rtaudio

underdoeg:
On arch linux the rtaudio includes are installed under /usr/include/rtaudio/ 

OF won't find them when compiling, unless set as a pkg-config library. Works with ubuntu too and I guess other systems provide the .pc file as well?


arturoc:
it seems this is failing in all the linux platforms we support, it might be that this got introduced in a later version but it's not in older ubuntu versions or debian. perhaps you can add a check that finds out if that pkg config exists by calling pkg-config and checking the status value, if that is 0 then you can add the pkg config safely

underdoeg:
Build servers are awesome :). Jep I only had a fairly recent version of ubuntu to test on. Should be fixed now. Let's wait if it compiles...

arturoc:
yeah, that worked, thanks!

PR #5076, opened 2016-04-27
Feature ofParameterGroup::remove(ofParameter)

underdoeg:
This adds remove functions to ofParameterGroup. I didn't really know how to handle generic parameters so right now it just compares type and name. Currently this won't work if two children have the same name but it looks like this is an issue with the parameter group anyways, so I added a warning.

I also tried to automatically remove a child from a parent when the parameter gets dereferenced. Like:
```
ofParameterGroup group;
ofParameter<int>* test = new ofParameter<int>();
group.add(test->set("test", 10));
delete test;
```
This won't work though because parameter groups store their children as shared_ptr. This makes sense if a user does not keep track of all parameters themselves. But I feel like we usually do and at least for me this resulted in some unexpected behavior where I wasn't able to get rid of my parameters in the gui.  It is not really an issue though...

### DIFF
--- a/libs/openFrameworks/types/ofParameter.cpp
+++ b/libs/openFrameworks/types/ofParameter.cpp
@@ -43,6 +43,10 @@ vector<string> ofAbstractParameter::getGroupHierarchyNames() const{
 	return hierarchy;
 }
 
+bool ofAbstractParameter::isReferenceTo(const ofAbstractParameter &other) const{
+	return getInternalObject() == other.getInternalObject();
+}
+
 ostream& operator<<(ostream& os, const ofAbstractParameter& p){
 	os << p.toString();
 	return os;
--- a/libs/openFrameworks/types/ofParameter.h
+++ b/libs/openFrameworks/types/ofParameter.h
@@ -52,10 +52,13 @@ public:
 	virtual bool isReadOnly() const = 0;
 	virtual shared_ptr<ofAbstractParameter> newReference() const = 0;
 
+	virtual bool isReferenceTo(const ofAbstractParameter& other) const;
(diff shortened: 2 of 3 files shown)


arturoc:
The way to know if it's exactly the same parameter would be to friend ofAbstractParameter to ofParameterGroup and have a private method that returns the internal pointer so it can be accurately compared by address.

Removing parameters from the group automatically doesn't make much sense imo since it might be that some parameters only live in the group. The most common case though is having parameters that are also declared somewhere else so we could optionally have something to do that like another vector of parameters that are added as weak_ptr or a second parameter group class that stores as weak ptr something like ofWeakParameterGroup

underdoeg:
You mean like a virtual method void* GetInternalObject()?

arturoc:
yeah that's it

underdoeg:
Ok. I've added those

arturoc:
looks great, going to leave the tests run on all platforms and will merge it. thanks!

underdoeg:
sounds good. thx

eduardfrigola:
Hi Philip and Arturo.
I found a bug in the remove function of `ofParameterGroup`. I don't know if here is the best place to put the bug, or open a new issue.
The bug is when removing parameters of the group `ofParameterGroup::remove(const string &name)`. The `parametersIndex` (the map that stores the position of each parameter in the` parameters` vector) is not updated when a parameter is removed.

Also, I think it should escape the name you passed, similar to the `ofParameterGroup::contains(const string& name)`.

arturoc:
can you create a new issue? it's easier to track it that way. thanks!

PR #5077, opened 2016-04-28
critical bugfix for ofReadOnlyParameter and isReferenceTo

underdoeg:
Sorry, I introduced a compiler error with read only parameters. This fixes it.

### DIFF
--- a/addons/ofxOpenCv/src/ofxCvBlob.h
+++ b/addons/ofxOpenCv/src/ofxCvBlob.h
@@ -37,6 +37,7 @@ class ofxCvBlob {
 
         //----------------------------------------
         void draw(float x = 0, float y = 0){
+            ofPushStyle();
             ofNoFill();
             ofSetHexColor(0x00FFFF);
             ofBeginShape();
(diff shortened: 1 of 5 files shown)


arturoc:
can we instead friend ofReadOnlyPArameter with ofParameter so it returns the correct value?

underdoeg:
I'd love to do this. But I didn't know how since ofReadOnlyParameter has not only the PArameterType but also the friend class Type. And afaik you cannot just go ```friend class ofReadOnlyParameter<>```. Or am I forgetting something?

We could have an ofAbstractReadOnlyParameter but it struck me as a little bit of an overkill...

underdoeg:
oh right... I got it now. Did not know about templated friend classes. Nice feature though. This should do it.

arturoc:
thanks, this now contains some changes that are in master already, i think it should be fine but if you can try to merge master into this branch to see if those changes go away?

underdoeg:
yes. I pulled before commiting and then it got merged automatically. I can clean it up but I think it should be fine.

arturoc:
thanks!

PR #5078, opened 2016-04-29
replace function pointer to std::function

2bbb:
now, we can't use `std::function` with capture like

```cpp
std::function<void(ofEventArgs &)> callback = [this](ofEventArgs &) { this->value++; };
ofAddListener(ofEvents().update, callback);
```

because, no capture lambda can be cast to static function. but if we capture something then it can't be cast.

this is simple problem. i only replace function pointer like `void(*)(ofEventArgs &)` to `std::function<void(ofEventArgs &)>`.

### DIFF
--- a/libs/openFrameworks/events/ofEventUtils.h
+++ b/libs/openFrameworks/events/ofEventUtils.h
@@ -65,45 +65,45 @@ void ofAddListener(ofEvent<void> & event, ListenerClass  * listener, bool (Liste
 }
 
 template <class EventType,typename ArgumentsType>
-void ofAddListener(EventType & event, void (*listenerFunction)(const void*, ArgumentsType&), int prio=OF_EVENT_ORDER_AFTER_APP){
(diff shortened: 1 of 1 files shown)


2bbb:
omg...
i will fix it.

2bbb:
i found `ofEvent::newListener();` method.
i think this PR isn't need soon.

arturoc:
:) yeah this won't work, ofRemoveListener needs a unique identifier for the function to be able to remove it from the event, for c functions and member functions (as well as lambdas without capture) that id comes from the function pointer but lamdas with capture don't have a function pointer so we create a numeric id that is returned by newListener and you can use to remove the listener from the event.

2bbb:
@arturoc 
sorry for my idleness… 😞 

but, i feel ofAddListener and ofRemoveListener have complexity that causes hard to intuitive use.
maybe this is caused by old school style code (like C++03), and maybe we want to use modern style.

so, i don't have good solution about this now.
however i wanna find good solution that doesn't cause conflict and destructive change.

if, i find it, then i post again.
and that time, also best regards.

thanks!

PR #5079, opened 2016-05-02
push and pop styling when drawing OpenCV blobs

brendan-w:
Currently, calling `ofxCvBlob.draw()` will leave the drawing color set to `0xff0099` (pink) with no fill. This PR simply adds push/pop style guards in the blobs drawing function, to prevent trampling the users styles.


arturoc:
thanks!

PR #5080, opened 2016-05-02
Bugfix/android vram memory leak

kkirby:
The camera's texture doesn't get deallocated, causing a pretty big memory leak when the app is paused/resumed several times. In theory, the texture should eventually deallocate automatically, but I haven't seen this occur.

### DIFF
--- a/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidVideoGrabber.java
+++ b/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidVideoGrabber.java
@@ -197,10 +197,11 @@ public class OFAndroidVideoGrabber extends OFAndroidObject implements Runnable, 
 			camera.setPreviewCallback(null);
 			try {
 				if (Build.VERSION.SDK_INT >= 11) {
-					camera.setPreviewTexture(surfaceTexture);
+					camera.setPreviewTexture(null);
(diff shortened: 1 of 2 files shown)


arturoc:
thanks for the fix, can you use ofTexture::clear instead of glDelete...

kkirby:
@arturoc Since the texture is an external texture, calling `clear` is essentially a NOOP, unless I'm missing something.

```c++
void ofTexture::clear(){
	if(!texData.bUseExternalTextureID){
		release(texData.textureID); // <-- This is never reached because
		                            //     bUseExternalTextureID is true.
	}
	texData.bUseExternalTextureID = false;
	texData.textureID  = 0;
	texData.bAllocated = false;
}
```

arturoc:
mmh so that texture is allocated by the OS or at least by android's camera class no? not sure why we should delete it ourselves if we haven't allocated

kkirby:
The texture is generated [here](https://github.com/openFrameworks/openFrameworks/blob/master/addons/ofxAndroid/src/ofxAndroidVideoGrabber.cpp#L115).

```c++
void ofxAndroidVideoGrabber::Data::loadTexture(){
	ofTextureData td;
	GLuint texId[1];
	glGenTextures(1, texId); // <--
	...
	...
}
```

arturoc:
oh i  see, thanks!

Issue #4981, opened 2016-03-09
OpenCV Linker error with nightly on VS


ofTheo:
Not sure why yet - but a lot of errors relating to OpenCv 

![buildcv](https://cloud.githubusercontent.com/assets/144000/13650018/a675605c-e60e-11e5-96e3-c99b31e9a5f1.jpg)

arturoc:
yes, somebody has reported this already, i think in the forum only. it seems like it would be trying to use ipp so it's probably just some missing/extra flag in the apothecary config.

prisonerjohn:
I just ran into this issue too. I'm not seeing any mention of opencv in apothecary, am I looking in the right place? https://github.com/openframeworks/openFrameworks/tree/master/scripts/apothecary/formulas

arturoc:
the most up to date version of apothecary is in https://github.com/openframeworks//apothecary but that doesn't contain addons yet, the addons formulas are in the addons folder of this repo. you can copy the opencv formula from ofxOpenCv to the apothecary repo and build it there.

PR #5081, opened 2016-05-03
Recompiled opencv for vs, including ippicv.

prisonerjohn:
Address #4981


arturoc:
thanks! did you finally used apothecary to recompile it? in that case can you send a PR with the fixed script to the apothecary repository?

prisonerjohn:
I did use apothecary. I don't think there was anything wrong with the script, I just manually copied over the ippicv stuff that it downloaded. I'll look and see if I can add that copy step in there so it happens automatically.

PR #5084, opened 2016-05-04
ofxiOSEAGLView - reset touches dictionary before the view appears

tallavi:
I have a hybrid c++/ObjectC app, so I have additional, non OF view controllers. I had a weird bug that if I have a finger down while going to a different view controller, that touch id is gone forever, and won’t be received again. This created issues with gesture recognition. I propose a small fix - to reset the touches dictionary in viewWIllAppear. I tested it in production scenarios and it works great. I saw no adverse affects for this. And it’s also a straight forward.


tallavi:
pinging @openframeworks/ios

danoli3:
Tested this out, works well and fixes that potential bug.

PR #5085, opened 2016-05-04
ofxiOSEventAdapter

tallavi:
My app is a hybrid c++/objC so I devised a way to easily register to ofEvents with an objC callback. It’s pretty much seamless for the end-user and looks like this:

```
//c++
static ofEvent<const bool> myEvent;

//ObjC

- (void)viewDidLoad {
    [super viewDidLoad];
    ofAddListener(myEvent, self, @selector(myCallBack:));
}

- (void) dealloc {
    [super dealloc];
    ofRemoveListener(myEvent, self, @selector(myCallBack:));
}

- (void) myCallBack:(const bool&) param{
}
```
    
Basically it creates a c++ callback internally and routes the event to the requested objC target. I use it in production scenarios all the time and it works flawlessly.

### DIFF
--- a/addons/ofxiOS/src/utils/ofxiOSEventAdapter.h
+++ b/addons/ofxiOS/src/utils/ofxiOSEventAdapter.h
@@ -1,0 +1,75 @@
+//
+//  ofxiOSEventAdapter.h
+//
+//  Created by Tal Lavi on 16/2/2015.
+//
+//
+
+#pragma once
+
+#import "ofxiOSExtras.h"
+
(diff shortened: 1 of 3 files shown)


tallavi:
Not sure why travis would fail. It compiles fine on my machine.

tallavi:
pinging @openframeworks/ios

danoli3:
Looks good thanks for adding this. 
Travis failed on getting a sub-module for the project generator. I've restarted it.

Issue #5087, opened 2016-05-06
String tests failing on MSYS2


oxillo:
The 'tolower' and 'toupper' tests on accentuated characters are failing on MSYS2 (on my system and on Appveyor)

```
[ warning] ofUtils: Couldn't create locale  using default, C
[  error ] tolower failed 
[  error ] test_eq(ofToLower("AbCéÉBbCcc"), "abcéébbccc")
[  error ] value1: ofToLower("AbCéÉBbCcc") is abcéÉbbccc
[  error ] value2: "abcéébbccc" is abcéébbccc
[  error ] C:/projects/openframeworks/tests/utils/strings/src/main.cpp: 58
[ warning] ofUtils: Couldn't create locale  using default, C
[  error ] toupper failed 
[  error ] test_eq(ofToUpper("AbCéÉBbCcc"), "ABCÉÉBBCCC")
[  error ] value1: ofToUpper("AbCéÉBbCcc") is ABCéÉBBCCC
[  error ] value2: "ABCÉÉBBCCC" is ABCÉÉBBCCC
[  error ] C:/projects/openframeworks/tests/utils/strings/src/main.cpp: 59
```

The other tests pass even if there is the "Couldn't create locale  using default, C" warning.

Should we test for accentuated characters as the result may depend on the locale? Shouldn't we force locale for the tests ?

dimitre:
I think I'm having similar issues while trying to convert strings with accents to upper or lower.
ofw 0.9.8 on OS X

oxillo:
Using C setlocale seems to correct the issue but I'm concerned about side effects.


`setlocale(LC_ALL,locale.c_str())`

arturoc:
we do something similar for linux if we detect a non utf8 locale so let's do this if it fixes the problem.

can you add a similar ifdef for windows non visual studio ater: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppRunner.cpp#L143-L156

```
#if defined(TARGET_WIN32) && !_MSC_VER
```

should detect windows but not visual studio

PR #5089, opened 2016-05-06
Appveyor : add Msys2 build

oxillo:
Add MSYS2 32 and 64b bits building.
64 bits currently fails.
32 bits fails on string tests (see issue #5087)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,19 +1,50 @@
 version: 1.0.{build}
 os: Visual Studio 2015 RC
+
+environment:
+  global:
+    APPVEYOR_OS_NAME: windows
+  matrix:
+  #MSYS2 Building
+    - BUILDER: MSYS2
+  #VisualStudio Building
+    - BUILDER : VS 
+    
 configuration: Debug
 platform:
 - x86
 - x64
 shallow_clone: true
-clone_depth: 10
-environment:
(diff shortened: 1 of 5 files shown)


arturoc:
appveyor seems to fail randomly cause the repo is too big, i believe it'll be solved once we remove the binaries. for the strings issues, in linux we try to set the default locale to something with utf-8. i think that or something similar should solve it:

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppRunner.cpp#L136-L149


can you give it a try?

oxillo:
Spent some hours on std::locale. My conclusion is that MSYS2 only supports "C" locale (not even C.utf8).
I'll set the default locale to "C" in ofAppRunner.cpp and make a MSYS2/C branch in the string tests. I'll submit a separate PR.

oxillo:
According to [this answer](http://stackoverflow.com/questions/4406895/what-stdlocale-names-are-available-on-common-windows-compilers), Mingw C++ doesn't support other locales than "C" and "POSIX" as it is not based on GNU C library.
Opened #5090 to investigate solutions.

arturoc:
just merged this, i've disabled the 64bit builds and the tests that needed a different locale until/if we figure something out

PR #5091, opened 2016-05-07
Fix #5087 : "C" only locale support

oxillo:
Define C_LOCALE_ONLY for systems that only support "C"/"POSIX" locale as MSYS2.
Update tests consequently.

Issue #5090, opened 2016-05-07
MSYS2 locale support


oxillo:
MSYS2 (and underlying Mingw C++)  only support "C" and "POSIX" locales. 
Need to find a solution for local support on MSYS2. Maybe use boost.locale or ICU?

PR #5092, opened 2016-05-08
MSYS2 locale support with boost::locale

oxillo:
Switch to boost::locale as MSYS2 only support "C" locale natively.
Fixes issues #5087 and #5090 
Replace PR #5091


arturoc:
I would like to not introduce more dependencies with boost since the idea is that it'll go away once std::filesystem is in every compiler

Issue #4809, opened 2016-02-11
ofThreadChannel : destroying and resetting


elliotwoods:
Hey!

It would be great if there's an option to reset an `ofThreadChannel` i.e. set closed = false

this would be used in cases where you are dynamically joining and setting up local threads
and you need to do the same with your thread channels

since ofThreadChannel doesn't support `=` assignment, and can't be reset, this means we have to store it in a pointer and recreate it each time

also. is there any reason why `.close()` isn't called from the destructor?

arturoc:
yeah i've needed that sometimes too, we can add a reset method, i haven't added it yet because other implementations i've seen didn't had anything similar so i wanted to make sure that there wasn't anything that would bite us after adding it in terms of thread safety... i've used a reset method in some custom implementations that i've done for other projects and i haven't found any problem so i guess it should be ok.

also not sure about calling close on destruction, i think you usually want to have full control over the thread shutdown if you leave it to RAII it's can be problematic if you aren't very careful, for example some members in the class might have been already destroyed when the channel is closed and the thread might try to access them so it's better to explicitly close the channel and wait for the thread on the destructor of your class

elliotwoods:
with destructor:
yes i'm aware that in many situations that's not going to work out and i presume that having destructor call close, users might not be aware that they generally need to call close themselves

if there's a way to reset the threadChannel then you wouldn't need to reinstantiate the channel in most cases so the destructor goes back to being used generally in automatic (unpredictable ways)

fingers crossed on reset!

and this is a great little class.

elliotwoods:
also we need a size(), or something to cope with backlogs
e.g. an optional argument on send(..., size_t maximumQueueSize)

i think size() is pretty basic and can allow the user to deal with it as they like a little
clear() would be nice too. and/or `trim(size_t)`

arturoc:
yes i didn't implement that either because i haven0t seen it in any implementation. right now this is using an std queue internally but it could use some non-locking thread safe queue and in that case most implementations won't have size. even now size is just a hint since even before than the method returned it could be a different size alredy.

about trim the way this is done in other implementations is by specifying a max size and a direction in which the channel drops new frames when it's full, that way you can implement things like drop the oldest or the newest frames.

elliotwoods:
setting the max size sounds great to me too

the issue with size and trim being called externally is that you'd have to lock twice
perhaps 

```c++
enum ofThreadChannel::TrimStrategy {
    RemoveOld,
[...]

PR #5097, opened 2016-05-21
ofThreadChannel - Clarify documentation, add locks and add queue size accessor.

bakercp:


arturoc:
thanks, i don't think the locks make much sense though since right after returning the value and before it might be used anywhere it might have changed already.

there's some discussion about having a size method here: https://github.com/openframeworks/openFrameworks/issues/4809 but we agreed that it's more useful to implement some new methods or a new version of ofThreadChannel that has a size limit and discards elements instead since the size reading is just a hint anyway

bakercp:
Got it!  Missed that discussion ... closing this.

Issue #5096, opened 2016-05-21
Segfault on exit


oxillo:
On MSYS2, segfault happens when exiting with the esc key.

danzeeeman:
Does this happen on all app? I've seen some segfaults on master when
closing apps with multiple threads and when they aren't stopped correctly
on OSX.

Can you document this bug a bit more with stack traces etc and steps to
reproduce?

Thanks!

On Saturday, May 21, 2016, Olivier XILLO <notifications@github.com> wrote:

> On MSYS2, segfault happens when exiting with the esc key.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/openframeworks/openFrameworks/issues/5096>
>


-- 
Sent from Gmail Mobile #autocorrect

oxillo:
Have it on several examples (may be all ?). Commenting the glfwHideWindow call in ofAppGlfwWindow::close() seems to solve the problem. But I need to dig deeper...

oxillo:
I suspect the glfw calbacks to be the reason. 
Disabling them solves the problem.
Will PR that tomorrow after more checks.
 
```C++
void ofAppGLFWWindow::close(){
	if(windowP){
                glfwSetMouseButtonCallback(windowP, nullptr);
                glfwSetCursorPosCallback(windowP, nullptr);
                glfwSetCursorEnterCallback(windowP, nullptr);
                glfwSetKeyCallback(windowP, nullptr);
                glfwSetWindowSizeCallback(windowP, nullptr);
                glfwSetWindowCloseCallback(windowP, nullptr);
                glfwSetScrollCallback(windowP, nullptr);
                glfwSetDropCallback(windowP, nullptr);
		//hide the window before we destroy it stops a flicker on OS X on exit.
		glfwHideWindow(windowP);
		glfwDestroyWindow(windowP);
		windowP = nullptr;
		events().disable();
		bWindowNeedsShowing = true;
	}
}
```

tgfrerer:
I believe that's fixed now that #5098 has been merged.

PR #5098, opened 2016-05-22
Unregister GLFW callbacks when closing GLFWAppWindow

oxillo:
bugfix #5096. The key_cb was trying to get the window being detroyed.
Tested on MSYS2 only.


arturoc:
thanks!

PR #5099, opened 2016-05-24
Feature: Update GLFW to 3.2+

tgfrerer:
# READY FOR TESTING 

+ updates to allow using latest GLFW from original repo
+ nothing changes on the client-side
+ the logic for resolving the unicode codepoint has moved into `ofAppGLFWWindow.cpp`, into a method called `keycodeToUnicode` 

---
Note: if you wanted to test this, you will need to update GLFW using apothecary. To do this, from within your openFrameworks root dir, call:
```
./scripts/apothecary update glfw
```
---
Currently, we're using a customised, patched version of GLFW for reasons historical and practical. 

Chief reason is the internal handling of keycodes and codepoints: In openFrameworks, a unicode codepoint is provided with each key event, while GLFW provides unicode codepoints only when responding to a **character** _- not a key -_ event.

This PR prepares the move to using the latest GLFW directly from @elmindreda's original repository instead of our current patched version, and moves the openFrameworks specific unicode/keypress logic into `ofAppGLFWWindow.cpp` where it can be easier maintained. 

`keycodeToUnicode`'s internal logic is OS-specific, and in this PR contains a Windows, and a Linux-specific solution, the former of which I have tested with an Austrian Keyboard, and Umlauts. The linux codepath is by @arturoc.

---

## Todo: 
- [x] add windows libs for glfw
- [x] add windows msys libs for glfw
- [x] update glfw header files to latest
- [x] add windows codepath for `keysToUnicode`
- [x] add linux codepath for `keysToUnicode`
- [x] add OS X codepath for `keysToUnicode`
- [x] add compiled osx glfw lib
- [x] update apothecary recipe

Merging this would leave one more yak shaved, on the path to Vulkan.


elmindreda:
For the X11 and OS X paths, it might be useful to have a look at the implementations of `_glfwPlatformGetKeyName`.

tgfrerer:
I've added an OS X path, it's probably worth testing this on a Spanish or Chinese or Japanese mac and windows keyboard.

My tests on an Autrian keyboard returned the correct unicode values for Umlauts, and curly brackets (which require combinations of special kezs). Thanks @elmindreda for the hint!

oxillo:
Some feedback on MSYS2 side.
Apothecary recipe and compilation are OK. 
With some minor change in config.msys2.default.mk, of can even use the official MSYS2 packages (3.1.2 actually) using pkgconfig.
I've seen (on the fontsExample) that '~' and '`' keys don't work anymore on my french keyboard. 
Voyels with ¨ or ^ and € symbol don't work (and didn't on previous version).

tgfrerer:
Merci Olivier!

I wonder if the mysys2 packaged glfw is version 3.2, but it's great to hear that using the official package from there could be possible. 

Was it at all possible to enter unicode characters using a french (i assume) keyboard? 
If you `ofLog()` out `key`'s value on `ofApp::keyReleased(int key)` does it say `233` for `é`? (see [here][code] for unicode table, note that `233` is in normal (decimal) and not hex notation).

--

The so-called dead keys '\`' will always be a bit erratic with how we read the keyboard in oF. That's becasue the unicode keypoint is not "finished" with one key, but composed together with the next key. But `ofKeyReleased` will trigger anyway...

If someone wanted to use dead keys -- or have solid unicode aware text-input really -- it's probably better to hook up the special GLFW callback that triggers when the OS registers text input: [`glfwSetCharCallback`][docs].

[code]: http://unicode-table.com/en/00E9/
[docs]: http://www.glfw.org/docs/latest/group__input.html#ga556239421c6a5a243c66fca28da9f742

arturoc:
yes è or similar will never work since it's 2 key strokes so it will report ` and then e but ` alone (or any other dead key) should work and it's working currently in all platforms i think.

oxillo:
Msys2 has GLFW 3.1.2 actually. I've used Apothecary recipe to build GLFW 3.2. The results are the same for both.
I confirm I can enter Unicode characters on my French (you guessed correctly) keyboard. é è à ç are single key characters and come out correctly.
When it comes to 'ï' 'ô', they require two strokes on my keyboard and I agree that it shouldn't work.
Dead keys alone used to work (print '~', '¨' or '^') but don't anymore in this PR. So, from arturo's statement, I think that is not what is expected.
I think oF should offer a way to work with 2 stroke characters (like a new event  hooked to glfwSetCharCallback as Tim suggested)

PR #5100, opened 2016-05-24
Android orientation event

HalfdanJ:
Copy of #4516, recreated after the android refactor so it can get merged.

- Implements `deviceOrientationChanged(ofOrientation newOrientation)` in ofxAndroidApp like in iOS that gets called whenever the orientation on the app is changed
- Implements support for `ofSetOrientation(OF_ORIENTATION_90_RIGHT)` and `ofSetOrientation(OF_ORIENTATION_180)` (these were implemented as their flipped versions before)

Test orientation event with this snippet in a `ofApp.h`: 

```{.cpp}
void deviceOrientationChanged(ofOrientation newOrientation){
    ofLogNotice("ANGLE")<<ofOrientationToDegrees(newOrientation);
}
```


HalfdanJ:
Disregard this PR, merged incorrectly

Issue #5034, opened 2016-04-06
ofImage: move constructor doesn't actually move


Spacechild1:
Hi, I'm currently trying to fix some move constructors and assignments of container classes (mostly adding the noexcept keyword to make them work with the STL containers).
I noticed that ofImage's move constructor actually calls clone(), which only takes const lvalue refs and simply copies the pixels and allocates a new texture.
Move assignements are missing at all.
I don't see any reason why there couldn't be real move semantics involved: If you pass an ofImage that is an rvalue reference, why not simply steal the texture and pixels? Given that both ofPixels and ofTexture themselves already support move semantics this seems like a rather obvious thing to do.
Is this avoided on purpose?
If not, I'll include a commit with revised move constructors and assignments for ofImage in my future pull request.

arturoc:
yes i'm taking a look and it seems that particular constructor was copied from the copy constructor but never adapted to really be a move constructor. if you can fix it that would be great.

btw, it seems you are generating some weird commits directly in the repo, not sure how, probably you are editing things directly in github? would be better if you could work directly in your machine and test things there for a something like this

thanks!

Spacechild1:
Yeah, I tried to do some small things directly in github, but it's cumbersome and confusing. I'll do everything on my desktop.

BrianSipple:
Just stumbled upon this issue now -- it looks like it can be closed after #5101 (cc @arturoc)

arturoc:
thanks!

PR #5101, opened 2016-05-24
implemented move constructor for ofImage

brendan-w:
Currently, `ofImage`'s move constructor performs a clone, as noted in #5034. This PR uses `std::move` to move pixel and texture data from the parent image. The parent is left `clear()`ed after the move.

I tested it with this app, on Fedora with gcc:

```c++
#pragma once
#include "ofMain.h"

	public:

    void setup(){
        a.load("of.png");
        ofSetFrameRate(1);
    };

	void draw(){
        ofBackground(0);
        ofImage b = a;

        ofImage c = std::move(b); // tests move constructor
        //ofImage c; c = std::move(b); // tests move assignment

        b.draw(0, 0); //should always draw black and print "ofGLRenderer: drawing an unallocated texture"
        c.draw(256, 0); //should draw correctly
    };

    private:
        ofImage a;
};
```

### DIFF
--- a/libs/openFrameworks/graphics/ofImage.cpp
+++ b/libs/openFrameworks/graphics/ofImage.cpp
@@ -648,6 +648,30 @@ ofImage_<PixelType>::ofImage_(ofImage_<PixelType>&& mom){
 
 //----------------------------------------------------------
 template<typename PixelType>
+ofImage_<PixelType>& ofImage_<PixelType>::operator=(ofImage_<PixelType>&& mom){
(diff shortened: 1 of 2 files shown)


arturoc:
thanks so much, can you change the move assignment operator to behave the same?

brendan-w:
Apologies. It never had one to begin with, so I've added it. Updated the test app above.

arturoc:
thanks!

PR #5102, opened 2016-05-24
handle negative framerate requests

brendan-w:
Currently, supplying a negative framerate to `ofSetFrameRate()` will produce the following segfault:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7ffacb0 in clock_gettime ()
```

This PR allows negative framerates to make time go backwards.

Just kidding.

Uses existing handler to mark the framerate as "not set".


arturoc:
:) thanks!

PR #5103, opened 2016-05-24
Android orientation event

HalfdanJ:
Copy of #4516  solving merge conflict after android refactor

- Implements `deviceOrientationChanged(ofOrientation newOrientation)` in ofxAndroidApp like in iOS that gets called whenever the orientation on the app is changed
- Implements support for `ofSetOrientation(OF_ORIENTATION_90_RIGHT)` and `ofSetOrientation(OF_ORIENTATION_180)` (these were implemented as their flipped versions before)

Orientation event can be tested with following gist in ofApp.h

```{.cpp}
	void deviceOrientationChanged(ofOrientation newOrientation){
            ofLogNotice("ANGLE")<<ofOrientationToDegrees(newOrientation);
        }
```


arturoc:
looks great, thanks!

PR #5105, opened 2016-05-26
small ofxGui, minimize, maximize fix

jonasfehr:
added a if to abort minimize/maximize function if allready done.
the issue was when calling of maximize() in a loop, the following parameter group was replaced wrongly.


jonasfehr:
Just found issues when trying to maximize/minimize in the gui afterwards, not ready to pull jet

PR #5111, opened 2016-05-30
fix ofLog padding so that it is correctly initialised between dynamically loaded exe's

elliotwoods:
hey!

here's a small fix which makes the static padding variable dll-safe
i.e. when you're dynamically loading a library which itself links openFrameworks

elliot


arturoc:
it would be even safer like:

```cpp
string & ofLog::getPadding() {
	static string * padding = new string;
	return *padding;
}
```

that way the access is correct even on destruction

elliotwoods:
fancy making the edit directly?
i've got a backlog of git work here to setup and it's a bit difficult to get back to that branch right now

arturoc:
ok, will do that

Issue #5043, opened 2016-04-09
Fix : background gradient should use viewport size


elliotwoods:
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofGraphics.cpp#L395

should use ofGetViewportWidth/Height not ofGetWidth/Height

Issue #4946, opened 2016-02-25
Bug? ofThreadChannel can have size of 0 when tryReceive continues


elliotwoods:
Hey!

![image](https://cloud.githubusercontent.com/assets/328294/13312918/a7e80a4c-dbdd-11e5-9d1c-1a2360235b57.png)
I'm getting crashes on this line because queue is empty

I have stable operation when I'm using tryReceive without the timeout, but it crashes on 5% of calls when I use this function instead.
To note, only 1 thread is receiving (so it's not like i'm unlocking elsewhere after i've emptied the queue). Also perhaps note that multiple threads are sending, but that doesn't seem to be unsupported operation.

Perhaps condition.wait_for can just return true for other reasons?
Anyway, the fix here is to check again if the queue is empty before entering the scope with swap

arturoc:
this structure is usually mpsc, multiple producer single consumer and i've really only tested it with spsc. 

can you try changing notify_all for notify_one in send? it seems like the condition would notify everyone, if there's multiple consumers then the queue could get emptied and when the next one tries to swap it crashes because it's been emptied already.

elliotwoods:
to clarify, i'm multiple producer, single consumer
which is why this is confusing (i can't imagine a situation where a lock would be released in another thread apart from when we add to the queue, i.e. queue should never be empty after an unlock from another thread)

elliotwoods:
testing here... if all the producers are turned off then it still crashes
so i presume `wait_for` returns true if the timeout has been hit

arturoc:
that's strange, accoding to the docs wait_for can only return != timeout if it actually didn't timeout or if there was an exception but in that case the exception is propagated if i understand it correctly: http://www.cplusplus.com/reference/condition_variable/condition_variable/wait_for/

elliotwoods:
right, and `cv_status::timeout == 1`, so we need to do `if(cv_status::timeout != condition.wait_for(...`

elliotwoods:
hmm.. wait, that's there already. ignore me! :)

arturoc:
:)

elliotwoods:
ok back to this
having the same problem again
according the the specification, a spurious wake would cause `wait_for` to return `no_timeout` even if no notification was called.

it's common in the specified use of `wait_for` to always ensure that your own condition is met before continuing because of spurious waking

http://www.cplusplus.com/reference/condition_variable/condition_variable/wait_for/

so the only safe implementation would be something like:

```c++
[...]

PR #5112, opened 2016-05-30
fixes to ofBackgroundGradient and ofThreadChannel

elliotwoods:
hey!
2 little fixes

1. ofBackgroundGradient currently uses ofGetWidth rather than ofGetViewportWidth, which means it doesnt' work with viewports.#5043
2. fix spurious wake in ofThreadChannel #4946


arturoc:
thanks!

PR #5118, opened 2016-06-08
added explanation for ofSetDrawBitmapMode options

roymacdonald:
I took this from the OF forum where @arturoc  explained each mode.
https://forum.openframeworks.cc/t/what-is-the-bitmap-drawing-mode-in-openframeworks/22923


arturoc:
thanks!

PR #5119, opened 2016-06-09
Fixes to enable building on Linux (openSUSE LEAP)

highperformancecoder:


arturoc:
where is that macro usually defined? will it allow to use pulse in other distibutions?

highperformancecoder:
In answer to your question, the macro is used within the RtAudio.h header, but seems to be an internal compilation flag of the RtAudio library!!

Also, unfortunately, the version information is a string, so cannot be used for excluding problematic enum labels. Damn people for not taking API evolution seriously!

Short of dumping the entire RtAudio 4.1.2 source code into the openFrameworks repository and building that from scratch on Linux systems, I'm not sure what else to do, as the API doesn't provide ways of selecting valid feature sets.

highperformancecoder:
I'm thinking the answer is to build a small utility c++ program that includes RtAudio.h with __LINUX_PULSE__ defined. If it succeeds, we know we can define that macro in the openFrameworks CFLAGS. This is a strategy often taken in configure scripts.

danoli3 (approved):

Issue #5123, opened 2016-06-12
ofTrueTypeFont copy/assignment issues


Paul92:
If an ofTrueTypeFont object has been initialized using load() and then assigned to another object of the same type, any attempt of drawing using the second object will create a segmentation fault error (even if isLoaded() returns true):

```
ofTrueTypeFont font;
ofTrueTypeFont font2;

void ofApp::setup(){
    font.load("sans-serif", 10);
    font2 = font;
}

void ofApp::draw(){
    font2.drawString("test", 10, 10);
}
```

If in the draw method, the font object is used to draw as opposed to font2, the application runs.

This behavior was observed on Archlinux kernel 4.4.1, make 4.1, gcc 5.3, the last version of openframeworks from the main branch (commit 8e08ba8).

PR #5124, opened 2016-06-13
Bugfix of true type font copying

Paul92:
The problem, described in issue #5123, was caused by the copy/assignment constructors of the ofTrueTypeFont that did not copy all the members to the target.

Since my familiarity with openFrameworks and C++ knowledge are limited, please check if my additions are relevant/do not have any side effects.

The bug was apparently fixed and did not cause any visible side effects. Tested on the small sample of code I've shown in the issue and on few examples (such as graphics/fontShapesExample and graphics/fontsExample).

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -581,8 +581,8 @@ ofTrueTypeFont & ofTrueTypeFont::operator=(const ofTrueTypeFont& mom){
 		ofAddListener(ofxAndroidEvents().reloadGL,this,&ofTrueTypeFont::reloadTextures);
 	}
 #endif
-	settings = mom.settings;
 	bLoadedOk = mom.bLoadedOk;
+
 	charOutlines = mom.charOutlines;
 	charOutlinesNonVFlipped = mom.charOutlinesNonVFlipped;
 	charOutlinesContour = mom.charOutlinesContour;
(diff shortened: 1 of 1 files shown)


bakercp:
If you've tested this and the appveyor/travis tests pass, this seems pretty solid from a code-review standpoint.  Thanks for the contribution.

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
```diff
@@ -620,9 +628,13 @@ ofTrueTypeFont::ofTrueTypeFont(ofTrueTypeFont&& mom)
 	glyphBBox = mom.glyphBBox;
 	letterSpacing = mom.letterSpacing;
 	spaceSize = mom.spaceSize;
+	fontUnitScale = mom.fontUnitScale;
 
 	cps = mom.cps; // properties for each character
+	settings = mom.settings;
+	glyphIndexMap = std::move(mom.glyphIndexMap);
 	texAtlas = mom.texAtlas;
+	stringQuads = mom.stringQuads;
```
arturoc: can you change this to stringQuads = std::move(mom::stringQuads)

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
```diff
@@ -649,11 +660,14 @@ ofTrueTypeFont & ofTrueTypeFont::operator=(ofTrueTypeFont&& mom){
...
 	spaceSize = mom.spaceSize;
+	fontUnitScale = mom.fontUnitScale;
 
 	cps = mom.cps; // properties for each character
-
+	settings = mom.settings;
+	glyphIndexMap = std::move(mom.glyphIndexMap);
 	texAtlas = mom.texAtlas;
-	face = mom.face;
+	stringQuads = mom.stringQuads;
```
arturoc: this to should use move instead of copy

arturoc:
hey, thanks! i've added a couple of comments, you could also just remove those lines since that variable is just a cache and not copying/moving it shouldn't have any effect

Paul92:
Done. Since you said is just a cache, I removed it from all constructors and it seems to work.

arturoc:
thanks!

PR #5128, opened 2016-06-14
Bug Fix for "ofPixels mirrorTo horizontal"

IanByun:
When mirrorTo-ing ofPixels horizontally, line 1179, which was
int pixelb = i;
previously, should be 
int pixelb = j*width + (width - 1 - i);.

Otherwise, results in very irregular behaviours as below: (grayscale images that look like strips on the left are mirrored versions of top-left-most image using previous ofPixels mirrorTo horizontally)
default

### DIFF
--- a/libs/openFrameworks/graphics/ofPixels.cpp
+++ b/libs/openFrameworks/graphics/ofPixels.cpp
@@ -1176,7 +1176,7 @@ void ofPixels_<PixelType>::mirrorTo(ofPixels_<PixelType> & dst, bool vertically,
 		int hToDo = height;
 		for (int i = 0; i < wToDo; i++){
 			for (int j = 0; j < hToDo; j++){
-				int pixelb = i;
+				int pixelb = j*width + (width - 1 - i);
 				int pixela = j*width + i;
 				for (int k = 0; k < bytesPerPixel; k++){
 					dst[pixela*bytesPerPixel + k] = pixels[pixelb*bytesPerPixel + k];


IanVative:
Also, I did not edited the code right now,
but isn't
int bytesPerPixel = getNumChannels();
on line 1162, type specific?

What I mean is, it will work when PixelType is uchar that is 8bit(1byte), because then number of channels will be the byte size per pixel,
but when if Pixel Type is say ushort that is 16bit(2byte), and the this pixel has 3 channels, real bytes per pixel will be 2byte*3channel = 6byte, not 3 which is the result calculated with the above, previous code.

I think
int bytesPerPixel = getNumChannels();
on line 1162 should be
int bytesPerPixel = getNumChannels() * sizeof(PixelType);

Please consider.

IanVative:
Also, I did not edited the code right now but,
int bytesPerPixel = getNumChannels();
on line 1162 is very misleading.

If I understand this right,
real bytes per pixel are equal to the number of channels if and only if the pixel is 8bit or 1byte.
However, if you review the following codes, what bytesPerPixel really mean in mirrorTo function is just the number of channels.

I urge those in charge to change the name of the variable, "bytesPerPixel", to such as "numChannels".

Please consider, and correct me if I got this wrong.

arturoc:
thanks, yes, it seems the functionality is correct is just the naming that is misleading. do you want to include the change in this PR too?

IanVative:
Yes. That would be nice.
But I am new to this GitHub and don't know how to include post-changes to the pull re quest.
Can you revise the PR on your side?

arturoc:
you just need to push any new commit and they'll appear right away in the PR

PR #5131, opened 2016-06-15
Bugfix of true type font size for empty string

Paul92:
I noticed that stringWidth and stringHeight methods produce weird results when called with an empty string parameter. Simple test program:

```
ofTrueTypeFont font;

void ofApp::setup(){
    font.load("sans-serif", 10);
}

void ofApp::draw(){
    std::cout << font.stringWidth("text") << " " << font.stringWidth("") << " " << font.stringWidth(" ") << "       ";
    std::cout << font.stringHeight("text") << " " << font.stringHeight("") << " " << font.stringHeight(" ") << std::endl;
}
```

(The result should be 27 0 0   9 0 0).

The problem was caused by the fact that an empty string produces a mesh with no vertices, so the corners of the bounding box are not changed (the defaults are limits of float). Solved by a simple check.

Also, the std::numeric_limits<float>::min() is not the minimum value that can be represented with a float, but rather the smallest positive value (equal to 1E-37 or less, according to the standard). The minimum value is -std::numeric_limits<float>::max().

Besides this, I've done a bit of refactoring to keep consistency (I didn't really like the fact that, while computing what is basically the limits of the bounding box, the minimum values were the position of the rectangle and the maximum values were separate variables).

Note that "  " is considered the same as "". I'd say it's fine this way (it's nothing drawn, so it has no dimension in either case) but it might be a good moment to think a bit whether this is really wanted or not (a space is still a character, not emptiness).

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -1205,8 +1205,8 @@ ofRectangle ofTrueTypeFont::getStringBoundingBox(const std::string& c, float x, 
 	    return ofRectangle(x,y,0,0);
 
 	ofRectangle bb(std::numeric_limits<float>::max(),std::numeric_limits<float>::max(),0,0);
-	float maxX = std::numeric_limits<float>::min();
-	float maxY = std::numeric_limits<float>::min();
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

PR #5133, opened 2016-06-17
Bugfix of true type font set space size

Paul92:
Ok, so after a bit of digging I found out that:

- long time ago, the spaceSize was initialized with 1 and space was treated as a separate case when drawing a character.
- for some reasons I don't fully understand, someone decided to change that. spaceSize was now initialized with 0 and there was a check (if space size is set, draw the standard space, otherwise draw the scaled one).
- after some time, at a refactor, that case was gone.

So, I reinserted that check and initialized spaceSize with 1 (thus, there is just one case, not 2).

A bit off-topic, there was an if with some weird indentation. I assumed there should have been an else.

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -1126,8 +1126,13 @@ void ofTrueTypeFont::iterateString(const string & str, float x, float y, bool vF
 				if(prevC>0){
 					pos.x += getKerning(c,prevC);// * directionX;
 				}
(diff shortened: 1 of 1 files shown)


Paul92:
The last commit is a bit unrelated to the issue but I needed the fix for my work and I tried to avoid a merge conflict. The problem was with updating the x value when drawing from right (you need to update then draw, rather than draw then update, due to the letters' anchor).

Before:
http://imgur.com/IJkSLGM

After:
http://imgur.com/i7hSENB

arturoc:
thanks!

Issue #4811, opened 2016-02-11
ofVideoGrabber with bUseTexture = false does not update


antimodular:
i'm on OS X 10.10 with OF 0.9.0 and Xcode 6.4

will trying to make a headless app via ofAppNoWindow i noticed when setting up ofVideoGrabber like so: 

```
vidGrabber.setup(camWidth,camHeight,false);
```

vidGrabber.isFrameNew() always returns 0.
looking in to void ofVideoGrabber::update() need bUseTexture to be true otherwise nothing happens.

is it even possible to use the camera image with bUseTexture = false?
thx

arturoc:
this sounds like a bug, even if there's no texture, the pixels are updated and the grabber should return true on isNewFrame. sounds like an easy fix, just moving out of that if wheerever the flag is set to true

antimodular:
in the hopes of helping here is a bit more info.

while in headless mode:
if vidGrabber.setup(camWidth,camHeight,**true**);
then app crashes here in line tex.push_back(ofTexture());
here 
````
    if( grabber->isInitialized() && bUseTexture ){
		if(!grabber->getTexturePtr()){
			for(int i=0;i<grabber->getPixels().getNumPlanes();i++){
				ofPixels plane = grabber->getPixels().getPlane(i);
				tex.push_back(ofTexture());
				tex[i].allocate(plane);
			}
		}
	}
````

while NOT in headless mode
if vidGrabber.setup(camWidth,camHeight,**false**);
everything works well and i can pass the ofVideoGrabber pixel to a texture

arturoc:
yes, if you are in headless mode you can't use opengl at all and any call to an opengl command, like creating a texture will crash the app since there's no opengl context.

you should use setUseTexture(false) in that case

antimodular:
i did try that with setUseTexture(false) before and after setup.
but that still causes a crash unless is i do setup(camWidth,camHeight,false)

here what crashes in headless mode
````
        vidGrabber.setUseTexture(false);
        vidGrabber.setDeviceID(deviceID);
        vidGrabber.setup(camWidth,camHeight); //,false);
        vidGrabber.setUseTexture(false);
````

arturoc:
yes that's true you need to call setup(w,h,false) otherwise it'll try to create a texture in setup and crash too

bakercp:
@antimodular is this fixed now with #5134?

stephanschulz:
yes.
thanks for your efforts.

[...]

PR #5134, opened 2016-06-18
ofVideoGrabber -- fix default bUseTexture behavior.

bakercp:
Currently if one calls

```cpp
ofVideoGrabber grabber;
grabber.setUseTexture(false);
// grabber.isUsingTexture() returns false.
```

then 
```cpp
grabber.setup(640, 480);
// grabber.isUsingTexture() returns true.
```
Rather than `setup()` assuming `true`, it should use the value that was (already) set by the user.  In order to avoid breaking existing setups, the default value of bUseTexture is set to true in the constructor.

This is related to #4811


arturoc:
thanks!

PR #5142, opened 2016-06-30
More stable easy cam

arturoc:
Before the relative y axis was used for rotating which made the camera movement really unstable and you always ended up with a tilted camera

3d softwares like blender or cinema 4d seem to just fix the y axis for the arcball movement to the up axis which makes the interaction much easier to control

There's still a method to enable the old behaviour

This also adds a method to disable the mouse inertia which is not very useful and somehow anoying if you want precise movements. I would even vote for disabling it by default but i've left it enabled by now


roymacdonald:
I kinda rewrote easycam completely some time ago and added the intertia thing, although more than following the 3d cad software standards I was aiming to not break the already implemented behaviour.
Yes, after long time using ofEasyCam, I can say that inertia is kind of annoying, I wouldn't mind if it got removed completely.

As for the axis thing I agree with you. Actually I like much more @elliotwoods ofxGrabCam. It is so much more easier to move around large scenes and rotations are much more intuitive.
I vote for removing/disabling inertia.

arturoc:
ok, i've disabled it by default by now. if people misses it we can enable it back later but i also think it's much easier to control without. going to merge it then

PR #5145, opened 2016-07-01
add ofBoundingBox

edap:
I've added a struct that contains the min and max vertices of a `of3dPrimitive`.
it can be used like this. Default values are initialized to (0.0, 0.0, 0.0).
When the primitive move, the bounding box does not get updated.

``` cpp
ofBoxPrimitive geometry;
geometry.computeBoundingBox();
cout << geometry.getBoundingBox().min.x << endl;
cout << geometry.getBoundingBox().max.y << endl;
```

### DIFF
--- a/libs/openFrameworks/3d/of3dPrimitives.cpp
+++ b/libs/openFrameworks/3d/of3dPrimitives.cpp
@@ -211,6 +211,20 @@ void of3dPrimitive::draw() const{
 	draw(OF_MESH_FILL);
 }
 
+ofBoundingBox of3dPrimitive::getBoundingBox() const{
+    ofBoundingBox boundingBox;
+    const vector<glm::vec3>& vertices  = getMesh().getVertices();
+    for(auto v : vertices){
+        if (v.x < boundingBox.min.x) boundingBox.min.x = v.x;
+        if (v.y < boundingBox.min.y) boundingBox.min.y = v.y;
+        if (v.z < boundingBox.min.z) boundingBox.min.z = v.z;
+         if (v.x > boundingBox.max.x) boundingBox.max.x = v.x;
+        if (v.y > boundingBox.max.y) boundingBox.max.y = v.y;
+        if (v.z > boundingBox.max.z) boundingBox.max.z = v.z;
+    }
+    return boundingBox;
+}
+
 //--------------------------------------------------------------
 void of3dPrimitive::drawNormals(float length, bool bFaceNormals) const{
     ofNode::transformGL(ofGetCurrentRenderer().get());
--- a/libs/openFrameworks/3d/of3dPrimitives.h
+++ b/libs/openFrameworks/3d/of3dPrimitives.h
@@ -60,6 +60,7 @@ public:
 
     void setUseVbo(bool useVbo);
     bool isUsingVbo() const;
+    ofBoundingBox getBoundingBox() const;
 protected:
 
     // useful when creating a new model, since it uses normalized tex coords //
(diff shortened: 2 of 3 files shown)


arturoc:
this sounds useful but it would be better if getBoundingBox did compute and returned the bounding box without storing it internally. computing it is a relatively expensive operation but i think it's better for the user to keep those values out of the geometry.

Also to declare a struct in c++ you don't need a typedef you can just do:

```
struct ofBoundingBox{
    ofVec3f min;
    ofVec3f max;
}
```

not sure though about the members of ofBoundingBox or if we could have a more generic class like ofBox, i remember there's a PR for a class similar to ofRectangle in 3d that might be useful for this

edap:
do you mean `ofBoxPrimitive`? or do you mean to create a new class ofBoundingBox that takes care of the things related to the bounding box?

edap:
I've delete the part where the variable was stored. Now the `getBoundingBox` method is returning the min and the max vertices.

```cpp
ofBoxPrimitive geometry;
cout << geometry.getBoundingBox().min.x << endl;
```

edap:
Is there anything that I can do to improve this pull request? Looks like travis failed because it can not retrieve the repository at that time.

underdoeg:
Looks like your code is ok. Only the ios travis job did not compile because of ```fatal: write error: No space left on device```

arturoc:
this should be using glm instead of ofVec3f and also not sure what's the best representation for a bounding box. we should check if for example center, size is not more useful and in any case have the possibility to check that through methods.

perhaps until it's clear what's more useful the member variables can be private, have a constructor for min max and have getter/setter methods for min max center and size

edap:
Sure, I will move to glm. Regarding the 2 other points:
1) "we should check if for example center, size is not more useful". Do you mean to have  `center` as a vector, and `sizes` {x:{min, max}, y:{min,max}, z:{min,max}} relative to the center? Or do you mean to return a ofBoxPrimitive? like in threejs https://github.com/mrdoob/three.js/blob/master/src/extras/helpers/BoundingBoxHelper.js
2)I was thinking that we did not wanted to store any variable internally, as in the second comment in this thread. Did I understand wrong? to me sounds reasonable have setters and getters.

bakercp:
Just one point on naming -- we currently use "bounding box" to refer to 2D boxes in `ofTrueTypeFont`, `ofPolyline` and `ofBitmapFont`.  These are represented with an `ofRectangle`.  I'd suggest that we make the difference clear (2d vs 3d?) or transparent ... That said, those 2d methods should probably be called `getBoundingRectangle` and ofRectangle should be based on `glm::vec2`, rather than `glm::vec3` ... which is kind of an artifact of previously being based on ofPoint, which was a typedef for `ofVec3f`.  

Anyway, just trying to bring clarity between a "box" 3d and a "rectangle" 2d.

edap:
- After @bakercp , as type name I think `ofBoundingBox` is a clear one.
- I have switched to glm::vec3.
- @arturoc  regarding the consideration about save the bbox as a member variable or not, it is not clear to me how to process, as in the first comment you suggest to not store it internally but in the next one you are talking about private member variable. This PR is 2 years old and it is difficult to remember how was the final decision.
- regarding the center, all of3primtives have the vertices relative to the center, the min and max position are relative to the center of the mesh.

danoli3:
We really needed to merge this stuff...

danoli3 (approved):
Clean.

danoli3:
Thank you @edap

PR #5146, opened 2016-07-01
cosmetic fix in simpleShaderInteraction example

mfandl:
while reading the shader examples, I noticed this line in examples/shader/03_simpleShaderInteraction/src/ofApp.cpp:

`shader.setUniform4fv("mouseColor", &mouseColor[0]);`

`mouseColor` itself is a pointer to the beginning of the array and I since I got a bit confused about why it's there, here is a PR where i changed `&mouseColor[0]` to `mouseColor`

hope I am correct about this


arturoc:
thanks!

PR #5147, opened 2016-07-04
ofGLUtils: ofEnable/DisableGLDebugLog: Enables OpenGL debug log

arturoc:
Shows internal GL errors on the console. The log depends on the specific
driver but is of great help catching GL errors


tgfrerer:
You might have to check against either the OpenGL version being > 4.3 (which is when the debug extension was added to core) or the debug extension being available: `ARB_debug_output` to be sure that debug output is possible. <https://www.opengl.org/wiki/Debug_Output>

PR #5162, opened 2016-07-15
Fix mouseScrolled()'s initial inertia behavior

saki7:
See the commit comment for more details and see how to reproduce.


arturoc:
thanks!

PR #5164, opened 2016-07-15
fix Tess crash on teardown

tgfrerer:
- in rare cases Tess has the ability to cause a crash on app teardown,
  because the desctructor attempts to free an internal tesselator object
  repeatedly.
  
  This patch addresses this by checking for the pointer holding the
  Tesselator to be valid before freeing the object. After free, it nulls
  the pointer so that an object already freed cannot be freed again.


arturoc:
this seems like it would be masking the real problem rather than fixing it. that variable shouldn't be accessed at all once the tessellator has been destroyed so the problem seems to be that the object is being destroyed twice which this won't solve but rather hide so it could be even more problematic. can you post a stack trace of the problem to see where it might be coming from?

Issue #5152, opened 2016-07-06
ofFastEvent broken in master


c-mendoza:
It seems that ofFastEvent is not working with the new event code.

```

template<typename T>

class ofFastEvent: public ofEvent<T,of::priv::NoopMutex>{

public:

	inline bool notify(const void* sender, T & param){

		if(ofFastEvent::enabled){

			for(auto & f: ofFastEvent::functions){

                if(f->notify(sender,param)){

					return true;
                }

			}
		}
		return false;
	}
};
```

`enabled` is no longer a class static, it is now part of the "self" ofData member. Same goes with `ofFastEvent::functions`

Something like this seems to let it compile, but I haven't fully tested it:

```
template<typename T>

class ofFastEvent: public ofEvent<T,of::priv::NoopMutex>{

public:
	inline bool notify(const void* sender, T & param){
		if(this->isEnabled()){
			for(auto & f: ofFastEvent<T>::self->functions){
				if(f->notify(sender, param)){
					return true;
				}
			}

		}
		return false;
	}
};
```

PR #5166, opened 2016-07-18
Fix for ofFastEvent

c-mendoza:
ofFastEvent was broken when the events system was updated. Fixes
https://github.com/openframeworks/openFrameworks/issues/5152


arturoc:
thanks!

PR #5168, opened 2016-07-21
Project changes for iOS/tvOS for new GLM changes

danoli3:
This fixes missing files and new glm headers.

Current problem with new refactor still is the following error. @arturoc 

> openFrameworks/addons/ofxiOS/src/app/ofAppiOSWindow.h:165:18: Virtual function 'getWindowPosition' has a different return type ('ofPoint' (aka 'ofVec3f')) than the function it overrides (which has return type 'glm::vec2' (aka 'tvec2<float, highp>'))

https://github.com/openframeworks/openFrameworks/commit/b6e5fab9ebcee4dcfe275d73c505f787264107d7


arturoc:
i was getting a not enough space error on the PR tests and though ios was ok. the errors on the window class are easily solved by changing ofPoint -> glm::vec2, can you take a look? otherwise i'll try to fix it tomorrow.

thanks!

danoli3:
Are we keeping the glm::vec2 datatype as the base type of the window or
using the alias?

Currently the forward declaration of ofVec2 is in the objective-c class
(which was aliased to ofPoint internally before), so it worked.

--- ah so basically I think changing the top class declaration to keep
ofPoint will solve this else I'm not sure forward declaring glm::vec2 will
work...


I'll give it a go tomorrow

PR #5177, opened 2016-07-28
Quick fixes

prisonerjohn:
* Commenting out call to `glLineWidth()` in programmable renderer. Calling it causes a GL error, which makes it a pain to debug other GL errors.
* `isnan()` appears to be defined for Visual Studio, implementing our own causes conflicts with external libs.


arturoc:
thought i had merged this to the glm branch... weird.

prisonerjohn:
Yeah me too, maybe you only merged it to your local `refactor-glm`...

arturoc:
yeah that might be it i then probably merged that branch from my computer instead of from github

Issue #3062, opened 2014-07-01
ofxKinect compilation broken on Codeblocks MinGW


pierrep:
I had to add changes some #defines in the code below in order to avoid a redefinition error. Seems like the MinGW constant has changed.

Old code:

```
// MinGW defines _SSIZE_T_DEFINED in sys/types.h when it defines ssize_t to be a long.
// Redefining it causes an error.
// MSVC does not define this.

#ifndef _SSIZE_T_DEFINED
#define _SSIZE_T_DEFINED
typedef long ssize_t;
#endif // _SSIZE_T_DEFINED

```

to this:

```
// MinGW defines _SSIZE_T_ in sys/types.h when it defines ssize_t to be a long.
// Redefining it causes an error.
// MSVC does not define this.


#ifndef _SSIZE_T_
#define _SSIZE_T_
typedef long ssize_t;
#endif // _SSIZE_T_
```

armadillu:
I'm having this back on when compiling a VS2015 project with ofxKinect on 64Bit.

here: https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxKinect/libs/libfreenect/platform/windows/unistd.h#L36

changing to 
```
#ifdef _WIN64
typedef __int64 ssize_t;
#else
typedef long ssize_t;
#endif
```
allows me to compile in both 32 and 64 bit

ofTheo:
Thanks for this fix!
Could you send a PR?

PR #5179, opened 2016-07-29
ofxKinect ssize_t typdef should be different for x64 and x86 on windows

armadillu:
As requested in #3062; I had trouble compiling ofxKinect on Windows x64 and this seems to fix.

### DIFF
--- a/addons/ofxKinect/libs/libfreenect/platform/windows/unistd.h
+++ b/addons/ofxKinect/libs/libfreenect/platform/windows/unistd.h
@@ -33,5 +33,9 @@
 // MSVC does not define this.
 #ifndef _SSIZE_T_
 #define _SSIZE_T_
+#ifdef _WIN64
+typedef __int64 ssize_t;
+#else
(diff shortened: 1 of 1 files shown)


arturoc:
perhaps we should check if this is not already fixed in libfreenect

armadillu:
seems like it's not handled?

https://github.com/OpenKinect/libfreenect/blob/master/platform/windows/unistd.h#L37

ofTheo:
@armadillu thanks for checking.  
we can merge this in, just so its not broken in OF. 

Though it would be good to submit a PR to libfreenect too if you are up for it.  
They generally merge simple stuff like this fairly quickly.  

I'm going to pull in the latest libfreenect for the next release and probably submit some PRs, so getting it fixed upstream would be good. 

Cheers!
Theo

armadillu:
Yep it makes sense; done!
https://github.com/OpenKinect/libfreenect/issues/476

PR #5180, opened 2016-07-30
ofColor tidying up

2bbb:
* change argument like `const float&` to `float`
* add left side scalar operator. (i.e. allow `2.0f * color;`)


arturoc:
thanks!

PR #5188, opened 2016-08-06
update ofVec3f.h new function "angleSigned"

sterlingcrispin:
new function that returns -180 to 180 degree value between vectors, also clarified documentation that the existing function is unsigned between 0 180


arturoc:
ofVec*f are deprecated. vector math is now done using glm, so better not keep adding stuff to the old classes

sterlingcrispin:
Oh okay trippy sounds good

arturoc:
thanks anyway, i'm going to close it, if i remember well glm already has something like this

Issue #5189, opened 2016-08-08
bug: Gstreamer pipeline is not closed when starting it was unsuccessful


hschwane:
Hi,

i am working on an open frameworks app (0.9.3) in Ubuntu which is supposed to display a presentation of different pictures and videos.  After running my open frameworks app for a few hours it crashed with a "to many open files error" I started investigating and found out there are tons of unix sockets opened and never closed again. 

How to reproduce:
Create a of app which tries to playback video using gstreamer. Then tell it to load a video that does not exist (or where the app has no reading rights). Everey time the app tries to load the video, a new unix socket will be created but never closed. Check using `lsof -r 1 -p xxxx | grep socket` where xxxx is your apps PID.

I looked into the code of ofGstVideoPlayer.cpp and think the error is in the load function:
(starting at line 205)
```
if(isInitialized()){
		gst_element_set_state (videoUtils.getPipeline(), GST_STATE_READY);
		if(!bIsStream){
			gst_element_get_state (videoUtils.getPipeline(), NULL, NULL, -1);
		}
		internalPixelFormat = OF_PIXELS_NATIVE;
		bIsAllocated = false;
		videoUtils.reallocateOnNextFrame();
		g_object_set(G_OBJECT(videoUtils.getPipeline()), "uri", name.c_str(), (void*)NULL);
		gst_element_set_state (videoUtils.getPipeline(), GST_STATE_PAUSED);
		if(!bIsStream){
			gst_element_get_state (videoUtils.getPipeline(), NULL, NULL, -1);
			return allocate();
		}else{
			return true;
		}
	}else{
		ofGstUtils::startGstMainLoop();
		return createPipeline(name) &&
				videoUtils.startPipeline() &&
				(bIsStream || allocate());
	}
```
in the "else"-part, a new piepeline is created, but if `startPipeline()` fails, `bLoaded` will remain false, so the created pipeline is never released.

I'm working on a fix, but it is kinda difficult because the pipeline itself belongs to the ofGstVideoUtils class and calling close wont do the trick (since loaded is false). Also, what else needs to be done at this point to properly clean up everything thats already loaded?

hschwane:
i added a close pipeline function which is called in the load function:

```
void ofGstUtils::closePipeline() {

	if(!bIsMovieDone && !bPaused && !isStream){
		std::unique_lock<std::mutex> lck(eosMutex);
		closing = true;
		gst_element_send_event(gstPipeline,gst_event_new_eos());
[...]

PR #5191, opened 2016-08-10
close gstreamer pipline after load() was not successful 

hschwane:
When trying to load a video that does not exist using the gstreamer video player on linux, a unix socket is created and never closed which can lead to crashing the app. This should fix it. 
See #5189 for details.

### DIFF
--- a/libs/openFrameworks/video/ofGstUtils.cpp
+++ b/libs/openFrameworks/video/ofGstUtils.cpp
@@ -44,9 +44,12 @@ GMainLoop * ofGstUtils::getGstMainLoop(){
 }
 
 void ofGstUtils::quitGstMainLoop(){
-	mainLoop->quit();
-	delete mainLoop;
-	mainLoop=0;
+	if(mainLoop)
+	{
+		mainLoop->quit();
+		delete mainLoop;
+		mainLoop=0;
+	}
 }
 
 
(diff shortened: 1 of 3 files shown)


arturoc:
Hi thanks for the fix. I think it's almost correct but there's an issue. You are shutting down the gst main loop when closing the pipeline but if there's another video running closing the main loop will stop that other video(s) too since the main loop is common to all the present pipelines.

I think you just need to not quit the main loop when closing the pipeline, can you fix that?

thanks

hschwane:
I don't think that will work, I remember getting a segmentation fault inside the main loop after closing the pipeline, thats why I closed the main loop in the first place.
I can't remember where exactly the segfault happened, I will post the stacktrace later, if I have time to try it once more.

arturoc:
there must be something else that you need to close first then we can't close the main loop cause there might be other instances running, we never do so when closing a video after loading correctly so surely there must be a way to do so when it's not been loaded

hschwane:
I looked into it again. The problem is that after closing the pipeline the main loop called ´gst_element_set_state(GST_ELEMENT(gstPipeline), GST_STATE_NULL);´ in line 630 (in the gstHandleMessage function). I simply changed it to:
´´´
if(gstPipeline)
{
	gst_element_set_state(GST_ELEMENT(gstPipeline), GST_STATE_NULL);
}

´´´
Now it sems to work fine.
Is this ok for you? I could not find a way to prevent the GST_MESSAGE_ERROR from being send after closing the pipeline, which would probably be a more beautiful solution...

PR #5195, opened 2016-08-12
bugfix for static ofFile and ofDirectory utils adding the data path even if disabled. Closes #5187

### LINKED ISSUE #5187: ofFile::open() ALWAYS calls ofToDataPath()
This makes many (most?) ofFile functions fail if bRelativeToData is false, such as doesFileExist(), copyFromTo(), removeFile() etc.

ofTheo:
This PR fixes the issue of ofFile and ofDirectory always adding ofToDataPath to the supplied path even if the bRelativeToData is false in the static ofFile/ofDirectory calls.  

Below is some test code which was failing in multiple places in the current master, which all passes with this PR.  

Nothing in this PR should affect typical usage or where bRelativeToData is set to true. 
Making these changes I did notice one potential existing bug in ofFile::create which I will point out in the diff. 

Note: I considered using a different class/method to replace the ofFile and ofDirectory calls, but felt like passing through the bRelativeToData was a safer way to go ( and with less changes )  than trying to rewrite a lot of the functionality.  It also has the added bonus of being able to open an ofFile/ofDirectory object and pass in a bool to disable the ofToDataPath call. 

Here is the test code: 

```
    cout << " does file test1  exist (should be 1) " << ofFile::doesFileExist("../../../test1.txt", false) << endl;
    cout << " does file false1 exist (should be 1) " << ofFile::doesFileExist("false1.txt", true) << endl;
    
    ofFile::moveFromTo("../../../test1.txt", "test1.txt", false, false);
    ofFile::moveFromTo("../../../test2.txt", "test2.txt", false, false);
    
    cout << " does file test1  exist (should be 1) " << ofFile::doesFileExist("test1.txt", false) << endl;
    cout << " does file test2  exist (should be 1) " << ofFile::doesFileExist("test2.txt", false) << endl;

    ofFile::copyFromTo("test1.txt", ofToDataPath("test1.txt"), false, false);
    ofFile::copyFromTo("test2.txt", ofToDataPath("test2.txt"), false, false);
    
    cout << " does file test1  exist (should be 1) " << ofFile::doesFileExist("test1.txt", true) << endl;
    cout << " does file test2  exist (should be 1) " << ofFile::doesFileExist("test2.txt", true) << endl;

    ofDirectory dir;
    dir.open("", true);
    dir.listDir();
    for(int d = 0; d < dir.size(); d++){
        cout << " files in data now are " << dir.getPath(d) << endl;
    }
    
    ofFile::removeFile("test1.txt", false);
    ofFile::removeFile("test2.txt", false);

    cout << " does file test1  exist (should be 0) " << ofFile::doesFileExist("test1.txt", false) << endl;
    cout << " does file test2  exist (should be 0) " << ofFile::doesFileExist("test2.txt", false) << endl;
```

with the PR this results in:

```
 does file test1  exist (should be 1) 1
 does file false1 exist (should be 1) 1
 does file test1  exist (should be 1) 1
 does file test2  exist (should be 1) 1
 does file test1  exist (should be 1) 1
 does file test2  exist (should be 1) 1
 files in data now are false1.txt
 files in data now are test1.txt
 files in data now are test2.txt
 does file test1  exist (should be 0) 0
 does file test2  exist (should be 0) 0
```

without the PR it results in:

```
 does file test1  exist (should be 1) 0
 does file false1 exist (should be 1) 1
[ error ] ofFile: moveTo(): source file does not exist
[ error ] ofFile: moveTo(): source file does not exist
 does file test1  exist (should be 1) 0
 does file test2  exist (should be 1) 0
[ error ] ofFile: copyTo(): source file ../../../data/test1.txt does not exist
[ error ] ofFile: copyTo(): source file ../../../data/test2.txt does not exist
 does file test1  exist (should be 1) 0
 does file test2  exist (should be 1) 0
 files in data now are false1.txt
[ error ] ofFile: remove(): file does not exist
[ error ] ofFile: remove(): file does not exist
 does file test1  exist (should be 0) 0
 does file test2  exist (should be 0) 0
```

### DIFF
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -353,10 +353,10 @@ ofFile::ofFile()
 ,binary(true){
 }
 
-ofFile::ofFile(const std::filesystem::path & path, Mode mode, bool binary)
+ofFile::ofFile(const std::filesystem::path & path, Mode mode, bool binary, bool bRelativeToData)
 :mode(mode)
 ,binary(true){
-	open(path, mode, binary);
+	open(path, mode, binary, bRelativeToData);
 }
 
 //-------------------------------------------------------------------------------------------------------------
@@ -435,9 +435,13 @@ bool ofFile::openStream(Mode _mode, bool _binary){
 }
 
 //------------------------------------------------------------------------------------------------------------
-bool ofFile::open(const std::filesystem::path & _path, Mode _mode, bool binary){
+bool ofFile::open(const std::filesystem::path & _path, Mode _mode, bool binary, bool bRelativeToData){
 	close();
-	myFile = std::filesystem::path(ofToDataPath(_path.string()));
+	if( bRelativeToData ){
+		myFile = std::filesystem::path(ofToDataPath(_path.string()));
+	}else{
+		myFile = std::filesystem::path(_path.string());
+	}
 	return openStream(_mode, binary);
 }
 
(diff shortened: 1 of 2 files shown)


ofTheo:
going to merge this in unless there are objections.

arturoc:
I would avoid having so many parameters, mostly booleans and with default values. it's really hard to understand what's going on and it's very error prone.

not sure what would be a better solution, perhaps a settings object or some function or class to specify paths that explicitly says if it's realtive to data or cwd.

something like?

ofFile(ofAsset(path))

would indicate that it's a path relative to data although we currently default to that so it should be the other way arround perhaps

ofFile(std::path(somepath))

means relative to cwd

arturoc:
perhaps even simpler:

```
file.openFromCWD(path)
```

ofTheo:
yeah - I like openFromCWD. 
was going to do something like that, but it seemed it would involve changing the internal code a bit more.  

I think it does make more sense though and I know we're trying to get out of the habit of having a million default arguments in the core :)

ofTheo:
okay closing this one - going to redo it with the openFromCWD approach. good now that the std::filepath changes are in.

Issue #5045, opened 2016-04-10
ofPixels, problem to allocate huge size of pixels.


anagma:
Hello,

I found a bug in bytesFromPixelFormat.

```
template<typename PixelType>
static int bytesFromPixelFormat(int w, int h, ofPixelFormat format){
	return w*h*pixelBitsFromPixelFormat<PixelType>(format)/8;
}
```

this code should be 

```
template<typename PixelType>
static int bytesFromPixelFormat(int w, int h, ofPixelFormat format){
	return w*h*(pixelBitsFromPixelFormat<PixelType>(format)/8);
}

```

e.g.
to try to allocate 10000w * 10000h, RGBA pixels, 
the order of calculation will be
10,000 * 10,000 = 100,000,000
100,000,000 * 32 = 3,200,000,000 <- overflow!!!

arturoc:
mmh, the max int in 32bit systems is 4294967296 so that won't overflow
but can you send a PR anyway? it could overflow in 32bits systems with
bigger images

2bbb:
isn't max of signed int 2147483647?

arturoc:
oh yeah sorry, i was calculating for uint

elliotwoods:
also shouln't we just use `size_t` ?

arturoc:
yes we should just use size_t

armadillu:
prepared a PR https://github.com/openframeworks/openFrameworks/pull/5197

armadillu:
actually let me prepare another one...

armadillu:
Ok I have something ready here https://github.com/armadillu/openFrameworks/commit/608274233cdb8ae6e9fa530e23f507862c513f25 - I manually walked through all the ints in ofPixel.cpp and ofPixel.h  and decided which one needed a bigger range. Basically anything that stores compound sizes (w x h) became size_t.

There is one issue with the []operator though; I seem to be forced to explicitly state the type:

```
ofPixels pix;
pix[0]; //compiler complains "use of overloaded operator '[]' is ambiguous (with operand types 'ofPixels_<char>' and 'int')"
pix[(size_t)0]; //compiler is happy.
```

You can see I had to cast here; https://github.com/armadillu/openFrameworks/commit/608274233cdb8ae6e9fa530e23f507862c513f25#diff-0b9c2d953cf54437297a3c630cb238ffL896

That's a bit annoying; any ideas?

bakercp:
[...]

PR #5197, opened 2016-08-12
replace int for size_t to avoid crashes when allocating very large ofPixels

armadillu:
as seen in  https://github.com/openframeworks/openFrameworks/issues/5045 trying to load very large images (12000 x 12000 for example) leads to an int overflow, issuing a new() on a negative number.

As requested, replaced int with size_t in all the relevant places I could find.

### DIFF
--- a/libs/openFrameworks/graphics/ofPixels.cpp
+++ b/libs/openFrameworks/graphics/ofPixels.cpp
@@ -55,8 +55,8 @@ static int pixelBitsFromPixelFormat(ofPixelFormat format){
 }
 
 template<typename PixelType>
-static int bytesFromPixelFormat(int w, int h, ofPixelFormat format){
-	return w*h*pixelBitsFromPixelFormat<PixelType>(format)/8;
+static size_t bytesFromPixelFormat(size_t w, size_t h, ofPixelFormat format){
+	return w * h * pixelBitsFromPixelFormat<PixelType>(format)/8;
 }
 
 static int channelsFromPixelFormat(ofPixelFormat format){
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -55,8 +55,8 @@ static int pixelBitsFromPixelFormat(ofPixelFormat format){
 }
 
 template<typename PixelType>
-static int bytesFromPixelFormat(int w, int h, ofPixelFormat format){
```
bakercp: Shouldn't `static int pixelBitsFromPixelFormat(ofPixelFormat format)` also return size_t?

armadillu:
sorry just realized i wasn't up to date

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -348,7 +348,7 @@ void ofPixels_<PixelType>::setFromAlignedPixels(const PixelType * newPixels, int
 		return;
 	}
 	allocate(width, height, _pixelFormat);
-	int dstStride = width * pixelBitsFromPixelFormat<PixelType>(_pixelFormat)/8;
```
bakercp: And why not size_t here as well?

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -348,7 +348,7 @@ void ofPixels_<PixelType>::setFromAlignedPixels(const PixelType * newPixels, int
...
+	int dstStride = width * pixelBitsFromPixelFormat<PixelType>(_pixelFormat) / 8;
 	const unsigned char* src = (unsigned char*) newPixels;
 	unsigned char* dst =  (unsigned char*) pixels;
 	for(int i = 0; i < height; i++) {
```
bakercp: and int here -> size_t too.

armadillu: yep let me fix those + the merge conflicts - preparing another PR

bakercp:
Also, `int ofPixels_<PixelType>::getPixelIndex`, etc.  Indices should all be size_t.

PR #5200, opened 2016-08-12
Convert all ofPixel dimensions from int to size_t

armadillu:
as discussed in https://github.com/openframeworks/openFrameworks/issues/5045 and in plenty of comments in https://github.com/armadillu/openFrameworks/commit/608274233cdb8ae6e9fa530e23f507862c513f25

Basically all ints are gone. 
I Had to change a couple of things in ofBitmapFont and ofVideoPlayer too.


libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -443,18 +443,18 @@ const PixelType * ofPixels_<PixelType>::getData() const{
...
 template<typename PixelType>
-void ofPixels_<PixelType>::allocate(int w, int h, int _channels){
+void ofPixels_<PixelType>::allocate(size_t w, size_t h, size_t _channels){
 	allocate(w,h,pixelFormatFromNumChannels(_channels));
 }
 
 template<typename PixelType>
-void ofPixels_<PixelType>::allocate(int w, int h, ofPixelFormat format){
+void ofPixels_<PixelType>::allocate(size_t w, size_t h, ofPixelFormat format){
 	if (w <= 0 || h <= 0 || format==OF_PIXELS_UNKNOWN) {
```
bakercp: Probably change to == 0 on w/ h, as some compilers may give warnings since w/h can now never be < 0.

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -858,7 +859,7 @@ ofPixels_<PixelType> ofPixels_<PixelType>::getPlane(int planeIdx){
 				plane.setFromExternalPixels(pixels+width*height,width/2,height/2,OF_PIXELS_V);
 				break;
 			case 2:
-				plane.setFromExternalPixels(pixels+int(width*height+width/2*height/2),width/2,height/2,OF_PIXELS_U);
+				plane.setFromExternalPixels(pixels+size_t(width*height+width/2*height/2),width/2,height/2,OF_PIXELS_U);
```
bakercp: Change to c++ (e.g. `static_cast<size_t>(width * ...)`) ?

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -871,7 +872,7 @@ ofPixels_<PixelType> ofPixels_<PixelType>::getPlane(int planeIdx){
 				plane.setFromExternalPixels(pixels+width*height,width/2,height/2,OF_PIXELS_U);
 				break;
 			case 2:
-				plane.setFromExternalPixels(pixels+int(width*height+width/2*height/2),width/2,height/2,OF_PIXELS_V);
+				plane.setFromExternalPixels(pixels+size_t(width*height+width/2*height/2),width/2,height/2,OF_PIXELS_V);
```
bakercp: static_cast ...

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -893,17 +894,17 @@ void ofPixels_<PixelType>::setImageType(ofImageType imageType){
 	if(!isAllocated() || imageType==getImageType()) return;
 	ofPixels_<PixelType> dst;
 	dst.allocate(width,height,imageType);
-	PixelType * dstPtr = &dst[0];
-	PixelType * srcPtr = &pixels[0];
-	int dstNumChannels = dst.getNumChannels();
-	int srcNumChannels = getNumChannels();
-	int diffNumChannels = 0;
+	PixelType * dstPtr = &dst[static_cast<size_t>(0)];
```
bakercp: I don't understand why casts are needed here ... are you still getting the warning now that the other bits are changed?

armadillu: hm let me try again - it was an error btw

libs/openFrameworks/video/ofVideoPlayer.cpp
```diff
@@ -77,7 +77,7 @@ bool ofVideoPlayer::load(string name){
         if(bUseTexture){
         	if(player->getTexturePtr()==nullptr){
 				if(tex.empty()) {
-					tex.resize(max(player->getPixels().getNumPlanes(),1));
```
bakercp: Nice!  I'd suggest prepending this with `std::max(` to be explicit.  Also, this should be tested on windows (Appveyor I guess) since std::min / std::max have windows problems ...

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -1322,23 +1323,23 @@ bool ofPixels_<PixelType>::resizeTo(ofPixels_<PixelType>& dst, ofInterpolationMe
...
 
-					int   dstIndex0 = (dsty*dstWidth + dstx) * bytesPerPixel;
+					size_t   dstIndex0 = (dsty*dstWidth + dstx) * bytesPerPixel;
 					float srcxf = srcWidth  * (float)dstx/(float)dstWidth;
 					float srcyf = srcHeight * (float)dsty/(float)dstHeight;
-					int   srcx = (int) MIN(srcWidth-1,   srcxf);
-					int   srcy = (int) MIN(srcHeight-1,  srcyf);
-					int   srcIndex0 = (srcy*srcWidth + srcx) * bytesPerPixel;
+					size_t   srcx = static_cast<size_t>(MIN(srcWidth-1,   srcxf));
+					size_t   srcy = static_cast<size_t>(MIN(srcHeight-1,  srcyf));
```
bakercp: if you are going to change other defines to std::min/max, I'd do it here too.

libs/openFrameworks/graphics/ofPixels.h
```diff
@@ -148,18 +148,18 @@ class ofPixels_ {
...
 	/// As a word of caution this reallocates memory and can be a bit
 	/// expensive if done a lot.
-	void cropTo(ofPixels_<PixelType> &toPix, int x, int y, int width, int height) const;
+	void cropTo(ofPixels_<PixelType> &toPix, size_t x, size_t y, size_t width, size_t height) const;
 
 	// crop to a new width and height, this reallocates memory.
-	void rotate90(int nClockwiseRotations);
-	void rotate90To(ofPixels_<PixelType> & dst, int nClockwiseRotations) const;
+	void rotate90(size_t nClockwiseRotations);
+	void rotate90To(ofPixels_<PixelType> & dst, size_t nClockwiseRotations) const;
```
bakercp: nClockwiseRotations should not be size_t.  Sometimes we rotate backwards :)

bakercp: Or at least we've always allowed it because of the while-loop / modulo ...

```
	int rotation = nClockwiseRotations;
	while (rotation < 0){
		rotation+=4;
	}
	rotation %= 4;
```

libs/openFrameworks/video/ofVideoPlayer.cpp
bakercp: Oh, whoops, I noticed you changed it back to the define, rather than the std::max.  I'd try to use the standard library std::min/max as I believe we're trying to cut down on our custom defines ...

libs/openFrameworks/graphics/ofPixels.cpp
armadillu: std::max() is not found by Xcode? even after adding #include <algorithm>

bakercp:
Seems like if we're going down this path (which I'm 100% for doing), we need to visit `ofImage` and `ofTexture`(e.g. https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofImage.h#L459)

libs/openFrameworks/graphics/ofPixels.cpp
armadillu: still needed, i don't see why either

libs/openFrameworks/graphics/ofPixels.cpp
bakercp: You added `#include <algorithm>`?

armadillu:
preparing a new one with those modifications

libs/openFrameworks/graphics/ofPixels.cpp
armadillu: yep #include <algorithm> - still no go

bakercp:
Just thinking about the different return type with have for getWidth()/getHeight() all over the API (a lot of it is defined by ofBaseDraws).  For ofPixels, I'd say that we should definitely use size_t because ofPixels are not "drawable" on their own.  They are a data structure and their internal constraints and structure are best defined by size_t when it comes to width / height.  It feels a little odd that ofImage would return a different type (i.e. `float`), but that makes perfect sense for a drawable situation (just like you would expect a texture to have a floating point w/ height).   Anyway, just processing this.  Hopefully not adding too much noise :)

libs/openFrameworks/graphics/ofPixels.cpp
armadillu: i see why now - the zero is read as "int" - if I cast the zero again it works fine
size_t minX = std::max(x, static_cast<size_t>(0));

bakercp: Just looking at this http://stackoverflow.com/a/31847973 -- if you use `0U` does that work?  Now I'm just trying to learn some c++ :)

armadillu: nope, still not accepted

armadillu: If we are force to state the type, I very much prefer
```
MAX(player->getPixels().getNumPlanes(),1) 
```
to
```
std::max(player->getPixels().getNumPlanes(),static_cast<size_t>(1))
```

armadillu:
I think all the inline comments have been resolved with this last one push.

I still think that operator[] becomes very ugly if you are forced to cast it.
```pixels[static_cast<size_t>(0)]```

I agree that expanding size_t to other areas makes sense for all pixel related units (i.e. ofTexture allocate w & h, etc) but might be too much to do in one shot - And although int is not ideal, the other areas of OF don't seem to be affected as badly as ofPixels was with large images (bc only of pixels has indexed access & direct bytecount mem alloc in the w x h fashion), so we can afford a slower transition.

**Comment by arturoc** [comment]
> I still think that operator[] becomes very ugly if you are forced to cast it.
> 
> ```cpp
> pixels[static_cast<size_t>(0)]
> ```

std::vector uses size_t as index for the [] operator, why do you need to cast it like that?

**Comment by armadillu** [comment]
yeah you would think so, but it won't compile and I couldn't find out why.

https://github.com/openframeworks/openFrameworks/issues/5045#issuecomment-239550183

**Comment by arturoc** [comment]
can you try removing this methods:

https://github.com/armadillu/openFrameworks/blob/9e4356b17227c423669b5291f7ec6d7399a58189/libs/openFrameworks/graphics/ofPixels.h#L212-L213

**Comment by armadillu** [comment]
yep, that works. If I remove what you mentioned and the implementation at:

https://github.com/armadillu/openFrameworks/blob/9e4356b17227c423669b5291f7ec6d7399a58189/libs/openFrameworks/graphics/ofPixels.cpp#L415-L423

the ofPixels[] operator cast to size_t is not required anymore, so stuff like this

```
PixelType * srcPtr = &pixels[0];
```
just works.

May be it's time to make the deprecation take effect?

**Comment by arturoc** [comment]
yes, go for it, i think most code will be adapted by now and if it's not the fix is pretty straight forward

**Comment by armadillu** [comment]
It's still required for in some places like max()
```
std::max(x, static_cast<size_t>(0));
```

but that's expected

**Comment by arturoc** [comment]
yeah that's ok, apart from that, this is breaking in some platforms, i think it might be an easy fix, if you can take a look at the output from the test servers? if there's something you can't figure out let me know and i'll take a look when i have a moment

**Comment by armadillu** [comment]
ok will try and report back

**Comment by arturoc** [comment]
by the way for simple casts like that i prefer:

```cpp
std::max(somesizet, size_t(0));
```
rather than the ultra long static_cast but it's ok if you want to leave it like that

armadillu:
Yeah me too - I prefer readability to "academic conformity"; I was following @bakercp 's advice.

arturoc:
:) in this case it's even completely correct cause it's not even an old c style cast it's just a size_t constructor which is perfectly fine to cast primitive types. static_ reinterpret and dynamic make the most sense when casting objects, for primitive types, numbers, there's really no ambiguity.

armadillu:
looks like it was only the Gst video player - lets see if this fixes it.

armadillu:
Ok seems like its compiling one all platforms.

arturoc:
thanks!

PR #5208, opened 2016-08-22
Refactor every load/save method and ofFilePath to use filesystem::path

arturoc:
Old code should work unmodified and it allows to use all the facilities of
filesystem path without having to convert to string before passing the path.

For example:

```cpp
std::string shaderPath = ofFilePath::join(somepath, "shader.vert");
ofShader shader;
shader.setupFromFile(GL_VERTEX_SHADER, shaderPath);
```

can now be:

```cpp
ofShader shader;
shader.setupFromFile(GL_VERTEX_SHADER, somepath / "shader.vert");
```

### DIFF
--- a/addons/ofxAndroid/src/ofxAndroidSoundPlayer.cpp
+++ b/addons/ofxAndroid/src/ofxAndroidSoundPlayer.cpp
@@ -43,7 +43,7 @@ ofxAndroidSoundPlayer::~ofxAndroidSoundPlayer(){
 
 
 //------------------------------------------------------------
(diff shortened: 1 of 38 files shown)


bakercp:
This looks great.  Only question -- if we're changing, can we make the parameters const references?  It seems to be a mix of by value and const references.

arturoc:
yeah i don't think it makes such a big difference to use one or the other but for uniformity sake would be good to use the same thing all over. if you want to send a PR to my branch? this is broken at the moment though so perhaps wait until i fix it. it might even be a PR after this is merged

PR #5209, opened 2016-08-22
Android - Fix compilation error on OS X

tallavi:
The ar command is apparently not configured to create compatible .a file and thus the linking fails on OSX.

I fixed the error by giving the path to ar explicitly. I hope this line is properly generalized and will work on other OSs. I only have access to OSX right now.


arturoc:
looks good and it compiles in the test server so i guess it's ok. thanks!

PR #5218, opened 2016-08-24
put apothecary in git submodule and update path in scripts

### LINKED ISSUE #5210: poco does not compile / apothecary not found
```
It seems you are running gcc 5 or later, due to incomatible ABI with previous versions
we need to recompile poco. This will take a while
Press any key to continue... ./install_dependencies.sh: line 210: cd: /home/nanu-c/Projekte/openFrameworks/scripts/linux/ubuntu/../../apothecary: No such file or directory
./install_dependencies.sh: line 211: ./apothecary: No such file or directory
chown: cannot access 'build/poco': No such file or directory
chown: cannot access '../../libs/poco': No such file or directory
```

In the branch 0.9.4 it's still working...

nanu-c:
closes #5210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ libs/openFrameworksCompiled/lib/*/*
 !libs/openFrameworksCompiled/lib/*/.gitkeep
 
 # apothecary
-scripts/apothecary/build
+scripts/apothecary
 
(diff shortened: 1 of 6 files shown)


arturoc:
thanks!

PR #5219, opened 2016-08-24
OFAndroid - support for showing the canvas on multiple activities

tallavi:
**Purpose:** 
Allow complicated applications to show GL content in more than one activity.

**How we did it:**
We decided that it’s just too complicated to actually create more than one canvas or instantiate more than one ofApp instance, so we took another route: recycle a single ofApp and canvas across multiple OFActivities. Because of this, the layout for OFActivity doesn’t contain the canvas anymore, instead it contains a place holder view, that we pass to OFAndroid, and OFAndroid injects the canvas as a child. The first activity that requests the canvas triggers the creation of the canvas. Subsequent activities reuse the existing canvas. When the last activity that used the canvas is destroyed, the canvas is also destroyed.

**Refactoring/improvements:**
1) We took more bits of code out of the over-bloated OFAndroid class and moved it to a more specific class (i.e. android life cycle)
2) We allowed ANY activity to hold the canvas, and not just OFActivity. OFActivity still exists and it’s still the easiest way to use OFAndroid, but now it’s just a wrapper for some calls to OFAndroid and you can use any activity you want. We use the direct way and not OFActivity since we want activities that extends AppCompatActivity.

**Full backwards compatibility:**
main_layout.xml file was changed in all examples. Now it contains a place holder view instead of the actual canvas, BUT we added automatic backwards compatibility. If old projects are used, where the layout contains the canvas view and not the place holder view, it will be detected and will switch to GL only mode. This part happens in OFActivity creation. The only case that is not fully backwards compatible are apps that added more views in the xml. These apps need to update main_layout.xml to contain the placeholder view (there’s a very clear logcat line for these cases).

**New example:**
We added an example called androidMultiOFActivitiesExample that has two activities. Both activities have a GL canvas with a simple animation (tap the canvas to change background color). You can see that even when changing activities, the background color is preserved, since it’s the same ofApp.

**Showing different content in each activity:**
We DID NOT add a mechanism to distinguish from the c++ side which activity is currently showing. We believe it breaks the generalized nature of openFrameworks. We ourselves have our own mechanism of notifying events to the c++ side and we use that. We figured that each app that want to show different content in different activities is complicated enough to have some sort of messaging service between the layers and such a service is out of scope for openFrameworks.


arturoc:
This looks great, i really don't have time right now to test it properly so i think i'm just going to merge it. Do you plan to post the ios version you told me about soon? in that case i can wait to see if we need to do something on that side that will need to be changed here too but if you think it's better to just merge it let me know

tallavi:
Oh, yeah, forgot to write about iOS. After finishing it, I realized that due to the ability of iOS to nest view controllers, there's actually no need to change anything in OF for that. I just made a simple view controller that recycles the same of view controller and made it part of my project, and here's the entire thing:

```
#import "GLViewController.h"
#include "ofxiOS.h"
#include "ofAppMap.h"

ofxiOSViewController* m_ofViewController = NULL;
int m_refCount = 0;

@implementation GLViewController

- (void)viewDidLoad {
    [super viewDidLoad];
    
    if (m_ofViewController == NULL) {
        ofxiOSGetOFWindow()->setOrientation( OF_ORIENTATION_DEFAULT );   //-- default portait orientation.
    
        m_ofViewController =
        [[ofxiOSViewController alloc] initWithFrame:[[UIScreen mainScreen] bounds]
                                             app:(new Trailze::ofAppMap())];
    }
    
    ++m_refCount;
}

-(void) viewWillAppear:(BOOL)animated {
    [super viewWillAppear:animated];
    
    if (m_ofViewController.parentViewController != self) {
        
        if (m_ofViewController.parentViewController != nil) {
            [m_ofViewController.view removeFromSuperview];
            [m_ofViewController removeFromParentViewController];
        }
        
        [self addChildViewController:m_ofViewController];
        [self.view insertSubview:m_ofViewController.view atIndex:0];
    }
}

-(void) dealloc {
    
    if (self == m_ofViewController.parentViewController) {
        [m_ofViewController.view removeFromSuperview];
        [m_ofViewController removeFromParentViewController];
    }
    
    --m_refCount;
    
    if (m_refCount == 0) {
        [m_ofViewController release];
        m_ofViewController = nil;
    }
    
    [super dealloc];
}

@end
```

Do you think that this should be the included inside ofxiOS, instead or in addition to the normal ofxiOSViewController?

arturoc:
i have never used ios so no idea but if it's that simple i guess we can merge this right away anyway. perhaps you can open a new issue pointing to this one but refering to ios in the title. i was going to notify here the people who usually maintain ios but since they will get any notification mentioning them is pretty useless.

thanks so much for all the work on this!

tallavi:
Done. My pleasure!

PR #5220, opened 2016-08-24
ofShader: add methods to change constant and defines values

arturoc:
in a shader with a define like:

```cpp
#define USE_SOME_FEATURE 1
```

you can now call:

```cpp
shader.setDefineConstant(USE_SOME_FEATURE, 0);
```

to disable it.

There's a similar method for const numerical or boolean values. in
this case the types are more strict similar to how setUniform* works.

This recompiles the shader so it's slow to use but it can be faster
than uniforms in several cases, usually when it's used for values
that won't change often and would introduce a conditional or a for
loop with bounds comming from a uniform which makes the compiler
unable to unroll the loop.

### DIFF
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -1217,6 +1217,7 @@ void ofShader::setConstantb(const string & name, bool value){
             auto & shader = shader_pair.second;
             shader.expandedSource = std::regex_replace(shader.expandedSource, re_define, "const bool " + name + "=" + (value?"true":"false") + ";");
         }
+#ifndef TARGET_OPENGLES
         if(bLoadedAsXFB){
             auto settings = TransformFeedbackSettings();
             settings.bindDefaults = boundDefaults;
@@ -1227,13 +1228,16 @@ void ofShader::setConstantb(const string & name, bool value){
             settings.bufferMode = xfbBufferMode;
             setup(settings);
         }else{
+#endif
             auto settings = Settings();
             settings.bindDefaults = boundDefaults;
             for(auto & shader_pair: shaders){
                 settings.shaderSources[shader_pair.first] = shader_pair.second.expandedSource;
             }
             setup(settings);
+#ifndef TARGET_OPENGLES
         }
+#endif
     }
 #endif
 }
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -200,7 +200,7 @@ bool ofShader::load(std::filesystem::path vertName, std::filesystem::path fragNa
 #endif
 	if(ofIsGLProgrammableRenderer()){
 		bindDefaults();
-	}
+    }
 	return linkProgram();
 }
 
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/gl/ofShader.cpp
```diff
@@ -1070,6 +1077,151 @@ void ofShader::setUniforms(const ofParameterGroup & parameters) const{
 		}
 	}
 }
+
+
+//--------------------------------------------------------------
+template<typename T>
+void ofShader::setDefineConstantTemp(const string & name, T value){
+#if (!defined(TARGET_LINUX) || defined(GCC_HAS_REGEX))
```
tgfrerer: Will these work on VS? If not, it could be good to have the parsing done in a platform - agnostic way- otherwise nothing not even an error message happens on programs compiled with compilers that don't support gcc_regex

arturoc: yes it will, it only won't work on old gccs, anything below 4.9 which almost nobody should be using by now, i was planning to add an error message in case it's not implemented though but completely forgot

tgfrerer:
+1 this can be very useful for übershader style scenarios. Shaderc (google'a open source shader precompiler) has a similar function, named `void shaderc_compile_options_add_macro_definition`, which like a default c-style compiler allows setting or overriding #defines.

I don't think it needs the ability to directly set consts though, as the #define code path already implements all needed functionality. So I guess the const methods could be removed for the sake of DRY, what do you think?

libs/openFrameworks/gl/ofShader.cpp
tgfrerer: Makes sense =) I thought I saw some subtle differences in how vs implements std::regex, I can do some testing on windows...

arturoc:
i find it pretty useful to have a more type strict way of specifying constants, i usually use defines to completely enable/disable blocks of code and constants for example for for loops so they are unrolled by the compiler.

libs/openFrameworks/gl/ofShader.h
```diff
@@ -127,6 +127,38 @@ class ofShader {
...
+    ///
+    /// Allows to change a define in the shader declared like
+    ///
+    /// const float somevalue=10;
+    ///
+    /// calling setConstant1f("somevalue", 20.f);
+    ///
+    /// Be aware that this recompiles the shader which can be
+    /// slow if done often.
+    void setConstant1f(const string & name, float value);
```
tgfrerer: It's not just that it recompiles the shader - it also globally invalidates the previous GL shader program object across all users of this shader if I understand correctly. This could be a quite surprising side-effect =)

I wonder if it wasn't possible to make these methods only callable before calling setup() - or to print out a performance warning otherwise - as some folks might be tempted to call these method in the draw() method repeatedly on a shader.

Maybe also changing the name from `setConstant..` to something more like `setSourceCodeConstant` might help to distinguish these methods from the classic "draw" `setUniform..` methods...

arturoc: >  it also globally invalidates the previous GL shader program object across all users of this shader

what do you mean? copies of the same ofShader?

> I wonder if it wasn't possible to make these methods only callable before calling setup()

I think is useful to allow to call them afterwards. it allows to test things easily but perhaps we could have a public recompile method or something like that that needs to be called afterwards, that would also allow to change several at the same time before recompiling but probably that would be confusing and people will forget to call recompile

changing the name to something more explicit is a good idea, usually you will change this from a gui or something so you mostly change them one by one. perhaps setConstant*AndRecompile? :)

Issue #5221, opened 2016-08-24
ofSerial is not listing any devices on OS X.


smallfly:
The ```ofSerial::buildDeviceList()``` is not working properly on OS X. No serial devices are listed.
The 'prefix match' strings needed to be changed to ```/dev/cu.``` and ```/dev/tty.```

Tested on OS X 10.11.6. I do not have an older version of OS X to test this.

arturoc:
This seems to be caused by ofDirectory returning an absolute path if it was created with an absolute path, so this:

```
ofDirectory dir("/dev");
```

will return absolute paths for the paths of it's entries while


```
ofDirectory dir(".");
```

will return local paths. seems it was returning local paths before no matter how ofDirectory was created and indeed itdoes so in wdows,is only unix OSs that are affected.

arturoc:
there's also some tests here https://github.com/openframeworks/openFrameworks/pull/5223 to check that the problem is solved on all platforms

mattfelsen:
Hi all. I just ran into this problem today on OSX. I looked over the discussion in https://github.com/openframeworks/openFrameworks/pull/5222 & https://github.com/openframeworks/openFrameworks/pull/5223 and understand the underlying problem with ofDirectory, and I'm not sure on the best approach to fix that.

However, as a quicker fix for this specific issue on both OSX and Linux that doesn't feel as "dirty" as the original fix, this seems to do the trick:

Replacing [this line in ofSerial.cpp](https://github.com/openframeworks/openFrameworks/blob/84d6d1a18cf22cc4e5ea5ea522abb3923738d1fa/libs/openFrameworks/communication/ofSerial.cpp#L173):
`std::string deviceName = entry.path();`
with this one:
`std::string deviceName = entry.getFileName();`
ensures that the prefix is always getting checked against only the file name, regardless of whether a relative or absolute path is returned by ofDirectory.

I'm happy to submit a PR if this feels ok, and doesn't require ofDirectory getting fixed first. What do you guys think? ping @bakercp as well

arturoc:
that sounds great, can you submit a PR?

PR #5222, opened 2016-08-24
Fixed the ofSerial::buildDeviceList() method on OS X

smallfly:
See the issue #5221

### DIFF
--- a/libs/openFrameworks/communication/ofSerial.cpp
+++ b/libs/openFrameworks/communication/ofSerial.cpp
@@ -152,8 +152,8 @@ void ofSerial::buildDeviceList(){
 
 	#ifdef TARGET_OSX
 
-		prefixMatch.push_back("cu.");
-		prefixMatch.push_back("tty.");
+		prefixMatch.push_back("/dev/cu.");
+		prefixMatch.push_back("/dev/tty.");
 
 	#endif
 


bakercp:
This may break Linux and is probably a result of the recent std::filesystem::path updates to the internals of ofDirectory.  Ideally, since we're listing the contents of /dev/ on both Linux / OSX, we shouldn't need this prefix to do matching.  There are a lot of stray /dev/ bits in there so perhaps that just needs to be cleaned up?

arturoc:
the changes to use filesystem path where just to the input arguments for methods not to any internals or return methods so it shouldn't have almost any side effects. in any case i've just tested and the problem was already there.

This probably comes from before that when we swapped ofFileUtils to use std::filesystem instead of poco

it seems that for ofDirectory created with absolute paths the paths of the entries are also absolute which wasn't like that before if that code was working.

I guess the correct fix would be to make ofDirectory  behave the way it did before not to fix ofSerial. I'm going to open a PR request only with the tests to check that the paths returned by ofDirectoy are local

smallfly:
With all the info you both brought to the discussion, it's clear that this should not be fixed in ofSerial, but in ofDirectory instead. Sorry for this PR which is in fact just a quick and dirty fix. Should this PR just be closed and the discussion moved to the issue #5221 ? 

I did test ofSerial on Windows before doing this PR, and it was working properly. The issue is only on OS X. Have not tested it on Linux though.

arturoc:
yes i'm going to close this, in windows, the com port is not a file device so it wouldn't be affected by this problem, in linux it happens the same. 

the weird thing is that in windows the directory doesn't have this problem either. i'll comment more in the issue

PR #5225, opened 2016-08-24
Latest libfreenect for ofxKinect and bugfixes for device reconnection. 

ofTheo:
This PR pulls in the latest from the libfreenect project.  
While testing I also noticed some bugs in the way that device reconnection was happening that could lead to the wrong device being restarted.  

There is also a fix in libfreenect now for better firmware upload for multiple devices that should fix multiple kinects not always opening correctly.  

Tested this pretty extensively on OS X with up to three Kinect of 1473, K4W and 1414 types. 

Would appreciate others trying this out, especially on Windows and Linux.


ofTheo:
doing a new one which doesn't introduce untracked files from libfreenect

ofTheo:
better PR with some additional fixes here: #5231

PR #5226, opened 2016-08-25
ofTexture: Added convenience method to draw subsection using ofRectangle parameter.

prisonerjohn:


ofTheo:
@prisonerjohn - only thing I would change would be better parameter naming. 
I had to read the comments to understand what each rectangle was for, whereas it could be implied better in the naming 

maybe?:
dstBounds -> drawBounds
srcBounds -> subsectionBounds

prisonerjohn:
Yeah that sounds good!

PR #5229, opened 2016-08-25
ofShader: add methods to setup defines values

arturoc:
This substitutes #5220 but instead of having a method which changes the constants after the shader has been setup this allows to set those values in the new settings objects. as pointed out by @tgfrerer  the way it was done before was confusing cause the methods resembled too much the way uniforms are set which could lead to people using this methods every frame getting considerable slowdowns.

Now to change a define one can do:

```cpp
shaderSettings.intDefines["SOME_DEFINE"] = 5;
shader.setup(shaderSettings);
```

usually the shaderSettings object is created while setting up the shader for the first time and kept around for reusing it afterwards

### DIFF
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -1217,6 +1217,7 @@ void ofShader::setConstantb(const string & name, bool value){
             auto & shader = shader_pair.second;
             shader.expandedSource = std::regex_replace(shader.expandedSource, re_define, "const bool " + name + "=" + (value?"true":"false") + ";");
         }
+#ifndef TARGET_OPENGLES
         if(bLoadedAsXFB){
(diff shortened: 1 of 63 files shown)


libs/openFrameworks/gl/ofShader.cpp
```diff
@@ -288,63 +295,92 @@ bool ofShader::setupShaderFromFile(GLenum type, std::filesystem::path filename)
...
 bool ofShader::setupShaderFromSource(GLenum type, string source, string sourceDirectoryPath) {
-    unload();
-    
+	return setupShaderFromSource({type, source, sourceDirectoryPath});
+}
+
+
+//--------------------------------------------------------------
+bool ofShader::setupShaderFromSource(ofShader::Source && source){
+	unload();
```
tgfrerer: Maybe this should be conditional - if bLoaded, unload()?

PR #5230, opened 2016-08-25
bugfix for static ofFile and ofDirectory calls not working if data path is set false. Closes #5187

### LINKED ISSUE #5187: ofFile::open() ALWAYS calls ofToDataPath()

ofTheo:
New PR for issue #5187
Adds openFromCWD to ofFile and ofDirectory. 

Tested pretty extensively with all calls which seem to be affected by the bRelativeToData arguments.  

Couple of small things - will point out in the diff.

### DIFF
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -435,6 +435,13 @@ bool ofFile::open(const std::filesystem::path & _path, Mode _mode, bool binary){
 	return openStream(_mode, binary);
 }
 
+//------------------------------------------------------------------------------------------------------------
+bool ofFile::openFromCWD(const std::filesystem::path & _path, Mode _mode, bool binary){
+	close();
+	myFile = std::filesystem::path(_path.string());
+	return openStream(_mode, binary);
+}
+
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -944,7 +991,7 @@ bool ofDirectory::create(bool recursive){
 
 //------------------------------------------------------------------------------------------------------------
 bool ofDirectory::exists() const {
-	return std::filesystem::exists(myDir);
+	return (myDir == "" || std::filesystem::exists(myDir));
```
ofTheo: I had to add this as ofDirectory::exists was failing when it was set to "".  
In the case of a directory it does exist if the path is "" as it is the cwd. 

So things like:

```
ofDirectory dir; 
dir.openFromCWD( ofFilePath::getEnclosingDirectory("test1.txt") ); 
dir.exists();
```
were failing. 

I think we didn't notice this in the past as a directory was never "" as we were always adding ofToDataPath to everything.

Issue #4828, opened 2016-02-16
waitForThread sometimes causes EXC_BAD_ACCESS


kylemcdonald:
I left a note here https://github.com/openframeworks/openFrameworks/commit/77f1f15019c4557ed09eb628b250c22dfacaa09e#commitcomment-16106595 but those notes are lost easily and @arturoc encouraged me to add an issue.

The bug I noticed is that sometimes calling thread.join() can fail because the `thread.__t_ == 0`. This means when you are (for example) closing an app and call `thread.waitForThread()` in your `ofBaseApp::exit()`, it will throw `EXC_BAD_ACCESS`.

There's a method called `thread.joinable()` which seems to be meant for handling this case, and since adding `if(thread.joinable()) {...}` to that block I haven't gotten any more crashes. But the crashes were rare (1 out of 10 times) so it might just be chance...

While it's still possible for multiple `join` attempts to create a conflict, I think in general, I think it can't hurt to add `if(thread.joinable()) {...}` (especially because the try / catch doesn't catch EXC_BAD_ACCESS).

In the future, I think the more complete solution would be to make sure that joining the thread requires locking the thread, so only one attempt is made at a time to join the thread.

If I can get a very minimal example of this behavior I will post it, but I hope this is helpful for now.

arturoc:
hey thanks, i've gone ahead and added the if(thread.joinable()) anyway, from reading the docs it seems that this might happen when the thread hasn't been initialized yet?

my concern is that at some point the thread can become not joinable while already having checked that but i don't either want to lock there cause it can enter into a deadlock.

anyway i'm going to keep this open and milestone it for 0.10 to review it in case it's still problematic for other people

kylemcdonald:
Just want to add an update to this, I'm still seeing it. Here is a simple example to reproduce the bug:

```c++
#include "ofMain.h"

class ofApp : public ofBaseApp, ofThread {
public:
    void setup(){
        startThread();
    }
    void exit(){
        if(isThreadRunning()){
            waitForThread();
        }
    }
    void threadedFunction(){
        while(isThreadRunning()) {
            sleep(5);
        }
    }
};

int main() {
    ofSetupOpenGL(1280, 720, OF_WINDOW);
    ofRunApp(new ofApp());
}
```

if it doesn't crash the first time, you might need to run it ten times. and try exiting after different amounts of time (1-10 seconds).

[...]

PR #5231, opened 2016-08-25
Bugfix ofxKinect / latest libfreenect

ofTheo:
This is a cleaner PR with the same functionality as #5225 with a couple of additional fixes. 

fixes:
- latest libfreenect with fix for multiple kinect devices not being able to be opened at the same time
- fix for device reconnection false positive due too much timing passing from the open call and the first update. 
- now will reopen if either the depth or color stream isn't coming in - before one could be failing and it would not try and get both coming in. 
- fix for device index not being correct 
- fix for a crash on ofxKinect::close() that was in master that seems to be related to #4828 

Should be good to merge assuming the checks pass.


armadillu:
I can't get ProjectGenerator to create a VS2015 project that will compile with ofxKinect, 32 or 64 bit. Did you try this patch on windows? It complains that all the libusb methods are not found (libusb_control_transfer, etc)

ofTheo:
@armadillu - I will give it a try on my machine here.  My guess is its not from this PR as none of the files changed should effect any of the libusb methods working on windows.  ( the changes are just to platform agnostic files ) 

Thanks for reporting this though! 
Theo

ofZach:
in terms of PG related issues, I'd look here https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxKinect/addon_config.mk#L119-L134 and compare a working vs non working vs file to see if this needs to be altered.

PR #5233, opened 2016-08-25
Update iterator types for planes to remove warnings and make better f…

bakercp:
…loat equality comparisons.


arturoc:
don't think ofFloatIsEqual makes sense here, the width of the player is always going to be an integer

bakercp:
Yeah, I see that, but as we're comparing drawing dimensions (floats) with pixels dimensions (size_t), I opted to compare via floating point rather than (potentially in some future implementation) truncating to std::size_t.  That said, there are more operations and (for now at least) it seems our video player impl are pixel-backed.

arturoc:
yes the player returns floats to help doing math with those values it's kind strange and i don't think they should be floats but in any case even if they are floats the numbers are always going to be integers so it's ok to just cast them like this. thanks!

PR #5238, opened 2016-08-28
Android - make fix - build correct ABIs when compiling release

tallavi:
Pretty self explanatory. I guess it was a copy paste typo. 

Tested to work correctly. Before the fix the resulting apk did not contain x86 and neon libs (as if the compilation was made with debug).


arturoc:
thanks!

PR #5244, opened 2016-09-03
fix for qtcreator 4.1

underdoeg:
Quick fix for qtcreator 4.1. I had to import missing modules into the  javascript helpers.

Note that I also get these deprecation warnings:

```
The item 'Transformer' is deprecated and will be removed in qbs 1.7.0.
Use the 'Rule' item instead.
```


arturoc:
thanks! will take a look into the transformer warning

Issue #5248, opened 2016-09-06
calling ofParameter::getFirstParent() in side ofParameterGroup::parameterChangedE listener


markkorput:
_note: I'm running OSX 10.11.16 and an up-to-date master branch (commit: 0a9927c)_

I was getting mysterious crashes while calling ofParameter::getGroupHierarchyNames inside an ofParameterGroup::parameterChangedE callback and managed to boil it down to the ofParameter::getFirstParent() call. Code to reproduce (it doesn't crash until you change the param value):

```c++
#include "ofMain.h"
#include "ofxGui.h"

class ofApp : public ofBaseApp{
    
public:
    void setup();
    void draw();
    void parameterChanged(ofAbstractParameter & parameter);
    
    ofxPanel gui;
    ofParameterGroup group;
    ofParameter<float> float1, float2;
};

void ofApp::setup(){
    group.setName("myParams");
    // group.add(float2.set("myFloat2",0.5));
    group.add(float1.set("myFloat",0.5));
    
    gui.setup(group);
    ofAddListener(group.parameterChangedE(),this,&ofApp::parameterChanged);
}

void ofApp::draw(){
    gui.draw();
}

void ofApp::parameterChanged(ofAbstractParameter & parameter){
//    vector<string> hierarchy = float1.getGroupHierarchyNames();
     ofParameterGroup p = float1.getFirstParent();
}

//========================================================================
int main( ){
	ofSetupOpenGL(1024,768, OF_WINDOW);
	ofRunApp( new ofApp());
}
```

What makes this bug more obscure is that it only crashes when the ofParameter instance that getFirstParent is called upon is the only param inside its group. If you uncomment the following line in ofApp::setup(), all functions normally.

```c++
// group.add(float2.set("myFloat2",0.5));
```

markkorput:
Hey @arturoc , thanks for spending time on this. Unfortunately your patch didn't stop my crashes. I [tried to apply your weak_ptr as reference approach on ofParameter's getFirstParent method as well](https://github.com/markkorput/openFrameworks/commit/99008a96b580230956f357c33d32ac38d02eaf13) without any luck.

I'm not sure where the invalid weak_ptr in the obj->parents vector comes from, but my call to float1. getFirstParent() removes it from within the parameterChangedE listener. At that point we've already started iterating over obj->parents inside ofParameter::eventsSetValue. This iterations then becomes corrupt.

I'll submit a new pull requests, but my solution simply involves running a preliminary cleanup.

PR #5249, opened 2016-09-06
don't perform "hierarchy-management" in ofParameter::getFirstParent()

markkorput:
Relates to issue #5248 

Note that I simply removed the code that seems to check the ParameterGroup hierarchy integrity from the getFirstParent function, which might be problematic? I'm unclear about why this happens here. Similar functionality can be found in the ofParameter::eventsSetValue function that is called every time a parameter's value is changed, which seems wildly inefficient.

Note that I know very little about all this weak_ptr/shared_ptr style coding, so it's not unlikely that I don't know what I'm talking bout.


markkorput:
This pull request is deprecated by #5250

PR #5250, opened 2016-09-06
ofParameter::getFirstParent only iterates over expired parent pointers until valid one is found

markkorput:
Fixes issue #5248  and deprecates pull request #5249 

Though it's not entirely clear why this fixes issue #5248, it seems to do so. It should also be a bit more efficient, as it doesn't needlessly iterate over all obj->parents.

### DIFF
--- a/libs/openFrameworks/types/ofParameter.h
+++ b/libs/openFrameworks/types/ofParameter.h
@@ -516,16 +516,27 @@ public:
 	void setParent(ofParameterGroup & _parent);
 
 	const ofParameterGroup getFirstParent() const{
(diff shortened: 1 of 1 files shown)


arturoc:
Thanks for looking into this, we really want to delete all the invalid listeners so going thorugh the whole collection is necesary, usually there's no more than a couple of listeners at most per parameter so it shouldn't be a performance problem and the way it's done should be actually faster than a while loop and erasing elements one by one.

I've just uploaded a change that fixes it for me but if you can take a look just in case?

thanks!

PR #5251, opened 2016-09-06
Feature soundplayer::isPaused

thomasgeissl:
added missing isPaused getter

### DIFF
--- a/libs/openFrameworks/sound/ofBaseSoundPlayer.h
+++ b/libs/openFrameworks/sound/ofBaseSoundPlayer.h
@@ -33,6 +33,7 @@ public:
 	virtual float getPosition() const = 0;
 	virtual int getPositionMS() const = 0;
 	virtual bool isPlaying() const = 0;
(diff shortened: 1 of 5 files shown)


arturoc:
Hey thanks, this is missing the implementations of the method for emscripten and ios so it'll break those platforms if we merge it. I think it should be an easy fix, if you can take a look? you can see the errors in the tests output

thomasgeissl:
oh, I just realised. i will have a look and add it.

thomasgeissl:
@arturoc would you mind having a quick look. It is untested.
Also, there is a problem with the iOS machine.
```sh
$ cd openframeworks/openFrameworks
320.64s$ git fetch origin +refs/pull/5251/merge:
remote: Counting objects: 120905, done.
remote: Compressing objects: 100% (34343/34343), done.
fatal: write error: No space left on device
fatal: index-pack failed
```

arturoc:
looks good, i've just restarted the ios tests hopefully they'll finish now

PR #5252, opened 2016-09-07
ofFbo: add methods to clear the different buffers + allow no color

arturoc:
useful when using MRT and need to clear each buffer with different colors.

also allow no color buffer without warning since it's perfectly valid to have an fbo with depth only

### DIFF
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -451,6 +451,33 @@ void ofFbo::clear() {
 #endif
 }
 
+
+//--------------------------------------------------------------
+void ofFbo::clearColorBuffer(const ofFloatColor & color){
+	glClearBufferfv(GL_COLOR, 0, &color.r);
+}
+
+
+//--------------------------------------------------------------
+void ofFbo::clearColorBuffer(size_t buffer_idx, const ofFloatColor & color){
+	glClearBufferfv(GL_COLOR, buffer_idx, &color.r);
+}
+
+//--------------------------------------------------------------
+void ofFbo::clearDepthBuffer(float value){
+	glClearBufferfv(GL_DEPTH, 0, &value);
+}
+
+//--------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


libs/openFrameworks/gl/ofFbo.cpp
```diff
@@ -451,6 +451,33 @@ void ofFbo::clear() {
...
+	glClearBufferfv(GL_DEPTH, 0, &value);
+}
+
+//--------------------------------------------------------------
+void ofFbo::clearStencilBuffer(float value){
+	glClearBufferfv(GL_STENCIL, 0, &value);
+}
+
+//--------------------------------------------------------------
+void ofFbo::clearDepthStencilBuffer(float depth, float stencil){
```
tgfrerer: i believe stencil shoud be an int value (see: https://www.opengl.org/sdk/docs/man/html/glClearBuffer.xhtml)

libs/openFrameworks/gl/ofFbo.cpp
```diff
@@ -451,6 +451,33 @@ void ofFbo::clear() {
...
+}
+
+//--------------------------------------------------------------
+void ofFbo::clearDepthBuffer(float value){
+	glClearBufferfv(GL_DEPTH, 0, &value);
+}
+
+//--------------------------------------------------------------
+void ofFbo::clearStencilBuffer(float value){
+	glClearBufferfv(GL_STENCIL, 0, &value);
```
tgfrerer: Stencil should probably be cleared with `int` value, and `glClearBuffervi` used instead, according to the spec: 
see: https://www.opengl.org/sdk/docs/man/html/glClearBuffer.xhtml
"If buffer is GL_STENCIL, drawbuffer must be zero, and value points to a single value to clear the stencil buffer to. Masking is performed in the same fashion as for glClearStencil. Only the *iv forms of these commands should be used to clear stencil buffers; be used to clear stencil buffers"

arturoc: it's fixed now, thanks!

PR #5253, opened 2016-09-07
Conditionally disable all poco related code using a constant in ofConstants

arturoc:
this will disable ofXml, ofURLFileLoader and anything using Poco::URI, (the only remaining uses of poco in the core)

as a side effect it allows to remove openssl too since only poco uses it which can be useful for applications published in the app/play store or similar to not have to update openssl everytime a new version is released.


bakercp:
Are we ready to create an `ofxPoco`?  If so, would it make sense to make it a core addon?  Or are we planning to keep Poco in for one more release?

arturoc:
not yet, we still depend on poco for ofURLFileLoader, uri parsing and ofXml.

for network stuff https://github.com/cpp-netlib/cpp-netlib seems the best alternative, it also contains a uri parsing lib but last time we tried it we had to revert back to poco cause it was causing xcode to crash, not sure if that's fixed already

for xml we could look into libxml which would also allow to use the original svgtiny in ofxSvg

PR #5254, opened 2016-09-07
cleanup ofParameter's parents before starting the notification iteration

markkorput:
Another shot at resolving #5248. See also #5250.

While writing the unittest I noticed that the issue doesn't occur if ofxGui isn't getting involved with the ofParameterGroup, so I guess the issue (#5248) might be inside ofxGui (I haven't been able to find the culprit yet). On the other hand, if ofParameter does all these weak_ptr validity checks, it should probably be resistant against situations where there is actually an invalid weak_ptr among the obj->parents. Hence, still the patch to ofParameter.h

### DIFF
--- a/libs/openFrameworks/types/ofParameter.h
+++ b/libs/openFrameworks/types/ofParameter.h
@@ -680,35 +680,19 @@ inline void ofParameter<ParameterType>::eventsSetValue(const ParameterType & v){
         // Notify all parents, if there are any.
         if(!obj->parents.empty())
         {
-
-            // This lambda will conditionally notify a parent if its child
-            // value has changed.
-            //
-            // If it was successful (i.e. the parent pointer is valid) the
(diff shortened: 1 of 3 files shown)


arturoc:
that makes a lot of sense this probably came from a recursive call, delaying the notification to a second loop should solve it, can you change:

```
for(auto parent = obj->parents.begin(); parent != obj->parents.end(); parent++){
    parent->lock()->notifyParameterChanged(*this);
}
```

to:

```
for(auto & parent: obj->parents){
    auto p = parent.lock();
    if(p){
        p->notifyParameterChanged(*this);
    }
}
```

it's safer to do that otherwise the parent might be invalidated between one loop and the next

markkorput:
Of course. Done.

arturoc:
thanks!

Issue #5262, opened 2016-09-13
ofThread : short lived threads can cause uncaught exceptions


armadillu:
I am getting sporadic crashes on very short lived threads; This code is enough to reproduce the issue; at about one crash per minute. I am seeing this happen on OS X 10.11.

```
class MyThread : public ofThread{
public:
    void threadedFunction(){}
};

void ofApp::update(){
    MyThread* t = new MyThread();
    t->startThread();
}

```

It seems that every now and then the thread fails to detach() and throws an exception.

https://github.com/openframeworks/openFrameworks/blob/88545f1e54c928830eae568b98e816df8170e298/libs/openFrameworks/utils/ofThread.cpp#L172

Wrapping the call in a try/catch keeps the app running.

```
try{
    thread.detach();
}catch(...){
    ofLogFatalError("ofThreadErrorLogger::exception") << "failed to detach!";
}
```

arturoc:
can you check which kind of exception is throwing? a debugger should tell you that

armadillu:
thread::detach failed: Invalid argument

armadillu:
i tried checking if its joinable() b4 detaching first, but it happens too.

arturoc:
yes it's better to check the exception since the thread could be joinable when checking but become not joinable after. can you send a PR?

armadillu:
fixed in https://github.com/openframeworks/openFrameworks/pull/5263

armadillu:
I am going to reopen this bc I finally found some time to really look into it.

When you launch several short-lived ofThreads too suddenly (i.e. spawn a few threads in a single frame), sometimes the thread gets in a weird state.

I have an example demonstrating the issue [here](https://gist.github.com/armadillu/c7337cc8433bffe318d1266d126522f6). I have seen the issue both on OSX and windows.

If you run the code above, you will see printouts like this:

```
[ error ] ofApp: ofThread is done but we will crash if we try to delete - LEAKING
```

every few seconds. This is as far as I got with this, basically the thread object is done (and it executed fine) but the std::thread itself is in a weird state, so if you try to delete it it leads to crash. 
You can modify the ```USE_OF_THREAD``` macro and set it to false to see how it doesn't happen with an alternative implementation.

I think it is related to the way the ofThread is launched with : 

```
thread = std::thread(std::bind(&ofThread::run,this));
``` 
[...]

PR #5263, opened 2016-09-13
Fix for #5262 - Short lived threads can cause exceptions

armadillu:
Try-catch around ofThread deteach(), and only detach if not joinable

### DIFF
--- a/libs/openFrameworks/utils/ofThread.cpp
+++ b/libs/openFrameworks/utils/ofThread.cpp
@@ -172,11 +172,7 @@ void ofThread::run(){
 	if(thread.joinable()){
 		try{
 			thread.detach();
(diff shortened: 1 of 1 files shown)


arturoc:
can you change it to just:

```
try{
	thread.detach();
}catch(...){}
```

if it's something that just happens and is really not an error we don't want an error to show up in the console and checking for joinable is not very safe in any case so better to just try to detach it and ignore the exception i think

armadillu:
yep

arturoc:
can you remove the check for `joinable()` too? it doesn't even make sense to join a thread from itself so an implementation might even return false there if called from the same thread

arturoc:
thanks!

Issue #5264, opened 2016-09-15
Xcode 8/macOS 10.12 removed QtKit support


gottfired:
Just updated to Xcode to the latest version. QtKit is not supported anymore with BaseSDK 10.12 which leads to compile errors.

ofZach:
relevant: 

https://forum.openframeworks.cc/t/quicktime-quicktime-h-file-not-found/24494/3
https://forum.openframeworks.cc/t/macos-10-12-sierra-xcode-8-quicktime-deprecation-workaround/24179/1

arturoc:
yeah there's a couple of fixes for latest versions of qtcreator and i
think we still haven't done a release for some library that wasn't
compiled for 32bit in osx so if someone can send a PR for this against
the 0.9.4 branch we can do a release right away

joaodafonseca:
Would be really helpful to have OF building in Xcode 8 as soon as possible. I'm in a middle of a IOS project and need to build it to IOS 10. Any ideas when is this going to happen? Thanks

ofZach:
@joaodafonseca is there an IOS issue that's different then the quicktime issue?

this is the workaround for osx which should fix things: 

https://forum.openframeworks.cc/t/macos-10-12-sierra-xcode-8-quicktime-deprecation-workaround/24179

joaodafonseca:
Hi @ofZach thanks for the reply. I don't get any issue with quicktime in IOS, this is what I got:
![screen shot 2016-09-15 at 15 09 20](https://cloud.githubusercontent.com/assets/1932216/18550946/a4d93a56-7b56-11e6-9ac9-7616b264aab1.png)

Any thoughts?

joaodafonseca:
Ok so I found a way to build it for IOS 

Targets -> Build Settings -> Build Active Architecture Only -> set to 'YES'

And now it Builds, no complains about the quicktime issue.

Sorry if this was off topic

ofTheo:
Made a PR for OS X here: #5267 If anyone with 10.12 can test that would be helpful!

ofTheo:
This is actually a bit more complicated than just adding the preprocessor guards. 
Currently we are using AVFoundation for video file playback, but we are using QTKit for video camera capture.  So with the above PR ( #5267 ) there is still an error that OF_VIDEO_PLAYER_QTKIT is not defined. 

[...]

PR #5267, opened 2016-09-16
preprocessor fixes for OS X 10.12 for QTKit and QuickTime

ofTheo:
Don't merge just yet. 
Running some checks on 10.11 / Xcode 8

Should address the issue in #5264

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -256,8 +256,11 @@ typedef TESSindex ofIndexType;
 		//on 10.6 and below we can use the old grabber
 		#ifndef MAC_OS_X_VERSION_10_7
 			#define OF_VIDEO_CAPTURE_QUICKTIME
-		#else
+		//if we are below 10.12 or targeting below 10.12 we use QTKit
+		#elseif !defined(MAC_OS_X_VERSION_10_12) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_10_12
 			#define OF_VIDEO_CAPTURE_QTKIT
+		#else
+			#define OF_VIDEO_CAPTURE_AVF
         #endif
 
 	#elif defined (TARGET_WIN32)
--- a/libs/openFrameworks/video/ofAVFoundationGrabber.h
+++ b/libs/openFrameworks/video/ofAVFoundationGrabber.h
@@ -1,0 +1,121 @@
+/*
+ *  AVFoundationVideoGrabber.h
+ */
+
+#pragma once
+#include "ofConstants.h"
+
+#ifdef OF_VIDEO_CAPTURE_AVF 
+
+//------
+#include "ofBaseTypes.h"
+#include "ofPixels.h"
+#include "ofTexture.h"
+#include "ofThread.h"
+
+#ifdef __OBJC__
+
+#import <Foundation/Foundation.h>
+#import <AVFoundation/AVFoundation.h>
+#import <Accelerate/Accelerate.h>
+#import <CoreMedia/CoreMedia.h>
+#import <CoreVideo/CoreVideo.h>
+
+class ofAVFoundationGrabber;
+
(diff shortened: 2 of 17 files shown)


ofTheo:
Okay this branch should build now for the 10.12 SDK. 
( I tested on 10.11 but building against 10.12 SDK which previously didn't work with 0.9.3 download ) 

The video grabber took a bit of work and there are a couple of small things marked in there as TODOs: but it seems to work well with the cameras I had available to test. 

Would be great if others could try this out. 
Ping @danzeeeman @danoli3 @i-n-g-o

danoli3:
I'll have a look! Awesome work Theo!

i-n-g-o:
hi

compiled without problems on 10.11 with XCode Version 8.0 (8A218a) and 10.12 SDK

running the grabber example with internal FaceTime camera crashes in ofAVFoundationGrabber when setting up the device:
```
device.activeVideoMinFrameDuration = CMTimeMake(1, framerate);
device.activeVideoMaxFrameDuration = CMTimeMake(1, framerate);
```

> An uncaught exception was raised
> The activeVideoMinFrameDuration passed is not supported by the device.  Use -activeFormat.videoSupportedFrameRateRange to discover valid ranges.

using one of `device.activeFormat.videoSupportedFrameRateRanges` does not improve the situation...
the only chance to make it run for me is to disable the two lines.

ofTheo:
Thanks for checking! 
Yes, that was the issue I ran into with the orbecc camera too. Had to comment those lines out, even though I was giving valid framerates. 

Going to look into this issue more as well as performance difference between specifying rgb vs bgra capture in the setup.

i-n-g-o:
there is something weird with min/max rates.

i set the framerate to 15, but it tries to set 29.97.
that is, because first entry in `supportedFrameRates` is 30fps. would need to check if desired rate is within min/max.

maybe better then to use `AVFrameRateRange` and set it's `minFrameDuration` and `minFrameDuration`?
this is from an example:
```
device.activeVideoMinFrameDuration = bestFrameRateRange.minFrameDuration;
device.activeVideoMaxFrameDuration = bestFrameRateRange.maxFrameDuration;
```

ofTheo:
Thanks @i-n-g-o ! 
I think I got a fix for the fps stuff based on your comment.  It works for me at least. 

I also fixed some issues where you could get a crash if you specified a capture size that wasn't supported by the device. Now it finds the closest matching dimensions to the size requested.  

I also switch the capture format back to BGRA and am now using the accelerate framework to do conversion to RGB.  This should support more devices, as some won't give you a raw RGB stream. 

I think this should be good to merge for the 0.9.4 hotfix release.  
Would be great if you and others could try it out with their setups. 

Cheers!
Theo

i-n-g-o:
nice. great work!

looks good to me. :)

ofTheo:
Thanks @i-n-g-o 
Does this branch now not crash for you?

Cheers!
Theo

i-n-g-o:
sorry, to have been too unspecific...
yes, the branch now does not crash anymore for me.

but, i had another look to verify that the desired framerate gets set...
it seems it needs a `ceil` instead of a `floor` for `range.maxFrameRate`:

`if( (floor(range.minFrameRate) <= framerate && ceil(range.maxFrameRate) >= framerate) ){`

otherwise e.g. framerate 15 does not get set.
in my case both `minFrameRate` and `maxFrameRate` gives 14.999993 when logging with NSLog. using floor for min/max let's it fail and the framerate does not get set.

cheers
io

ofTheo:
Thanks @i-n-g-o - that changes also worked for my devices too.  
Okay, I think this is in a pretty good place to merge. 
Will merge tomorrow with 0.9.4 branch

heaversm:
This merge should have resolved the issues with the videoGrabber and videoPlayer examples, correct? I am still having a problem building these projects on OSX 10.12 with XCode 8.3:

screen shot 2017-04-06 at 10 42 45 am

PR #5269, opened 2016-09-19
enable RTS/CTS handshake and enable raw mode on serial port

edovino:
without this ofArduino and ofSerial don't work (on F24, at least)

### DIFF
--- a/libs/openFrameworks/communication/ofSerial.cpp
+++ b/libs/openFrameworks/communication/ofSerial.cpp
@@ -361,6 +361,10 @@ bool ofSerial::setup(string portName, int baud){
 		options.c_iflag &= (tcflag_t) ~(INLCR | IGNCR | ICRNL | IGNBRK);
 		options.c_oflag &= (tcflag_t) ~(OPOST);
 		options.c_cflag |= CS8;
+		#if defined( TARGET_LINUX )
+			options.c_cflag |= CRTSCTS;
+			options.c_lflag &= ~(ICANON | ECHO | ISIG);
+		#endif
 		tcsetattr(fd, TCSANOW, &options);
 
 		bInited = true;


bakercp:
Thanks @edovino.  While we're on the topic -- @arturoc a few people have asked before, and I'm wondering if the next release would be a good time move [ofxSerial](https://github.com/bakercp/ofxSerial) to become a core addon?  It already supports RTS/CTS, etc.  @arturoc and a lot of other features that keep getting tacked onto the core ofSerial object.  At the very least, just backing `ofSerial` with https://github.com/wjwwood/serial (which is what ofxSerial does) would be a good step in the right direction.  It's a very solid library.

arturoc:
@edovinohave you tried it in osx? if not can you ifdef the changes to linux?

@bakercp i don't think this change would have been solved if we had used the library you point out instead of raw serial. from what i understand this kind of changes need to be done from time to time cause changes in the kernel or in arduino make it necessary to use different settings for the serial communication to work, so even if we exposed those arguments we would still need to add this settings somewhere since most people won't figure out the correct settings.

We could add have a settings object which passed to ofSerial has the correct serial settings to work with arduino, something like ofSerial::ArduinoSettings or ofArduino::SerialSettings to make the api cleaner and perhaps expose those settings in ofSerial as methods if it makes sense to change them after the port has been opened

i'm checking the library anyway and it seems not to be very standard (it's not in any linux distro or mac repos like homebrew...) it's not been maintained since last year although arguably you don't need much maintenance for something like serial comm but it also has some weird implementation like it's locking a mutex on every read/write operation which seems unnecessary or a setup method with around 8 arguments.

i would try to avoid adding more libraries unless is something hard to implement or makes maintenance much easier and in this case it seems we would just need to have some settings objects for setup which doesn't seem like it would be much work to add to the current implementation.

if you would want to take a look at that, that would be really useful.

edovino:
@arturoc no, didn't check on os x - I'll dust of my macbook air & try tonight. Some of my students have macbooks, and apparently they don't seem to need the change (although I saw some comments in the forum that suggested otherwise)
If I can't find anything definitive, I'll ifdef it for Linux
And yes, it does seem to be kernel-related: I had no trouble running without this patch on F23, but on recent F24 this suddenly broke.

edovino:
replaced by pull request #5280

PR #5270, opened 2016-09-20
Add ofSkew function

Furkanzmc:
Add a function to support skewing.

### DIFF
--- a/libs/openFrameworks/graphics/ofGraphics.cpp
+++ b/libs/openFrameworks/graphics/ofGraphics.cpp
@@ -336,6 +336,19 @@ void ofRotateZRad(float degrees){
 }
 
 //----------------------------------------------------------
+void ofSkew(const ofVec2f& skew){
+    const float skewMatArray[16] = {
+        1, tanf(ofDegToRad(skew.x)), 0, 0,
+        tanf(ofDegToRad(skew.y)), 1, 0, 0,
+        0,  0,  1, 0,
+        0,  0,  0, 1
+    };
+    ofMatrix4x4 skewMatrix(skewMatArray);
+    const ofMatrix4x4 currentMatrix(ofGetCurrentRenderer()->getCurrentMatrix(ofMatrixMode::OF_MATRIX_MODELVIEW));
+    ofGetCurrentRenderer()->loadMatrix(currentMatrix * skewMatrix);
+}
+
+//----------------------------------------------------------
 void ofLoadIdentityMatrix (void){
 	ofGetCurrentRenderer()->loadIdentityMatrix();
 }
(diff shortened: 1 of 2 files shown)


Furkanzmc:
I don't think that failed check is about me. But if it is, please tell me.

arturoc:
Thanks for your contribution. I'm really not sure if we want to include more global functions like this so i'll let others give their opinion.

In any case some things that are wrong:

- ofMatrix4x4 and ofVec2f are now deprecated, any new functionality should be using glm::mat4 and glm::vec2
- The way this is done is completely different to other functions like ofTranslate or ofRotate. this should be done in the renderer itself not getting the matrix from the renderer and pushing it again.

As i said i'm not sure we want this as a global function so perhaps wait to get an approval before putting more work on it.

thanks!

Furkanzmc:
Although I'm using openFrameworks I haven't contributed before so I wasn't aware of the deprecated classes, sorry for that. I'll wait for the others. Thanks.

bakercp:
I think this is a valuable contribution (in fact I just had a student ask if a function like this existed the other day).  It makes sense to me that we would have this functions to bring it into parity with CSS3, etc.  That said, I'd recommend following our other recent patterns post glm and have the following functions ...

```c++
void ofSkewRad(const glm::vec2& skew);
void ofSkewDeg(const glm::vec2& skew);
void ofSkewXRad(float angle);
void ofSkewYRad(float angle);
void ofSkewXDeg(float angle);
void ofSkewYDeg(float angle);
```

tgfrerer:
Like @arturoc, I'm not too sure about adding it as a global method. Ideally, we keep the core relatively light, and global methods to a minium.

My main concern is that i'm not too sure it's clear what is meant by `ofSkew`: is it the model, or the camera - or both the camera and model matrices that are being skewed?

The current code will modify the matrix stack by (mathematically) pre-multiplying a skew matrix onto the modelview matrix. So the vertex calculation looks like (in GLSL matrix multiplication order, which is the reverse of the openFrameworks matrix multiplication order):

`vertex = projectionMatrix * skewMatrix * viewMatrix * modelMatrix * vec4(vertexAttributePosition);`

So the attribute first gets transformed by model, then view, then skew, then projection, until it ends up in projection space.

This could lead to unexpected results if you modify the current camera in code following such a call, as it modifies the "view" part. 

Also, with skew applied, there is a very high chance that your modelViewMatrix ends up non-orthogonal, which means you have to be super careful when calculating normal matrices, for example. Introducing it could bring a bit of technical debt down the line, in making sure all matrix stack operations interact properly, once skew is applied.

I think if we add skew, or if we modified the matrix stack in such a way, it is probably best to do it on the surface - in `ofApp` code - so that it's clearer to see what's going on.

Furkanzmc:
I'm not familiar with the renderers so I have to take the time to get myself familiar, which I'm happy to do. But it's going to take a few days, I'm busy with work these days. 😄 But If nobody takes it up, I'm happy to implement this in a few days. Thanks for your opinions.

PR #5279, opened 2016-09-24
QtCreator fixes for MSYS2

oxillo:
Add some checks on PATH when using QTCreator with MSYS2.


arturoc:
That looks great thanks so much! i was looking into using rule instead of transformer some days ago but was missing the multiplex so it wouldn't work.

We should go through all the other msys2 pr's you have at some point and merge them into master.

thanks!!

PR #5280, opened 2016-09-24
on linux, set serial port to raw mode

edovino:
this makes ofArduino & ofSerial work on F24 linux. OS X doesn't appear to need this (anymore?)


edovino:
as suggested by @arturoc, only ifdefd for linux. OS X doesn't appear to need this anymore.

arturoc:
thanks!

PR #5297, opened 2016-10-04
Use requested pixel type, not the current (allocated) member type.

bakercp:
This is a typo / bug as it ignores the requested `_pixelFormat` in the switch statement.


arturoc:
thanks!

PR #5301, opened 2016-10-07
add some const to ofxOsc

2bbb:


addons/ofxOsc/src/ofxOscMessage.h
```diff
@@ -48,10 +48,10 @@ class ofxOscMessage
 	void clear();
 
 	/// return the address
-    std::string getAddress() const { return address; }
+    const std::string &getAddress() const { return address; }
```
arturoc: can you change the return types back to std::string? it doesn't add much and forces the internal implementation to have that particular variable as an instance variable string

arturoc:
thanks!

PR #5303, opened 2016-10-08
Make Run is not a target on Ubuntu

SpacePossum:
The target for `make` is case sensitive (at least on Ubuntu).

There for the current `make Run` fails as the target is `run`;
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk#L69

Tried on;
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial


bakercp:
It seems like you might be trying to merge the master branch into 0.9.4, rather than you single commit?

arturoc:
yes, also the target is run not Run and that's what the documentation says as well no?

I'm going to close this anyway since the pull request is not correct but feel free to keep commenting here if you still think there's something wrong as it is

SpacePossum:
Indeed I picked the wrong target branch :(

>yes, also the target is run not Run and that's what the documentation says as well no?

it is not; http://openframeworks.cc/setup/linux-install/ states `make Run`
I'll try a new PR soon ;)

arturoc:
ah ok, no then what we need is to change the documentation, can you send a PR to ofSite instead?

bakercp:
@SpacePossum did have a useful commit here https://github.com/openframeworks/openFrameworks/pull/5303/commits/b7ea52e6ef777f267e11d507b1981b0c78a93d0d, it just got clouded by the wrong target.  @SpacePossum, could you resubmit this clean commit AND submit a fix to ofSite :)

Thanks for the sharp eyes!

arturoc:
oh, ok, sorry i didn't understood, thought you wanted to change the makefiles not the readme. go for it then. thanks!

PR #5306, opened 2016-10-10
Typos

SpacePossum:


arturoc:
thanks!

SpacePossum:
thank you @arturoc :)
I was looking into what happend here btw. https://github.com/openframeworks/openFrameworks/pull/5306/files#diff-957be4ca71de2a401a9191c572196480L52 not sure if that change was OK (happend during a rebase I think)

arturoc:
it's probably just a breakline or something like that shoulnd't be a problem

PR #5322, opened 2016-10-22
Refactor filesystem

arturoc:
substitutes this PR:

as agreed there the filesystem now looks like:

```
 addons/
 apps/
 docs/
 examples/
 build/
  |_projects
  |_lib
 src/
 tests/
 scripts/
```

where libs now is gone and only downloaded from the ci servers (will be present in the web download) and only contains the external dependencies

build now contains the projects in projects and the compiled OF lib in lib

the openframeworks source itself is in src

all the changes have been done with git mv so old PRs should still work but i'll try to merge something onto this new structure before merging to test it it breaks things to much in which case it might be not worth it


ofZach:
I am pretty supportive of this refactor (especially tests!), some questions to throw out there -- 

a) does it make sense to have examples in build?  Or projects outside of build?   Or a different name?  I think "build" as a folder name might be a little opaque as to what it has inside it.     IMO build sounds like the end result of something (like .o files go there) vs something more active like project files which you'd be interacting with alot.    Just thinking about consistency where examples are easy to find but your own projects are a level deeper.  

b) does it make sense to move templates into it's own folder?  I've always felt like it's a little odd having it in the script folder -- maybe it's cleaner to seperate it out or does that leave too many root folders?  

c) if we are doing a refactor of folders, does it make sense to move to separate folders for project files (ie, xcodeproj, sln, etc) since this can get kind of ridiculous now for cross platform development? this is a change we can do further down the line, but I think it would clean things up a bit and we've discussed it earlier when we've run into conflicts.

arturoc:
tests are already there :) https://github.com/openframeworks/openFrameworks/tree/master/tests  they run every time we do a commit on the 3 desktop platforms

a) build is the build system for OF itself (project files and makefiles) and also the build results, .o files and also the OF library itself, it's not related with your projects

b) yeah probably it does, the main problem with moving templates or anything in scripts is that there's a lot of tools out of this repo that depend on that location, like the CI servers, the project generator, all the IDE wizards, the package generation scripts... so moving them is kind of a pain.

c) yes but same problem as with templates really.

All this makes me think that it's probably a better idea to leave this for next release that way we'll have a  whole release cycle to test and fix stuff rather than merging it right before doing a major release

ofZach:
ah sorry I misunderstood about projects :)    sounds good to hold off on b/c at least and do a larger re-organization at once...

PR #5326, opened 2016-10-24
Fix missing openssl includes for Poco.

bakercp:
In the current version, the project generator from 0.9.6 does not include search paths for openssl, leading to failures using SSL features.


arturoc:
thanks!

PR #5331, opened 2016-11-01
include Xinerama in linux of.qbs

tgfrerer:
qt composer project file was missing Xinerama for linux compiles. Xinerama is already included in makefiles. This only affects builds through the Qt gui, which may otherwise fail linking.


arturoc:
thanks!

PR #5335, opened 2016-11-04
changed ofToFloat to ofToDouble in ofxXmlSettings

MatillaMartin:
It is parsing a double, it should convert to double, or we will lose precision


arturoc:
thanks!

PR #5340, opened 2016-11-07
rename check to check_op in ofParameter SFINAE.

2bbb:
Xcode has terrible macro `check` in "AssertMacros.h".
if we include AssertMacros.h, then ofParameter got error.
and, AssertMacros.h is include too deep layer.

other way,
`#undef check`
or,
```
#define MACRO_DISABLER
	template <class T> yes check MACRO_DISABLER (T const&);
	no check MACRO_DISABLER (no);
...
	static const bool value = sizeof(check MACRO_DISABLER (stream >> x)) == sizeof(yes);
...
	static const bool value = sizeof(check MACRO_DISABLER (stream << x)) == sizeof(yes);
...
```

i think, rename is simple most better way.

### DIFF
--- a/libs/openFrameworks/types/ofParameter.h
+++ b/libs/openFrameworks/types/ofParameter.h
@@ -363,22 +363,22 @@ namespace priv{
 	no operator << (const anyx &, const anyx &);
 	no operator >> (const anyx &, const anyx &);
 
-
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

PR #5341, opened 2016-11-09
Feature/fix some typo

edap:
I've just fixed two small typos and remove whithespaces. Also, the pdf linked in the header is not existing anymore. Can these resource http://www.bantherewind.com/wrap-your-mind-around-your-gpu be a valid replacement?


arturoc:
thanks!

Issue #5344, opened 2016-11-14
ofxGui save and load are broken with new ofXml api.


roymacdonald:
Hi,
with the current HEAD I cannot save to xml as the ofSerialize and ofDeserialize were not added to the new implementation of ofXml. I have a working version of for it. Should I PR?

Also saving to json works but loading is half working as it only loads basetype ofParameters. It will not load ofColor or ofVecs. Any clues on this?
@arturoc 

all the best!

roymacdonald:
Ok. so Here's the fix
https://github.com/roymacdonald/openFrameworks/commit/120760f2725363fb1b33ff866f218d3b42690b28
The problem with saving to json that the nested groups were not being parsed. I've also added this fix in the commit. 
PR?

arturoc:
yes please can you send a PR? thanks!

PR #5345, opened 2016-11-14
Fixed ofxGui save and load.

roymacdonald:
-Added ofSerialize and ofDeserialize methods to ofXml, using the new ofXml api.
-Fixed ofDeserialize in ofJson, as it was not reading the nested groups.

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -1,9 +1,7 @@
 #include "ofxBaseGui.h"
 #include "ofImage.h"
 #include "ofBitmapFont.h"
-#if OF_USE_POCO
 #include "ofXml.h"
-#endif
 #include "ofJson.h"
 using namespace std;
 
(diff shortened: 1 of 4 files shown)


roymacdonald:
https://github.com/openframeworks/openFrameworks/issues/5344

roymacdonald:
Hi @arturoc did you see the CI results for this PR? it seems that on windows when compiling with VS there are some library linking issues.
Also on linux, if fails only when compiling for armv6 as the SAX lib is not found. 
I have very little experience with OF on both this platforms so I'm not sure how to help on this.
all the best!

arturoc:
yes i'm working on it, linux is fixed now, so if you rebuild it it should be ok, windows needs the debug version for one of the new libraries we use but i'm traveling and my windows partition has somehow expired so it'll be take a while for me to fix it

roymacdonald:
great! I have windows on bootcamp and parallels. I can try to figure it out. Any advice? shouldn't it be done via apothecary?
best!

PR #5347, opened 2016-11-15
fixing cameraLensOffsetExample call to inverse()

agoodney:
There are a bunch of small errors that keep some of the examples from building cleanly. This is the first of a series of PRs to fix these problems.


agoodney:
Don't want two commits in one PR...

Issue #5315, opened 2016-10-18
DoesDirectoryExist can crash the app


Taprik:
I'm experiencing crash when I call the doesDirectoryExist function. The path searched does not exist and the app crash.
The setup : windows7, OF 0.9.3.
It's not crashing the app with windows 10.
The solution is to use this version of doesDirectoryExist
`bool ofDirectory::doesDirectoryExist(const std::string& _dirPath, bool bRelativeToData){
    std::string dirPath = _dirPath;
    try {
        if (bRelativeToData) {
            dirPath = ofToDataPath(dirPath);
        }
        return std::filesystem::exists(dirPath) && std::filesystem::is_directory(dirPath);
    }
    catch (std::exception & except) {
        ofLogError("ofDirectory") << "doesDirectoryExist(): couldn't find directory \"" << dirPath << "\": " << except.what()<<endl;
        return false;
    }
}`

dimitre:
closed by https://github.com/openframeworks/openFrameworks/issues/5315

PR #5348, opened 2016-11-16
Avoid crash of doesDirectoryExists function

Taprik:
This function makes the app crash on win7. I put a try catch to avoid
the crash.
issue #5315


Taprik:
Also adding a sort by date function.
Sorry for the mess, it's the first time I made a pull request

arturoc:
thanks, this seems to be breaking msys2 though. looks like the stat strcut doesn't exist there.

can you use instead:
```
auto ta = std::filesystem::last_write_time(a);
auto tb = std::filesystem::last_write_time(b);
return ta < tb;
```

for the time comparison function?

Taprik:
It works perfectly for me (of 0.9.3)
Thanks Arturo

arturoc:
can you add it to the PR?

arturoc:
looks good, can you also make sortByDate behave the same as sort? where if the files list is empty it'll call listDir before? thanks!

arturoc:
thanks! i'll let the test finish and merge it later

PR #5351, opened 2016-11-17
simple fix to get inverse

agoodney:
There are a number of examples that do not build cleanly. This is the first in a series of PRs to fix the examples.


arturoc:
hey, thanks! can you use glm::inverse instead? not sure how it even works using inverse directly since we don't include the glm namespace anywhere

agoodney:
Updated to use glm::inverse().

arturoc:
thanks!

PR #5352, opened 2016-11-17
fix to get the mesh verticies

agoodney:
mesh.getVerticies() doesn't return the correct vector type. This fix takes the resulting vector and creates the correct type for the remainder of the code.

### DIFF
--- a/examples/3d/modelNoiseExample/src/ofApp.cpp
+++ b/examples/3d/modelNoiseExample/src/ofApp.cpp
@@ -105,12 +105,8 @@ void ofApp::drawWithMesh(){
 	float liquidness = 5;
 	float amplitude = mouseY/100.0;
 	float speedDampen = 5;
-    auto t_vect = mesh.getVertices();
-    vector<ofVec3f> verts;
-    for(unsigned int i = 0;i < t_vect.size(); i++){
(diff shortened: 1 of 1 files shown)


examples/3d/modelNoiseExample/src/ofApp.cpp
```diff
@@ -105,7 +105,12 @@ void ofApp::drawWithMesh(){
 	float liquidness = 5;
 	float amplitude = mouseY/100.0;
 	float speedDampen = 5;
-	vector<ofVec3f>& verts = mesh.getVertices();
+    auto t_vect = mesh.getVertices();
```
tgfrerer: I fear this won't fix the issue you've found. Note that `verts` was a _reference_ to a vector of vertices, i.e. a shorthand to directly manipulating the vertices inside the mesh. 

Could you try and check if the following line fixes the compile issues:

`auto & verts = mesh.getVertices();`

This wouldn't need the "manual" conversion.

agoodney: Yes, that fixes it. And, LOL, now I see what the demo is supposed to do.

arturoc:
thanks!

PR #5355, opened 2016-11-18
initilize verticies without using set()

agoodney:
Open to comment on the type that I ended up using for the templated vector type. Is there an of* type that will yield the same result?

### DIFF
--- a/examples/gl/billboardExample/src/ofApp.cpp
+++ b/examples/gl/billboardExample/src/ofApp.cpp
@@ -16,10 +16,8 @@ void ofApp::setup() {
 	for (int i=0; i<NUM_BILLBOARDS; i++) {
 
 		billboardVels[i].set(ofRandomf(), -1.0, ofRandomf());
-		billboards.getVertices()[i].set(ofRandom(-500, 500),
-													ofRandom(-500, 500),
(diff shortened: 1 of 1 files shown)


arturoc:
you can just do:

```
billboards.getVertices()[i] = glm::vec3(ofRandom(-500, 500),
							      ofRandom(-500, 500),
							      ofRandom(-500, 500));
```

or

```
billboards.getVertices()[i] = {ofRandom(-500, 500),
                                             ofRandom(-500, 500),
                                             ofRandom(-500, 500)};
```

better to use the last one since that would be compatible with ofVec3f too.

btw, you can make OF use ofVec* by default by uncommenting line 926 in ofConstants.h:

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofConstants.h#L926

right now the examples haven't been ported yet to use glm and will probably remain like that for the next release which will release with that line in ofConstants uncommented so the default still is ofVec. 

we still have to decide on that but in any case any fixes to the examples have to remain compatible with the old syntax by now and not introduce a mix of glm + ofVec* 


thanks!!

agoodney:
Updated to {} style.

arturoc:
thanks!

PR #5361, opened 2016-11-22
Missing texCoordVarying in vert shader example

MatillaMartin:
Added the copy from vert to frag for clarification of use in the example


arturoc:
i think the correct thing would be to remove that variable from the fragment since it's not use at all?

MatillaMartin:
true, is there another shader example that shows how to send the variable
from the vertex to the fragment?

MatillaMartin:
ok yep, there are many examples in examples/shader! Modifying pull request

arturoc:
thanks!

PR #5373, opened 2016-12-01
Chip dependency scripts

zealtv:
This branch contains installation scripts for the CHIP single board computer.  They are the same as the Debian scripts with mesa related packages removed.


arturoc:
thanks!

PR #5375, opened 2016-12-01
assimp: add missing .dll copy step to vs

tgfrerer:
Adds a missing step to the vs addon_config for assimp so that the project generator knows to copy the assimp dlls into the app's bin directory when generating or updating a project which uses the `ofxAssimpModelLoader` addon. This depends on a [fix in apothecary ](https://github.com/openframeworks/apothecary/pull/64)which assures that the 32 and 64 bit libs for assimp have different names.


arturoc:
this isn't necesarry anymore since the PG now copies all the dlls from the file system

PR #5384, opened 2016-12-04
Add 12mb/s communication over ofSerial.cpp

mcanet:
12mb/s (baud rate 12000000) communication it is the Teensy maximum serial speed communication(http://www.pjrc.com/teensy/benchmark_usb_serial_receive.html). I tested in linux and windows and works good.


arturoc:
thanks mar!

PR #5388, opened 2016-12-06
Handle return and enter key presses properly

xseignard:
This PR translates keypresses from `KEY_ENTER` and `KEY_KPENTER` properly to the right OF value.

See: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L1287-L1292

And: https://forum.openframeworks.cc/t/keyboard-layout-problem/25248


arturoc:
thanks!

Issue #5386, opened 2016-12-05
ofxGui ofxLabel draws outside ofxPanel when the string is too long


npisanti:
When adding a `ofParameter<string>` or `ofxLabel` with a label too long for the `ofxPanel` width the label overflows the border and keeps drawing outside the `ofxPanel`. It is simple to fix by changing `ofxLabel` `generateDraw()` method like this:

```cpp
void ofxLabel::generateDraw(){
    
    bg.clear();

    bg.setFillColor(thisBackgroundColor);
    bg.setFilled(true);
    bg.rectangle(b);

    string name;
    if(!getName().empty()){
    	name = getName() + ": ";
    }
    name += (string)label;    
    
    // resize the string inside the max width
    if(font.isLoaded()){ // using font
        while( font.stringWidth(name) > (getWidth() - textPadding*2.0f) ){
            name.resize(name.size()-1);
        }  
    }else{ // using bitmap font
        int max = (getWidth() - textPadding*2.0f) / 8.0f;
        name.resize(max);
    } 
    
    textMesh = getTextMesh(name, b.x + textPadding, b.y + b.height / 2 + 4);

}
```

arturoc:
hey thanks! can you send a pull request?

npisanti:
just did it! =)

PR #5390, opened 2016-12-08
fixes issues #5386, ofxLabel overflowing when the string is too long

npisanti:
as the title says, it fixes the displaying of an ofxLabel when the label is too long for the ofxPanel boundaries


arturoc:
thanks!

PR #5391, opened 2016-12-10
fix call to .addVerticies()

agoodney:


arturoc:
thanks, this is not correct though since it's only adding 1 point instead of all of them. the correct fix would be to port ofxOpenCv to use glm but we are not sure yet if next release will have ofVec still as default by enabling OF_USE_LEGACY_MESH in line 926 of ofConstants.h or we'll port the missing examples and addons to use glm instead.

going to close this since the fix still has to be decided but thanks anyway

PR #5392, opened 2016-12-10
.serialize() appears to be deprecated

agoodney:


arturoc:
hey, thanks, the problem with this PR is that it removes the serialization completely. the correct fix would be to use:

```
ofSerialize(settings, parameters)
```

and ofDeserialize. can you add those?

agoodney:
So I'm pretty sure my change doesn't fix this. At least in Xcode it seems like the right header file to find the XML version of ofSerialize/ofDeserialize isn't being included. The JSON version is found though.

Any advice?

arturoc:
i think the problem is that you are not up to date with latest master where those functions are already there

arturoc:
going to merge this anyway since it seems to have passed the tests

PR #5400, opened 2016-12-18
iOS EAGLView - prevent flicker by synchronously drawing the first frame when starting animation

tallavi:
A small bug, mostly affecting rare scenarios (like mine) with multiple view controllers drawing OF content. This one line fix draws the first frame right away, so that the previous content is not shown.


danoli3:
Tested working!

PR #5401, opened 2016-12-18
Android - OFAndroidLifeCycle - cleanup

tallavi:
Cleanup of old stack management code. Not needed now that we have reference count for gl activities.


arturoc:
thanks!

PR #5407, opened 2016-12-28
AVFoundationGrabber: Avoid integer and floating point comparison

r9y9:
`dimentions.width` and `dimentions.height` are `int32_t`-valued, so it doesn't need to be floating point values.


danoli3:
Nice! Yeah definitely was wrong datatype.

```
typedef struct {
    int32_t width;
    int32_t height;
} CMVideoDimensions;
```

arturoc:
looks good to me, thanks!

PR #5411, opened 2017-01-05
Update PLATFORMS.md

brightredchilli:
Hi there, I just cloned openFrameworks and discovered that I needed to run download_libs in the scripts folder. I found this deep in an oF thread(https://forum.openframeworks.cc/t/0-9-4-release-candidate-1/24627?page=2) and thought someone might benefit from this.

Let me know if this is not the appropriate place, I'm happy to move, but I do think it will be helpful for newbies like myself.


brightredchilli:
My bad, I keep missing the notes in main READMEs!  Now that I see it it is clear as day.

PR #5425, opened 2017-01-24
Add GLM_META_PROG_HELPERS to mimic previous ofVec2f/3f/4f DIM functio…

bakercp:
…nality.

Allows us to call:

```
glm::vec2::components // etc
```
to get a get the number of components.  Works the same as `ofVec2f::DIM`, etc.

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -67,6 +67,7 @@ enum ofTargetPlatform{
 #include <unordered_map>
 #include <memory>
 
+#define GLM_META_PROG_HELPERS
 #define GLM_SWIZZLE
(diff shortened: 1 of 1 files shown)


arturoc:
you can already use size() to get the number of coordinates, is that not enough?

bakercp:
`glm::vec*::size()` is not static like `ofVec*f::DIM`, so can't deduce template parameters.  Using it here for example:

https://github.com/bakercp/ofxSpatialHash/blob/master/libs/ofxSpatialHash/include/ofx/KDTree.h#L17-L36

bakercp:
@arturoc any additional thoughts on this?

bakercp:
@arturoc thanks!

PR #5427, opened 2017-01-25
s/face/vertex in a comment about normals

hamoid:
This comment could imply that normals are per-face instead of per-vertex, and it ends up in the documentation.


arturoc:
thanks!

PR #5430, opened 2017-01-26
Fixes MAKEFILEDEBUG functionality broken by ec057361d170bd6b3ca64e54a…

bakercp:
Commit ec057361d170bd6b3ca64e54a32e89caaacd4fd5 broke `MAKEFILEDEBUG` functionality.

For some reason, that commit replaced _all_ spaces with tabs.  Commands (`$(info ...)`) that are indented with tabs before the first target cause the script to fail.  There are still other places where, with the right combination of variables commit ec057361d170bd6b3ca64e54a32e89caaacd4fd5 will cause failures, but I'm taking a minimal approach to undoing that commit.


arturoc:
spaces are not legal in makefiles so that commit was removing them which is correct but as you say the way it was done will break other stuff. we can revert it back and then remove the spaces later

bakercp:
@arturoc status?  Is it "later" yet?

arturoc:
i think this has been fixed meanwile but feel free to rebase and merge if not

PR #5434, opened 2017-01-30
debug specific defines (preprocessor macros)

markkorput:
This implement the PROJECT_DEFINES_DEBUG variable in the config.project.mk so you can specify PROJECT_DEFINES (preprocessor definitions) in your project's config.mk to distinguish in-code between Debug and Release builds, like so;

> PROJECT_DEFINES_DEBUG=DEBUG

In the code you would then perform #ifdef to enable/disable Debug-specific code like so;

> #ifdef DEBUG
>    performSomeKindOfDebugOperationLikeGatheringPerformanceStatistics();
> #endif

Relates to forum topic;
https://forum.openframeworks.cc/t/determing-whether-debug-or-release-is-active/7249/12

### DIFF
--- a/libs/openFrameworksCompiled/project/makefileCommon/config.project.mk
+++ b/libs/openFrameworksCompiled/project/makefileCommon/config.project.mk
@@ -330,6 +330,10 @@ OF_PROJECT_LIBS += $(PROJECT_ADDONS_LIBS)
 OF_PROJECT_DEFINES := $(PROJECT_DEFINES)
 OF_PROJECT_DEFINES += $(PROJECT_ADDONS_DEFINES)
 
+ifeq ($(findstring Debug,$(MAKECMDGOALS)),Debug)
+	OF_PROJECT_DEFINES += $(PROJECT_DEFINES_DEBUG)
+endif
+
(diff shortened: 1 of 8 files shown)


bakercp:
Is this needed since, according to @arturoc 

```
right now when using makefiles or qtcreator DEBUG is defined when using the debug target and NDEBUG when using release
```

Here's code you can use to test it 

```
#if defined(DEBUG)
std::cout << "IN DEBUG MODE (DEBUG)" << std::endl;
#endif

#if defined(NDEBUG)
std::cout << "IN NDEBUG MODE (RELEASE)" << std::endl;
#endif
```

tgfrerer:
Yes, `NDEBUG` is pretty universally defined for release builds - partly, because `assert.h` depends on it. See also this discussion on stackoverflow: http://stackoverflow.com/questions/2290509/debug-vs-ndebug

To check for release builds you can use: 

```
#if defined(NDEBUG)
// NDEBUG means "Not Debug", i.e. Release
std::cout << "Compiled and running in Release mode" << std::endl;
#else 
std::cout << "Compiled and Running in Debug mode" << std::endl;
#endif
```

PR #5435, opened 2017-01-30
s/alphabetically/by date/ in sortByDate comment

hamoid:


arturoc:
@bakercp this is for sortByDate not plain sort, so yes i'd say that it's correct.

arturoc:
thanks!

PR #5448, opened 2017-02-13
Fix documentation in ofVec3f

aramgre:
The current description is only correct for unit vectors. I also removed a reference to "inner product" in favor of "scalar product", which is more helpful to understanding the difference between dot and cross product.


arturoc:
thanks!

PR #5452, opened 2017-02-16
ofSerial: only check filenames, not full paths, in buildDeviceList()

mattfelsen:
Fix ofSerial not listing any devices on OSX (and possibly Linux) as per issue/discussion https://github.com/openframeworks/openFrameworks/issues/5221


arturoc:
thanks!

PR #5458, opened 2017-02-22
ofImage: encode path to URI before parsing

mattfelsen:
As per discussion in https://github.com/openframeworks/openFrameworks/issues/5455, `ofImage` was failing to load when the file path has spaces. According to [the discussion here](http://stackoverflow.com/questions/497908/is-a-url-allowed-to-contain-a-space), and probably elaborated in [RFC 3986](https://tools.ietf.org/html/rfc3986) which I haven't read in detail, spaces are not allowed and *must* be encoded to `%20` first.

This PR uses the `uriparser` library to first encode the path string, and then use that when trying to parse with `uriParseUriA` as previously implemented. The code was taken directly from the [example related to converting filenames to URIs](http://uriparser.sourceforge.net/doc/html/#filenames), and uses an `#ifdef` to check on using the platform-appropriate calls to either `uriUnixFilenameToUriStringA()` or `uriUriStringToWindowsFilenameA()`

### DIFF
--- a/libs/openFrameworks/graphics/ofImage.cpp
+++ b/libs/openFrameworks/graphics/ofImage.cpp
@@ -184,17 +184,29 @@ template<typename PixelType>
 static bool loadImage(ofPixels_<PixelType> & pix, const std::filesystem::path& _fileName, const ofImageLoadSettings& settings){
 	ofInitFreeImage();
 
-	UriParserStateA state;
-	UriUriA uri;
 	auto uriStr = _fileName.string();
+	const int bytesNeeded = 8 + 3 * strlen(uriStr.c_str()) + 1;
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

can you change:

```
char* absUri = (char*) malloc(bytesNeeded * sizeof(char));
if(uriParseUriA(&state, absUri)!=URI_SUCCESS){
```

with

```
std::vector<char> absUri(bytesNeeded);
if(uriParseUriA(&state, absUri.data())!=URI_SUCCESS){
```

and remove the free at the end?

mattfelsen:
Sure thing, just made the change (and had to do the same update to the `uri*FilenameToUriStringA()` calls) - looks good on OSX & Windows

arturoc:
thanks!

PR #5459, opened 2017-02-23
Added mapping functions in ofRectangle for a vec2 and an ofRectangle

MatillaMartin:


bakercp:
Can you explain the intended use of such a function? Also, if you are going to make a map function, you should give the user access to the ofMap clamp parameter which had a default value.

MatillaMartin:
The use if this method is to map normalized values into a specific rectangle. For example, you want to make a series of boxes resolution independent, meaning that they will scale correctly when the window is expanded etc. With this method you can define the layout of the boxes normalized between 0 and 1, and you can map them to the layout rectangle to find their correct position for that layout.

Added the clamp parameter, default to false

arturoc:
hey, that seems useful.

sorry to make you change this again, but can you add another method for clamp instead of adding a parameter, so instead of map(..., bool clamp) it would be map and mapClamp, both do the same but one also clamps, the clamp version can even call the non clamped and clamp afterwards.

we are trying to ad least not add more bool parameters unless it's something really clear since the syntax is usually really confusing

libs/openFrameworks/types/ofRectangle.h
```diff
@@ -754,6 +754,16 @@ class ofRectangle{
     /// \returns The height of the rectangle.
     float getHeight() const;
 
+    
+    /// \brief Maps a normalized coordinate into this rectangle
+    /// \returns The mapped coordinate
```
bakercp: Could you add `\param` descriptions to the docs? e.g.

```cpp
/// \param coef ...
/// \param clamp ...
```

bakercp: Also, in addition to the brief, could you paste a short paragraph (like the above documentation) that expands upon the brief just a bit.

arturoc:
thanks!

bakercp:
Nice one @MatillaMartin 👍

Issue #5460, opened 2017-02-24
fboAlphaMask shader example dark edges


roymacdonald:
This issues appeared in the forum
https://forum.openframeworks.cc/t/fboalphamask-dark-edges-examples-shader-07-fboalphamask/25929/4
for which I found the solution.
Should I PR this fix?

arturoc:
yeah sure

PR #5462, opened 2017-02-25
fboAlphaMask example fix.

roymacdonald:
-Fixed issue where blending gave darker edges.
-Moved all the masking into the shader.
-Removed one fbo so it is easier to understand.

fixes issue
https://github.com/openframeworks/openFrameworks/issues/5460


arturoc:
thanks!

Issue #5463, opened 2017-02-26
Window title disappears after toggling fullscreen


valillon:
The window title set by `ofSetWindowTitle()` disappears after calling `ofToggleFullscreen()` to fullscreen (`OF_FULLSCREEN`) and then toggling back again (`OF_WINDOW`). Testing with OSX 10.12, Xcode 8.2.1 and OF 9.8.

tgfrerer:
Tested with oF from git at revision fb2da472128fc3c77c33ccbf959d55241fe52709

* windows: can't reproduce
* macOs: bug confirmed

PR #5469, opened 2017-03-03
MacOS: restore window title when restoring window

tgfrerer:
+ fixes an issue where when restoring a window from
  fullscreen on MacOS the window title would be lost. The
  restore appears only to affect MacOS.

+ keeps glfw window title in sync with window
  `settings.title`

Closes issue #5463


arturoc:
thanks!

PR #5471, opened 2017-03-03
Updates compileOF.sh for Linux to solve `chown` error

xuv:
Very small fix. Solving errors when compiling using a different user than the one logged in.

Example, compilation would produce error in this case:
 1. log into a machine using a certain user `user1`
 2. use sudo powers from user1 to switch to `user2`
 3. launch `./compileOF.sh` as user2
 4. breaks when reaching the line `chown -R $ID:$GROUP_ID ../lib/*` because the `$ID` returned by `logname` earlier in the script is `user1` and not `user2` who is calling the script. 

Using `whoami` instead of `logname` fixes this error without breaking the script for other common cases.

### DIFF
--- a/scripts/linux/compileOF.sh
+++ b/scripts/linux/compileOF.sh
@@ -9,9 +9,6 @@ else
         LIBSPATH=linux
 fi
 
-ID=`whoami`
-GROUP_ID=`id --group -n ${ID}`
-
 pushd `dirname $0` > /dev/null
 SCRIPTPATH=`pwd`
 popd > /dev/null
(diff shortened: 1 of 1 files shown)


arturoc:
this was done like this cause the install_dependencies script which is run under sudo was running compile_OF at the end so it needed the original user to change the owner. now install_dependencies doesn't run compile_OF anymore so it's safe to just remove this lines completely from compile_OF.sh.

can you change the PR to that?

thanks!

xuv:
@arturoc Thx for the explanation. I have modified the code to reflect this. I went further than just removing that one last line and also removed all the calls to the same type of action. I'll be happy to revert that change if you think this is wrong.

arturoc:
thanks!

PR #5476, opened 2017-03-07
ofxOsc updates

danomatika:
This is a work-in-progress cleanup/bugfix PR for various things I've noticed with ofxOsc:

* ofxOscMessage: inconsistent type conversions (sometimes we do, sometime we don't arbitrarily)
* ofxOscSender & ofxOscReceiver:  no way to access current host & port info
* formatting is all over the place

As this is a WIP PR, let's discuss but please wait until I/we all agree on the changes before merging.

Details So Far
--------------

* with permission of original author, Damian Stewart, removed explicit license headers and kept simple attribution
* cleaned up and made overall formatting consistent
* added light commenting for all classes & member functions
* added support for "N" nil type as ofxOscArgNone
* only included required headers instead of ofMain.h
* removed "using namespace std" and  removed use of std:: for primitive types (string & int types)
* fixed midi message, time tag, & rgba color int/uint conversion to/from oscpack issues by changing them to unsigned values to match the corresponding oscpack types
* removed bad argument type asserts in favor of better error prints

### ofxOscMessage

* fixed wrong type when adding Timetag arg
* numeric type conversion now works between all numeric types
* removed warning prints for numeric type conversions which do not lose precision (i.e. float -> int32 OK, but int64 -> int32: might be an issue) and kept them for string conversions (slower)
* added getRemoteHost() and deprecated getRemoteIp() (use of *Host used across API)
* promoted shutdown() to public clear() member function
* added getTypeString() for multiple argument type checking:
~~~
    int i = 0; float f = 0.0; string s = "";
    if(message.getTypeString() == "ifs") {
        i = message.getArgAsInt32(0);
        f = message.getArgAsFloat(1);
        s = message.getArgAsString(2);
    } 
~~~
* added stream conversion function for easy printing, prints address and arguments separated by spaces

### ofxOscSender

* added getHost() & getPort()
* promoted shutdown() to public clear() member function
* fixed RgbaColor type not being sent to outgoing packet stream
* fixed wrong type values being sent for non-primitives as oscpack types were not being used (midi message, nil, infinitum, time tag, etc)

### ofxOscReceiver

* added getPort()
* promoted shutdown() to public clear() member function
* added isListening()
* listener can be restarted with current port value when calling setup() with a 0 
* added setup() start arg to allow setting port value without starting thread

### DIFF
--- a/addons/ofxOsc/src/ofxOscBundle.cpp
+++ b/addons/ofxOsc/src/ofxOscBundle.cpp
@@ -1,31 +1,5 @@
-/*
- 
- Copyright (c) 2007-2009, Damian Stewart
- All rights reserved.
- 
- Redistribution and use in source and binary forms, with or without
- modification, are permitted provided that the following conditions are met:
- * Redistributions of source code must retain the above copyright
- notice, this list of conditions and the following disclaimer.
- * Redistributions in binary form must reproduce the above copyright
- notice, this list of conditions and the following disclaimer in the
- documentation and/or other materials provided with the distribution.
- * Neither the name of the developer nor the
- names of its contributors may be used to endorse or promote products
- derived from this software without specific prior written permission.
- 
- THIS SOFTWARE IS PROVIDED BY DAMIAN STEWART ''AS IS'' AND ANY
- EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
- WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
- DISCLAIMED. IN NO EVENT SHALL DAMIAN STEWART BE LIABLE FOR ANY
- DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
- (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
- LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
- ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
- (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
- SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
-
+// copyright (c) openFrameworks team 2010-2017
+// copyright (c) damian stewart 2007-2009
 #include "ofxOscBundle.h"
 
 //--------------------------------------------------------------
--- a/addons/ofxOsc/src/ofxOscBundle.cpp
+++ b/addons/ofxOsc/src/ofxOscBundle.cpp
@@ -1,5 +1,5 @@
 // copyright (c) openFrameworks team 2010-2017
-// copyright (c) damian stewart 2007-2009
+// copyright (c) Damian Stewart 2007-2009
 #include "ofxOscBundle.h"
 
 //--------------------------------------------------------------
--- a/addons/ofxOsc/src/ofxOscBundle.cpp
+++ b/addons/ofxOsc/src/ofxOscBundle.cpp
@@ -3,23 +3,17 @@
 #include "ofxOscBundle.h"
 
 //--------------------------------------------------------------
-ofxOscBundle::ofxOscBundle() {}
-
-//--------------------------------------------------------------
-ofxOscBundle::~ofxOscBundle() {}
-
-//--------------------------------------------------------------
-ofxOscBundle::ofxOscBundle(const ofxOscBundle& other){ 
+ofxOscBundle::ofxOscBundle(const ofxOscBundle &other){ 
 	copy(other);
 }
 
 //--------------------------------------------------------------
-ofxOscBundle& ofxOscBundle::operator=(const ofxOscBundle& other){
+ofxOscBundle& ofxOscBundle::operator=(const ofxOscBundle &other){
 	return copy(other);
 }
 
 //--------------------------------------------------------------
-ofxOscBundle& ofxOscBundle::copy(const ofxOscBundle& other){
+ofxOscBundle& ofxOscBundle::copy(const ofxOscBundle &other){
 	if(this == &other) return *this;
 	for(int i = 0; i < (int)other.bundles.size(); i++){
 		bundles.push_back(other.bundles[i]);
@@ -37,12 +31,12 @@ void ofxOscBundle::clear(){
 }
 
 //--------------------------------------------------------------
-void ofxOscBundle::addBundle(const ofxOscBundle& bundle){
+void ofxOscBundle::addBundle(const ofxOscBundle &bundle){
 	bundles.push_back(bundle);
 }
 
 //--------------------------------------------------------------
-void ofxOscBundle::addMessage(const ofxOscMessage& message){
+void ofxOscBundle::addMessage(const ofxOscMessage &message){
 	messages.push_back(message);
 }
 
--- a/addons/ofxOsc/src/ofxOscBundle.cpp
+++ b/addons/ofxOsc/src/ofxOscBundle.cpp
@@ -73,7 +73,7 @@ ofxOscMessage& ofxOscBundle::getMessageAt(int i){
 // friend functions
 //--------------------------------------------------------------
 std::ostream& operator<<(std::ostream &os, const ofxOscBundle &bundle) {
-	os << bundle.getMessageCount() << " messages "
-	   << bundle.getBundleCount() << " bundles";
+	os << bundle.getMessageCount() << " message(s) "
+	   << bundle.getBundleCount() << " bundle(s)";
 	return os;
 }
--- a/addons/ofxOsc/src/ofxOscBundle.cpp
+++ b/addons/ofxOsc/src/ofxOscBundle.cpp
@@ -28,37 +28,76 @@
 
 #include "ofxOscBundle.h"
 
+//--------------------------------------------------------------
+ofxOscBundle::ofxOscBundle() {}
 
-ofxOscBundle::ofxOscBundle()
-{
+//--------------------------------------------------------------
+ofxOscBundle::~ofxOscBundle() {}
+
+//--------------------------------------------------------------
+ofxOscBundle::ofxOscBundle(const ofxOscBundle& other){ 
+	copy(other);
 }
 
-ofxOscBundle::~ofxOscBundle()
-{
+//--------------------------------------------------------------
+ofxOscBundle& ofxOscBundle::operator=(const ofxOscBundle& other){
+	return copy(other);
 }
 
-ofxOscBundle& ofxOscBundle::copy( const ofxOscBundle& other )
-{
-	for ( int i=0; i<(int)other.bundles.size(); i++ )
-	{
-		bundles.push_back( other.bundles[i] );
+//--------------------------------------------------------------
+ofxOscBundle& ofxOscBundle::copy(const ofxOscBundle& other){
+	if(this == &other) return *this;
+	for(int i = 0; i < (int)other.bundles.size(); i++){
+		bundles.push_back(other.bundles[i]);
 	}
-	for ( int i=0; i<(int)other.messages.size(); i++ )
-	{
-		messages.push_back( other.messages[i] );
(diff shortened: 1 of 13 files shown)


danomatika:
O, I've added most of the main things I was looking to update. Two more things that come to mind:

1. Adding ability to set incoming & outgoing buffer sizes, this is listed a TODO in ofxOscSender. This could be done by setting the default to a static const class variable and adding an argument with this as a default value to ofxOscSender:::sendMessage() and ofOscSender::sendBundle(). Another simpler option is a settable static value to be used by all senders. Same could go for ofxOscReceiver.
2. I'd like to add a tostring method for ofxOscMessage for easy debug printing. I have the code in another project and can bring it in. Printing a message would print the address and the arguments separated by spaces.

danomatika:
Ok, I'm pretty much done and have been using this work in a project.

danomatika:
*ping* anyone?

Looks like the Travis build is failing on linuxarmv7 due to some configuration issue.

arturoc:
hey dan this looks great, thanks! i went to merge it when you said it was done but linuxarm7 is failing i have to look into it but i think it this should be ok anyway. i'll take a closer look tomorrow and merge it.

danomatika:
I looked at the build log but it seems to fail before compiling, so I think it's a config issue.

arturoc:
yeah it's some package name that has changed name or something in archlinux

danomatika:
I added one more commit which adds a start argument to ofxOscReceiver setup(). This allows you do set the internal port value *without* starting the receiver, which you can then start later by leaving port to 0:

    receiver.setup(1234, false); // sets port but doesn't start listening
    ...
    receiver.setup(); // starts listening with current port value since defaults are setup(0, true);

It would probably make more sense to use "start()" and "stop()" instead of "setup()" and "clear()" but the commenting should make things clear enough. In anywise, the original functionality is preserved while allowing for more flexibility ie. my need for being able to start and stop without having to use a printer and keep a copy of the communication settings around somewhere.

danomatika:
As mentioned in my previous post, one thing we could consider updating would be to add a start() function and rename clear() to close() or stop():

    bool setup(int port=0, bool start=true);

    /// start listening on the current port
    /// \return true on success
    bool start();

    /// stop listening and close the port
    void stop();

The setup() function would then still retain it's current functionality but allow for more explicitly starting and stopping with better function naming:

    receiver.setup(1234, false); // sets port but doesn't start listening
    ...
    receiver.start(); // starts listening with current port value
    ...
    receiver.stop(); // stop listening

addons/ofxOsc/src/ofxOscBundle.cpp
```diff
@@ -1,64 +1,77 @@
...
-
+// copyright (c) openFrameworks team 2010-2017
+// copyright (c) Damian Stewart 2007-2009
 #include "ofxOscBundle.h"
 
+//--------------------------------------------------------------
+ofxOscBundle::ofxOscBundle() {}
+
+//--------------------------------------------------------------
+ofxOscBundle::~ofxOscBundle() {}
```
arturoc: do we need destructor and copy constructor/operator? all this class has is a couple of vectors so it seems unnecesary and would make maintenance harder

addons/ofxOsc/src/ofxOscBundle.h
```diff
@@ -1,64 +1,50 @@
...
-    ofxOscBundle& getBundleAt( int i ) { return bundles[i]; }
-	const ofxOscMessage& getMessageAt( int i ) const { return messages[i]; }
-    ofxOscMessage& getMessageAt( int i ) { return messages[i]; }
+	/// \return the message at the given index
+	ofxOscMessage& getMessageAt(int i);
 	
 private:
 		
-	std::vector< ofxOscMessage > messages;
-	std::vector< ofxOscBundle > bundles;
```
arturoc: we are moving towards using std:: explicitly in headers to at some point remove the global include for std so better to leave these as they were

addons/ofxOsc/src/ofxOscMessage.h
```diff
@@ -1,136 +1,232 @@
...
-			  m - 4 byte midi packet (8 digits hexadecimal)
-			  T - TRUE (no value required)
-			  F - FALSE (no value required)
-			  N - NIL (no value required)
-			  I - IMPULSE, act as a trigger (no value required), previously named INFINITUM
-			  t - TIMETAG, an OSC timetag in NTP format, encoded in the data section
-
-	See : http://cnmat.berkeley.edu/system/files/attachments/Nime09OSCfinal.pdf
-	*/
-    std::int32_t getArgAsInt( int index ) const;
```
arturoc: same here, better to leave the explicit std::

addons/ofxOsc/src/ofxOscReceiver.h
```diff
@@ -1,101 +1,83 @@
...
+	/// enabled (true by default)
+	///
+	/// if port is set to 0, receiver will try to use current port value to
+	/// restart listening
+	///
+	/// if start is true, receiver will start listening (default)
+	/// if start is false, port value is set without starting
+	///
+	/// \return true on success
+	bool setup(int port=0, bool start=true);
```
arturoc: we are trying to remove bools as parameters for cases like this the syntax is confusing in the sense that it's hard to understand what

```
setup(1234,false)
```

means

if there's plans or there could be more parameters in the future, better to have a settings struct like:

```
ofxOsx::Settings settings;
settings.port = 1234;
settings.start = false;
osc.setup(settings);
```

if these 2 parameters seem like the only 2 there will ever be then better to have an additional method like:

```
setup(port);
setupDontStart(port)
```

addons/ofxOsc/src/ofxOscSender.h
```diff
@@ -1,91 +1,66 @@
...
-
-class ofxOscSender
-{
+/// \class ofxOscSender
+/// \brief OSC message sender which sends to a specific host & port
+class ofxOscSender{
 public:
+
 	ofxOscSender();
 	ofxOscSender(const ofxOscSender & mom);
```
arturoc: now that emplace_back exists perhaps we could remove copy constructors from classes like this were they only existed to be able to put them in a vector (it doesn't make much sense to copy a class like this otherwise). but we can leave that for another moment

addons/ofxOsc/src/ofxOscBundle.cpp
danomatika: That's a leftover from the original implementation. Easy to remove.

danomatika: Actually, I need the stub constructor for making bindings, so I'm leaving that but just in the header.

danomatika: Done. I also added destructors for ofxOscReceiver and ofxOscSender which call their clear() functions.

addons/ofxOsc/src/ofxOscReceiver.h
danomatika: I tried it this way in order to ensure backward compatibility, it's def not the best way. I don't think there is a need for a struct.

I think the best would be to have the following really:

receiver.setPort(1234);
receiver.start();

and we could keep setup() for backward compatibility which simply does both of those internally.

addons/ofxOsc/src/ofxOscMessage.h
danomatika: Done. These should be fine as long as #include <cstdint> is set somewhere.

addons/ofxOsc/src/ofxOscSender.h
danomatika: Ok. So we'd add a move constructor at that point?

addons/ofxOsc/src/ofxOscReceiver.h
danomatika: Ok, I split the functionality within the setup() function into two more functions which allows for more manual control without overloading setup(). This also keeps the original functionality of setup() the same for backwards compatibility:

~~~
/// set up the receiver with the port to listen for messages on
	/// and start listening
	///
	/// multiple receivers can share the same port if port reuse is
	/// enabled (true by default)
	///
	/// \return true if listening started successfully
	bool setup(int port);
	
	/// set the receiver port manually
	///
	/// does not *start* listening, restarts receiver if already running
	///
	/// multiple receivers can share the same port if port reuse is
	/// enabled (true by default)
	///
	/// \returns true if port set successfully
	bool setPort(int port);
	
	/// start listening manually using the current port value
	///
	/// note: this is not required if you called setup()
	///
	/// \return true if listening started or was already running
	bool start();
	
	/// stop listening, does not clear port value
	void stop();
~~~

I also renamed clear() to stop() as I think that more accurately fits it's function.

danomatika: A follow-up to this might be to also add the following to ofxOscSender:

    void setPort(int port);
    void setHost(const std::string &host);

This would make updating the current settings easier. As it is now, I'm currently calling setup() using one of the (new) getters:

    // set the host but keep the current port
    sender.setup("localhost", sender.getPort());

danomatika: Hah, I went ahead and added them. They are simply convenience wrappers for ofxOscSender::setup().

danomatika:
Added stream operators for bundle, receiver, and sender. Also made status getters const.

danomatika:
I tested all 3 osc examples. Working fine for me on macOS 10.12.3. I updated the chat example to remove deprecations.

arturoc:
i'm not very convinced about the setters for port/host, it can be problematic or at least ambiguous to have parameters that are mandatory to properly setup the object in separate methods:

- someone who doesn't know much about networks might for example just set the ip and try to send stuff.
- it introduces mutable state that can be problematic, what happens if you change the ip once the object has been started? right now it's restarting the whole thing but it might not be obvious from the api.
- looking through the cpp now it looks like setting the port or the host is reseting up the object which would fail if you haven't set it up before right? if the host == "" and you call setPort() it'll try to setup but will probably even crash on passing a nullptr on the host.c_str() no?

there's another 2 methods that i introduced recently that have a similar problem, enableBroadcasting/disableBroadCasting, i'd say we remove all this setters and change them with a struct with default values:

```
class ofxOscSender{
struct Settings{
    Settings(string host, int port);
    bool startOnSetup = true;
    bool broadcast = true;
    bool reuseport = true;
private:
    string host;
    int port;
    friend class ofxOscSender;
}
}
```

we stil keep the current setup(host,port) and add a new one that gets the settings object

this removes all mutable methods from the class which makes maintainance easier and allows us to introduce new parameters in the future without having to add new setup methods or extra parameters to the current one

arturoc:
also we are introducing this settings pattern in other places where there's need for more parameters to the current setup/load method or it was already too long so doing the same here helps making the api more uniform all over

arturoc:
> Do you have an example I can look at? I'm not 100% convinced about using C-style structs but I can understand being able to remove the tons of getter/setter boilerplate code.

we are using this pattern in several places already:

ofTruetypeFont: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofTrueTypeFont.h#L118-L147
ofImage: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofImage.h#L117-L128
ofFbo: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofFbo.h#L208-L229
of*Window: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofWindowSettings.h
ofGLFWWindow: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.h#L15-L47

in general this pattern kind of remedies the lack for named parameters in c++. while in other languages you can do:

```
something.setup(port:1234, host:"localhost", start:false);
```

there's no such thing in c++ but you can still use a struct like the above to allow to specify the parameters by name, remove the need for extra long parameters lists and also provide some defaults.

danomatika:
>in general this pattern kind of remedies the lack for named parameters in c++. while in other languages you can do

Oh I definitely agree on that. It's one of the things I like about Obj-C/Swift.

I have a commit coming now that adds exception handling to the sender with an error print.

danomatika:
Latest commit adds ofxOscReceiverSettings and ofxOscSenderSettings structs, setups which take the settings struct, and removes the explicit setters while keeping the core getters (getPort, getHost).

Now  using a manual start/stop on the receiver is:

~~~
ofxOscReceiverSettings settings;
settings.port = 1234; // set port
settings.start = false; // but don't start listening
receiver.setup(settings);
...
receiver.start(); // starts listening with current port value
...
receiver.stop(); // stop listening
~~~

arturoc:
just had a chance to fix the problem with linux armv7 and all the tests are passing. the settings changes look perfect too. thanks for looking into this!

danomatika:
Thanks!

Issue #5483, opened 2017-03-14
gles/customEGLWindowSettingsExample dumps core


johnkingsley:
The gles/customEGLWindowSettingsExample example dumps core.
This is on the Raspberry Pi 3.
The problem is that the window setup in main.cpp is out of date. Patch to follow.

PR #5484, opened 2017-03-14
Fix for gles/customEGLWindowSettingsExample core dump

johnkingsley:
Old way of setting up the window caused a core dump.
This is a fix for #5483


arturoc:
thanks!

Issue #5485, opened 2017-03-14
gles/customEGLWindowSettingsExample isn't transparent


johnkingsley:
This example sets the opacity to 127 in main.cpp:
	settings.eglWindowOpacity = 127;

Yet when the example is run, it's not transparent.
Patch to come.

PR #5486, opened 2017-03-14
Settings for EGLWindow were getting lost.

johnkingsley:
When setup(const ofWindowSettings &) was called, it was expected
that the inherited setup(const ofWindowSettings &) method would be called.
But it was not.

Instead, a temporary of type ofGLESWindowSettings was being created
and setup(const ofGLESWindowSettings &) was being called.
When the temporary was created, the EGL settings were lost.

By adding a "using" statement, the proper method is called.

This is a fix for #5485

### DIFF
--- a/libs/openFrameworks/app/ofAppEGLWindow.h
+++ b/libs/openFrameworks/app/ofAppEGLWindow.h
@@ -60,8 +60,10 @@ public:
 	static bool needsPolling(){ return true; }
 	static void pollEvents();
 
+	using ofAppBaseGLESWindow::setup;
 	void setup(const Settings & settings);
 	void setup(const ofGLESWindowSettings & settings);
+
 	void update();
 	void draw();
 	void close();


libs/openFrameworks/app/ofAppEGLWindow.h
```diff
@@ -62,6 +62,12 @@ class ofAppEGLWindow : public ofAppBaseGLESWindow, public ofThread {
 
 	void setup(const Settings & settings);
 	void setup(const ofGLESWindowSettings & settings);
+	// Need to use ofAppBaseGLESWindow::setup(const ofWindowSettings &)
+	// otherwise a parameter of type ofWindowSettings will be automatically
+	// converted to a temporary of type ofGLESWindowSettings and the above
+	// setup(const ofGLESWindowSettings &) will be called -- not what we want!
```
bakercp: This looks good to me, but I'd suggest removing the comments.

johnkingsley: Will do!  Thanks for the feedback.

arturoc:
thanks!

PR #5498, opened 2017-03-19
ofxAndroid - bug fix in pause/resume that caused random crashes

tallavi:
We found an issue caused by supposedly incorrect handling of onPause/onResume.

It seems that glView.onPause/ onResume should be called synchronously and not later in another thread, and that it's important to call onPause before removing the view and call onResume only after adding the view.


tallavi:
@openframeworks/android 

Hi @arturoc :-)

danzeeeman:
Interesting and makes since cause you need to pause the view before you
destroy it and onResume gets called when it's created no matter what so if
it is dynamically added and removed to the view stack this would cause
issues.

arturoc:
thanks!!

PR #5502, opened 2017-03-21
Need proper cleanup when EGL window is destroyed

johnkingsley:
Otherwise you get an error "terminate called without an active exception" because
the thread hasn't been joined.  This causes an abort() which causes even more trouble!


arturoc:
thanks!

PR #5511, opened 2017-03-29
Type fixes for OSC, additional documentation.

bakercp:
A few of the review items slipped through on #5476.  Here are the fixes.


bakercp:
@arturoc ping

arturoc:
going to merge this but i wonder if this kind of changes are really worth except in a few cases.

i've just fixed a problem with ofMesh save cause a similar change, where all types where converted to size_t, introduced a bug since some types needed to be the exact type they were before and not size_t.

I understand this in cases like ofPixels size, where using an int limited the size a lot, but in a case like this where the index refers to number of parameters of a single osc message it's really not necessary and creates warnings if users use int instead of size_t which can be less intuitive for beginners

PR #5513, opened 2017-03-29
 Fix descriptions for ofVec2f

aramgre:
Similar to documentation fixes made for ofVec3f in #5448, forgot to see if I should also update ofVec2f.


arturoc:
thanks!

PR #5524, opened 2017-04-07
set shader defines to 0 if not using tex, col

tgfrerer:
+ materials shaders expect HAS_COLOR and HAS_TEXTURE to be
  defined and to be either 0, or 1.

Closes issue #5523


bakercp:
This fixes the issue for me.  Thanks @tgfrerer !

PR #5526, opened 2017-04-07
Global time modes

arturoc:
Introduces 3 different time modes that change how the following
functions work:

```cpp
ofGetLastFrameTime,
ofGetElapsedTime{Millis, Micros, f}
```

**System**: Works as it was until now, the most precise monotonic clock
available in each platform is used to provide timming functions. Useful
for realtime applications

**Filtered**: Still uses the system clock but filters the times returned
by `ofGetLastFrameTime`, useful to remove small variations in the timing
which can make time based animations jittery.

**FixedRate**: Instead of a real clock, the time functions return a fixed
step. Every frame the clock is incremented by a step that can be set
globally. Useful for offline rendering.

This new functionality allows to have an application run in realtime
with time based animations and later change it with just one function
call for offline rendering.

There's some new functions to set the different time modes:

```cpp
ofSetTimeModeSystem()
ofSetTimeModeFixedRate(uint64_t stepNanos = ofGetFixedStepForFps(60))
ofSetTimeModeFiltered(float alpha = 0.9)

ofGetFixedStepForFps();
```

And a new function and time for getting the system time and integrate
the timming functions with std::chrono:

```cpp
ofTime ofGetCurrentTime()
```


bakercp:
Very nice.

bakercp:
I'm not sure if this relates, but I keep running into situations where a purely event-driven loop (as opposed to a timer-driven loop) would be useful.  Should that be a part of this conversation?

arturoc:
not sure what you mean, can you perhaps explain it with an example?

bakercp:
Ah, sorry.  So for example, if I wanted to only draw a new image to the screen when an image was uploaded to an openFrameworks web-server (e.g. ofxHTTP) or some other event occurred (e.g. a key pressed).  Currently, I believe, the only option for simulating this behavior is setting a delay in the update / draw loop or setting a low framerate (e.g. `ofSetFrameRate(1)`) and essentially polling.  I suppose one could do something similar with a blocking call to `ofThreadChannel::receive()` or (something similar) in the update loop, but basically I want arbitrary control over the ofApp::update / ofApp::draw sequence that is not pinned to a timer, but could be triggered on a given call to "loop()" or something.

bakercp:
So maybe it would be somthing like:

```c++
enum ofTimeMode{
	System,
	FixedRate,
	Filtered,
        Trigger
};
```

And one would be required to call a `loop()` function manually if it was set in `ofTimeMode::Trigger` mode.

arturoc:
ah i see, that's probably more related to being able to have a custom ofMainLoop. someone opened an issue about that some weeks ago but the fact that we need some templated functions make it harder to create a base class.

having a custom main loop is independent from this functionality i think. you can have a main loop that only loops when there's an event and still want system time or fixed rate time

bakercp:
Got it, that sounds like a better path to pursue -- thanks for the clarification and nice work on this PR.

PR #5529, opened 2017-04-08
ofXml: Feature - get name of node

c-mendoza:
Adds the ability for ofXml to obtain the name of the current node (i.e. the name of the tag). This is exposed via ofXml::getName(), returns a String, and is implemented via pugi's xml_node::name() funcion.


arturoc:
thanks!

PR #5535, opened 2017-04-11
Fix broken function signature.

bakercp:
This is probably not covered by a test case, else it would have been found.

Basically, cpp doesn't match .h.


arturoc:
oh thanks, yeah this won't throw an undefined symbol unless it's used by an executable

bakercp:
It passed tests, so I merged ...

arturoc:
perfect, thanks, yeah i was waiting on the tests

PR #5537, opened 2017-04-12
adds README to computeShaderParticlesExample

frauzufall:
Not completely sure if I got this right. 
- I don't understand the usage of `particlesBuffer2`, what is it good for?
- This comment and the computation at this point is also not clear to me https://github.com/openframeworks/openFrameworks/blob/master/examples/gl/computeShaderParticlesExample/src/ofApp.cpp#L71-L76

Can someone else help? Otherwise it should be fine as it is since I did not write anything in the README about these parts of the code.


arturoc:
we use 2 buffers so while opengl is doing the compute in one it can draw the previous one, then we swap them. it's just a double buffer to make things faster by introducing 1 frame delay.

and yeah the comment is kind of confusing. dispatchCompute just tells how many parallel computations are going to be done. part of it is specified in the compute shader source: https://github.com/openframeworks/openFrameworks/blob/master/examples/gl/computeShaderParticlesExample/bin/data/compute1.glsl#L59  since there the local size is already 1024 for the x dimension specifying 1 in the dispatchCompute call means 1024 parallel computations.

since we want to calculate the position of all the particles we need to issue as many computations as there are particles:

dispatchCompute(particles.size()/1024,1,1)

would do so but because it's an integer number if particles.size() is not a multiple of 1024 it would not issue enough computations so we add 1024

frauzufall:
Thanks, that was helpful! I updated the README.

arturoc:
looks good, one thing: the compute shader is a flocking algorithm (like in a flock of birds) which involves cohesion, repulsion... so perhaps it's easier to look for references if you name it like that instead of or apart from mentioning the individual parts of the algorithm

frauzufall:
Awesome. I learn a lot doing these READMEs.

arturoc:
thanks!

PR #5538, opened 2017-04-12
adds README to computeShaderTextureExample

frauzufall:
(if the single PRs for each example bug anyone, I can also collect them, but it's easier to discuss this way if something has to be changed)


arturoc:
not at all it's much easier to review and discuss this way

Issue #5522, opened 2017-04-07
ofLight crashes on destruction w/ programmable renderer.


bakercp:
While I realize programmable renderer doesn't support old-fashioned lights / materials, `ofLight` is currently crashing on destruction when using the programmable renderer.

On `master` branch, OSX, xcode, etc.

Steps to reproduce:

```c++
#include "ofMain.h"
#include "ofApp.h"

int main( ){
    ofGLWindowSettings settings;
    settings.setGLVersion(3,2);
    ofCreateWindow(settings);
    return ofRunApp(std::make_shared<ofApp>());
}
```
```c++
#pragma once

#include "ofMain.h"

class ofApp : public ofBaseApp{

public:
    void setup(){
        ofSetLogLevel(OF_LOG_VERBOSE);
        light.setPosition(0, 0, 500);
    }
    void update() { }
    void draw() {
        light.enable();
        light.disable();
    }

    ofLight light;

};
```

![ofapprunner_cpp_and_main_cpp](https://cloud.githubusercontent.com/assets/300484/24784736/33b720a4-1b1a-11e7-88c4-f157e09a79d0.png)

tgfrerer:
I've tracked this down to `ofLight` calling the renderer when `ofLight` gets destroyed. 

The bad access is due to calls from [ofLight#L80-84](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofLight.cpp#L80-L84)

On application teardown, it can't be guaranteed that a renderer is still alive, and if the renderer has been destroyed before an `ofLight` this would lead to `BAD_ACCESS`, as a destroyed renderer can't be de-referenced.

It seems these renderer calls might be better placed inside the renderer itself, rather than in the destructor of ofLight, as they change state in the renderer (which should own the lights), and don't touch de-allocation/destruction per se. Perhaps we shouldn't need these calls in the first place, as ofLight doesn't seem to be an object which is allocated/deallocated, rather than enabled/disabled.

I'd suggest we remove global calls to the renderer from `ofLight`'s destructor.

PR #5539, opened 2017-04-12
fix issue with ofLight crashing during app exit

tgfrerer:
Upon destruction `ofLight` called a global method to retrieve the current renderer. In most cases this is legit, but upon app teardown the renderer might have already been destroyed before to the `ofLight` object, which could lead to `BAD_ACCESS` errors when closing apps using `ofLight`. 

As this is an issue within `ofLight`, it affects the programmable as well as the fixed function renderers.

+ This PR suggests using a `weak_ptr` inside the `ofLight`'s `data` member to keep a non-owning link back from the light to the renderer.
+  This allows the light to check whether the renderer is still alive before calling methods on it.
+  This should also accelerate access to the renderer as the renderer itself doesn't need to be resolved on every call, only the `weak_ptr` locked.

Closes Issue #5522

### DIFF
--- a/libs/openFrameworks/gl/ofLight.cpp
+++ b/libs/openFrameworks/gl/ofLight.cpp
@@ -77,11 +77,13 @@ ofLight::Data::Data(){
 
 ofLight::Data::~Data(){
 	if(glIndex==-1) return;
-	ofGetGLRenderer()->setLightAmbientColor(glIndex,ofColor(0,0,0,255));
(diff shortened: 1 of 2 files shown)


arturoc:
thanks!

arturoc:
btw, if you write Closes #5522 or Fixes #5522 instead of Closes Issue.... in the PR comment or in the commit message the issue gets automatically closed when merging

PR #5545, opened 2017-04-13
Fix segfault when starting gl/gpuParticleSystemExample

frauzufall:
The example crashes on my machine during startup because `posPingPong.src` and `posPingPong.dst` are null pointers at the beginning. This PR sets these pointers during allocation.

Also ofApp.h was not UTF-8 encoded (I could not edit it at first) so I changed that too. Is that the right encoding? Not sure..


arturoc:
thanks!

PR #5552, opened 2017-04-18
Adds missing spaces in example README headlines

frauzufall:
Many of the headlines looked like this `###Learning Objectives` instead of `### Learning Objectives` so they did not appear as headlines on github.


arturoc:
thanks!

Issue #5554, opened 2017-04-18
ofXml::loadFromBuffer in documentation, but not in source code


frauzufall:
I was looking for a way to load data from ofBuffer to ofXml and found this function: http://openframeworks.cc/documentation/utils/ofXml/#show_loadFromBuffer But ofXml does not appear to include such a function. I guess the documentation needs to be updated, but I also would really like to use this function. Do I have to do that with ofxXmlSettings or is there another way in ofXml?

arturoc:
is this with 0.9.8 or master? ofXml has changed quite a bit in master and this function is porbably missing but should be pretty straight forward to bring it back

frauzufall:
I am using the master branch. Is the documentation (http://openframeworks.cc/documentation/utils/ofXml/) related to master or to the last release version? It refers to the old version. 

I added the function and filed a PR, please have a look :)

bakercp:
Fixed w/ #5555

PR #5555, opened 2017-04-18
#5554 ofXml: adds load(const ofBuffer&)

frauzufall:
In the old ofXml it was called `loadFromBuffer`, but I named it the same way `ofImage` does and added an `load(const ofBuffer& buffer)` function.

### DIFF
--- a/libs/openFrameworks/utils/ofXml.cpp
+++ b/libs/openFrameworks/utils/ofXml.cpp
@@ -23,6 +23,17 @@ bool ofXml::load(const std::filesystem::path & file){
 	}
 }
 
+bool ofXml::load(const ofBuffer & buffer){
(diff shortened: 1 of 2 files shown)


frauzufall:
Sorry the first commit did not really work, now it should.

arturoc:
didn't remembered that we have parse already, perhaps we don't need load(ofBuffer) then since you can just easily call `parse(buffer.getText())`? not very opposed to merge it if you think it's easier to find or use this way

frauzufall:
Like you I also didn't have the `parse` function in my mind at the beginning, I just came across it in the source code when I added the other function.. It's as easy to use but not too intuitive to find. Especially when looking at `ofImage` where you can use `load` with `ofBuffer`. Long story short, I am fine with keeping it the way it is but I would prefer to merge it :-)

Issue #5553, opened 2017-04-18
ofImage loading URL synchronously does not work


frauzufall:
The example `examples/input_output/imageLoaderWebExample` throws this error after startup: 

   ofImage: loadImage(): couldn't load image from ""http://images.wildmadagascar.org/pictures/bemaraha/tsingy_forest.JPG""

The asynchronous image load of the same URL does work. Hitting the spacebar displays the image.

frauzufall:
I tried to get more information and as I see it, [this line in `ofImage`](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofImage.cpp#L207) should be called but isn't reached because `scheme` is empty.

frauzufall:
Something else is off:
The URL file loader produces an endless reload if the network is unavailable or the URL not legit.
This behavior is reproducible by disabling network connectivity and  running the example `input_output/imageLoaderWebExample`. It will print an endless stream of these messages:
```
-1  for request tsingy_forest
-1  for request tsingy_forest
-1  for request tsingy_forest
```
It seems as if `ofLoadURLAsync` loads a URL again and again if it receives an error code `-1`. It nearly freezes my system. This doesn't happen on purpose, does it? I also find no way to stop it.

bakercp:
Related https://github.com/openframeworks/openFrameworks/issues/1312

PR #5556, opened 2017-04-18
regularExpressionExample: removing Poco..

frauzufall:
.. by switching from Poco::RegularExpression to std::regex.

It is still not working because of https://github.com/openframeworks/openFrameworks/issues/5553 but the example does not crash anymore because Poco is missing. I checked the image URLs the new ode is getting, seems to be fine.


arturoc:
thanks! i'll take a look into the url issue

PR #5563, opened 2017-04-20
Replace assumed OF path to a more robust one relative to $OF_PATH

armadillu:
The rsync command fails if your project is not at the standard depth. (i.e. OpenFrameworks/apps/myApps")


arturoc:
thanks!

PR #5568, opened 2017-04-21
Adds a text box to ofxGui

arturoc:
brings @fx-lange [ofxInputField](https://github.com/fx-lange/ofxInputField/) into ofxGui with some additions:

- full utf8 support (if ofxGui is setup with a ttf font with support for the used characters)
- clipboard (copy/paste/cut) support
- pressing esc goes back to previous value
- ctrl-a selects the full text
- blinking cursor!
- less mutable state + optimized rendering

also ofxInputField is now used by other gui elements:

- ofxSlider now shows an input field to allow to change the value when right clicked
- adding a string parameter to a gui group now shows as an input field instead of a label


yty:
cool! When can I merge?

PR #5573, opened 2017-04-25
Fix for Qt Creator in macOS

### LINKED ISSUE #5572: Qt Creator Mac: Error parsing ofApp.qbs
The error is:
```
(...)/openFrameworks/libs/openFrameworksCompiled/project/qtcreator/ofApp.qbs:35: error: TypeError: Result of expression 'CORE.includes' [undefined] is not an object.
```

Which means that no apps can be compiled. It seems like the problem began in [this commit].(https://github.com/openframeworks/openFrameworks/commit/62346e0f5f6c25fb24650adf31bb6743d7f95787)

I would love to help, but unfortunately I'm new at Qt Creator.

c-mendoza:
Changes to qbs files so that Qt Creator can compile again in macOS. Only tested in 10.11, someone should test in Linux. Fixes #5572

### DIFF
--- a/libs/openFrameworksCompiled/project/qtcreator/modules/of/of.qbs
+++ b/libs/openFrameworksCompiled/project/qtcreator/modules/of/of.qbs
@@ -48,11 +48,11 @@ Module{
 
     Probe{
         id: CORE
-        property stringList includes
-        property stringList cflags
-        property stringList ldflags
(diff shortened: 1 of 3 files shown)


c-mendoza:
Upon further testing I realize that this fix doesn't work.

PR #5584, opened 2017-04-30
Converts spaces to tabs in .mk files

frauzufall:


arturoc:
mmh, ok thanks, it's probably some spaces in the OF makefile then (that's usually what that error means). I'll let this tests finish and merge it. thanks!

frauzufall:
Found a couple more.. Adding "ofxCv" to any app will still not compile though.

libs/openFrameworksCompiled/project/b_release -a
```diff
@@ -0,0 +1,249 @@
+fredda
```
arturoc: i think you've commited this file accidentally

frauzufall: argh.. sorry..

arturoc:
thanks!!

PR #5586, opened 2017-05-01
Fix warning about unused function.

johnkingsley:
isDeviceArduino() is only used on TARGET_OSX


bakercp:
Looks good.

I use a slightly more generic sorting algorithm in ofxSerial that attempts to achieve something similar on all platforms.

https://github.com/bakercp/ofxSerial/blob/8086059d4de58620b7bc45d67a7388b32a7c4627/libs/ofxSerial/src/SerialDeviceUtils.cpp#L82-L111

bakercp:
Thanks!

PR #5587, opened 2017-05-02
Add basic support for ortho zooming to ofEasyCam

mattfelsen:
I took a stab at adding zooming support to ofEasyCam when in ortho mode. It feels a little dirty by using `ofPushMatrix()/ofPopMatrix()` and `ofScale()` - I feel like the proper way to do it would be by manipulating the viewport or similar in `begin()` but it works as a start at least.

In addition, there are a couple issues like...
- no separate sensitivity setting. Maybe it should use the existing sensitivityZ?
- "zoom" between perspective (ie position) and ortho (ie scale) are not in sync

Anyway, just throwing this out there to see what others think..
ping @ofZach re: https://twitter.com/zachlieberman/status/859185792681476096 hehe

### DIFF
--- a/libs/openFrameworks/3d/ofEasyCam.cpp
+++ b/libs/openFrameworks/3d/ofEasyCam.cpp
@@ -40,6 +40,20 @@ void ofEasyCam::begin(ofRectangle _viewport){
 	}
 	viewport = getViewport(_viewport);
 	ofCamera::begin(viewport);
+
+	if (getOrtho()) {
+		ofPushMatrix();
+		ofScale(orthoScale);
+	}
+}
+
+//----------------------------------------
+void ofEasyCam::end(){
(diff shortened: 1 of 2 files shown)


arturoc:
we shouldn't be doing global calls from inside the camera, can't you instead add the transformations to the modelview matrix?

roymacdonald:
because of this very same issue I made this addon. http://github.com/roymacdonald/ofxInfiniteCanvas
At the moment it doesn't inherit from ofCamera (cant remember why) although it works really well. I use it a lot. I'll take a look if it is possible to integrate into ofEasyCam.

roymacdonald:
Hi, Ive made this PR https://github.com/openframeworks/openFrameworks/pull/5604
cheers

arturoc:
this is solved now with #5604

Issue #5589, opened 2017-05-04
ofxColorPicker fixes


roymacdonald:
It is so good to see the color picker as well as the input field as part of ofxGui!

I fixed ofxColorPicker's following issues

-The color point (over the color wheel) would not display the correct color immediately after setup. 
-The color point would not update on mousepressed but only on mouse dragged. Now it updates on both.

I have done a PR with these fixes.

roymacdonald:
Pull request
https://github.com/openframeworks/openFrameworks/pull/5590

PR #5590, opened 2017-05-04
ofxColorPicker fixes.

roymacdonald:
-Now the color point is set correctly on setup.
-update color point on mouse pressed and mouse released.

Issue:
https://github.com/openframeworks/openFrameworks/issues/5589

### DIFF
--- a/addons/ofxGui/src/ofxColorPicker.cpp
+++ b/addons/ofxGui/src/ofxColorPicker.cpp
@@ -155,22 +155,25 @@ ofxColorPicker_<ColorType> * ofxColorPicker_<ColorType>::setup(ofParameter<ofCol
 	this->color.makeReferenceTo(parameter);
 
 	listener = color.newListener([this](ofColor_<ColorType> & c){
-		if(bSettingColor){
-			return;
-		}
-
-		ofFloatColor cf = c;
-		float hue, saturation, brightness;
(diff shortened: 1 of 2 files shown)


arturoc:
thanks roy, one minor thing, can you change:

```cpp
	listener = color.newListener([this](ofColor_<ColorType> & c){
		colorChanged();
        });
       colorChanged();
```

with:

```cpp
        colorChanged = [this](ofColor_<ColorType> & c){
		if(bSettingColor){
			return;
		}
		ofFloatColor cf = color.get();
		float hue, saturation, brightness;
		cf.getHsb(hue,saturation,brightness);
		colorScale = brightness;
		colorRadius = saturation;
		colorAngle = 360.f * hue;
		setNeedsRedraw();
	});
	listener = color.newListener(colorChanged);
	colorChanged();
```

and remove the colorChanged method? it's only used here so it doesn't make much sense to expose it everywhere else.

Or if you prefer the method then just change the listener syntax from


```cpp
	listener = color.newListener([this](ofColor_<ColorType> & c){
		colorChanged();
        });
```

to:

```cpp
	listener = color.newListener(this, &ofxColorPicker<ColorType>::colorChanged);
```

roymacdonald:
Hi @arturoc ahh right. I'll leave it as a local function.
I'll do it asap.

roymacdonald:
done!

Issue #5588, opened 2017-05-02
audioOut(ofSoundBuffer) reports incorrect samplerate


kylemcdonald:
```c++
#include "ofMain.h"

class ofApp : public ofBaseApp {
public:
    ofSoundStream ss;
    bool logged = false;
    void setup() {
        ofSoundStreamSettings settings;
        settings.sampleRate = 48000;
        settings.numInputChannels = 0;
        settings.numOutputChannels = 2;
        settings.setOutListener(this);
        ss.setup(settings);
    }
    void update() {
    }
    void draw() {
    }
    void audioOut(ofSoundBuffer& buffer) {
        if(!logged) {
            ofLog() << buffer.getSampleRate();
            logged = true;
        }
        buffer.fillWithTone();
    }
};

int main() {
    ofSetupOpenGL(1280, 720, OF_WINDOW);
    ofRunApp(new ofApp());
}
```

Expected output: 48000. Actual output: 44100.

The sample rate is definitely being set, because the pitch of the tone changes depending on the sample rate setting.

This is on macOS 10.12, XCode 8.3.2, using OF version https://github.com/openframeworks/openFrameworks/commit/63d20a5b5f3c4181f83febf2ad441921332f29e0

Temporary solution: save the output before setting it and re-use it later for any other calculations.

roymacdonald:
fixed here
https://github.com/openframeworks/openFrameworks/pull/5591

PR #5591, opened 2017-05-04
Fix audio sample rate issue #5588

roymacdonald:
-soundstream in and out buffer's samplerate is set at setup.

https://github.com/openframeworks/openFrameworks/issues/5588


arturoc:
thanks!

kylemcdonald:
thanks roy!

PR #5592, opened 2017-05-04
Adding support for OF_KEY_SPACE

MAKIO135:
Using OF_KEY_SPACE instead of GLFW_KEY_SPACE to keep homogeneity, and more readable than `' '`

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -1444,6 +1444,9 @@ void ofAppGLFWWindow::keyboard_cb(GLFWwindow* windowP_, int keycode, int scancod
 			break;
 		case GLFW_KEY_KP_EQUAL:
 			key = codepoint = '=';
+			break; 
+		case GLFW_KEY_SPACE:
+			key = codepoint = OF_KEY_SPACE;
 			break;
 		default:
 			codepoint = keycodeToUnicode(instance, scancode, mods);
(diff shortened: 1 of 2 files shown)


arturoc:
we consider space a character which means that you can just check it using `' '` as in:

```cpp
if(key==' ')
```

MAKIO135:
;) I thought about it before suggesting this.
Yet, i think this is less error prone/faster to read `OF_KEY_SPACE` than `' '`, don't you think ?
It also looks more homogeneous when your key checks are `OF_KEY_UP`, `OF_KEY_DOWN`, `OF_KEY_RIGHT`, ... but `GLFW_KEY_SPACE`

One could also do:
```
#define OF_KEY_SPACE 32
```
or:
```
#define OF_KEY_SPACE ' '
```

bakercp:
In my mind, the logic of the defined keys in OF is for codepoints that are unprintable / control characters.  "space" is in the gray zone.  As an educator, I'd prefer to be explicit so I'm all for adding `OF_KEY_SPACE`.  That said, I'm assuming we don't want / need to add all keys (e.g.  http://www.glfw.org/docs/latest/group__keys.html).

Also, if we're going for consistency, we should double check if `KEY_SPACE` should be added here:

https://github.com/openframeworks/openFrameworks/blob/2639bba55070c8efea0c873d023d7982817556fa/libs/openFrameworks/app/ofAppEGLWindow.cpp#L1520-L1657

Perhaps we should also go ahead and update examples to use it.

MAKIO135:
Regarding `ofAppEGLWindow.cpp`, i based my modifications according to @ofTheo 's PR: `added support for tab key OF_KEY_TAB / '\t' #2254` https://github.com/openframeworks/openFrameworks/pull/2254 but double checking should be better.

Tabulation character could be checked the same way:  `if ( key == '	' )`, yet for the same reasons it is better to have `OF_KEY_TAB`

bakercp:
I think the `OF_KEY_TAB` comparison is pretty compelling.  I know that I've had students get confused about the `key == ' '` thing before, so I'd support this.

danoli3 (approved):
moved away from defines but I'll merge this

danoli3:
Merged via #8095

Issue #5440, opened 2017-02-04
glm::vec3 to glm::vec2 and vice-versa


kylemcdonald:
It seems like glm is happy converting vec3 to vec2 via truncation, but it won't automatically convert vec2 to vec3 by adding a zero. I know this has been a contentious topic in the past (e.g., vec3 to vec4 might require a one instead of a zero) but I'm running into a lot of code that will break because of this. For example, adding `ofMesh::addVertex(ofVec2f)` now fails.

If there are no plans to change this behavior, it would be great to add it as another note in the http://openframeworks.cc/learning/02_graphics/how_to_use_glm/ tutorial!

bakercp:
Agreed.  I'm 101% excited about glm, but some extra documentation here would be good.  I've run into this a lot with `ofPolyline` in the last days. In the past we have been very non-strict with our use of 2d and 3d points.  For instance, our `ofRectangle` and `ofPolyline` are built on 3d vectors, but many of their member operations only are only valid for 2d operations

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofPolyline.h#L370-L378

What does `bool ofPolyline::inside(const glm::vec3 & p) const;` mean :) ?

And `ofPolyline::fromRectangle(const ofRectangle& rect)` doesn't actually take the dubious `ofRectangle::z` component into account.

... and lots of warnings may not be a good substitute for stricter compile-time errors

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/types/ofRectangle.h#L14-L15

Anyway, I'd argue that we should make `ofRectangle` a 2D rectangle and we should make an `ofPolyline2f` (or something), to make all of this clearer, but admittedly this does make oF less readable.  But with the right combination of `enable_if`, etc and the newly-templated ofPolyline may make this pretty easy to implement.

roymacdonald:
Hi,
@kylemcdonald +1 on this.
I'm not sure if I'm excited as you are @bakercp  with glm as it gave me a huge head ache on a recently made project.
As for what you mention of the polyline, although it is not correct in one sense it is super handy on the other. I really like ofRectangle, I use it a lot and I have never faced a problem because its position is a 3d vec . I'm not sure if making it 2d will bring problems but it will certainly break code that expects a 3d vec.

`bool ofPolyline::inside(const glm::vec3 & p) const;` this checks if the passed vec falls inside the polyline, just like in ofRectangle. :)

[...]

PR #5593, opened 2017-05-05
Progress on #5440.

bakercp:
Adds better backward compatibility during the `glm::vec*` / `ofVec*` transition.

### DIFF
--- a/libs/openFrameworks/3d/ofMesh.h
+++ b/libs/openFrameworks/3d/ofMesh.h
@@ -654,6 +654,21 @@ public:
 
 	/// \}
 
+#ifndef OF_USE_LEGACY_MESH
+	OF_DEPRECATED_MSG("Use glm::vec2.", ofMesh_(ofPrimitiveMode mode, const vector<ofVec2f>& verts));
+	ofMesh_(ofPrimitiveMode mode, const vector<glm::vec2>& verts);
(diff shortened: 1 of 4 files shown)


libs/openFrameworks/3d/ofMesh.h
```diff
@@ -654,6 +654,21 @@ class ofMesh_{
 
 	/// \}
 
+#ifndef OF_USE_LEGACY_MESH
+	OF_DEPRECATED_MSG("Use glm::vec2.", ofMesh_(ofPrimitiveMode mode, const vector<ofVec2f>& verts));
```
bakercp: Well, technically, we aren't _adding_ new deprecated methods ... these are all methods that currently exist in 0.9.* and have not been gracefully deprecated, but rather hidden behind a somewhat unintuitive `OF_USE_LEGACY_MESH` def.  I say unintuitive based on my experiences with many students this past semester trying to mix old and new addons. Huge amounts of time were consumed trying to figure out the right combination of using  `OF_USE_LEGACY_MESH` or converting addons, to glm.  

So, basically I'd strongly argue for formally deprecating these by default to ease the transition, especially with the very heterogeneous state of addons.

bakercp:
@arturoc Do you have additional thoughts on this?

PR #5594, opened 2017-05-05
glm transition - deprecations implicit conversions, remove implicit conversions from the core.

bakercp:
This is related to #5440, the effort to ease the glm transition.

I looked into deprecating entire classes (e.g. `ofVec2f`, etc), but it created a huge mess of deprecation warnings, so instead I opted to just add deprecation warnings in those places where glm/of conversions were happening implicitly. 

I also discovered a few remaining places where `ofVec*`  was being used explicitly and a few interesting places in `ofNode` where when adding an `glm::vec3 + glm::vec4` resulted in an intermediate conversion to `ofVec3f`. 

Currently this shouldn't produce any deprecations warnings from the core itself, but only with non-conforming examples and user code that still employs `ofVec*`.

### DIFF
--- a/libs/openFrameworks/3d/ofNode.cpp
+++ b/libs/openFrameworks/3d/ofNode.cpp
@@ -132,9 +132,9 @@ void ofNode::setParent(ofNode& parent, bool bMaintainGlobalTransform) {
 		clearParent(bMaintainGlobalTransform);
 	}
 	if(bMaintainGlobalTransform) {
-		auto postParentPosition = position - parent.getGlobalPosition();
+		auto postParentPosition = position.get() - parent.getGlobalPosition();
 		auto postParentOrientation = orientation.get() * glm::inverse(parent.getGlobalOrientation());
-		auto postParentScale = scale / parent.getGlobalScale();
+		auto postParentScale = scale.get() / parent.getGlobalScale();
 		parent.addListener(*this);
 		setOrientation(postParentOrientation);
 		setPosition(postParentPosition);
@@ -597,7 +597,7 @@ void ofNode::orbitDeg(float longitude, float latitude, float radius, const glm::
 	p = q * p;							   // rotate p on unit sphere based on quaternion
 	p = p * radius;						   // scale p by radius from its position on unit sphere
 
-	setGlobalPosition(centerPoint + p);
+	setGlobalPosition(centerPoint + p.xyz());
 	setOrientation(q);
 
 	onOrientationChanged();
@@ -620,7 +620,7 @@ void ofNode::orbitRad(float longitude, float latitude, float radius, const glm::
 	p = q * p;							   // rotate p on unit sphere based on quaternion
 	p = p * radius;						   // scale p by radius from its position on unit sphere
 
-	setGlobalPosition(centerPoint + p);
+	setGlobalPosition(centerPoint + p.xyz());
 	setOrientation(q);
 
 	onOrientationChanged();
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -756,7 +756,7 @@ void ofAppGLFWWindow::setFullscreen(bool fullscreen){
         GLFWmonitor** monitors = glfwGetMonitors(&monitorCount);
  
 		int currentMonitor = getCurrentMonitor();
-		ofVec3f screenSize = getScreenSize();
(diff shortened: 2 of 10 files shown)


libs/openFrameworks/3d/ofNode.cpp
```diff
@@ -597,7 +597,7 @@ void ofNode::orbitDeg(float longitude, float latitude, float radius, const glm::
 	p = q * p;							   // rotate p on unit sphere based on quaternion
 	p = p * radius;						   // scale p by radius from its position on unit sphere
 
-	setGlobalPosition(centerPoint + p);
+	setGlobalPosition(centerPoint + p.xyz());
```
bakercp: Without this, an intermediate ofVec3f conversion was being used.

libs/openFrameworks/3d/ofNode.cpp
```diff
@@ -132,9 +132,9 @@ void ofNode::setParent(ofNode& parent, bool bMaintainGlobalTransform) {
 		clearParent(bMaintainGlobalTransform);
 	}
 	if(bMaintainGlobalTransform) {
-		auto postParentPosition = position - parent.getGlobalPosition();
+		auto postParentPosition = position.get() - parent.getGlobalPosition();
```
bakercp: Without this, an intermediate ofVec3f conversion was being used.

libs/openFrameworks/app/ofAppGLFWWindow.cpp
```diff
@@ -756,7 +756,7 @@ void ofAppGLFWWindow::setFullscreen(bool fullscreen){
         GLFWmonitor** monitors = glfwGetMonitors(&monitorCount);
  
 		int currentMonitor = getCurrentMonitor();
-		ofVec3f screenSize = getScreenSize();
```
bakercp: Leftover ofVec.

libs/openFrameworks/graphics/ofGraphics.cpp
```diff
@@ -1138,14 +1138,14 @@ void ofVertices( const vector <glm::vec2> & polyPoints ){
 //----------------------------------------------------------
 void ofVertices( const vector <ofVec3f> & polyPoints ){
 	for( const auto & p: polyPoints ){
-		ofGetCurrentRenderer()->getPath().lineTo(p);
+		ofGetCurrentRenderer()->getPath().lineTo(glm::vec3(p.x, p.y, p.z));
```
bakercp: Use explicit conversions throughout.

libs/openFrameworks/math/ofMatrix3x3.h
```diff
@@ -42,11 +42,11 @@ class ofMatrix3x3 {
 				float _g=0.0, float _h=0.0, float _i=0.0 );
 
 
-	ofMatrix3x3( const glm::mat3 & mat) {
+	OF_DEPRECATED_MSG("Use glm::mat3.", ofMatrix3x3( const glm::mat3 & mat)) {
```
bakercp: Encourage the use of glm::* without breaking things.

libs/openFrameworks/math/ofVectorMath.h
```diff
@@ -243,88 +243,88 @@ namespace glm {
 
 //--------------------------------------------------------------
 inline glm::vec3 operator+(const glm::vec3 & v1, const ofVec3f & v2){
-	return v1 + glm::vec3(v2);
+	return v1 + glm::vec3(v2.x, v2.y, v2.z);
```
bakercp: Remove implicit conversions.

arturoc:
i'm not so sure about this, people are going to use ofVec for a long while and although i see how it's really useful for catching where you are still using ofVec i think it can also be a problem cause a lot of addons are using ofVec and most people are not going to port them right away (if at all) so this is going to throw a million warnings in most projects

even most examples still use ofVec right now

perhaps there could be a flag to enable this? this is only header code so you can just do:

```cpp
#define OF_DEPRECATE_OFVEC
#include "ofMain.h"
```

or define that in the project files, makefiles... while compiling the core, examples... so we can catch usage of ofVec

bakercp:
Well, the intention was to "encourage" people to migrate via deprecation warnings since we are deprecating ofVec.  Deprecation warnings are indeed annoying, but that is their purpose no?  Keeping oF in limbo for more than a CV release or two is going to be a bit of a maintenance pain.  Most people probably won't read the documentation, but the will see the warnings :)

bakercp:
@arturoc Any thoughts on this?

PR #5598, opened 2017-05-09
ofParameter advanced features

arturoc:
Some new features for `ofParameter`, mostly useful in conjunction with `ofxGui`:

- ofParameter allows to set a logarithmic scale (only numeric ones) which then translates in sliders with logarithmic scale. Can be directly use in the header like:

```cpp
ofParameter<float> volume{"volume", 0, 0, 1000, ofParameterScale::Logarithmic);
```

- `ofSetMutuallyExclusive(p1,p2,p3,...)`: allows to pass an arbitrary number of parameters and they will be set as mutually exclusive so when one is set to true the rest are set to false which allows to create option groups in guis really easy just by setting the corresponding parameters as mutually exclusive. It would probably be useful to be able to do the same with an `ofParameterGroup` or a collection of `ofParameter<bool>`

- `ofParameterLink(p1,p2,p3...)`: allows to pass an arbitrary number of parameters that will "track" each other so if any of them changes all the rest are set to the same value


tgfrerer:
Super useful - one remark about naming the naming of `[set|get]PctScaled`: I assume "Pct" refers to %, which would mean a number between 0..100, i.e. a fraction of 100. How about `[get|set]NormalizedScaled`? 

"Normalized" by convention means 0..1 range, which is the range the getter/setters appear to expect.

arturoc:
yeah that makes sense, will change it in a bit...

roymacdonald:
this looks great! Super useful.
I've been dealing with something where parameter link would have been super useful, although I've had problems with parameters in different threads, so, Is this new function thread safe?  From what I can read it doesn't seem to be. 
best

PR #5599, opened 2017-05-09
Update glm (DON'T MERGE) -- Just watch tests.

bakercp:
@arturoc I think this is the way to do it ... but I'm not quite sure how / if this change will trickle down to the MSYS tests.  We'll see!

It may go without saying, but don't merge this because it points to a test branch of apothecary and will get confusing.

Just keep an eye on the appveyor and other tests, particularly to see if MSYS builds with updated glm.


bakercp:
Everything passes, except the armv7, which has been filling sure to that tar issue that popped up again.  Can you confirm that this actually compiled with the new glm?

arturoc:
yeah arm7 has been failing in the last couple of tests cause there's some new dependencies we need to add.

i think this looks correct, can you send a new PR to apothecary with the change for glm?

PR #5602, opened 2017-05-12
ofAppQtWindow

wasawi:
Initial approach to Qt window.
So far it renders properly but key events might not work well.
Still many events to be implemented.
Opening 2 windows is not working yet.


arturoc:
hey this is great but i think it's more adequate to publish it as an addon, we would need to add all of qt to openframeworks which is quite big.

jordiexvision:
hey thanks! yes Qt is too big.. maybe an addon with Qt install steps? that could be the easiest right?

bakercp:
I'd vote for QT addon with install steps, or even better apothecary scripts for installation!

PR #5604, opened 2017-05-13
Feature ortho easy cam

roymacdonald:
Added a better interaction for orthographic projection.

- In ortho mode drag with left mouse button translates the camera as opposed to the perspective mode which rotates. I think that it is more intuitive and better this way.
- Ortho mode zooming via scroll or right drag. It zooms into the point below the mouse.
- Removed a lot of redundant maths.
- Mouse callbacks only set state. all calculations are carried out in the update cycle. Makes code cleaner.
- Added a boolean parameter to the enable... methods. This makes it easier for toggling.
- Updated ofEasyCamExample README.
- Renamed several private members with more self descriptive names.

There are a few things that we should discuss.

- On ortho mode you can rotate dragging either (mouse left button + 'translation key') or mouse middle button. Translation key is defaulted to 'm'.
The question is, should we change this default translation key to something like the alt or shift key? These would be a bit less arbitrary and more intuitive and more like most CAD software. 
- Is it OK having different mouse interactions on each mode? I think that it is better and more intuitive.

best


roymacdonald:
https://github.com/openframeworks/openFrameworks/issues/5605

roymacdonald:
By the way, I completely failed at producing a small size animated gif to update the one of this examples readme.
Any tips for achieving this?

roymacdonald:
How is in charge of the 3D area? Is it still @elliotwoods ? Are we still using "section leaders"?
@arturoc @ofTheo @ofZach 

Best!

libs/openFrameworks/3d/ofEasyCam.cpp
```diff
@@ -132,33 +148,32 @@ char ofEasyCam::getTranslationKey() const{
...
 
 //----------------------------------------
-void ofEasyCam::enableMouseInput(){
-	if(!bMouseInputEnabled && events){
-		listeners.push_back(events->update.newListener(this, &ofEasyCam::update));
-		listeners.push_back(events->mouseDragged.newListener(this, &ofEasyCam::mouseDragged));
-		listeners.push_back(events->mousePressed.newListener(this, &ofEasyCam::mousePressed));
-		listeners.push_back(events->mouseReleased.newListener(this, &ofEasyCam::mouseReleased));
-		listeners.push_back(events->mouseScrolled.newListener(this, &ofEasyCam::mouseScrolled));
+void ofEasyCam::enableMouseInput(bool e){
```
arturoc: why this? we don't use this pattern anywhere else, things are called either set*(bool) or enable/disable but we don't use enable(bool) anywhere so i would avoid having it here

arturoc:
I wouldn't have different interactions for perspective/ortho it would be confusing no?

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: I just had it like this and I like the enable(bool) pattern. But right, if it is not used anywhere else I can remove it.

roymacdonald:
At least I like to use the ortho camera for 2D drawing, so having the leftmouse clic to translate is really handy, as I would rarely rotate it.
Maybe a function that allows to switch this behavior might be the solution?
something like `setInvertedMouseButtons(bool)` ?
cheers

libs/openFrameworks/3d/ofEasyCam.cpp
```diff
@@ -12,23 +12,41 @@ static const unsigned long doubleclickTime = 200;
...
+	sensitivityTranslate = {1,1,1};
+	sensitivityRot = {1,1,1};
+}
+//----------------------------------------
+void ofEasyCam::enableOrtho(){
+	if (!getOrtho()) {
+		prevFarClip = getFarClip();
+		prevNearClip = getNearClip();
+	}
+	setNearClip(-10000);
```
tgfrerer: I'm not sure if setting the far and near clip here as a side-effect can't get confusing - it doesn't seem to be what you ask for when you enable orthographic procection. 

Imagine what happens if someone thinks they know what they are doing and changes the near clip before enabling orthographic projection - they would find themselves surprised to be stuck with hard-coded -10000, 10000 which might be tricky to debug because it's a bit surprising =)

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -219,22 +222,21 @@ class ofEasyCam : public ofCamera {
...
-
-    /// \brief The previous camera orientation.
-	glm::quat prevOrientation;
-
+	
+	/// \name On Press cache
+	/// \{
+	/// \brief camera properties when the mouse is pressed.
+
+	glm::vec3 onPressAxisX;
```
tgfrerer: More a cosmetic comment: `onSomething` might suggest that these variables have something to do with events - or callbacks, where this naming pattern is used frequently. If I understand this section correctly, these members are here to capture previous interactive camera state. 

In other parts of the openFrameworks codebase, we're using "previousSoundSo" or "lastSoundSo", and it might make things a bit more easy to read if consistent.

tgfrerer:
First of all, there are some lovely features in this PR - like zoom in ortho and scrolling etc. I'm looking forward to seeing this merged =)

*
I've tried out the updated example and noticed that it appears out of step with the latest update to this PR - I'll send your PR a PR with a few snags I had to apply to get it to compile... 

*
Something to consider with method names such as `setInvertedMouseButtons(bool)` is that they imply that anyone coming across a camera knows the *non-inverted* mouse button settings, and inverts them in their mind whilst reading the rest of the code. 

I tend to prefer methods which are more explicit in their naming, and ideally provide a as much context as possible to their meaning. When I tried the `ofEasyCam` example, not all mouse interactions appear *inverted* - only *middle* and *left* mouse buttons.

Anyway, having different interactions per perspective treatment seems a bit counter-intuitive, and I'd agree with @arturoc, it's probably more straightforward to keep mouse interactions the same regardless of perspective type.

If someone wanted to remap mouse buttons specifically for their application, that would be nice to allow, too, of course.

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: Right, it is super arbitrary. The reason I had was that the defaults were clipping the scene that I made for the example app. I'll remove this and maybe a good idea is to put it in the example. I didn't see the fact that it can be hard to debug, which could become very annoying.

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: right. 
actually there were some of these vars called lastSomething that were used to hold the values of Something when the mouse gets pressed. In particular there were prevMouse, which was storing the mouse position of the previous frame, and lastMouse which was storing the position the mouse had when it got pressed. this naming pattern seemed really confusing, so I decided to change the lastMouse for onPressMouse. 
Although I see your point. Maybe `lastPressMouse` would be better?

roymacdonald:
Hi @tgfrerer 
these problems you had to compile the example were related to the use of `cam.enableSomething(bool)` calls?
I pushed a commit fixing it. If you had other problems please let me know.

You are right about `setInvertedMouseButtons(bool)`. It is really not method good naming.

I was trying out something a bit more flexible that would allow to remap mouse buttons.
I will post later some code as I have some doubts about which way to go in terms of style.
best

tgfrerer:
@roymacdonald re: compile errors: yes indeed. I sent a pull request to your pull request with the fixes, (https://github.com/roymacdonald/openFrameworks/pull/1/files) but it seems you've fixed these already, so not to worry.

roymacdonald:
@tgfrerer  Ha,  I just merged your PR. thanks. :)

roymacdonald:
Hey @arturoc @tgfrerer 
So I implemented a way that allows to set custom mouse interactions very easily. Please check it and comment.
As a side effect this made the code much more cleaner and less error prone.

I have a doubt. This changes leave the updateMouse(...), updateTranslation() and updateRotation() private methods getting called only once each from ofEasyCam::update.
Is it worth keeping these methods or should I place all these methods code inside the update? I'm for the latter as i think it will reduce the number of lines and make it a bit less redundant. 
best

roymacdonald:
for customizing the ofEasyCam / mouse interaction.

There this public enum in ofEasyCam
```
enum easyCamTransformType{
		EASYCAM_TRANSFORM_NONE,
		EASYCAM_TRANSFORM_ROTATE,
		EASYCAM_TRANSFORM_TRANSLATE_XY,
		EASYCAM_TRANSFORM_TRANSLATE_Z,
		EASYCAM_TRANSFORM_SCALE
	};

```
there are these public methods

	void addInteraction(easyCamTransformType type, int mouseButton, int key = -1);
	void removeInteraction(easyCamTransformType type, int mouseButton, int key = -1);
	bool getHasInteraction(easyCamTransformType type, int mouseButton, int key = -1);
	bool getHasInteraction(int mouseButton, int key);
	void removeAllInteractions();

the current mouse intearction setting is set by these lines, (which are now in the ofEasyCam's constructor)

	addInteraction(EASYCAM_TRANSFORM_ROTATE, OF_MOUSE_BUTTON_LEFT);
	addInteraction(EASYCAM_TRANSFORM_TRANSLATE_XY, OF_MOUSE_BUTTON_LEFT,doTranslationKey);
	addInteraction(EASYCAM_TRANSFORM_TRANSLATE_Z, OF_MOUSE_BUTTON_RIGHT);
	addInteraction(EASYCAM_TRANSFORM_TRANSLATE_XY, OF_MOUSE_BUTTON_MIDDLE);

calling addInteraction with no third parameter means the mouse pressed while no keys are pressed.
Looking forwards for your comments.
best

roymacdonald:
Any comments on this?
@arturoc Any chances to merge it?
All the best!

libs/openFrameworks/3d/ofEasyCam.cpp
```diff
@@ -266,49 +269,65 @@ bool ofEasyCam::getInertiaEnabled() const{
...
-		if(ABS(moveZ) >= minDifference){
-			bIsBeingScrolled = true;
-		} else {
-			bIsBeingScrolled = false;
-		}
-
-		if(ABS(moveX) <= minDifference && ABS(moveY) <= minDifference && ABS(moveZ) <= minDifference){
+		translate *= drag;
+//		bDoZoom = (ABS(translate.z) >= minDifference);
+		if(ABS(translate.x) <= minDifference && ABS(translate.y) <= minDifference && ABS(translate.z) <= minDifference){
```
bakercp: Can I request using `std::abs`, rather than the `ABS` macro, as we're working on deprecating `ABS` etc.

See #5624 for conflicting changes.

libs/openFrameworks/3d/ofEasyCam.cpp
```diff
@@ -336,26 +355,26 @@ ofRectangle ofEasyCam::getControlArea() const {
...
-				bInsideArcball = false;
+		lastPressAxisX = getXAxis();
+		lastPressAxisY = getYAxis();
+		lastPressAxisZ = getZAxis();
+		lastPressPosition = ofCamera::getGlobalPosition();
+		lastPressOrientation = ofCamera::getGlobalOrientation();
+
+		currentTransformType = EASYCAM_TRANSFORM_NONE;
+		if (events) {
+		for (auto& i: interactions) {
```
bakercp: `const auto&` or `auto`, as we aren't changing `i`

libs/openFrameworks/3d/ofEasyCam.cpp
```diff
@@ -376,83 +395,106 @@ void ofEasyCam::mouseReleased(ofMouseEventArgs & mouse){
...
+			bApplyInertia = true;
+		}
+		lastPressPosition = ofCamera::getGlobalPosition();
+		lastPressAxisZ = getZAxis();
+		if (getOrtho()) {
+			translate.z = sensitivityScroll * mouse.scrollY / ofGetHeight();
+			mouseAtScroll = mouse;
 		}else{
-			vFlip =  1;
+			translate.z = mouse.scrollY * 30 * sensitivityTranslate.z * (getDistance() + FLT_EPSILON)/ area.height;
```
bakercp: What is `30`?

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -83,11 +79,12 @@ class ofEasyCam : public ofCamera {
...
 	/// \param value Scales the xyz axes rotation factor by these values.
 	void setRotationSensitivity(float x, float y, float z);
-    
+	void setRotationSensitivity(const glm::vec3& sens);
+	
     /// \brief Set the input sensitivity of the translation.
     /// \param value Scales the xyz axes translation factor by these values.
     void setTranslationSensitivity(float x, float y, float z);
-		
+	void setTranslationSensitivity(const glm::vec3& sens);
```
bakercp: I'd change `sens` to `sensitivity` for clarity.

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: sure!

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -163,52 +160,70 @@ class ofEasyCam : public ofCamera {
 	/// Returns the area bounds used for mouse control.
 	ofRectangle getControlArea() const;
 	
+	
+	enum easyCamTransformType{
```
bakercp: Can we Change the name to `EasyCamTransformType` (to fix caps) or just `TransformType`, since it is already in the `ofEasyCam` class?

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: no prob!

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -163,52 +160,70 @@ class ofEasyCam : public ofCamera {
...
+		EASYCAM_TRANSFORM_NONE,
```
bakercp: Likewise here, can we change these to `TRANSFORM_TYPE_NONE`, `TRANFORM_TYPE_ROTATE`, etc?

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: I don't know. I was there from before.

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: ok

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -163,52 +160,70 @@ class ofEasyCam : public ofCamera {
...
+		EASYCAM_TRANSFORM_ROTATE,
+		EASYCAM_TRANSFORM_TRANSLATE_XY,
+		EASYCAM_TRANSFORM_TRANSLATE_Z,
+		EASYCAM_TRANSFORM_SCALE
+	};
+
+	void addInteraction(easyCamTransformType type, int mouseButton, int key = -1);
+	void removeInteraction(easyCamTransformType type, int mouseButton, int key = -1);
+	bool getHasInteraction(easyCamTransformType type, int mouseButton, int key = -1);
```
bakercp: `getHas*` reads strangely -- can we just say `hasInteractionType(...)` or something?

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -163,52 +160,70 @@ class ofEasyCam : public ofCamera {
...
 	ofNode target;
 
 	bool bEnableMouseMiddleButton = true;
 	bool bApplyInertia = false;
-	bool bDoTranslate = false;
-	bool bDoRotate = false;
-	bool bDoScrollZoom = false;
-	bool bIsBeingScrolled = false;
+	
+//	bool bDoTranslate = false;
```
bakercp: Just remove if commented?

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: ha, sure! I also spotted that earlier today.

libs/openFrameworks/3d/ofEasyCam.h
```diff
@@ -243,7 +257,28 @@ class ofEasyCam : public ofCamera {
...
+		interaction():mouseButton(0), key(-1), transformType(EASYCAM_TRANSFORM_NONE){}
+		interaction(easyCamTransformType type, int _mouseButton, int _key = -1):mouseButton(_mouseButton), key(_key), transformType(type){}
+		int mouseButton;
+		int key;
+		easyCamTransformType transformType;
+	};
+
+	vector< interaction > interactions;
+	
+
```
bakercp: Remove extra spaces?

libs/openFrameworks/3d/ofEasyCam.cpp
bakercp: Magic numbers!  🦄

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.cpp
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: done

libs/openFrameworks/3d/ofEasyCam.h
roymacdonald: done

roymacdonald:
@bakercp just made those fixes. thanks for reviewing.
have a nice day

tgfrerer:
I've just tested these latest changes, thanks @roymacdonald, and sorry for not getting there earlier - and it looks grand! Ortho and Perspective now have the same mouse interaction, and ortho zoom is rather nice =) I think it's ready to merge...

roymacdonald:
Anything for merging this PR?
@arturoc @ofZach @ofTheo @elliotwoods 

best!

arturoc:
thanks!

mattfelsen:
Wow, I totally missed all this conversation on this PR - thanks for taking my basic/rough sketch all the way! 😄

libs/openFrameworks/3d/ofEasyCam.cpp
mattfelsen: @roymacdonald @bakercp Yea...magic numbers 😕  There's another one somewhere (`0.035` I think?) that just made the zooming in ortho feel right and somewhat match the sensitivity for the perspective camera ¯\_(ツ)_/¯

PR #5608, opened 2017-05-15
remove dependency ofMatrixStack<->ofFbo

tgfrerer:
+ `ofMatrixStack` currently depends on `ofFbo`, although it really only calls methods from `ofFbo`'s base class, `ofBaseDraws` - `ofMatrixStack` exclusively queries into the RenderSurface's geometry.

The proposed changes update `ofMatrixStack` to require just an `ofBaseDraws`-compatible object to be passed as target render surface.

This allows other renderers, such as the vulkan renderer, to use this object, without having access to `ofFbo`.


arturoc:
thanks!

PR #5609, opened 2017-05-15
Making sure json file exists before reading from it

prisonerjohn:
otherwise it crashes


bakercp:
Perhaps just use `ofLoadJson()` to keep things consistent?

https://github.com/openframeworks/openFrameworks/blob/d7fd67ea348ed05e7a7c52d50d0a85d129a7bc42/addons/ofxGui/src/ofxBaseGui.cpp#L194

prisonerjohn:
Good call, just updated it.

arturoc:
hey sorry just saw this, i think this was done without using ofLoadJson cause that would log an error message if trying to load a .json that doesn't exist but in this case it shouldn't be an error

prisonerjohn:
How come it shouldn't be an error, if the file we're trying to load doesn't exist?

arturoc:
oh sorry, i though this was when trying to load before saving not on the load method itself.

PR #5611, opened 2017-05-18
Fixes for ofBitmapFont getBoundingBox();

roymacdonald:
- incorrect use of numeric_limits<float>::min() caused wrong results. Replaced for numeric_limits<float>::lowest()
- Added  ofDrawBitmapMode mode = ofGetStyle().drawBitmapMode, bool vFlipped = ofIsVFlipped() arguments to getBoundingBox. This allows to get the correct mesh and doesn't break previous way of use of this function.

### DIFF
--- a/libs/openFrameworks/graphics/ofBitmapFont.cpp
+++ b/libs/openFrameworks/graphics/ofBitmapFont.cpp
@@ -418,11 +418,11 @@ ofMesh ofBitmapFont::getMesh(const string & text, int x, int y, ofDrawBitmapMode
 		if(text[c] == '\n'){
 
 			sy += lineHeight*newLineDirection;
-			if(mode == OF_BITMAPMODE_SIMPLE) {
+//			if(mode == OF_BITMAPMODE_SIMPLE) {
 				sx = x;
-			} else {
-				sx = 0;
-			}
+//			} else {
+//				sx = 0;
+//			}
 
 			column = 0;
 		} else if (text[c] == '\t'){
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/graphics/ofBitmapFont.cpp
```diff
@@ -418,11 +418,11 @@ ofMesh ofBitmapFont::getMesh(const string & text, int x, int y, ofDrawBitmapMode
 		if(text[c] == '\n'){
 
 			sy += lineHeight*newLineDirection;
-			if(mode == OF_BITMAPMODE_SIMPLE) {
+//			if(mode == OF_BITMAPMODE_SIMPLE) {
```
bakercp: Can we remove of the commented parts?

libs/openFrameworks/graphics/ofBitmapFont.cpp
```diff
@@ -474,20 +474,19 @@ const ofTexture & ofBitmapFont::getTexture() const{
...
-ofRectangle ofBitmapFont::getBoundingBox(const string & text, int x, int y) const{
+ofRectangle ofBitmapFont::getBoundingBox(const string & text, int x, int y, ofDrawBitmapMode mode, bool vFlipped) const{
     if(text.empty()){
         return ofRectangle(x,y,0,0);
     }
 
-	const ofMesh & mesh = getMesh(text,x,y);
-	glm::vec2 max(numeric_limits<float>::min(),numeric_limits<float>::min());
+	const ofMesh & mesh = getMesh(text,x,y, mode, vFlipped);
+	glm::vec2 max(numeric_limits<float>::lowest(),numeric_limits<float>::lowest());
```
bakercp: Nice find.

libs/openFrameworks/graphics/ofBitmapFont.h
```diff
@@ -22,7 +23,7 @@ class ofBitmapFont{
 	~ofBitmapFont();
 	ofMesh getMesh(const string & text, int x, int y, ofDrawBitmapMode mode=OF_BITMAPMODE_MODEL_BILLBOARD, bool vFlipped=true) const;
 	const ofTexture & getTexture() const;
-	ofRectangle getBoundingBox(const string & text, int x, int y) const;
+	ofRectangle getBoundingBox(const string & text, int x, int y, ofDrawBitmapMode mode = ofGetStyle().drawBitmapMode, bool vFlipped = ofIsVFlipped()) const;
```
bakercp: `ofGetStyle().drawBitmapMode` looks funny, but I'm assuming it's the only way to do this to preserve backward compatibility right?

bakercp: ... to clarify "looks funny" .. to call a member variable of a global instance as a default parameter in a function.  Typically I avoid non-constant default parameters because they are ... not constant :)

libs/openFrameworks/graphics/ofBitmapFont.cpp
roymacdonald: I didn't remove this because I really don't understand why it it is there in first place although the most logical thing is that it shouldn't be there. If there's no problem removing it I will.

libs/openFrameworks/graphics/ofBitmapFont.h
roymacdonald: Ha, right. There seems to be no other way to get the current bitmap mode. And right, it also looked funny to me, but it is there to keep backwards compatibility.

libs/openFrameworks/graphics/ofBitmapFont.cpp
roymacdonald: :)

roymacdonald:
@arturoc @bakercp any chance to merge this PR?

Issue #5500, opened 2017-03-20
ofTessellator error on Raspberry Pi


kr15h:
Compiling ofxPiMapper on Raspberry Pi 3 with Raspbian Jessie. The libs/openFrameworksCompiled project compiles without problems, but I get this output when failing to compile the ofxPiMapper/example from the master branch.

```
/home/pi/openframeworks-master/libs/openFrameworksCompiled/lib/linuxarmv6l/libopenFrameworks.a(ofTessellator.o): In function `ofTessellator::init()':
ofTessellator.cpp:(.text+0x74): undefined reference to `tessNewTess'
/home/pi/openframeworks-master/libs/openFrameworksCompiled/lib/linuxarmv6l/libopenFrameworks.a(ofTessellator.o): In function `ofTessellator::operator=(ofTessellator const&)':
ofTessellator.cpp:(.text+0xec): undefined reference to `tessDeleteTess'
/home/pi/openframeworks-master/libs/openFrameworksCompiled/lib/linuxarmv6l/libopenFrameworks.a(ofTessellator.o): In function `ofTessellator::performTessellation(ofPolyWindingMode, ofMesh_<glm::tvec3<float, (glm::precision)0>, glm::tvec3<float, (glm::precision)0>, ofColor_<float>, glm::tvec2<float, (glm::precision)0> >&, bool)':
ofTessellator.cpp:(.text+0x128): undefined reference to `tessTesselate'
ofTessellator.cpp:(.text+0x1f0): undefined reference to `tessGetVertexCount'
ofTessellator.cpp:(.text+0x200): undefined reference to `tessGetElementCount'
ofTessellator.cpp:(.text+0x288): undefined reference to `tessGetVertices'
ofTessellator.cpp:(.text+0x2b8): undefined reference to `tessGetElements'
/home/pi/openframeworks-master/libs/openFrameworksCompiled/lib/linuxarmv6l/libopenFrameworks.a(ofTessellator.o): In function `ofTessellator::tessellateToMesh(ofPolyline_<glm::tvec3<float, (glm::precision)0> > const&, ofPolyWindingMode, ofMesh_<glm::tvec3<float, (glm::precision)0>, glm::tvec3<float, (glm::precision)0>, ofColor_<float>, glm::tvec2<float, (glm::precision)0> >&, bool)':
ofTessellator.cpp:(.text+0x400): undefined reference to `tessAddContour'
[...]

PR #5616, opened 2017-05-18
Download correct RPI libs when arch is not spec'd

bakercp:
This is a problem that people working with the master branch (not the nightlies or releases) encounter when following the instructions.

Related
https://github.com/openframeworks/openFrameworks/issues/5500
https://github.com/bakercp/ofxIO/issues/22
https://forum.openframeworks.cc/t/raspberry-pi-3-arm6-or-arm7/26800/4


arturoc:
thanks!

PR #5215, opened 2016-08-23
Const corrections for std::filesystem::path parameters.

bakercp:
Fixes proposed in #5208.

PR #5618, opened 2017-05-19
Const corrections for std::filesystem::path, convert additional std::string to std::filesystem::path

bakercp:
Replaces https://github.com/openframeworks/openFrameworks/pull/5215


bakercp:
Replaced by #5626

PR #5620, opened 2017-05-19
Fix unsigned / signed comparison warnings in ofVideoPlayer.

bakercp:
These warnings are seen in linux in particular.


arturoc:
thanks!

PR #5624, opened 2017-05-21
Deprecate macros (e.g. PI, RAD_TO_DEG, etc) and remove from the core, fixes #5163.

### LINKED ISSUE #5163: Remove evil constants/macros defined in ofConstants.h
These are definetely evil:

```cpp
#ifndef PI
	#define PI       3.14159265358979323846
#endif

#ifndef TWO_PI
	#define TWO_PI   6.28318530717958647693
#endif

#ifndef M_TWO_PI
	#define M_TWO_PI   6.28318530717958647693
#endif

#ifndef FOUR_PI
	#define FOUR_PI 12.56637061435917295385
#endif

#ifndef HALF_PI
	#define HALF_PI  1.57079632679489661923
#endif

#ifndef DEG_TO_RAD
	#define DEG_TO_RAD (PI/180.0)
#endif

#ifndef RAD_TO_DEG
	#define RAD_TO_DEG (180.0/PI)
#endif

#ifndef MIN
	#define MIN(x,y) (((x) < (y)) ? (x) : (y))
#endif

#ifndef MAX
	#define MAX(x,y) (((x) > (y)) ? (x) : (y))
#endif

#ifndef CLAMP
	#define CLAMP(val,min,max) ((val) < (min) ? (min) : ((val > max) ? (max) : (val)))
#endif

#ifndef ABS
	#define ABS(x) (((x) < 0) ? -(x) : (x))
#endif
```

## Suggestions

- `PI`
  - I suggest a `constexpr` value with `M_PI`.

- `TWO_PI`
  - Trivial change.

- `M_TWO_PI`
  - Trivial change.

- `FOUR_PI`
  - Trivial change.

- `HALF_PI`
  - Trivial change.

- `DEG_TO_RAD`
  - This should be a `constexpr` function.

- `RAD_TO_DEG`
  - This should be a `constexpr` function.

- `MIN`
  - This should use `std::min()`.

- `MAX`
  - This should use `std::max()`.

- `CLAMP`
  - This should be a `constexpr` function.

- `ABS`
  - This should use `std::abs`.
  - We need a float version which uses `std::fabs` too.

bakercp:
I think I've gotten all of them, the waiting on CI.

Fixes #5163


bakercp:
@arturoc no?

ofTheo:
I would be in favor of this (with a bit of double checking it doesn't break anything) but I think we should have enable deprecated macros by default for a few versions.  

Is it possible to do an inline warning when people use the old macros like PI or ABS?
That really helps getting people to switch over.

arturoc:
yes feel free to rebase and merge not sure why it wasn't merged before sorry

bakercp:
I will work on rebasing this today.

ofTheo:
@bakercp if you are able to rebase and merge would be awesome. Otherwise @dimitre or I can take a stab at it. 
Thanks again for this!! 🙂

bakercp:
Hi @ofTheo  - unfortunately I don't have the bandwidth to merge this at the moment.

ofTheo:
No worries @bakercp, thanks for letting us know. 👍
I can pull this into another branch and do a PR. 

To @ofZach's point I think we will need to update examples too ( see: https://github.com/openframeworks/openFrameworks/search?p=3&q=PI ) 

So I can take care of that. 
I think I'll aim to deprecate first ( so leave the legacy macro define for the next release ) then we can comment it out for the release after that. 

Thanks for all of this!

dimitre:
I'll be closing this one as the similar one was recently merged
- https://github.com/openframeworks/openFrameworks/pull/7918

PR #5626, opened 2017-05-21
Switched from std::string to std::filesystem::path in Shader::Source

prisonerjohn:


bakercp:
Some overlap with https://github.com/openframeworks/openFrameworks/pull/5618/files perhaps you could merge my PR into this one to catch all of your fixes too?

prisonerjohn:
Here we go! Sorry for the delay, thought this would be more complicated than it actually was :)

bakercp:
Neat!  Closing my PR.  Thanks!

PR #5632, opened 2017-05-22
Fixes #757 (set newly allocated pixels to 0).

### LINKED ISSUE #757: ofPixels::allocate should init new memory
in my opinion, it would make sense for ofPixels::allocate to initialize the new memory to 0. even for large images, i don't think this is such an expensive operation.

i discovered that it is not initialized to 0 when i had a really weird bug -- some code that gave me different results depending on whether i loaded another image before allocating or not. i was expecting OF to act like opencv, which also sets new memory to 0. this would be in line with ofVec*, ofColor, and a few other things, which are all allocated to 0.

bakercp:


bakercp:
@arturoc what do you think of this one?

arturoc:
I'm not sure about this, I think it's nice to have the option to get uninitialized memory. Perhaps we could have some kind of method that allocates without initialization?

bakercp:
I was just trying to create PRs for easy "bugs", so I don't feel strongly about this one in particular.  @kylemcdonald do you still feel strongly about this?

bakercp:
ping @kylemcdonald @arturoc

bakercp:
@arturoc ?

PR #5637, opened 2017-05-24
ofSoundPlayer::getDurationMS

arturoc:
Implemented for openal and fmod, missing iOS and android

### DIFF
--- a/libs/openFrameworks/sound/ofBaseSoundPlayer.h
+++ b/libs/openFrameworks/sound/ofBaseSoundPlayer.h
@@ -3,7 +3,7 @@
 
 #include "ofConstants.h"
 #include "ofFileUtils.h"
-
+#include "ofLog.h"
 
 
 //----------------------------------------------------------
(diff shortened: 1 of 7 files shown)


bakercp:
In other places we refer to milliseconds as "millis" (i.e ofGetElapsedTimeMillis()). should we use the same convention throughout?

arturoc:
yeah but since there's other methods in ofSoundPlayer that use MS i think it's better to use this and then perhaps send a different PR that deprecates them and changes them with millis

bakercp:
Why add a new method just to deprecate it?  Why not just deprecate the old methods and change them now?  Adding this feature with a corrected name will encourage people to use the un-deprecated methods.

hiroMTB:
If we change the name of all related time function, it will break many old project.
How about 2 step update like,

1. implement getDuration()
2. change all related function name

I remember JUCE or some another library has enum like TIME_MS, TIME_SEC etc in order to specify returning time format . But this need more discussion.

hiroMTB:
Here is short summary. 
MS is used in media(sound and video) class.
Millis is used for function related to system.

MS
ofSoundBuffer::getDurationMS()
ofSoundBuffer::getDurationMicros()
ofSoundBuffer::getDurationNanos()
ofBaseSoundPlayer::setPositionMS(int ms)
ofBaseVideoPlayer::getDuration() (return in seconds)

Millis
ofGetElapsedTimeMillis()
ofTime::getAsMilliseconds()
ofTime::getAsMicroseconds()
ofTime::getAsNanoseconds()
ofTime::getAsSeconds()

dimitre:
I'll be closing this one since an equivalent was recently merged
- https://github.com/openframeworks/openFrameworks/pull/7747

PR #5655, opened 2017-06-09
Update vs template for when multiple addons with identical class names are used.

bakercp:
Fixes https://github.com/openframeworks/projectGenerator/issues/159
Fixes https://github.com/bakercp/ofxJSONRPC/issues/10

See also https://github.com/bakercp/ofxJSONRPC/issues/10

### DIFF
--- a/scripts/templates/vs/emptyExample.vcxproj
+++ b/scripts/templates/vs/emptyExample.vcxproj
@@ -97,6 +97,7 @@
       <WarningLevel>Level3</WarningLevel>
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <CompileAs>CompileAsCpp</CompileAs>
+      <ObjectFileName>$(IntDir)/%(RelativeDir)/</ObjectFileName>
     </ClCompile>
     <Link>
       <GenerateDebugInformation>true</GenerateDebugInformation>
@@ -117,6 +118,7 @@
       <WarningLevel>Level3</WarningLevel>
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <CompileAs>CompileAsCpp</CompileAs>
+      <ObjectFileName>$(IntDir)/%(RelativeDir)/</ObjectFileName>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
     </ClCompile>
     <Link>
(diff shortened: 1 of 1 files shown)


armadillu:
This solution can lead to getting the object files all over the place. In my case, I'm including files that are several directories above the project;

```
<ObjectFileName>$(IntDir)/%(RelativeDir)/</ObjectFileName>
```

For example, I have some addons outisde the OF folder itself which leads to my include paths looking like this `..\..\ExternalAddons\ofxScreenSetup\src`... Which leads to this path for the obj files `"MyOfProject/obj/../../ExternalAddons/ofxScreenSetup\src/*.obj"`. I end up with two random folders inside my project folder; one called ExternalAddons and one called OpenFrameworks. 

All of this weird path arrangements are to be able to keep OF as a submodule inside a project repo; in our company we need to be able to clone projects easily, so we came up with a git repo structure where the project itself is in the main repo, and OF and whatever addons are required are added to the repo as submodules. This means we can't put anyhting inside the OF folder, this having this ../../ paths all over the place.

armadillu:
I propose we change it to :

```
 <ObjectFileName>$(IntDir)/$(Platform)/%(Directory)/</ObjectFileName>
```
Which leads to this obj file setup:
(For this example, building debug and x64)

```
MyProjectDir/
   MyProject.sln
   MyProject.vcproj
   ...
   obj/
      Debug/
         x64/
            Users/
               oriol/
                  Desktop/
                     MyProject/
                        apps/
                           MyProjectDir/
                              src/
                                 main.obj
                                 *.obj
                        OpenFrameworks/
                           addons/
                              oxOsc/
                                 src/
                                    *.obj

```

This create really deep folder structures, but it makes sure everything is contained within obj. 
Also, it allows you to preserve obj files for different architectures (ie changing 64bit and 32 bit wont lead to a full recompile).

armadillu:
@bakercp @arturoc Do you guys see any issues with that? Otherwise I will prepare a PR...

bakercp:
I have no problems with any solution that continues to solve the original problem. The one issue I'd watch out for is paths getting too long.  I've run into a number of very odd errors on windows that turns out to be the result of very long file paths (e.g. https://github.com/elliotwoods/ofxKinectForWindows2/issues/72).  So just something to keep an eye on.

armadillu:
Hmm that is a good catch - ill try find a solution that doens't potentially create this problem

PR #5666, opened 2017-06-14
add libglm-dev to dependencies for ubuntu

admsyn:
Adds `libglm-dev` to the dependencies script

Only Ubuntu (since it's what I have available), but it seems to be missing on Debian as well https://github.com/openframeworks/openFrameworks/blob/master/scripts/linux/debian/install_dependencies.sh#L38


arturoc:
Thanks Adam but we are using libglm as a local library by now in linux since there's still some changes that make things incompatible from one version to the next, once glm reaches 1.0 we'll move it to a package and remove it from apothecary for all the linux distros.

I'm going to close this by now

PR #5667, opened 2017-06-14
Update install-from-github guide

admsyn:
Minor tweaks, added an example of how to use the CLI PG to generate example makefiles.


arturoc:
thanks!

PR #5668, opened 2017-06-14
scale number of make jobs to number of processors on linux

admsyn:
Essentially `make -j8` vs `make -j1` as default.

Is there reason for it to be a default of 1? I imagine the rPI might have a little difficulty, but it doesn't seem like that was part of the reasoning from the [original commit](https://github.com/openframeworks/openFrameworks/commit/10a5baaf15fb7ea44a7a2cba6eca0d5aac529b7c)

### DIFF
--- a/scripts/linux/compileOF.sh
+++ b/scripts/linux/compileOF.sh
@@ -14,7 +14,14 @@ SCRIPTPATH=`pwd`
 popd > /dev/null
 
 BUILD="install"
-JOBS=`nproc --all`
+
+# if nproc exists, use it to scale the number of jobs to number of processors
+if hash nproc 2>/dev/null; then
+  JOBS=`nproc --all`
+else
+  JOBS=1
+fi
+
 while getopts tj: opt ; do
 	case "$opt" in
 		t)  # testing, only build Debug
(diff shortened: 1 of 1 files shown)


bakercp:
Three comments - 

1)  Raspberry Pi 3 always fails for me (runs out of memory / maxes out disk i/o) when using all 4 cores to compile.  Usually using 3 cores will work.  Sometimes only two work.  I'm hesitant to recommend not defaulting to maximal # cores.  It may also cause issues with CI, though we may already manually specify `j`.  The underlying `make` doesn't automatically max out cores, but lets the user decide.  Perhaps we should continue to follow suit?  

That said I'm all for better defaults ... but perhaps make them more sensitive to lighter hardware?  Maybe `nproc - 1`. :)

2) The -j option allows users to specify if they want to use all of them.

3) If we change, probably should update the msys2 script to match https://github.com/openframeworks/openFrameworks/blob/ec0cc4418c51f04dded1fb1917ac2bcd5bc2da39/scripts/msys2/compileOF.sh#L10-L18?

Cheers!

admsyn:
It should be scaling down to available memory instead of processors if we're going that route, but that seems too clever for this bash script.

arturoc:
one possible solution would be to use:

```
nproc --ignore=2
```

newer computers will still use most cores and older ones and small boards will use between 1 and 2

Issue #5647, opened 2017-06-01
gcc 7.1.1 breaks OF OF 0.9.8


riccardolardi:
I just updated gcc and gcc-lib packages on Arch Linux and cannot make any OF project anymore. It seems to be an issue related to nlohmann's json.hpp. See here: https://github.com/nlohmann/json/issues/590

bakercp:
Seems to be related to `using namespace std`.

Here's the location:

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofConstants.h#L435

Here is the open issue related to removing using namespace.

https://github.com/openframeworks/openFrameworks/issues/2581

dimitre:
potentially closed by https://github.com/openframeworks/openFrameworks/pull/7005

PR #5671, opened 2017-06-15
Remove using namespace std

Paul92:
As discussed in #5647, GCC 7.1 seems to break OF due to the inclusion of ```using namespace std```. Actually, the problem seems to be a conflict with a [json](https://github.com/nlohmann/json) library, but it fixed by removing ```using namespace std``` from the core library, which is known is a bad practice and seems to be on the TODO list for quite a while (#2581).

This PR removes the last occurrences of this in the core library and fixes the compilation problem with GCC 7.1.

It breaks projectGenerator (a PR will follow soon).

I created this pull request to get feedback on this. IT IS NOT READY FOR MERGING YET.

More testing should be done to make sure something else doesn't break, including with other versions of GCC. Also, there are still some occurrences of ```using namespace std``` in some addons, which I am happy to remove if there is interest in this.

Another problem which should be considered is backwards compatibility. There is quite a lot of OF code present and some of it probably relies on the presence of ```using namespace std``` in the core library. A decision should be made in this sense. My suggestion (not yet added to this PR) is to add a ```using namespace std``` at the end of a header file for the core library (```ofMain.h```?) such that the code that uses the library is not impacted and the library can compile under the new GCC.

### DIFF
--- a/addons/ofxAndroid/src/ofAppAndroidWindow.cpp
+++ b/addons/ofxAndroid/src/ofAppAndroidWindow.cpp
@@ -20,7 +20,6 @@
 #include "ofGLProgrammableRenderer.h"
 #include "ofGLRenderer.h"
 #include "ofBaseTypes.h"
-using namespace std;
 
 static bool paused=true;
 static bool surfaceDestroyed=false;
@@ -37,7 +36,7 @@ static ofAppAndroidWindow * window;
 
 static ofOrientation orientation = OF_ORIENTATION_DEFAULT;
 
-static queue<ofTouchEventArgs> touchEventArgsQueue;
+static std::queue<ofTouchEventArgs> touchEventArgsQueue;
 static std::mutex mtx;
 static bool threadedTouchEvents = false;
 static bool appSetup = false;
@@ -121,9 +120,9 @@ void ofAppAndroidWindow::setup(const ofGLESWindowSettings & settings){
 	glesVersion = settings.glesVersion;
 	ofLogError() << "setup gles" << glesVersion;
 	if(glesVersion<2){
-		currentRenderer = make_shared<ofGLRenderer>(this);
+		currentRenderer = std::make_shared<ofGLRenderer>(this);
 	}else{
-		currentRenderer = make_shared<ofGLProgrammableRenderer>(this);
+		currentRenderer = std::make_shared<ofGLProgrammableRenderer>(this);
 	}
 
 	jclass javaClass = ofGetJNIEnv()->FindClass("cc/openframeworks/OFAndroid");
@@ -229,7 +228,7 @@ ofCoreEvents & ofAppAndroidWindow::events(){
 	return coreEvents;
 }
 
-shared_ptr<ofBaseRenderer> & ofAppAndroidWindow::renderer(){
+std::shared_ptr<ofBaseRenderer> & ofAppAndroidWindow::renderer(){
 	return currentRenderer;
 }
 
@@ -252,8 +251,8 @@ Java_cc_openframeworks_OFAndroid_setAppDataDir( JNIEnv*  env, jobject  thiz, jst
 {
 	jboolean iscopy;
 	const char *mfile = env->GetStringUTFChars(data_dir, &iscopy);
-	__android_log_print(ANDROID_LOG_INFO,"ofAppAndroidWindow",("setting app dir name to: \"" + string(mfile) + "\"").c_str());
-    ofSetDataPathRoot(string(mfile)+"/");
+	__android_log_print(ANDROID_LOG_INFO,"ofAppAndroidWindow",("setting app dir name to: \"" + std::string(mfile) + "\"").c_str());
+    ofSetDataPathRoot(std::string(mfile)+"/");
     env->ReleaseStringUTFChars(data_dir, mfile);
 }
 
@@ -374,7 +373,7 @@ Java_cc_openframeworks_OFAndroid_render( JNIEnv*  env, jclass  thiz )
 
 	if(!threadedTouchEvents){
 		mtx.lock();
-		queue<ofTouchEventArgs> events = touchEventArgsQueue;
+		std::queue<ofTouchEventArgs> events = touchEventArgsQueue;
 		while(!touchEventArgsQueue.empty()) touchEventArgsQueue.pop();
 		mtx.unlock();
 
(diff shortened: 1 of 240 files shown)


Paul92:
Ok, an update on this.

I removed ```using namespace std``` from addons too and made sure CI is happy (at least as happy as possible - there is a test falling on travis and it doesn't seem to be my fault).

As far as I can tell, my job is done. It would be great if I could get some feedback on this and if there is something else that needs to be done.

tgfrerer:
Hi @Paul92 that's great work - yes, there have been various pushes at getting namespaces to work within oF, most notably the issue #2581 you referenced. I believe the general consensus was enthusiastic about using namespaces, and there was a bit of concern when best to add this to oF, as it might break people's projects. 

Perhaps it's time to prioritise this for the next major release?

---
I've been reading through some of the changes you propose, and i would, to keep the number of lines which are touched lower, and to keep the code more readable, leave some `using namespace std;` in the code where there is no risk of namespace clashes. 

As long as the `using` directive is put into a `.cpp` file it affects only the current *translation unit*, meaning the `.cpp` file where the directive is placed in. If we take the `using namespace std;` out of `ofConstants.h` but put it into some `.cpp` files within oF we can keep changes lower. We'd still have to change all header files to prefix `std::` where we use elements from the STL...

Paul92:
Hi @tgfrerer. I'm aware that I did sort of a bundled commit where I squashed everything, partially due to the fact that I had only one environment available and had to rely on the output from CI. Once we decide what is needed and what not, I will create another PR with nicer commits.

I do believe the project should move towards explicit ```std::``` everywhere at some point.

The minimal change I suggest for now (with a touch of movement in the direction mentioned above) would be to add ```std::``` in headers and ```using namespace std;``` in all ```.cpp``` files where is required, along with one ```using namespace std;``` at the end of ```ofMain.h```. And also, the scope of this can be reduced only to the core library/affected addons. From my experience as a user of OF, this should not break user's code.

An even smaller change would be to add ```using namespace std;``` at the end of ```ofConstants.h``` (and some minor collateral changes). I think this should fix the ```gcc 7.1``` build issue.

Since we all seem to agree that we should move from ```using namespace std``` at some point, the question is how gradual and when would you like the (full) migration to happen.

bakercp:
I'm for this update 100%.

My main concerns are:
- minimize impact to 3rd party addons. It sounds like we have a solution to that. That said, a lot of addons are probably going to need maintenance due to the glm updates, so now it's as good a time as any.
- minimize impact to existing pull requests as some high quality ones have been in the queue for a very long time. The solution there is probably too get them merged ASAP.

bakercp:
Regarding code readability, to some extent that is preference and since it's in the core, and the benefit of fully qualified namespaces could mean less work in the future, I'd ultimately vote for fully qualified throughout.  ... As long as we can minimize damage to existing PRs.

Paul92:
@bakercp  I pretty confident that there will be (lots of) conflicts no matter what with this kind of changes. However, I'm not sure whether delaying this would help - there will always be a pending PR or a fork somewhere. Even worse, leaving this PR pending for too long and accepting many changes into the core would make this one impossible to merge. A possible meet in the middle solution is suggested above, by reducing the scope of this PR. But then, at some point these changes need to be done.

Paul92:
Hi guys,

I am a bit confused about the decision on this matter. The current options, in my opinion, are:

- Merge everything
- Move the ```using namespace std;``` in ```ofConstants.h``` at the end - minimal changes
- Move the ```using namespace std;``` in ```ofMain.h``` - adds explicit ```std::``` in the whole core, the impact on user code is limited
- Don't do anything and go home :smile: 

I'm open to any other suggestions.

tgfrerer:
I'm for limiting the amount of change in core, because it introduces a lot of headache for PR's which are pending.

I'd suggest using `using namespace std` at the top of `.cpp` files where there are no clashes (for readability, and because it touches less lines of code), and to fully qualify the core header files with `std::` prefix.

We could then phase in "`std::` everywhere" (i.e. in the core .cpp files too) once pending PRs are merged for the next major release (0.10, i believe).

I'd remove `using namespace std` from any global header file, that's `ofMain.h`, or `ofContants.h` - we can add it to examples in their ofApp.h header if we want to keep them less intimidating for beginners - or change all example user code to use std::everywhere =)

bakercp:
That sounds reasonable.  One problem I run into consistently teaching OF is when students try to use addons, they rarely work out of the box.  Mostly this is because they are not maintained.  I'd guess that removing using std:: is going to add even more addons to the "Don't work" list.  I have some ideas for things like "curated" addons that help with the larger problem, but do you all have ideas for how we can help new users discover and fix problems related to no longer having using std:: in the core?

I for one, have been teaching my students to use fully qualified std:: from day one of teaching :) but that doesn't mean they automatically understand what to do if they see "unknown" vector or something.  Maybe we could we add a quick paragraph the the README or docs somewhere as a part of this effort?

Paul92:
@tgfrerer I agree with your suggestion, I will make a PR with that in the following days. Also, I think we need to have a ```using namespace std;``` in a header file somewhere (```ofApp.h``` maybe) such that user code is not broken.

@bakercp I don't think there is an easy solution to fixing broken addons. The "curated" list of addons really should be the "maintained" list of addons (if works now and it is not maintained, will stop working at some point). What I think would help users would be a changelog of backwards compatibility disruptions. For example, removing ```using namespace std``` is expected to break some addons - there should be an entry saying that due to this change, the following types of errors are to be expected and this is the fix. This way, when an addon is broken, there is a troubleshooting list that people can use to quickly find the issue.

arturoc:
@Paul92 this is great! we've talked about this a lot before but it always seemed like a huge amount of work.

I think having `using namespace std` in ofMain.h is a good idea and if i remember well has been discussed before it would leave most code working and still be easy to opt out of it.

Perhaps we could add some new files: `of3d.h, ofCommunication.h, ofEvents.h, ofGl.h, ofGraphics.h, ofMath.h, ofSound.h, ofTypes.h, ofUtils.h, ofVideo.h` ofApp.h to not clash with the examples header and because it's probably a mix of other things anyway, could be divided in ofEvents,h, ofUtils.h and perhaps ofWindow.h.

This would allow a finer grained include system which would not include std in the globla namespace

Paul92:
@arturoc Having module level include files as you suggest is a really nice idea - I can make a separate PR for it.

The thing that worries me currently is that addons might be broken. This can be seen in CI output of #5676, where I limited the number of lines changed by adding the ```std::``` qualifier only in header files, addons are broken since they rely on having ```using namespace std``` in a header which they directly include. It's not a big deal to fix the core addons, but I am worried about user addons. Unfortunately, I don't think there is an easy fix to this.

arturoc:
if we have `using namespace std` in ofMain.h i think most addons will work right away. some addons in the core include individual headers but most user addons and almost all core addons just include ofMain.h

PR #5674, opened 2017-06-21
Add logger channel getters (closes #5558)

### LINKED ISSUE #5558: Feature request: ofGetLoggerChannel() / ofLog::getChannel()
There's no way to retrieve the currently-set logger channel - only a setter, no getter.

A getter would be very nice, if, for instance, you want use a custom logger to that _also_ forwards calls to whatever channel's already set. E.g. it'd allow something like this:

```
shared_ptr<CustomLoggerChannel> channel(new CustomLoggerChannel);
channel->setForwardingChannel(ofGetLoggerChannel()); // forward to existing channel
ofSetLoggerChannel(channel);
```

daniel-at-yoke:
As discussed in the issue, this adds getters for the currently set logger channel. Specifically an `ofGetLoggerChannel()` function and a `ofLog::getChannel()` method. This allows channels to be swapped in and out more easily, or for custom channels to continue forwarding to whatever channel was already set.

### DIFF
--- a/libs/openFrameworks/utils/ofLog.cpp
+++ b/libs/openFrameworks/utils/ofLog.cpp
@@ -251,6 +251,14 @@ void ofSetLoggerChannel(shared_ptr<ofBaseLoggerChannel> loggerChannel){
 	ofLog::setChannel(loggerChannel);
 }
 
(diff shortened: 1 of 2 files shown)


arturoc:
thanks!

PR #5675, opened 2017-06-23
Add a getCamera() method to get the camera used

kkirby:
This would be helpful if you need to re-use the camera elsewhere in your project.


kkirby:
I don't think there is any risk in adding this. The person grabbing the camera instance just needs to know not to close it ... Too bad there isn't something like const-ness in Java. If we feel this is a deal then we could create an immutable wrapped around the Camera, but I feel like that's overkill.

arturoc:
thanks!

PR #5676, opened 2017-06-23
Removed using namespace std from headers

Paul92:
Here is the promised pull request in #5671 . It's removing the ```using namespace std``` from all header files, added it to ```.cpp``` files that required it.

Looking forward to hear your feedback.

P.S. Not ready for merging yet - some addons are broken


arturoc:
just went through it and everything seems to be working fine. thanks so much for all this work!

PR #5681, opened 2017-06-30
refactor: move ofTtfSettings into ofTrueTypeFont

tgfrerer:
+ following the same pattern we use with ofFbo and other objects

*Example:* Create a `Settings` object, and pass it to `load()` - here we are instructing an Icon font to load specific unicode ranges:

	ofTrueTypeFont myFont;

	ofTrueTypeFont::Settings settings("MaterialIcons-Regular.ttf", 26);
	settings.antialiased = true;
	settings.addRange({0xe889,0xe889}); // history
	settings.addRange({0xe034,0xe039}); // play/pause buttons
	
	myFont.load(settings);

The hope is to make the Settings object more discoverable, which should help especially with IDEs that have code completion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,31 +1,3 @@
- OF 0.9.9
-========                          
-
-#### change key
-          + added
-          - removed
-          / modified
-
-CORE
-----
-
-### graphics
-    / refactor ofTtfSettings -> ofTrueTypeFont::Settings
-
-
-PLATFORM SPECIFIC
------------------
-
-### osx
-
-### linux
-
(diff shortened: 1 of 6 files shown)


arturoc:
This seems to be failing cause ofxGui is using the previous syntax

tgfrerer:
I've updated ofxGui - also grep'ed `examples` and `addons` to see if there were any other uses of ofTtfSettings - seems the 4 locations in ofxGui is all. I've updated the CHANGELOG for good measure.

arturoc:
thanks!!

PR #5691, opened 2017-07-08
Fix OF_ROOT path

baydam:


arturoc:
thanks! one question, in which case does one work and not the other? seems like both would depend on running the script from it's own folder no?

i've found that the more robust way to do this is to use something like:

```
OF_ROOT=$(cd $(dirname $0)/../..; pwd -P)
```

which returns the absolute path 2 levels below the path the script is in no matter where it's run from. if you want to change it to that?

baydam:
Thanks you I will take note

PR #5697, opened 2017-07-18
vs: update download_libs.ps1

### LINKED ISSUE #5695: Errors running 'download_libs.ps1'
Hi,

I have a bunch of errors when running the PowerShell script to download all the libs (current Master). I must be doing something wrong but I can not find what. [Here ](https://www.dropbox.com/s/jdu1srycs44y8o2/download_libs-LOG.txt?dl=0)is the entire log.

The 1st error being:

> Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
At C:\Users\Hugues\Development\openFrameworks\scripts\vs\download_libs.ps1:18 char:5

Thanks,

Hugues.

tgfrerer:
+ adds vs_ver as a selectable option to the powershell download script, which allows you to select which version of the libraries to download.
+ adds a comment at top of script to explain how paramters may be specified

Default paramter value for vs_ver is 2015, which will load the libraries compiled with the visual studio 2015(== 140) toollset.

Call the script with:

    powershell.exe -f download_libs.ps1 -vs_ver 2017

To download & install the libraries compiled with the visual studio 2017(== 141) toolset.

This should also fix an issue where the old download_libs.ps1 script would not work, as the libraries archives on the server now have a slightly different naming scheme.

Fixes #5695

### DIFF
--- a/scripts/vs/download_libs.ps1
+++ b/scripts/vs/download_libs.ps1
@@ -1,11 +1,11 @@
 # You may override default parameters for this script by specifying 
 #       -paramName paramValue
 # When calling this script. For example: 
-#       download_libs.ps1 -vs_ver 2017
+#       download_libs.ps1 -platform vs2017
 # Which will load the visual studio 2017 libraries
 param(
     [String]$ver="master",
-    [String]$vs_ver="2015"
+    [String]$platform="vs2015"
     )
 $currentPath = $pwd
 $scriptPath = split-path -parent $MyInvocation.MyCommand.Definition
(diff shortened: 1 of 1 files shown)


arturoc:
in the sh version i changed this to use --platform vs2015 or vs2017 instead of vs only can you change it to that instead so both behave the same?

tgfrerer:
sure - that's done.

arturoc:
thanks!

PR #5701, opened 2017-07-21
Bugfix for ofxKinect not working on VS. Closes #5700

### LINKED ISSUE #5389: Add 64bit libs for ofxKinect for Windows
I think the current INF files only work for 32bit builds of OF on Windows. 
It would be good to bundle 64bit, so people can use ofxKinect with both platforms. 

See the end of this issue:
https://github.com/openframeworks/openFrameworks/issues/2452

And forum post:
https://forum.openframeworks.cc/t/solved-original-kinect-for-windows-windows-10-64bit/22840/11
https://forum.openframeworks.cc/t/ofxkinect-addons-doesnt-linking-libusb0-dll/22922/6

### LINKED ISSUE #5700: ofxKinect libusb fix for Visual Studio
Currently ofxKinect in OF master doesn't work with Visual Studio. 
This is because libfreenect uses latest libusb which breaks the Windows libusb emulation layer. 

Using the latest libusb from:
https://github.com/libusb/libusb/ 

And applying this PR:
https://github.com/libusb/libusb/pull/246 

Then building ( in Visual Studio 2015 with the latest DDK / WDK ) results in a working lib for ofxKinect that only requires the libusb header and .lib to run. 

All the code in:
https://github.com/openframeworks/openFrameworks/tree/master/addons/ofxKinect/libs/libusb-win32 is now no-longer required. 

NOTE: you need to install the libusbK drivers via Zadig for the device.  But that is already much easier than the all .inf approach. 

So the ofxKinect structure can be quite a bit simpler as a result.  

I am going to work on a PR to the current master to get things working for VS. 

Then look at what else needs changing.
Like MSYS2 etc.

ofTheo:
Going to look into getting this stuff into apothecary - but this will at least fix ofxKinect in master if thats helpful in the meantime. 

Closes #5700 
This PR also closes #5389


arturoc:
Thanks Theo, yes i would really prefer to remove all the binaries from this addon since it's the only thing that still has binary libraries, every other addon and core library is now compiled through apothecary. It's a little bit more of work the first time but will make maintenance easier in the future and the size of the repo is much smaller now mostly when cloning with depth=1

Another solution as i told in the dev list would be to move the addon to it's own repo and just keep the libraries there

ofTheo:
@arturoc - totally makes sense! 
going to take a stab at it later today as it should be easier overall in the long term. 

just got to get my brain back into apothecary world :)

arturoc:
:) thanks, let me know if you have any doubt

ofTheo:
@arturoc here is the VS apothecary script which generates libs which work with ofxKinect. 
https://github.com/openframeworks/apothecary/pull/74 
Going to work on adding in OS X next.  I believe other platforms are handles separately. 

It might be still good to merge this PR in #5701  ( once I update it to not include libusb ), as it removes all the files no longer needed in ofxKinect.

arturoc:
sure, feel free to merge it once you remove the binaries, also if you can create a new PR without the binaries at all in the history that would be great

PR #5704, opened 2017-07-24
Fixes "libglfw3-devel not found" on Fedora.

ShadowMitia:
And moved from `yum` package manager to `dnf` package manager,
which is supposed to replace it.


arturoc:
thanks!

PR #5706, opened 2017-07-24
Removing compiled libs from ofxKinect closes #5700 

### LINKED ISSUE #5700: ofxKinect libusb fix for Visual Studio

ofTheo:
Libusb will come via apotechary now. 
VS nows uses libusb not libusb-win32

This PR replaces #5701
Closes #5700


ofTheo:
So I think this should be good, now that libusb is being built by apothecary. 
@arturoc do you want me to change download_libs.sh to include libusb for vs / osx for ofxKinect? 

I think we would need to make that change to allow the CI above not to error. 
( as described here: https://github.com/openframeworks/apothecary/issues/76 )

arturoc:
yes if you can do that change that would be great. 

also right now windows is disabling ofxKinect in the tests so if you can remove this lines from the install script for the ci servers:

https://github.com/openframeworks/openFrameworks/blob/master/scripts/ci/vs/install.sh#L11-L14

with that windows will compile ofxKinect too 

thanks!

ofTheo:
I think both this PR and this PR ( #5707 ) need to be merged for things to build fine for ofxKinect. 
@arturoc  Feel free to merge both if it looks good to you.

Issue #5703, opened 2017-07-21
ofEasyCam::disableMouseInput() doesn't work in setup()


neilmendoza:
Calling ofEasyCam::disableMouseInput() in setup() doesn't work. Mouse input get's reenabled when ofEasyCam::begin() is called the first time.

roymacdonald:
thanks for pointing this out. I'll take a look at it. 
cheers

roymacdonald:
fixed by https://github.com/openframeworks/openFrameworks/pull/5710.
please close this issue.
cheers

PR #5709, opened 2017-07-28
ofEasyCam. fix disabling mouse interaction at start.

roymacdonald:
This should fix this issue https://github.com/openframeworks/openFrameworks/issues/5703

### DIFF
--- a/libs/openFrameworks/3d/ofEasyCam.cpp
+++ b/libs/openFrameworks/3d/ofEasyCam.cpp
@@ -176,7 +176,7 @@ void ofEasyCam::setEvents(ofCoreEvents & _events){
 
 	// we need a temporary copy of bMouseInputEnabled, since it will 
 	// get changed by disableMouseInput as a side-effect.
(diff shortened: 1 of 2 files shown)


roymacdonald:
This fix is actually very simple. The `setEvents(...)`method was checking if the `events` pointer was valid as a way to determine if it was the first time  `setEvents(...)` was being called, thus allowing the automatic events registration upon startup. The problem is that if `disableMouseEvents()` is called before the events registration; during setup, the events pointer will still be null thus rendering the `bMouseInputEnabled` flag useless during the first run of setEvents(...).
So, the idea is to forget about checking the events pointer validity, and instead have the `bMouseInputEnabled`bool set to` true` upon construction.  When `setEvents(...)` gets called the fissrt time it will register automatically the events. Hence if `disableMouseEvents()` is called during `setup()` it will set `bMouseInputEnabled = false;` thus disabling the automatic registration. 
It is a bit of a hack but I think that this is better than adding a new flag just to make the initial registration.
Any comments?

PR #5710, opened 2017-07-28
ofEasyCam. fix disabling mouse interaction at start. 

roymacdonald:
This should fix this issue #5703 

This fix is actually very simple. The setEvents(...)method was checking if the events pointer was valid as a way to determine if it was the first time setEvents(...) was being called, thus allowing the automatic events registration upon startup. The problem is that if disableMouseEvents() is called before the events registration; during setup, the events pointer will still be null thus rendering the bMouseInputEnabled flag useless during the first run of setEvents(...).
So, the idea is to forget about checking the events pointer validity, and instead have the bMouseInputEnabledbool set totrue upon construction. When setEvents(...) gets called the fissrt time it will register automatically the events. Hence if disableMouseEvents() is called during setup() it will set bMouseInputEnabled = false; thus disabling the automatic registration.
It is a bit of a hack but I think that this is better than adding a new flag just to make the initial registration.
BTW, I made another PR from the wrong branch in my repo.

### DIFF
--- a/libs/openFrameworks/3d/ofEasyCam.cpp
+++ b/libs/openFrameworks/3d/ofEasyCam.cpp
@@ -176,7 +176,7 @@ void ofEasyCam::setEvents(ofCoreEvents & _events){
 
 	// we need a temporary copy of bMouseInputEnabled, since it will 
 	// get changed by disableMouseInput as a side-effect.
-	bool wasMouseInputEnabled = bMouseInputEnabled || !events;
+	bool wasMouseInputEnabled = bMouseInputEnabled;// || !events;
 	disableMouseInput();
 	events = &_events;
 	if (wasMouseInputEnabled) {
--- a/libs/openFrameworks/3d/ofEasyCam.h
+++ b/libs/openFrameworks/3d/ofEasyCam.h
@@ -183,7 +183,7 @@ private:
 	bool bApplyInertia = false;
 	
 	bool bInsideArcball = false;
-	bool bMouseInputEnabled = false;
+	bool bMouseInputEnabled = true;
 	bool bDistanceSet = false;
 	bool bAutoDistance = true;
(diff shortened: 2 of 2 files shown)


arturoc:
thanks!

neilmendoza:
I think this might be creating another problem. The ofEasyCam::update() method is never called now so the distance is not set automatically.

roymacdonald:
@neilmendoza so `ofEasyCam::update()`never gets called when you disable the mouseInput, right? It makes sense.
I'll take a look at it during the weekend.
best

neilmendoza:
Cool, thanks! Yeah, initialising the cam settings probably needs to be separated from updating them somehow so there's a first run function that's not dependent on mouse interaction; either update can be called without updating position based on mouse or first run function is not dependent on update.

roymacdonald:
Yes, it will need a separate initialization. otherwise it will become a bit messy. I will take a look at this during the weekend.
best!

arturoc:
can we change this to a small state machine? this class seems to be really problematic cause it has a ton of variables to track the mutable state. something like:

```
enum State{
Initializing,
MouseEnabled,
MouseDisabled,
...?
} state = Initializing;


void update(){
    switch (state){
    case Initializting:
        // register mouse events
        state = MouseEnabled;
    break;
    case MouseEnabled:
        // handle mouse input
    break;
    case MouseDisabled:
        // probably nothing
    break;
    case ...?
    }
// common code
}

void enableMouse(){
     // register mouse events;
    state = MouseEnabled;
}
void disableMouse(){
     // unregister mouse events;
    state = MouseDisabled;
}
```

something like that should remove a lot of the variables we use to track the state and make things less error prone

roymacdonald:
Yeah, that's a good idea.
I'll take a look at it.
best

PR #5707, opened 2017-07-26
Add libusb to download_libs.sh and allow VS to include ofxKinect

ofTheo:

PR #5711, opened 2017-07-28
Merge kinect PRs

arturoc:
Merges #5707 and #5706 to test hat everything is building properly before merging to master


ofTheo:
Hmm. 
So the osxmake build is failing. 
It can't find libusb which isn't being included in libs/ 

Digging into it I see this at the top of the build: 
https://travis-ci.org/openframeworks/openFrameworks/jobs/258488095

```
elif [ "$TARGET" == "osxmake" ]; then scripts/ios/download_libs.sh; 
else scripts/$TARGET/download_libs.sh; fi
```
which seems a bit weird 

That leads to:
https://github.com/openframeworks/openFrameworks/blob/master/scripts/ios/download_libs.sh
```
../dev/download_libs.sh -p tvos $@
../dev/download_libs.sh -p ios -n $@
```

So it ends up running the iOS target for download_libs which isn't currently set to pull libusb in. 
See:https://github.com/arturoc/openFrameworks/blob/e9ceab6375b346dd30bc22af1fa80e02fa053a02/scripts/dev/download_libs.sh#L190

We could add a makeosx argument for download_libs.sh or remove ofxKinect from makeosx. 

Curious why its using iOS as opposed to OSX though.

arturoc:
i just fixed that but it wasn't really the problem, that line was never being executed. 

the problem was that the library in apothecary was being renamed to usb-1.0.0.lib which is not a valid library name in osx so the makefiles weren't picking it up, i've just changed it to usb-1.0.0.a which should fix it. i'll restart this once the libraries are in the server again

ofTheo:
Doh! So sorry.  
Totally missed that. :(

arturoc:
np! it seems to be working now in all platforms, thanks!

PR #5712, opened 2017-07-28
feature: draw camera frustum

tgfrerer:
+ add method to draw camera frustum into scene, so that a camera may be previewed showing its frustum. This is useful to debug camera views, and also allows you to preview near and far planes for cameras. Inline comments explain the math behind the draw operation.

Proposed method name & signature: `void ofCamera.drawFrustum()`

Minimal demo app shows how this feature may be used: https://gist.github.com/tgfrerer/28b2bf3ad689e4f8cc778644c9d6595e

Screengrabs from demo app: 
camera with perspective frustum
perspective camera view
camera enabled ortho


arturoc:
that's great thanks!

PR #5718, opened 2017-08-03
Fix ofVec2f to glm::vec2 conversion issue

hamoid:
Without this change the program fails to compile on master


bakercp:
This looks good.  As an aside, I'm thinking the CI tests should be catching stuff like this.

arturoc:
thanks!

PR #5721, opened 2017-08-09
Fixed bash syntax error in download_libs.sh

khusseini:
There was a small syntax error in the `download_libs.sh` which is fixed in this PR.


khusseini:
Hm... I am not sure why Travis is failing but it seems unrelated to my fix?

arturoc:
yes sometimes the network tests on osx fail for some reason. thanks!

PR #5723, opened 2017-08-09
Add script verifying NDK install

HalfdanJ:
Added a script that is verifying the NDK version, and giving better errors when wrong version is installed. Making it more user friendly.

### DIFF
--- a/docs/android_studio.md
+++ b/docs/android_studio.md
@@ -9,14 +9,14 @@ The Android distribution of openFrameworks is setup to work with Android Studio 
 
 Setting up openFrameworks with Android Studio is fairly straightforward. The basic steps are:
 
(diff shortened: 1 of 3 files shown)


danzeeeman:
I want to see how much work it is to update to the latest NDK so we can just use the one from Android Studio.  But this looks interesting

HalfdanJ:
Yeah, I agree, but we will have this issue all the time, so thought it was good to make a user friendly error

HalfdanJ:
Okay, need to work some more on this. It works in android studio that creates the local.properties file, but not from commandline (and CI).

HalfdanJ:
Ok, fixed it

arturoc:
thanks!

PR #5726, opened 2017-08-14
Added quotes around paths in post-build script

prisonerjohn:
otherwise would get an error if the OF path has spaces in it.


arturoc:
thanks!

PR #5728, opened 2017-08-18
Fix Qt Creator macOS

### LINKED ISSUE #5572: Qt Creator Mac: Error parsing ofApp.qbs

bakercp:
This is being discusses here:

https://github.com/openframeworks/openFrameworks/issues/5572

and will fix #5572 when it is complete.

- [X] Update `cpp.minimumMacosVersion` and fix missing quotes.
- [ ] Add libcurl linker / header flags to core, rather than in the project .qbs (as noted in the thread).
- [ ] Figure out the `$(PRODUCT_NAME)` error.
- [ ] Copy libfmodex.dylib and the the Resources folder to the .app bundle.
- [ ] Fix release compiler (missing symbols error)

@c-mendoza
@Daandelange 
@arturoc

### DIFF
--- a/libs/openFrameworksCompiled/project/qtcreator/ofApp.qbs
+++ b/libs/openFrameworksCompiled/project/qtcreator/ofApp.qbs
@@ -47,7 +47,7 @@ CppApplication{
 
     Properties{
         condition: of.platform === "osx"
-        cpp.minimumOsxVersion: 10.8
+        cpp.minimumMacosVersion: "10.9"
     }
 
     Probe{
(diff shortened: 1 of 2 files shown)


bakercp:
QT Creator 4.5 is now working on macos 10.13 with the latest master branch. Closing this. Thanks @arturoc

PR #5730, opened 2017-08-24
Allow setting attributes on the root XML node

hamoid:
Apparently this might be a copy/paste error. 
It results in a new `<:anonymous/>` tag when trying to add attributes to the root XML node.

Discussion here: https://forum.openframeworks.cc/t/27585


arturoc:
thanks!

PR #5737, opened 2017-08-30
Fixed polylineBlobExample type mismatch. Fixes #5713

### LINKED ISSUE #5713: polylineBlobsExample compilation error
when compiling polylineBlobsExample with MSYS2, I have the following error :

```
D:/mygit/openFrameworks/examples/graphics/polylineBlobsExample/src/ofApp.cpp: In member function 'virtual void ofApp::update()':
D:/mygit/openFrameworks/examples/graphics/polylineBlobsExample/src/ofApp.cpp:42:46: error: no matching function for call to 'ofPolyline_<glm::tvec3<float, (glm::precision)0> >::addVertices(std::vector<ofVec3f>&)'
    cur.addVertices(contourFinder.blobs[i].pts);
                                              ^
In file included from D:/mygit/openFrameworks/libs/openFrameworks/graphics/ofPolyline.h:527:0,
                 from D:/mygit/openFrameworks/libs/openFrameworks/gl/ofGLRenderer.h:3,
                 from D:/mygit/openFrameworks/libs/openFrameworks/ofMain.h:48,
                 from D:/mygit/openFrameworks/examples/graphics/polylineBlobsExample/src/ofApp.h:3,
                 from D:/mygit/openFrameworks/examples/graphics/polylineBlobsExample/src/ofApp.cpp:1:
D:/mygit/openFrameworks/libs/openFrameworks/graphics/ofPolyline.inl:60:6: note: candidate: void ofPolyline_<T>::addVertices(const std::vector<T>&) [with T = glm::tvec3<float, (glm::precision)0>]
 void ofPolyline_<T>::addVertices(const vector<T>& verts) {
      ^~~~~~~~~~~~~~
```

Paul92:


hamoid:
Hi! This PR seems to deprecate ofPoint https://github.com/openframeworks/openFrameworks/pull/5634 Should it use glm::vec2 instead?

Paul92:
Hi!
Indeed, I forgot about the deprecation of ofPoint. However, I think the problem lies deeper (and actually I don't think this is a good solution for the problem).

Not using ofPoint at all is not possible. By not touching the library, there are 2 options: either convert the blobs from ofxOpenCv to glm vectors or use ofPoint everywhere. I feel like both of them are ugly.

A better solution would be to remove ofPoint from ofxOpenCv (especially since it's deprecated). I'm willing to do this if you agree with it, but i'm afraid it's going to break backwards compatibility.

A compromise would be to have something like getGLMBlob or something which returns the blobs as glm vectors.

arturoc:
Ideally we would convert all addons to use glm instead of ofVec*/Point
at some point but i'm not sure doing it now is the best idea since we
want to do a ralease soon. perhaps let's just keep this solution by now
and then port it to use glm::vec3 afterwards

There's plans in any case to integrate ofxCv into the official addons
and deprecate ofxOpenCv too so leaving it as it is might not be so
problematic

PR #5743, opened 2017-09-07
Replace “wget” with “curl -L --retry 20 -O --progress”.

### LINKED ISSUE #5371: download_libs.sh on OSX requires wget - replace with curl
wget doesn't ship with OSX this needs to be installed via brew or macport

bakercp:
Fixes #5371 

Replace `wget $URL` with `curl -L --retry 20 -O --progress $URL`

`-L` follows redirects (the default for `wget`).
`--retry 20` retries failed connections 20 times (the default for `wget`).
`-O` saves the target file with the name on the server (the default for `wget`).
`--progress` Uses a `wget`-like progress bar rather than default curl "meters".

### DIFF
--- a/scripts/ci/addons/install.cmd
+++ b/scripts/ci/addons/install.cmd
@@ -9,7 +9,7 @@ if "%BUILDER%"=="VS" (
     set PG_OF_PATH=%APPVEYOR_BUILD_FOLDER%\..\openFrameworks
     %MSYS2_PATH%\usr\bin\bash -lc "scripts/vs/download_libs.sh --silent"
     echo "Downloading projectGenerator from ci server"
-    %MSYS2_PATH%\usr\bin\bash -lc "wget http://ci.openframeworks.cc/projectGenerator/projectGenerator-vs.zip -nv"
+    %MSYS2_PATH%\usr\bin\bash -lc "curl -L --retry 20 -O --progress http://ci.openframeworks.cc/projectGenerator/projectGenerator-vs.zip -nv"
     %MSYS2_PATH%\usr\bin\bash -lc "unzip -qq projectGenerator-vs.zip"
     rm projectGenerator-vs.zip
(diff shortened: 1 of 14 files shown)


arturoc:
curl, doesn't come installed by default in at least ubuntu and msys2 or other windows bash terminals so this is fixing one platform but breaking at least 2 or 3. perhaps we can detect which one is present by checking the return code of `which wget` and `which curl` and use one or the other?

danoli3:
Yeah definitely use which and choose, curl being preferable. I'll fix this up for you @bakercp and PR into your branch

danoli3:
Looks like ```command -v``` is the safest bet. 
https://stackoverflow.com/a/677212 

Which has issues on different platforms. I'll set it up with this and a if statement.

danoli3:
Pull request sent @bakercp  https://github.com/bakercp/openFrameworks/pull/26

bakercp:
I just merged @danoli3's stuff.  Thanks sir!

danoli3:
I've been reviewing the developer download process for the libs, I'll make a new PR for adding some commands similar to PS1 clickable that Windows has, except for OSX related lib downloading using command.
Kinda annoying to have to shell and ./download_libs.sh when for simplicity could just be a click on the command script, and all the command script does is call the ./downloader.sh script.

arturoc:
some tests are now failing aparently cause curl is to verbose. can you perhaps run it in silent mode or default to wget if it's available

danoli3:
Can do silent by default on curl, will fix that.

Isn’t it we want curl priority and fallback on wget.

arturoc:
i replaced curl with wget a while ago cause it was more robust so i'd prefer to leave wget as default if it's present

danoli3:
Yeah okay, I'll switch that around.

danoli3:
I'll start this one again with head

bakercp:
@arturoc So is this a no?

ofTheo:
I have run into some issues using curl instead of wget.
I tried doing some of the above, but just ended up installing wget on osx. 

 I think it might be best to close this.

PR #5745, opened 2017-09-13
removes some obvious warnings and errors.

schmidh:
Just getting rid of some warnings and errors under gcc 6. Actually only errors are the includes in CreEPS.cpp. Gcc does not like these includes in namespaces. Everything else is pretty obvious. Mostly changing from "int" to "unsigned int".

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
@@ -253,7 +253,7 @@ void ofxAssimpModelLoader::loadGLResources(){
             
             ofxAssimpTexture assimpTexture;
             bool bTextureAlreadyExists = false;
-            for(int j=0; j<textures.size(); j++) {
+            for(unsigned int j=0; j<textures.size(); j++) {
                 assimpTexture = textures[j];
                 if(assimpTexture.getTexturePath() == realPath) {
                     bTextureAlreadyExists = true;
@@ -513,7 +513,7 @@ unsigned int ofxAssimpModelLoader::getAnimationCount(){
     return animations.size();
 }
 
-ofxAssimpAnimation & ofxAssimpModelLoader::getAnimation(int animationIndex) {
+ofxAssimpAnimation & ofxAssimpModelLoader::getAnimation(unsigned int animationIndex) {
     animationIndex = ofClamp(animationIndex, 0, animations.size()-1);
     return animations[animationIndex];
 }
@@ -585,7 +585,7 @@ void ofxAssimpModelLoader::setTime(float time) {
 }
 
 // DEPRECATED.
-float ofxAssimpModelLoader::getDuration(int animationIndex) {
+float ofxAssimpModelLoader::getDuration(unsigned int animationIndex) {
     if(!hasAnimations()) {
         return 0;
     }
@@ -603,7 +603,7 @@ unsigned int ofxAssimpModelLoader::getMeshCount() {
     return modelMeshes.size();
 }
 
-ofxAssimpMeshHelper & ofxAssimpModelLoader::getMeshHelper(int meshIndex) {
+ofxAssimpMeshHelper & ofxAssimpModelLoader::getMeshHelper(unsigned int meshIndex) {
     meshIndex = ofClamp(meshIndex, 0, modelMeshes.size()-1);
     return modelMeshes[meshIndex];
 }
@@ -830,9 +830,9 @@ ofMesh ofxAssimpModelLoader::getMesh(string name){
 }
 
 //-------------------------------------------
-ofMesh ofxAssimpModelLoader::getMesh(int num){
+ofMesh ofxAssimpModelLoader::getMesh(unsigned int num){
 	ofMesh ofm;
-	if((int)scene->mNumMeshes<=num){
+	if(scene->mNumMeshes<=num){
 		ofLogError("ofxAssimpModelLoader") << "getMesh(): mesh id " << num
 		<< " out of range for total num meshes: " << scene->mNumMeshes;
 		return ofm;
@@ -865,8 +865,8 @@ ofMesh ofxAssimpModelLoader::getCurrentAnimatedMesh(string name){
 }
 
 //-------------------------------------------
-ofMesh ofxAssimpModelLoader::getCurrentAnimatedMesh(int num){
-	if((int)modelMeshes.size()<=num){
+ofMesh ofxAssimpModelLoader::getCurrentAnimatedMesh(unsigned int num){
+	if(modelMeshes.size()<=num){
 		ofLogError("ofxAssimpModelLoader") << "getCurrentAnimatedMesh(): mesh id: " << num
 			<< "out of range for total num meshes: " << scene->mNumMeshes;
 		return ofMesh();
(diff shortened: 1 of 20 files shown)


addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
```diff
@@ -253,7 +253,7 @@ void ofxAssimpModelLoader::loadGLResources(){
             
             ofxAssimpTexture assimpTexture;
             bool bTextureAlreadyExists = false;
-            for(int j=0; j<textures.size(); j++) {
+            for(unsigned int j=0; j<textures.size(); j++) {
```
tgfrerer: Using `size_t` consistently for all comparison against `.size()`  would be nice... I know on most systems `unsigned int` is probably the same type as `decltype(textures.size())`, but it might not always be.

addons/ofxOsc/libs/oscpack/src/ip/posix/UdpSocket.cpp
```diff
@@ -502,7 +502,7 @@ class SocketReceiveMultiplexer::Implementation{
                 if( FD_ISSET( breakPipe_[0], &tempfds ) ){
                     // clear pending data from the asynchronous break pipe
                     char c;
-                    read( breakPipe_[0], &c, 1 );
+                    if(read( breakPipe_[0], &c, 1 )) {};
```
tgfrerer: Uh, this looks a bit odd ... `read()` should be able to stand alone, no? Does gcc print a warning for this line?

addons/ofxOsc/libs/oscpack/src/ip/posix/UdpSocket.cpp
```diff
@@ -557,7 +557,7 @@ class SocketReceiveMultiplexer::Implementation{
 		break_ = true;
 
 		// Send a termination message to the asynchronous break pipe, so select() will return
-		write( breakPipe_[1], "!", 1 );
+		if(write( breakPipe_[1], "!", 1 )) {};
```
tgfrerer: Uh, this looks a bit odd ... `write()` should be able to stand alone, no? Does gcc print a warning for this line?

examples/input_output/systemSpeakExample/src/ofApp.cpp
```diff
@@ -31,17 +31,17 @@ void ofApp::threadedFunction() {
 
 		#ifdef TARGET_OSX
 			string cmd = "say -v "+voice+" "+words[step]+" ";   // create the command
-			system(cmd.c_str());
+			if(system(cmd.c_str())) {};
```
tgfrerer: Uh, this looks a bit odd ... should be able to stand alone, no? Does gcc print a warning for this line?

examples/input_output/systemSpeakExample/src/ofApp.cpp
```diff
@@ -31,17 +31,17 @@ void ofApp::threadedFunction() {
...
 		#endif
 		#ifdef TARGET_WIN32
 			string cmd = "data\\SayStatic.exe "+words[step];   // create the command
 			cout << cmd << endl;
-			system(cmd.c_str());
+			if(system(cmd.c_str())) {};
```
tgfrerer: Uh, this looks a bit odd ... should be able to stand alone, no? Does gcc print a warning for this line?

examples/input_output/systemSpeakExample/src/ofApp.cpp
```diff
@@ -31,17 +31,17 @@ void ofApp::threadedFunction() {
...
 		#endif
 		#ifdef TARGET_LINUX
 			string cmd = "echo "+words[step]+"|espeak";   // create the command
 			cout << cmd << endl;
-			system(cmd.c_str());
+			if(system(cmd.c_str())) {};
```
tgfrerer: Uh, this looks a bit odd ... should be able to stand alone, no? Does gcc print a warning for this line?

addons/ofxOsc/libs/oscpack/src/ip/posix/UdpSocket.cpp
schmidh: Yes, unused return value.

addons/ofxOsc/libs/oscpack/src/ip/posix/UdpSocket.cpp
schmidh: Yes, unused return value.

examples/input_output/systemSpeakExample/src/ofApp.cpp
schmidh: Yes, unused return value.

examples/input_output/systemSpeakExample/src/ofApp.cpp
schmidh: Yes, unused return value.

examples/input_output/systemSpeakExample/src/ofApp.cpp
schmidh: Yes, unused return value.

addons/ofxOsc/libs/oscpack/src/ip/posix/UdpSocket.cpp
schmidh: Maybe a macro would be better.

addons/ofxOsc/libs/oscpack/src/ip/posix/UdpSocket.cpp
tgfrerer: I see ... since this code is imported straight from [oscpack](https://github.com/mkalten/oscpack/blob/master/ip/posix/UdpSocket.cpp#L537) i would leave it as-is here, and send a pull request to oscpack. Otherwise, if oscpack ever gets updated, and we re-import the library, any changes might get lost.

schmidh: Good point. Cool you know the codebase so well.

I will send a PR to oscpack.

examples/input_output/systemSpeakExample/src/ofApp.cpp
tgfrerer: For code which part of examples, it could make sense to add something to use the return value. I'm just not sure an if-statement is the clearest way of expressing the intent. 

How about: 

`auto status = system(cmd.c_str());` ?

schmidh: Will lead to another gcc warning. Unused variable.

schmidh: What about **std::ignore = system(cmd.c_str());** ?

Looks idiomatic although it isn't really.

schmidh: **std::ignore** is C++11. Probably not what we want. I am not sure.

Note: (void) fn(...), static_cast, and so on do not work. I even used a template, but might be overkill:

```
template <typename T>
void UnusedReturnValue(T&& /*No name*/) { /*Empty*/ }
```
Then:
```
UnusedReturnValue(system(cmd.c_str()));
```

I am afraid we have to live with that warning in gcc6. Gcc6 should just accept **(void) system(...)** that would be easiest and also upward compatible.

tgfrerer: Hmm... probably best to just live with the "unused return value" warning - it's one of the most benign warnings anyway... 

Putting the "unused return values" aside  - if the `unsigned` were changed to `size_t`, this PR would be a good candidate to merge =)

schmidh:
Closing this PR while opening a new one.

PR #5747, opened 2017-09-16
changes types to please gcc6.

schmidh:
Mostly type changes to make gcc6 a happy camper. CreEPS.cpp had includes in namespaces which gcc6 does not like any more. For system commands in systemSpeakExample I casted the result to (void) which is idiomatic c/c++ to subpress unused return value warnings. But gcc6 ignores it! At least one can see that everything is meant to be right when you compile the code. There seems to be a discussion in the gcc community how to handle this warning. Here is the official documentation https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html saying that it should work. And here is an old ticket which has been updated lately: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425

### DIFF
--- a/addons/ofxNetwork/src/ofxTCPClient.cpp
+++ b/addons/ofxNetwork/src/ofxTCPClient.cpp
@@ -298,7 +298,7 @@ int ofxTCPClient::receiveRawMsg(char * receiveBuffer, int numBytes){
 	int posDelimiter = findDelimiter(tmpBuffReceive.getData(),tmpBuffReceive.size(),messageDelimiter);
 	if(posDelimiter>0){
 		memcpy(receiveBuffer,tmpBuffReceive.getData(),posDelimiter);
-		if(tmpBuffReceive.size() > posDelimiter + (int)messageDelimiter.size()){
+		if(tmpBuffReceive.size() > (unsigned int) posDelimiter + messageDelimiter.size()){
 			memcpy(tmpBuff,tmpBuffReceive.getData()+posDelimiter+messageDelimiter.size(),tmpBuffReceive.size()-(posDelimiter+messageDelimiter.size()));
 			tmpBuffReceive.set(tmpBuff,tmpBuffReceive.size()-(posDelimiter+messageDelimiter.size()));
 		}else{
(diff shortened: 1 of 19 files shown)


schmidh:
Actually, I think return values of *getNumInputChannels* and *getNumFrames* should be changed from size_t to unsigned int. Please let me know.

addons/ofxNetwork/src/ofxTCPClient.cpp
```diff
@@ -298,7 +298,7 @@ int ofxTCPClient::receiveRawMsg(char * receiveBuffer, int numBytes){
 	int posDelimiter = findDelimiter(tmpBuffReceive.getData(),tmpBuffReceive.size(),messageDelimiter);
 	if(posDelimiter>0){
 		memcpy(receiveBuffer,tmpBuffReceive.getData(),posDelimiter);
-		if(tmpBuffReceive.size() > posDelimiter + (int)messageDelimiter.size()){
+		if(tmpBuffReceive.size() > (unsigned int) posDelimiter + messageDelimiter.size()){
```
bakercp: size_t here?

bakercp:
Besides my small comment, this looks good to me.  Casting the return value to `void` feels odd, but it isn't "wrong".

addons/ofxNetwork/src/ofxTCPClient.cpp
schmidh: You mean the *posDelimiter*? Looks strange to change a delimiter position to size_t.

schmidh:
Casting a return value to void is idiomatic C/C++. Please see the link to the documentation.

See also e.g. https://stackoverflow.com/a/689716/365425:
> ... covers the motivation for this, to explicitly show other "developers" that you know this function returns but you're explicitly ignoring it.

arturoc:
@schmidh even if it's common practice it might look strange to beginners who are the main target for examples so perhaps actually using the return value as the warning suggests is a better solution?

something like:

```
if(system(...) != 0){
    ofLogError() << "error while running ...";
}
```

we also have ofSystem which runs the command and returns the output instead so we could just use that

schmidh:
Oh yes, I like that a lot!

schmidh:
Using `ofSystem(cmd.c_str())` now.

arturoc:
thanks

Issue #5751, opened 2017-09-17
ofCamera bug introduced with switch to glm


roymacdonald:
I found out a bug in ofCamera's lens offset. When ofMatrix4x4 was being used its `translate(...)` method applied a post mult over the matrix, while glm::translate(glm::mat4, glm::vec3) works as premult.
The fix is quite simple.
It's here https://github.com/roymacdonald/openFrameworks/tree/fix-ofCamera_lensOffset_glm_math
the specific line that changed is
https://github.com/roymacdonald/openFrameworks/blob/fix-ofCamera_lensOffset_glm_math/libs/openFrameworks/3d/ofCamera.cpp#L159

Should I PR?
@arturoc @bakercp @elliotwoods 

Cheers

bakercp:
Yes, PR 👍

PR #5752, opened 2017-09-18
FIX: ofCamera. Fixed wrong maths done with lens offset. 

roymacdonald:
This bug was introduced in the GLM implementation.
Issue: https://github.com/openframeworks/openFrameworks/issues/5751


arturoc:
thanks!

PR #5754, opened 2017-09-19
adds missing OpenGL math library (libglm-dev).

schmidh:
Just for Ubuntu.


arturoc:
we are including glm by now since it's still in 0.x versions so it has very frequent breaking changes and it would be difficult to maintain OF working for more than one version. once glm arrives at 1.0 and the major distributions use that we'll include the library from the system instead of packing it with the ones OF includes but by now is just safer to include it ourselves

Issue #5731, opened 2017-08-25
reduce compile times by using more specific header includes


tgfrerer:
Lately, a full (from clean) compile of openFrameworks seemed to take 3min30s on a reasonably fast 12 core machine. Also, linker and code generation times seem quite high, and it appears to take about 10 sec from changing a single line of code (inserting a line break) in ofApp.cpp to be able to see a simple app run. 

Wouldn't it be great to have a really fast change-compile-run cycle?

I did some measurements using VS2017, and set the following diagnostic compiler switch: 

* `/Bt+` (added to openFrameworksLib property page under "Command Line" -> "Additional Options")

This enables compile time measurement for individual compilation units (i.e. `.cpp` files). On average, a single file takes about 30s to compile, some files, like `ofAppRunner.cpp`, take about 48sec. 

I believe this is partly due to the way we're currently including headers:

`ofConstants.h` is included nearly everywhere:
![ofconstants](https://user-images.githubusercontent.com/423509/29707726-32eb9a4c-897e-11e7-8dee-31f79aeae308.png)

And `ofConstants.h` currently does more than its name suggests it should do: apart from setting constants, it includes glm, opengl, and tesselator. And since this header is included nearly everywhere - these headers in turn get included nearly everywhere. 

There's a switch in VS which allows you to see which headers get included in compilation: it's under 
"C/C++" -> "Advanced" -> "Show Includes" (set to Yes)

Hit compile and my bet is you'll be amazed at what you see =)

---

I found the following immediate steps could reduce comile times:

1) remove "wholesale" include of "glm/ext.hpp" (this is something glm itself recommends) from `ofConstants.h` and replace it with:

```
#define GLM_META_PROG_HELPERS
#define GLM_FORCE_SWIZZLE
#define GLM_FORCE_SIZE_FUNC
#include "glm/glm.hpp"
#include "glm/gtc/type_ptr.hpp"
#include "glm/gtc/matrix_transform.hpp"
#include "glm/gtc/quaternion.hpp"
#include "glm/gtx/quaternion.hpp"
#include "glm/gtx/matrix_transform_2d.hpp"
#include "glm/gtx/vector_angle.hpp"
```

This means only the glm headers we actually use are included. This saves about 7sec per file on average.

2) remove `boost::filesystem` and replace it with `std::experimental::filesystem` on systems/compilers which support this (saves 5 sec per file on avg)

[...]

PR #5756, opened 2017-09-20
change color consts to constexpr so they are calculated at compile time 

### LINKED ISSUE #5750: initialization state of OF in ctors/dtors:
i am using of 0.9.8 on win10. let's say that i want to share a sqlite database between apps. 
as i want to share paths to the DB in constructors and destructors, i realized that several OF functions are not yet or no more initialized in ctors or dtors:

- the ofLog() functionality
- the ofFilePath:: functionality (boost paths are crashing)
- ofToDataPath()

i am talking about something like:
```
	someClassCtor()
	{
		string sPath = ofFilePath::getAbsolutePath("data/../../../../../apps/logger-sqlite/", false);
		sPath = ofFilePath::getCurrentExeDir() + "data/../../../../../apps/logger-sqlite/";
                spath = ofToDataPath(../../../../../apps/logger-sqlite/");
		sPath = ofFilePath::removeTrailingSlash(sPath) + "/";
		return sPath;
	}
```
above functions may not be valid in ctors and dtors. if ofDataPath() has an issue it creates another issue calling the unitialized ofLog().

i am repeatedly running into this issue and have not figured out a workaround that works on all platforms ( i use win and linux/rpi). it is especially an issue when i work with threads and dtors. and i wish to keep the relative path system of OF, so i can share the apps between platforms and OF versions.

any hints? thx!

arturoc:
fixes #5750 at the risk of making #5731 even worse


johanjohan:
thx arturo

jvcleave:
could these be moved to a core ofxAddon?

arturoc:
@jvcleave do you mean the constants? i think those should go in the core since they are pretty useful.

In any case the only possible solution i see without breaking user code, is turning the consts into constexpr won't work cause using a consexpr member of the own class type doesn't work in any compiler except gcc.

The original problem was caused cause of using variables declared directly in the file instead of a function or class which is known to be really problematic (see: (de)initialization fiasco) so i don't think is worth to try to solve it by breaking code and would instead recommend to declare variables inside a class or function or somehow have lazy initialization for statics through functions that only initializes those values the first time they are sued.

jvcleave:
@arturoc I find them useful as well but they have been problematic for compile times (particularly in the early RPi days)

Was just a thought - they are pretty easy to #ifdef out

arturoc:
are you sure that the problem is those constants and not ofColor itself being templated?

jvcleave:
I would have to double check but I remember just commenting out the color names to solve it

Issue #5749, opened 2017-09-17
ofOpenALSoundPlayer::setPositionMS(int ms) set wrong position when streaming audio


hiroMTB:
probably this is the problem.

```
ofOpenALSoundPlayer::setPositionMS(int ms){

        ...

	if(streamf){
		sf_seek(streamf,float(ms)/1000.f*samplerate*channels,SEEK_SET);
		stream_samples_read = 0;    <- should change position after seek
	}else{

        ...

```

hiroMTB:
I noticed we should pass frame number without multiple of channel numbers.
see this link.
http://www.mega-nerd.com/libsndfile/api.html#seek

So correct code will be something like,

...

	if(streamf){
		stream_samples_read = sf_seek(streamf,float(ms)/1000.f*samplerate,SEEK_SET);
		stream_samples_read *= channels;
	}else{

        ...

arturoc:
can you send a pull request?

hiroMTB:
done.
sorry PR is the one to my own fork. ;-)

PR #5760, opened 2017-09-27
Pass correct frame number to sf_seek

hiroMTB:
Fixes ofOpenALSoundPlayer::setPositionMS(int ms) set wrong position
when streaming audio #5749


hiroMTB:
This fix effects only follwing condition

On Linux (RPi as well)
Your music related project which use ofSoundPlayer::setPosition() or setPositionMS()
Non mp3 file playback (wav, aiff, etc using libsndfile internally)
Streaming

arturoc:
thanks!

Issue #5761, opened 2017-10-01
ofTrueTypeFont not loading default system fonts


roymacdonald:
ofTrueTypeFont will not load system fonts, either by explicitly passing their name or by using `OF_TTF_SANS`, `OF_TTF_SERIF` or `OF_TTF_MONO`.

The fix is quite simple. 
I'll PR asap.

(The guideline says that I have to make an issue before PRing. In cases like this, is it really needed to have an issue? Isn't enough just with the PR?)

arturoc:
yes a PR is enough, an issue is only needed if you plan to send a new feature or a fix that is not obvious and might introduce some side effects

roymacdonald:
great! 
gracias.

PR #5762, opened 2017-10-01
Fix: ofTrueTypeFont. Was not loading system fonts.

roymacdonald:
ofTrueTypeFont will not load system fonts, either by explicitly passing their name or by using OF_TTF_SANS, OF_TTF_SERIF or OF_TTF_MONO

issue https://github.com/openframeworks/openFrameworks/issues/5761


arturoc:
isn't it enough to just not do that copy?

roymacdonald:
> isn't it enough to just not do that copy?

no, because when you use OF_TTF_... the local var `fontname` is modified, and the passed var `_fontname` is const

roymacdonald:
well actually the copy if for when you pass the name of the font instead of OF_TTF_SERIF, ...

roymacdonald:
I've added another commit, where the `fontname`copy is only done when no file was found in the data folder. Adding more stuff just to avoid this copy seemed to me a bit to much as it adds several more lines of code.

arturoc:
ok, thanks!

PR #5770, opened 2017-10-06
fix ofCamera::drawFrustum

tgfrerer:
Maintenance PR for feature #5712, where cameras may draw their frustum into the scene for debugging.

* `drawFrustum` would foolishly assume ofRectMode to be set to its default value, `OF_RECTMODE_CENTER` when drawing the frustum caps.  This might lead to frustum caps being drawn incorrectly when, somewhere before, rectmode had been set to something other than `OF_RECTMODE_CORNER`.  This is fixed now - style is pushed before drawing the frustum, set correctly for drawing the frustum, and restored to user specified values after drawing the frustum, using
ofPush/PopStyle.

+ also adds parameter to optionally explicitly define a viewport from which to calculate the frustum to be drawn - this brings the method more in line with other camera-specific methods, which allow optional
specification of a viewport.

By default, the current viewport state is used, as before.

### DIFF
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -253,7 +253,7 @@ void ofCamera::setRenderer(shared_ptr<ofBaseRenderer> _renderer){
 	renderer = _renderer;
 }
 
-void ofCamera::drawFrustum() {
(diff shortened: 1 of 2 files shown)


arturoc:
thanks!

PR #5771, opened 2017-10-08
Added OSX Commands for OSX related downloading

danoli3:
- (similar to PS1 for Windows (clickable))
- Added tvOS build seperate to iOS
- Removed download tvOS in iOS download_libs script

### DIFF
--- a/scripts/dev/download_libs.command
+++ b/scripts/dev/download_libs.command
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+here="`dirname \"$0\"`"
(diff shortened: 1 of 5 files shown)


arturoc:
this makes the tvos tests fail, not sure if we need 2 separate scripts for this no? the official download for ios also includes tvos so it seems logical that downloading the libraries also work like that. this will also break the package creation as it is

i'm not opposed to having it separated but it seems like it will introduce a lot of issues for such a small change

PR #5772, opened 2017-10-08
OF Root Linux Resolver Update

danoli3:
Script Resolver Revision based on #5691

Includes cherry pick of @baydam original solution

Fixes #5691 / updates it.


arturoc:
thanks!

PR #5779, opened 2017-10-11
updating openCV android example

danzeeeman:
this pairs with the updates to apothecary to build openCV with android-r15
https://github.com/openframeworks/apothecary/tree/android-r15


danzeeeman:
I've updated a bunch of stuff @arturoc I think this is ready to merge for the most part

arturoc:
I've tried to make the ci tests pass again on this once we have the opencv libraries in place and it seems like it's not working.

I've then tried to merge current master into it just in case but still doesn't work complaining that android-24 can't be found. So i've updated the travis config to install android-24 instead of 19 but now when compiling it complains that it can find android-19!

I guess there's a project referencing an sdk and a different one referencing another or something like that.

Also there's some things that don't belong in this PR like changes to the osx project file or some commits to subrepos like the project generator and apothecary.

HalfdanJ:
Can we close this one over #5883 ?

PR #5780, opened 2017-10-12
Set mouseEvent.x and y values based on absolute limits of the axis

kr15h:
The Adafruit 5" 800x480 Backpack did not give reasonable touch values and I discovered that in ofAppEGLWindow the way how relative and absolute input positions are handled is the same. The absolute values depend on the device min and max absolute x and absolute y axis settings. This fix stores the min and max values of the x and y axes, as well as sets the mouse event x and y values according to those.


arturoc:
thanks!

PR #5782, opened 2017-10-17
ofAVFoundationGrabber not unlocking image buffer

### LINKED ISSUE #5769: videoGrabberExample -> error -> Finalizing CVPixelBuffer 0x6180001217c0 while lock count is 1.
I am on macOS 10.12.6 with OF 0.10.0.
When running the videoGrabberExample example at 1920x1080 cam resolution I get the following print out in the console. 
The same Logitech c920 at 320x240 does not give this message. I get the same result on a Logitech Brio and other usb cameras.

tried it with of_v20170714_osx_nightly.zip
```
Finalizing CVPixelBuffer 0x600000121720 while lock count is 1.
Finalizing CVPixelBuffer 0x600000121400 while lock count is 1.
Finalizing CVPixelBuffer 0x608000120dc0 while lock count is 1.
Finalizing CVPixelBuffer 0x608000120b40 while lock count is 1.
Finalizing CVPixelBuffer 0x610000121360 while lock count is 1.
Finalizing CVPixelBuffer 0x6180001217c0 while lock count is 1.
Finalizing CVPixelBuffer 0x600000121720 while lock count is 1.
Finalizing CVPixelBuffer 0x618000120780 while lock count is 1.
Finalizing CVPixelBuffer 0x6180001206e0 while lock count is 1.
Finalizing CVPixelBuffer 0x6080001205a0 while lock count is 1.
Finalizing CVPixelBuffer 0x6180001214a0 while lock count is 1.
```

and many more

JordanParsons:
This should fix #5769. It seems to be an error that is only reporting on newer (10.12.6) osx versions.

After some digging, I found that _ofAVFoundationGrabber.mm_ has a call to `CVPixelBufferLockBaseAddress(imageBuffer,0); ` without a matching unlock call. By placing a call to `CVPixelBufferUnlockBaseAddress(imageBuffer, kCVPixelBufferLock_ReadOnly);` on line 323 of _ofAVFoundationGrabber.mm_  I have suppressed that same warning. I'm not sure if that is actually fixing the issue, or if it's just suppressing the warning.

I know this is a minor change, but the level of logging this error produces is very frustrating if you are streaming video like in videoGrabberExample.

Passes local builds of 64 and 32bit versions with examples.

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationGrabber.mm
+++ b/libs/openFrameworks/video/ofAVFoundationGrabber.mm
@@ -317,7 +317,9 @@ didOutputSampleBuffer:(CMSampleBufferRef)sampleBuffer
 					
 					}
 				}
-				
+			
+			// Unlock the image buffer
+			CVPixelBufferUnlockBaseAddress(imageBuffer, kCVPixelBufferLock_ReadOnly);	
 				
 			}
 		}


arturoc:
thanks!

PR #5789, opened 2017-10-26
check pkg-config of crosscompile target for glfw3 and rtaudio

krisrok:
add checks if this is a crosscompile build.
without it the pkg-config of the host system is queried leading to wrong assumptions for crosscompilation.


arturoc:
thanks!

PR #5800, opened 2017-11-08
Fix missing std:: namespace in ofxOpenCv.

bakercp:
This addon was not compiling in certain circumstances w/ master and other addons.  Adding std:: fixes this.


bakercp:
Any objections to this?

arturoc:
thanks!

PR #5801, opened 2017-11-08
update: make `OF_DEPRECATED_MSG` allow complex template type.

2bbb:
now, we can't use like below:

```
OF_DEPRECATED_MSG("deprecated", template <typename type> typename std::enable_if<std::is_same<type, int>::value, type>::type f(type));
```

this PR makes it possible.


arturoc:
thanks!

PR #5806, opened 2017-11-11
Enable EL6 builds

smokhov:
I have a need of ofx to be buildable on EL6 environment (primarily CentOS 6 and Scientific Linux 6).
We used to use Fedora deps for the most part, but that changed in the summer to use `dnf`, and
other things, which are not supported on EL6. So I've adapted `fedora` to `el6` dependencies
installation for the initial build. Please consider for inclusion.


arturoc:
thanks!

PR #5807, opened 2017-11-11
use CURLcode, get error description from curl

i-n-g-o:
test with e.g.:
```
ofHttpResponse resp = ofLoadURL("http://foo.bar");
ofLog() << "code: " << resp.status;
ofLog() << "error: " << resp.error;
```


arturoc:
thanks!

PR #5811, opened 2017-11-20
add release number to pkg names, so correct version will be picked up

edovino:
... as per the instructions on https://rpmfusion.org/Configuration


arturoc:
thanks!

PR #5813, opened 2017-11-26
Qbs fixes qtcreator 441 osx

i-n-g-o:
i could not test the changes on other systems.
think the changes deal with newer version of qbs, so i guess it is fine on linux/win as well
but would be good if this would be checked on linux/win before merging...


arturoc:
thanks!

PR #5814, opened 2017-11-27
Added api parameter to ofSoundStream::getMatchingDevices()

bslote:
Currently `ofSoundStream::getMatchingDevices()` calls `ofSoundStream::getDeviceList()` with no arguments, resulting in `ofSoundStream::getDeviceList()` defaulting to `ofSoundDevice::Api::DEFAULT` which means searches are restricted to ASIO devices on Windows.

This PR simply adds an optional `api` param to  `getMatchingDevices()` that gets passed along to `getDeviceList()` so that the search works for all APIs.


arturoc:
thanks!

PR #5815, opened 2017-11-30
ofSoundBuffer: fixed wrong interpolation algorithm

roymacdonald:
The interpolation algorithm was only taking the first channel as it was not using the correct index


arturoc:
thanks!

PR #5821, opened 2017-12-15
Refactor optimize includes

arturoc:
- forward declare as much as possible
- include external libraries in .h only when we really want to expose them to users
- reduce size ofConstants
- subdivide ofBaseTypes per module

reduces compilation times to around 2/3 of current master in most platforms and solves issues with autocomplete being really slow in some ides


bakercp:
This made a big difference.  The biggest change that I'm dealing w/ now is addons that used `ofBaseTypes.h`.  Probably good to make a note of that in the changelog before we forget.

arturoc:
by big difference, you mean it fixed what it's supposed to or that it broke a lot of code? :)

i imagined ofBaseTypes could be problematic, i was thinking on adding ofBaseTypes.h back which would include the new files but wouldn't be included from ofMain.h do you think that would be useful?

bakercp:
It made a big difference in speed! ☺️

I think your solution of adding ofBaseTypes back in would be good as long as it doesn't hurt the compilation speed.

Issue #901, opened 2012-02-08
easier filled ofPolyline, or easy ofPolyline->ofPath conversion


kylemcdonald:
to draw a filled ofPolyline right now, you need to use beginshape/endshape. but i think a way of adding ofPolylines to an ofPath would be really good -- even a conversion from ofPolyline to ofPath might make sense.

hamoid:
With ofxClipper calculating the intersection between two polylines there's no simple way to draw the results as a filled shape. As a quick solution I added this method to ofPath:

```cpp
void ofPath::set(std::vector<ofPolyline> & lines) {
    commands.clear();
    polylines = lines;
    mode = POLYLINES;
    cachedTessellation.clear();
    flagShapeChanged();
}
```
I use it like this:
```cpp
    std::vector<ofPolyline> intersection = clipper.getClipped(ClipperLib::ClipType::ctIntersection);
    ofPath result;
    result.set(intersection);
    result.setFilled(true);
    result.setFillColor(ofColor::darkGray);
```
I'm not sure I understand @arturoc above when mentioning 1 polyline only, but if accepting a vector is too much it would be nice to be able to add multiple polylines to the ofPath like this:
```cpp
    std::vector<ofPolyline> intersection = clipper.getClipped(ClipperLib::ClipType::ctIntersection);
    ofPath result;
    for(auto & line : intersection) {
      result.addPolyline(line);
    }
```

PR #5822, opened 2017-12-16
new constructor for ofPath

brightredchilli:
Hi openFrameworks, this is my first pr! 💃  

This should address #901

I tested that any mutation to the polyline passed into the ofPath would not change the source polyline.

That is:

```cpp
    ofPolyline aLine;
    aLine.addVertex(0, 0);
    aLine.addVertex(100, 100);
    ofPath aPath(aLine);

    aPath.lineTo(120, 120); // does not affect aLine
    aLine.addVertex(30, 400); // does not affect aPath
```

Would love advice, testing tips and pointers!


brightredchilli:
I just realized that I didn't implement a constructor that takes vector <ofPolyline> - i can do that if that's heading down the right path.

libs/openFrameworks/graphics/ofPath.h
```diff
@@ -39,6 +39,8 @@ class ofPath{
 	/// \brief Create a new ofPath instance.
 	ofPath();
 
+	ofPath(ofPolyline &polyline);
```
bakercp: I think `const ofPolyline& polyline` would make more sense because we don't want to imply that the `ofPath` might change the polyline passed by reference.

bakercp: Additionally, please add documentation of the new constructor a e.g.

```c++
/// \brief Construct a new ofPath from the given ofPolyline.
///
/// A single polyline will be constructed from the given polyline and the ofPath mode will be set to `POLYLINES`.
///
/// \param polyline The polyline to that defines the new ofPath.

Comment (brightredchilli) [reply libs/openFrameworks/graphics/ofPath.h]: will do!

**Comment by arturoc** [comment]
Hey thanks for your contribution. The constructor shouldn't change the default mode to polylines and instead add the polyline vertex by vertex using lineTo. perhaps as an optimization there could be another constructor that took a polyline and the mode as parameters and just added the polyline if the mode is polylines.

also it would be useful to have an addPolyline method in case people don't won't to create the path but just add a polyline to an existing path

**Comment by brightredchilli** [comment]
I like the addPolyline idea! addPolyline internally either pushes to polylines or does the lineTo work depending on mode. For now I decided it would be best not to add the mode param, considering the use case in the thread(and mine as well). 

```cpp
// user didn't need to learn the difference between COMMANDS and POLYLINES, and still don't need to 
ofPath path(polyline);
path.scale();
```
Still happy to do it, but just wanted to let you know where it's coming from.

brightredchilli:
@arturoc this should be ready for re-review.

ofZach:
I find myself going back and forth between polyline and path alot (for tessellation, etc) so I'm supportive of this PR.

libs/openFrameworks/graphics/ofPath.cpp
```diff
@@ -59,6 +59,22 @@ ofPath::ofPath(){
...
 
+//----------------------------------------------------------
+ofPath::ofPath(const ofPolyline &polyline) : ofPath(){
+	addPolyline(polyline);
+}
+
+//----------------------------------------------------------
+void ofPath::addPolyline(const ofPolyline &polyline){
+	if (mode == COMMANDS) {
+		for (ofVec3f vertex : polyline) {
```
tgfrerer: Maybe we can use `auto vertex` instead of `ofVec3f vertex`, and later just `lineTo(vertex)`. This should make this method more independent of the vertex type.

arturoc:
yes it seems useful but haven't merged it yet cause i think there's several problems with it, i'll try to write a review when i have a moment. having a test that shows that it behaves properly would be useful too.

in any case i think we shouldn't merge it until 0.10 is released since we are just finishing with a few remaining issues with the CI tests but everything else is pretty much closed so we should only merge bugfixes until the release is out

PR #5832, opened 2018-01-02
add missing ofBaseApp.cpp file into xcode prj

armadillu:
As discussed in these comments, https://github.com/openframeworks/openFrameworks/commit/dcdc4f4a2cc8d0523998db438b2a3fa33f4fc86a#diff-3370fe6b5470b2a40b86861e869911b3L1 add ofBaseApp.cpp to the OF XcodeProject so that it can be compiled.


arturoc:
thanks!

PR #5833, opened 2018-01-06
Added make_preferred() to paths where necessary.

prisonerjohn:
Some absolute paths were returning a flipped separator on Windows at the drive level (e.g. `C:/some\\folder\\path\\`. Although this is no problem in most cases, it doesn't work with some Windows specific stuff, like setting the default path when using `ofSystemLoadDialog(...)`.

Here's a better description of a similar problem: https://stackoverflow.com/questions/27110973/boost-filesystem-canonical-path-is-not-valid-after-conversion-to-const-char


arturoc:
thanks!

PR #5838, opened 2018-01-07
add ccache to ci tests to make them faster

arturoc:


HalfdanJ:
Cool! Was doing something very similar last night in my fork. 

I can recommend adding a ccache -s somewhere after everything has run. It's a nice way to see in the log if the ccache is even hit.

HalfdanJ:
One other comment, I've noticed that caching the NDK might not be so effecient anymore. On apothecary it often times out uploading, and downloading the zipped file from the travis cache (S3) or from Google isn't that different. I'm not sure if adding ccache means that it now re-uploads the entire NDK on each run. I've disabled the caching on apothecary now, and its more reliable.

armadillu:
For reference, ive been using ccache after seeing this (https://forum.openframeworks.cc/t/speeding-up-compilation-with-ccache-xcode-tip/23492/7) for more than a year with zero drawbacks and lots of speed up. 

As for using it in CI, not sure it will help much as in my experience it only helps when recompiling (ie the second time you compile the same file). In fact, when compiling a first time, it slows you down a bit (or a lot depending on your i/o speed) as it needs to write the cache to disk. If the CI keeps the ccache files around every time its run, they sure... otherwise, if each ci run spawns a new machine, I don't see how it will help?

HalfdanJ:
Both travis and appveyor are caching the ccache files, so they are re-used from run to run.

armadillu:
hmm this is super handy then, I need to look into how you guys set this up - it would help loads with my ci setup as well... @bakercp might be interested as well?

HalfdanJ:
Travis is so handy that you can just add 
```
cache: ccache
```
to the yaml. Arturo has here been adding some flags to get it to work on some more platforms, but on apothecary this one line enabled it for osx and linux with no other changes

https://docs.travis-ci.com/user/caching/#ccache-cache

arturoc:
yes, linux (the one where it works best) is down from 25 minutes to 4. i'll add it to addons scripts as soon as i get it properly working for all platforms so it should work for all addons right away

PR #5848, opened 2018-01-15
Script Downloaders : if Wget then use else curl

danoli3:
Reset to latest from https://github.com/openframeworks/openFrameworks/pull/5743

Prioritizes wget and fallback to curl


arturoc:
happy to merge this but it's failing on a couple of platforms. i think the error in linux arm 7 is already solved in master so try pulling master into your branch. but linux 64 seems a real error.

danoli3:
Rebased lets see how it goes!

arturoc:
mmh, don't see any extra commits. btw one thing i realized some days ago is that we might need some way of telling the script to work in silent mode and to set the output file name since the syntax for curl and wget for this is slightly different so we can't just directly pass all the parameters to the binary

bakercp:
Just to underline this -- it would be _really_ nice if we can support curl as the default downloader on OSX.  I had an entire class installing a shallow clone of the master branch the other day (the stable version and nightlies were causing problems) and realized too late that they didn't have wget installed.  Anyway, I support this!

danoli3:
Maybe this is the solution. Just lets do this for OSX scripts for now. Migrate the rest later.

arturoc:
can we perhaps have a downloader script that does this? we can then call it from everywhere else instead of defining the same thing in every other script.

it can be really basic right now and we can tweak it later if we need more options

danoli3:
Okay I'll fix this up, as I just had this happen to me again testing out libs.

danoli3:
Okay @arturoc I've re-done this PR with dev/downloader.sh script!

arturoc:
thanks!, it's still making some platforms fail the tests though

danoli3:
Oh my gosh, I worked on getting this finally merged... lets see if all tests are good !

danoli3:
Okay @arturoc Good to Merge this finally... 

omfg @bakercp sorry this took so long to figure out

scripts/dev/downloader.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+downloader() { 
+    if command -v curl 2>/dev/null; 
+    then 
+    	curl -LO --retry 20 -O --progress $1 $2 $3; 
+    else 
+        wget $1 $2 $3 2> /dev/null; fi;
```
arturoc: You can pass all arguments using $@ instead of just passing the first 3.

But also i'm not sure we should pass any arguments at all like this since arguments to wget and curl are different so this won't work most of the times.

For example some of the scripts calling this use -q which for wget is the intended quiet so there's not so much output but for curl means "Disable .curlrc" which in this case is probably not a problem but in the case of other arguments might have unintended consequences.

I would just add the arguments we use by now, i think rn only -q, and parse them in the script then pass the corresponding arguments to wget or curl and fail if the arguments passed are not in the list of the ones we pass

scripts/dev/release.sh
```diff
@@ -65,7 +69,7 @@ mkdir -p /var/www/versions/v${version}
 mv *.tar.gz /var/www/versions/v${version}
 mv *.zip /var/www/versions/v${version}
 
-wget http://openframeworks.cc/release_hook.php?version=${version} -O /dev/null
+downloader http://openframeworks.cc/release_hook.php?version=${version}
```
arturoc: This will leave temporary files around that we don't want, -O (for output to a file) should be an option of the downloader script

scripts/ios/download_generator
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
arturoc: Can you add the .sh extension? we use that in every script and i think it makes it easier to distinguish that it's a bash script

scripts/ios/download_libs
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
arturoc: Again, can you add the extension? I think this one actually existed already with the .sh extension so it's probably just a matter of editing that one instead of adding a new scrtipt

scripts/osx/download_generator
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
arturoc: .sh extension

scripts/osx/download_libs
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
arturoc: .sh extension

scripts/dev/downloader.sh
danoli3: Yeah great idea... didn't know about $@. Awesome.

scripts/ios/download_generator
danoli3: Yeah this is actually intentional... On OSX you can embed scripts to be executable.. so this makes this basically a double-click and run option rather than the classic terminal ->./run_script.sh.

screen shot 2018-08-12 at 2 49 12 pm

The OS reads the first line and know how to run it. I can post-fix add ".command" if its ambiguous. Let me know what you think

arturoc: can't .sh be double clicked as well?

bakercp:
@danoli3 Are you interested in fixing this?  If not, I would be interested in adding the file extensions and moving this forward.

arturoc:
Happy to merge this once it's working but it seems everything is failing at the moment? Also would be good to merge current master into this since linux tests were disabled until now

danoli3:
Awesome on this now, lets see if I can get it working across ci

danoli3:
Oh wow Curl `progress` has changed to `progress-bar`

danoli3:
@ofTheo  Close unsure why macOS would fail tests in actions ?

ofTheo:
@danoli3 Yeah that looks unrelated. 🤔
I just re-triggered the build job - hopefully if it works this time.

ofTheo:
Looks like it just needed re-running. No errors now.

danoli3:
Doing some final tests and lets merge this

scripts/dev/downloader.sh
```diff
@@ -0,0 +1,34 @@
...
+}
+
+downloader() { 
+
+	if [ -z "$1" ]
+	then printDownloaderHelp; fi
+
+    SILENTARGS="";
+    if [ "$2" == "-s" ]; then
+    	`sh $CATALINA_HOME/bin/startup.sh`
```
arturoc, line 22: What is this doing? Won't it fail on other OSs but macos?

ofTheo: Yeah totally confused about the startup and shutdown commands. seems related to Apache Tomcat?  
@danoli3 what is the use of those calls? Thanks!

danoli3:
I legit have no idea, I never wrote that code from memory.
Let’s remove them and get this in

ofTheo (approved):
Thanks @danoli3 

This looks all good to me with that last change. 👍
I will let @arturoc do final approval though.

arturoc:
The way it's used by running the script to get the function instead of directly calling the script to download is a bit weird but going to merge it anyway. I guess we can change it later it it proves difficult to use

Issue #2812, opened 2014-02-13
ofHideCursor() not working on Arch Linux and Ubuntu 14.04


ghost:
It simply doesn't :) Basic template project with ofHideCursor() in setup() call - cursor stays visible.

PR #5860, opened 2018-01-29
Fixing hideCursor() for Fullscreen on Linux

PTS93:
This should behave well on all other platforms as well.
On Linux the cursor does not disappear in fullscreen, guessing this has to do with how GLFW_CURSOR_HIDDEN is handled internally, in window mode it work fine but in fullscreen GLFW_CURSOR_DISABLED is necessary.

This was also discussed here: https://github.com/openframeworks/openFrameworks/issues/2812


arturoc:
thanks!

Issue #5861, opened 2018-01-29
windowMode not being passed on to ofAppGLFWWindow


PTS93:
I discovered that the windowMode variable in ofAppGLFWWindow.cpp is always OF_WINDOW because of the constructor [setting it to that](https://github.com/openframeworks/openFrameworks/blob/f4341627bf22c5569979dbc8da1336a352afaf92/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L52) by default but the variable is never changed by the settings from ofBaseWindow(?). Many instances of the code work with settings.windowMode (which works) some only use the local windowMode variable (which does not work because its always OF_WINDOW).
See f.e.:
https://github.com/openframeworks/openFrameworks/blob/f4341627bf22c5569979dbc8da1336a352afaf92/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L220

https://github.com/openframeworks/openFrameworks/blob/f4341627bf22c5569979dbc8da1336a352afaf92/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L496

Because of that many functions are basically broken for OF_FULLSCREEN and OF_GAME_MODE.

I would fix it by having settings.windowMode assigned to the local windowMode variable in the setup and then from there on only use windowMode throughout the rest of the code.
Before I do that and make a pull request I wanted to ask here first though if that is the proper approach for OF or not, maybe there is a different intention there that I'm missing.

arturoc:
i think i would remove the redundant local variable and only use the one in settings that way nobody changing the code can do this mistake again

PTS93:
great, I will create pull request then.

arturoc:
great thanks! i'm going to reopen this until you send the PR. If you include Fixes and the number of this issue in your  PR commit or in the message the issue will get closed automatically once we merge the PR

PTS93:
Should this maybe changed for https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGlutWindow.cpp
and possibly others as well? In ofAppGlutWindow it is using the local variable and assigning it the  value of settings.windowMode in setup. There it is at least consistent after that and no mixed usage.

arturoc:
which functions didn't work originally for you? i'm testing your changes but some things are broken. for examples starting a window in full screen doesn't work now. i'm fixing them but would be good to know what was the original issue in the first place

PTS93:
[...]

PR #5862, opened 2018-01-29
Fix windowMode not affecting some functions

PTS93:
See this issue for more details: https://github.com/openframeworks/openFrameworks/issues/5861

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -49,7 +49,7 @@ ofAppGLFWWindow::ofAppGLFWWindow()
 	bWindowNeedsShowing	= true;
 
 	orientation 		= OF_ORIENTATION_DEFAULT;
-
+	windowMode		= OF_WINDOW;
 	ofAppPtr			= nullptr;
 
     pixelScreenCoordScale = 1;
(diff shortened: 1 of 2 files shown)


arturoc:
thanks! please be sure to check that everything works after this change. this has been like that for a while and i'm concerned that some logic might be depending on having 2 copies of this value. Perhaps check that all 3 modes still work correctly and that values like the window size from ofGetWindowSize are returned correctly in all 2 modes and after calling ofToggleFullscreen back and forth

PTS93:
I compiled and tested it on Linux. As I removed the declaration of the local variable as well it should have thrown an error on compile time.
I also went through every use of the variable and it only ever checked what mode it was in by only using either way but never both of them.
I can't test on other platforms atm unfortunately.

PTS93:
So turns out I tested with an old version..
In the latest master compilation failed because the static function resize_cb is acessing the local var:
https://github.com/openframeworks/openFrameworks/blob/7e62212d5845ac521ce4302b10c6b0a08060b78b/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L1510

I know fixed it by adding the local var again and assigning it the value from the settings in setup.
I will add this commit here but let me know if this is how you want to proceed. It seems it makes sense to do the reverse by using the local var everywhere if the other window classes do the same aswell. Or we need to refactor a bit more to fix functions like resize_cb as well.

arturoc:
I'm going to close this since it's currently breaking some things and there's been no answer from the @PTS93 

PR #5875 should fix the original issue

PTS93:
Oh I didn't see that the the last commit still caused build errors. Thats odd, it was compiling fine with a fresh Master clone.

PR #5879, opened 2018-02-08
fix download_libs.ps1 script

tgfrerer:
+ script was deleting temporary libs folder "32", "64" repeatedly meaning only the libs from the very last .zip slice of libs would survive the cull.

Since the temporary folders "32/64" get deleted automatically as they are part of the "cull list" when preparing the libs directory for upgrade this additional delete routine is not needed in the first place.

+ also added an echo to instruct users on how to install libs for visual studio 2017, instead of the default, visual studio 2015.

+ also added an echo telling users that that the script finished successfully.

### DIFF
--- a/scripts/vs/download_libs.ps1
+++ b/scripts/vs/download_libs.ps1
@@ -1,11 +1,11 @@
 # You may override default parameters for this script by specifying 
 #       -paramName paramValue
 # When calling this script. For example: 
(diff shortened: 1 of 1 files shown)


arturoc:
thanks! perhaps we should make vs2017 default already?

tgfrerer:
yes, i was thinking so, too... i'll update the PR.

Issue #5878, opened 2018-02-08
regularExpressionExample - image doesnt show due to redirect


kkshmz:
Hi everyone, 
```
Xcode 8.3.3
macOS Sierra
Macbook Pro late 2016
Google Chrome
```
trying to figure out why the image doesnt show on the regular expression example.
Seems to be that I get redirected to the new Google Images webUI.
I understand that the program is wanting to go through the pages and scrape the ids one by one, but I think this is a google problem rather than a oF problem. 
Am I wrong?

```
searching for http://www.google.com/search?q=openframeworks&tbm=isch&sout=1&tbs=isz&&start=0&sa=N
<HTML><HEAD><metahttp-equiv="content-typ
e"content="text/html;charset=utf-8">
<TI
TLE>302Moved</TITLE></HEAD><BODY>
<H1>30
2Moved</H1>
Thedocumenthasmoved
<AHREF="
http://www.google.co.jp/search?q=openfra
meworks&amp;tbm=isch&amp;sout=1&amp;tbs=
isz&amp;start=0&amp;sa=N&amp;gws_rd=cr&a
mp;dcr=0&amp;ei=QeB7WqDkJ8H-0gS7oYaQCQ">
here</A>.
</BODY></HTML>
```

arturoc:
@edap I think you created this example originally? can you take a look? perhaps we can host the images ourselves so API changes won't break it

edap:
Sure!

arturoc:
great, if you need to host anything in the OF servers perhaps we can send a PR to ofSite with the content

edap:
Can I change this example with that one proposed here https://github.com/openframeworks/openFrameworks/pull/4992/files ? I find pretty strange that in order to explain regular expression you have to introduce `ofHttpResponse`

arturoc:
yeah that makes sense

edap:
I was adding that example when I've realized that it is already present!
in the folder:
https://github.com/openframeworks/openFrameworks/tree/master/examples/strings
we have two example:
stdRegulaExpressionExample
and 
regularExpressionExample

[...]

PR #5880, opened 2018-02-08
remove http request from regex example

edap:
as disccussed in https://github.com/openframeworks/openFrameworks/issues/5878


arturoc:
thanks!

Issue #3139, opened 2014-08-11
ofDrawBox seems to map textures mirrored


arturoc:
![screenshot from 2014-08-11 20 52 03](https://cloud.githubusercontent.com/assets/48240/3881800/f8cca368-2190-11e4-80e9-c050c59f0f90.png)

tgfrerer:
I've looked a bit closer at this issue and found the following: 

1. ofDrawBox and ofDrawPlane both have the same texture orientation. 
2. In ofBoxExample, the logo appears flipped 
3. ofBoxExample is drawn through a camera and does not call setupPerspective, the camera will therefore not flip the viewport (as ofCamera is constructed with `vFlip=false` by default)
4. when drawing ofPlane and ofBox without a camera the logo appears correct (axis show: Y-down, X-right)

From this i would conclude: 

1. adding `cam.setupPerspective(true)` to `ofApp::setup()` in ofBoxExample will fix the flipped apparance of the logo, as this sets vflip for camera.

Note that ofCamera has no member initialisers - and the ofCamera constructor sets `vflip` to `false` by default. On the other hand, `ofCamera::setupPerspective()` has `vFlip=true` as default argument.

Perhaps we should move `ofCamera`'s initialisers to member initializers, so that initial values for these fields are visible from the header file already?

tgfrerer:
I believe we may consider this re-closed with #5884

PR #5884, opened 2018-02-08
Fix texcoords primitives

tgfrerer:
Flips texture coordinates for all oF primitives so that textures seen through a default camera (without view flipping) appear correct. 

Addresses #3139 

ofboxexample

* updates of3DPrimitivesExample to use a default camera
capture


ofTheo:
be good to have @NickHardeman take a look. 
I wonder if this will break a lot of stuff as people have been coding around this? 

I agree though that textures should look correct with a default ofCamera with positive Y up.

NickHardeman:
Having the textures render correctly with the camera y up axis makes sense. 👍

arturoc:
thanks!

PR #5888, opened 2018-02-09
Modernise advanced3D example

tgfrerer:
modernise advanced3DExample

+ remove redundant sub-class of ofCamera, OrthoCam
+ clarify some variable names
+ clean up camera assignments
+ mention built-in frustum draw method in code comment


arturoc:
thanks!

PR #5893, opened 2018-02-12
fix box primitive winding order

tgfrerer:
Aligns box primitive triangle winding order so that it is consistent with the winding order of other primitives. That is, when viewed through an `ofCamera`, and `GL_CULL_FACE` is `enabled` with `glFrontFace(GL_CW)` it will look correct, the front faces visible, the back faces invisible.

All other oF primitives appear to follow this convention.


arturoc:
yes! thanks!

PR #5896, opened 2018-02-13
ofImage: add return value to save image

arturoc:


libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -366,7 +366,7 @@ static void saveImage(const ofPixels_<PixelType> & _pix, const std::filesystem::
 		ofPixels pix3 = _pix;
 		pix3.setNumChannels(3);
 		saveImage(pix3,_fileName,qualityLevel);
-		return;
+		return false;
```
hamoid: Should this be `return saveImage(pix3,_fileName,qualityLevel);`?

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -411,53 +412,55 @@ static void saveImage(const ofPixels_<PixelType> & _pix, const std::filesystem::
...
 	}
 
 	if(format==OF_IMAGE_FORMAT_JPEG && (_pix.getNumChannels()==4 || _pix.getBitsPerChannel() > 8)){
 		ofPixels pix3 = _pix;
 		pix3.setNumChannels(3);
 		saveImage(pix3,buffer,format,qualityLevel);
-		return;
+		return false;
```
hamoid: return always false?

Issue #5899, opened 2018-02-19
Images not loading on Android


martinell:
We are having problems loading images on Android from a URL string.

While debugging we found out that in this part of the ofLoadImage:

https://github.com/openframeworks/openFrameworks/blob/22a42144bc57393bd2f12b607f44e756fa06a1f1/libs/openFrameworks/graphics/ofImage.cpp#L181-L207

The URL scheme comes empty.

If we replace `absUri.data()` with `uriStr.c_str()` in the line below:

https://github.com/openframeworks/openFrameworks/blob/22a42144bc57393bd2f12b607f44e756fa06a1f1/libs/openFrameworks/graphics/ofImage.cpp#L197

Then it works but we are not sure this is the correct fix.

When debugging it, we can see that absUri holds correct values (we can inspect it and read the URL from inside it).

To add more information, we are compiling with the NDK 16 and using the clang compiler (by default in this version of the NDK). Also, we replaced the build configuration with our own cmake files.

arturoc:
can you try changing:

std::vector<char> absUri(bytesNeeded);

with

std::vector<char> absUri(bytesNeeded, 0);

martinell:
Hi Arturo,

This did not work. We actually debugged it and the vector is initialised correctly. Also, we checked the output of the method **uriUnixFilenameToUriStringA** and there is no errors.

The commit @acampayo referenced this issue from fixes it for us.

arturoc:
have you tried that still works when loading images from a file instead of an uri? it seems in that case the uri check would fail

kkirby:
Can you check that the vector has a null termination at the end when filled by uriUnixFilenameToUriStringA? Also, @arturoc isn't a vector here a little overkill? Wouldn't it make more sense to just do `char absUrl[bytesNeeded]`?

martinell:
Yes, checking the data we didn't see anything out of the ordinary. It copied all the characters from the url and the rest was filled up with 0's.

kkirby:
@martinell What does the URL look like?

arturoc:
if i remember well if we don't use it and we get a file path then the next check:

```
if(uriParseUriA(&state, absUri.data())!=URI_SUCCESS){ 
```

would fail for files and we wouldn't had a way to know if we got a malformed uri or just a correct file path

That's why I was asking if you had tried your changes when loading images from the filesystem instead of a url

martinell:
[...]

PR #5901, opened 2018-02-21
Images not loading on Android

acampayo:
Related to #5899


martinell:
This PR fixes #5899

arturoc:
This fix will still fail with absolute windows paths i think. I've sent my own fix here https://github.com/openframeworks/openFrameworks/pull/5926 adding a test to really check that every case is working on every platform. will keep this open by now just in case until we see if the tests pass

arturoc:
Thanks for reporting this, I just merged my PR which should fix it without breaking other platforms let me know if it's still a problem.

PR #5902, opened 2018-02-21
Add unimplemented methods to ofTexture on Android

acampayo:
I have added the implementation of the following methods (which are defined for Android in ofGLUtils) to ofTexture class to avoid missing method implementation issues on Android:

```
void ofRegenerateAllVbos() {}
void ofReloadAllImageTextures() {}
void ofReloadAllFontTextures() {}
```

### DIFF
--- a/libs/openFrameworks/gl/ofGLUtils.cpp
+++ b/libs/openFrameworks/gl/ofGLUtils.cpp
@@ -832,20 +832,6 @@ shared_ptr<ofBaseGLRenderer> ofGetGLRenderer(){
 }
 #endif
 
-#if defined(TARGET_ANDROID)
-void ofRegenerateAllVbos();
-void ofRegenerateAllTextures();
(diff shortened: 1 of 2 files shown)


arturoc:
i think this is just a left over from the previous implementation for this methods, it's surely safe to just remove all of them since there's no references to them anywhere else. can you send a new PR just removing lines 835 - 847 in ofGLUtils? that is:

```cpp
#if defined(TARGET_ANDROID)
void ofRegenerateAllVbos();
void ofRegenerateAllTextures();
void ofReloadAllImageTextures();
void ofReloadAllFontTextures();

void ofReloadGLResources(){
	ofRegenerateAllTextures();
	ofRegenerateAllVbos();
	ofReloadAllImageTextures();
	ofReloadAllFontTextures();
}
#endif
```

arturoc:
thanks!

PR #5909, opened 2018-02-27
Fixes for drawing BGRA textures on iOS.

martinell:
Fixes to enable drawing of BGRA textures on iOS.


libs/openFrameworks/gl/ofGLUtils.cpp
```diff
@@ -655,6 +659,7 @@ int ofGetNumChannelsFromGLFormat(int glFormat){
 	switch(glFormat){
 	case GL_RGB:
 		return 3;
+	case GL_BGRA:
```
arturoc: i think you are missing an ifdef here otherwise this fails in other opengles platforms

martinell: Done

ofTheo:
If this still looks good going to add to 0.11.1

PR #5912, opened 2018-02-28
Fix GL_INVALID_OPERATION on OpenGL ES 2.0 +

danoli3:
Fix Open GL ES 1.0/1.1 command being sent into Open GL ES 2/3.

According to Analyser this is a severe error and happening every frame multiple times.

screen shot 2018-02-20 at 11 48 50 am
screen shot 2018-03-01 at 12 41 16 am

Fixed by checking against major GL version.

### DIFF
--- a/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
@@ -1202,29 +1202,6 @@ void ofGLProgrammableRenderer::setAttributes(bool vertices, bool color, bool tex
 	if(wasColorsEnabled!=color){
 		if(currentShader) currentShader->setUniform1f(USE_COLORS_UNIFORM,color);
 	}
-#if defined(TARGET_OPENGLES) && !defined(TARGET_EMSCRIPTEN)
-	if(getGLVersionMajor() == 1) {
-		if(vertices){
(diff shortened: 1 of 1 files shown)


arturoc:
we only use the programable renderer in gl es 2 so if this is not correct we should just remove those calls completely. if i remember well we still needed this in opengl es 2 though since we are not using custom shader attributes in vbos but the fixed pipeline ones so i'm not so sure.

have you tried that disabling this doesn't break drawing with color or texture coordinates?

danoli3:
I tested with color on a vbo mesh and it works.

Those commands don't exist in ES2+ so I guess remove then if this isn't touched by ES1.

danoli3:
Okay yeah tested:
Works!
screen shot 2018-03-02 at 12 40 05 pm

I've updated this with removing those commands

danoli3:
@arturoc this is safe to merge. Those commands do nothing right now except throw those errors all the time.

danoli3:
screen shot 2018-03-10 at 12 33 03 pm

arturoc:
ok, yeah we are not really using this i was confused cause we are using already custom attributes but not VAOs. looking into this has actually lead me to discover it seems we are drawing things twice atm in opengl es when drawing an ofMesh.

thanks!!

Issue #5464, opened 2017-02-27
iOS calls draw() 3x before setup()


kylemcdonald:
I noticed I was getting a `texture has not been allocated` allocated and some other very confusing crashes. After stripping it way down I got to this point:

```c++
#include "ofMain.h"
#include "ofxiOS.h"

class ofApp : public ofxiOSApp {
public:
    void setup() {
        ofLog() << "setup";
    }
    void draw() {
        ofLog() << "draw";
    }
};

int main() {
    ofiOSWindowSettings settings;
    settings.enableRetina = true;
    settings.enableDepth = true;
    settings.enableAntiAliasing = true;
    settings.numOfAntiAliasingSamples = 4;
    settings.enableHardwareOrientation = true;
    settings.enableHardwareOrientationAnimation = false;
    settings.glesVersion = OFXIOS_RENDERER_ES1;
    settings.windowMode = OF_FULLSCREEN;
    ofCreateWindow(settings);
    return ofRunApp(new ofApp);
}
```

This code prints:

```
[notice ] draw
[notice ] draw
[notice ] draw
[notice ] setup
[notice ] draw
...
```

This behavior is consistent. I worked around this by adding a `bool ready = false;` near the top of ofApp, `ready = true;` at the end of `setup()` and `if(!ready) return;` at the top of `draw()`.

I couldn't find this documented anywhere, so apologies if this is the expected behavior on iOS :)

Running on an iPad Pro model ML0N2LL/A, iOS version 10.2.1, built with Xcode 8.2.1 and openFrameworks master (currently at 0c0a45dd6644c000c328302eb90bb7a4d52e2452 ). Project is set to build in debug mode (the default).

kylemcdonald:
I also get this same behavior on an iPhone 6s (iOS version 10.2.1 built with XCode 8.2.1).

jvcleave:
doesn't appear to be an issue in 0.9.8
Simple fix is to change:
https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxiOS/src/core/ofxiOSViewController.mm#L63

`[self.glView performSelector:@selector(setup) withObject:nil afterDelay:0];`

to

`[self.glView setup];`

arturoc:
this should be fixed now with #5918

PR #5918, opened 2018-03-02
iOS/tvOS - Fixes Update and Draw being called before Setup

danoli3:
This is a strange one really, it might have been around for a while (iOS 11, maybe 10). I noticed it recently and just guarded my own code with a bool, however after running a few old examples I see it breaks those so this is a critical.

Call stack: (bottom to top)
screen shot 2018-03-02 at 12 44 14 pm

viewDidLoad calls startAnimation which calls draw view and this is now calling a first call on Creation (something that didn't happen previously in iOS).
screen shot 2018-03-02 at 12 45 07 pm
(First called from here)

Just added a bool to guard calling the draw function that notify's update/draw if setup hasn't been called.


hamoid:
Related to https://github.com/openframeworks/openFrameworks/issues/5464 ?

PR #5920, opened 2018-03-02
Use the textureMatrix supplied from SurfaceTexture

kkirby:
This fixes a bug where on some devices the camera preview wouldn’t appear correctly. In my case it had a green line a few pixels tall along the bottom edge of the preview.

Also, in my instance the texture matrix flipped the image, so I had to enable bFlipTexture. I don’t know if this is device dependent or not. I tested on two separate devices, a Saumsung Tab 2 and Nexus 5X.

### DIFF
--- a/addons/ofxAndroid/src/ofxAndroidVideoGrabber.cpp
+++ b/addons/ofxAndroid/src/ofxAndroidVideoGrabber.cpp
@@ -106,6 +106,14 @@ void ofxAndroidVideoGrabber::Data::update(){
 	env->CallVoidMethod(javaVideoGrabber, getTextureMatrix, matrixJava);
 	jfloat* cfloats = env->GetFloatArrayElements(matrixJava, 0);
 	ofMatrix4x4 mat(cfloats);
+	if(mat(0,0) == -1){
+		mat.scale(-1,1,1);
+		mat.translate(1,0,0);
+	}
+	if(mat(1,1) == -1){
+		mat.scale(1,-1,1);
+		mat.translate(0,1,0);
+	}
 	texture.setTextureMatrix(mat);
 }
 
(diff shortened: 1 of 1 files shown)


kkirby:
Is there anyone who can test this on their own device to make sure it doesn't break the preview? Given the nature of the change and how device manufactures could dictate the value pulled from `getTransformMatrix(float*)` I'm unsure if `bFlipTexture` is required in 100% of cases.

HalfdanJ:
I'm getting the error `Error:(108, 2) error: unknown type name 'ofMatrix4x4'`, guess you need to add `#include "ofMatrix4x4.h"` or something like that. Otherwise looks good to me. I'm curious what devices this was an issue on?

kkirby:
Thanks. It was a Samsung Tab S2 with a preview resolution of 1920x1080.

HalfdanJ:
Looks good to me

arturoc:
We had this a while ago but it seems this commit: https://github.com/kkirby/openFrameworks/commit/43ec1b043ea2fd5c6d6cdca2a9f2190c819b0ca2#diff-7a722a68bbce49b9e3d296fb08507ed6 removed it.

in the original solution we were baking the flip in the matrix:

https://github.com/kkirby/openFrameworks/blob/c9cd3e912691d5286c3b6c84d6dd83de055e89d0/addons/ofxAndroid/src/ofxAndroidVideoGrabber.cpp#L232-L236

can you do that instead?

kkirby:
@arturoc That's interesting. I wonder why it got removed.

@HalfdanJ Are you able to provide insight into why you removed the code?

arturoc:
i can imagine it was just an oversight but perhaps you had any problem with that code @halfdanj?

arturoc:
thanks!

kkirby:
@arturoc Hey, just a heads up, I hadn't really tested my changes and whether or not my if conditionals are solid and accurate. Basically I was checking to see if the matrix coming from the Surface Texture was flipping either x or y and to reverse it since I had an instance where the x was flipped and the y wasn't. So, not really sure if we should have merged that in. Though you may have looked at it and agreed with my code. Let me know. Otherwise, feel free to revert until I have more time to test. Currently the machine I do most of my work on is being held hostage by IT so I won't be making any progress on testing anytime in the coming days.

arturoc:
ah ok, no worries we can fix it later with a new PR. did you tried the original solution i posted i remember that used to work fine

PR #5926, opened 2018-03-08
ofImage: Fix loading from a URI

### LINKED ISSUE #5899: Images not loading on Android

arturoc:
Fixes #5899


martinell:
For what we hace seen, Unix paths will pass as a success through uriParseUriA.

martinell:
Also, it does not check for "file://..." I'm not sure if it's necessary, tough.

arturoc:
yes that's it unix paths will pass the text but not windows ones, probably the unix conversion from file -> uri won't be necesary but it doesn't hurt.

Not sure about file:// don't think anyone will use that syntax

PR #5930, opened 2018-03-18
Fixes potential leak of memory pointed to by 'uniformName'

gilbertfrancois:


gilbertfrancois:
The change adds a missing `delete[]` statement at the end of the function. It is tested on macOS 10.13.

arturoc:
thanks!

PR #5933, opened 2018-03-20
Add translate, rotate and scale to ofPolyline

hamoid:
Based on ofPath methods.

To consider for both ofPath and ofPolyline:
* scale uses float but translate and rotate uses glm::vec
* it might nice to be able to set the origin for rotate and scale, instead of using (0,0,0)
* rotating in 2D is not convenient: you need to specify glm::vec3(0.0, 0.0, 1.0) as axis. Add `rotate(float az)` for 2D rotations?
* az == azimuth? use `deg` instead? or `rotateDeg()` + `rotateRad()`?

### DIFF
--- a/libs/openFrameworks/graphics/ofPolyline.h
+++ b/libs/openFrameworks/graphics/ofPolyline.h
@@ -349,6 +349,20 @@ public:
     /// points removed.
 	void simplify(float tolerance=0.3f);
 
+        /// \}
+	/// \name Transform polyline
+	/// \{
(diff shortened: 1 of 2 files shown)


hamoid:
Change suggested here: https://forum.openframeworks.cc/t/how-do-i-scale-a-ofpolyline-or-convert-ofpolyline-to-ofpath/28170/7

arturoc:
thanks!!

Issue #5157, opened 2016-07-07
ofImage::grabScreen not works for enableAntiAliasing = true 


datangguo:
Hey guys,

ofImage::grabScreen not works when settings.enableAntiAliasing = true for window, it works when turn antialiasing off, I searched related topic, but no luck. I'm using of_iOS_0.93 for iPhone 5S with iOS 9.3.2

Any suggestion? thanks!

datangguo:
The default: (grabScreen works great)
```
settings.enableAntiAliasing = false; // enables anti-aliasing which smooths out graphics on the screen.
settings.numOfAntiAliasingSamples = 0; // number of samples used for anti-aliasing.
```

and my code: (grabScreen fail to work)
```
settings.enableAntiAliasing = true; // enables anti-aliasing which smooths out graphics on the screen.
settings.numOfAntiAliasingSamples = 4; // number of samples used for anti-aliasing.
```
This is how I grab:
```
void ofApp::touchUp(ofTouchEventArgs & touch){
    
    string fileName = "grab.png";

    ofImage img;
    img.grabScreen(0,0,ofGetWidth(),ofGetHeight());
    img.save(ofxiPhoneGetDocumentsDirectory() + fileName);
    
    cout << ofxiPhoneGetDocumentsDirectory() + fileName << endl;
}
```

datangguo:
in short: ofImage::grabScreen returns **BLACK** when multiSample enabled for AntiAliasing, OF / iOS.

martinell:
I was dealing with this too and found a solution that works [here](https://stackoverflow.com/questions/35016720/glresolvemultisampleframebufferapple-generate-gl-invalid-operation-on-ios-ope):

I borrowed the code and translated it into C++

`
        GLint wid;
        GLint hei;
        glGetRenderbufferParameteriv(GL_RENDERBUFFER, GL_RENDERBUFFER_WIDTH, &wid);
        glGetRenderbufferParameteriv(GL_RENDERBUFFER, GL_RENDERBUFFER_HEIGHT, &hei);
        GLint def;
        glGetIntegerv(GL_FRAMEBUFFER_BINDING_OES, &def);
        
        // init non-multisampled frame buffer
        GLuint framebuffer = 0;
        GLuint colorRenderbuffer = 0;
        
        glGenFramebuffersOES(1, &framebuffer);
        glBindFramebufferOES(GL_FRAMEBUFFER_OES, framebuffer);
        
[...]

PR #5935, opened 2018-03-23
Fixes screen grab when multisample anti-aliasing is enabled for iOS devices.

martinell:
Related to this issue: #5157

### DIFF
--- a/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
@@ -15,6 +15,10 @@
 #include "ofNode.h"
 #include "ofVideoBaseTypes.h"
 
+#ifdef TARGET_OF_IOS
+#import "ofAppiOSWindow.h"
+#endif
+
 using namespace std;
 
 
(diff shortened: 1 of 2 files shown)


arturoc:
Thanks for this fixes, but i don't think this is correct. I'm not 100% sure what's going on in the new code but it's creating an fbo every frame and not even deallocating it later. It's also doing get calls which can be slow things down a lot just to get the size of the framebuffer which could be read from the window.

In general i think this is something that should at least be done through ofFbo so there's no memory. Or perhaps during the creation of the window. If i remember correctly on iOS you set up a framebuffer at the very beginning so perhaps that's the place to do this but not sure

martinell:
You're right. I was just testing code to get a fix. I'll try to move the code  to the setup part of the glview.

arturoc:
cc @danoli3

martinell:
I just saw this is already done here:

https://github.com/openframeworks/openFrameworks/blob/ffd8fdd96e761d008bcc19e4c1b21c6069d1c863/addons/ofxiOS/src/gl/ES1Renderer.m#L53-L69

I moved this to a method:

`- (void)resolveMultisampleFramebufferIfNeeded `

and propagated it to the ofAppiOSWindow to be called directly from the renderer when capturing the screen.

Does this sound more reasonable?

PR #5938, opened 2018-03-23
Preload the 3d model resources to memory

acampayo:
The aim of these modifications is to allow to preload the 3D model resources to memory from a background thread, improving the 3D model loading process performance.


arturoc:
this looks good, i'll wait until 0.10 is released to merge it though since it changes a bunch of stuff and we are only merging bugfixes until the release is done.

Thanks!

PR #5939, opened 2018-03-25
Added shader formatting to unique shader (for Raspberry Pi).

prisonerjohn:


arturoc:
why is this needed? if i remember well the unique shader doesn't need formating cause is specific to the raspberry so the syntax is fixed

prisonerjohn:
The syntax is fixed, but the [unique shaders](https://github.com/openframeworks/openFrameworks/blob/0ba4a06fb5ec6ba042a32db429247aab0bc65307/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp#L2126-L2175) are prefixed with the [default shader headers](https://github.com/openframeworks/openFrameworks/0ba4a06fb5ec6ba042a32db429247aab0bc65307/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp#L1845-L1859). Those have placeholders like `%extensions%` and `%glsl_version%` which need to be replaced in the [shaderSource(...)](https://github.com/openframeworks/openFrameworks/blob/0ba4a06fb5ec6ba042a32db429247aab0bc65307/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp#L2293) method.

arturoc:
ah ok, yeah some people have been having this error lately,thanks!

PR #5941, opened 2018-03-28
Change ofPath: flagShapeChanged() and hasChanged() from private to public

c-mendoza:
ofPath::flagShapeChanged() could be really useful when paths are changed externally (the workaround is to call ofPath::getCommands() only so that ofPath::flagShapeChanged is called), and ofPath::hasChanged doesn't get called at all anywhere in OF, so making it public doesn't hurt the internals. Also, making it public mirrors the behavior of ofMesh::hasChanged.

### DIFF
--- a/libs/openFrameworks/graphics/ofPath.h
+++ b/libs/openFrameworks/graphics/ofPath.h
@@ -393,10 +393,12 @@ private:
 	void addCommand(const Command & command);
 	void generatePolylinesFromCommands();
 
+public:
 	// only needs to be called when path is modified externally
 	void flagShapeChanged();
 	bool hasChanged();
 
+private:
 	// path description
 	//vector<ofSubPath>		paths;
 	std::vector<Command> 	commands;


arturoc:
the only way to change a path is by calling getCommands no? making this public can lead to think that this needs to be called whenever we change the path or something

c-mendoza:
Well, if you change a command or a point externally you have to call getCommands for the path to notice the change, and that is just odd semantically. We could note in the documentation that under most circumstances flagShapeChanged doesn't have to be called by the user, only when ofPath's data is modified externally.

arturoc:
can you put an example of how you would use it? in principle there's no way to change a path without it flagging itself as changed and if there is it should be fixed so the path automatically changes the flag.

c-mendoza:
If you change the commands yourself. I made a UI ofPath editor that can add, insert, move, delete points, and the most convenient thing to do was to give each path handle a pointer reference to the command they represented. Thus, the commands are updated outside of the path, and ofPath::getCommands() needs to be called to update ofPath's internals. 

Also, if you were to insert a point at an arbitrary position in the path (i.e. not at the end) you would have to insert a command yourself into the commands vector... which would then require you to call getCommands to get ofPath to use the new command:

```C++            
path->getCommands().insert(pos, std::move(command));
// Need to have the path update itself, but flagShapeChanged() is private:
path->getCommands();

```

To be honest, the whole getCommands() thing doesn't bother me that much, I can live with the workaround, although it is semantically weird, especially when there is already a method that does exactly what it says it does (flagShapeChanged). 

All that said, how about making hasChanged() public? This is a much bigger issue for me.

arturoc:
you are already calling getCommands once when doing:

```
path->getCommands().insert(pos, std::move(command));
```

which flags the path as changed already so there's no need to call it again, no?

c-mendoza:
Ah, true.

arturoc:
That's what i mean you shouldn't need to flag the path as changed at all it auto detects when it's been changed by it self. Same with other objects in OF.

should we close this then?

c-mendoza:
But not if you are holding a reference to the command outside of the vector and you then change it... in that case you HAVE to call getCommands() for ofPath to use the changes.

How about hasChaged? I can submit a different PR for that if you are ok with that change.

arturoc:
Is not a very good idea to call a method on an object while holding a mutable reference to it and i can imagine it would be a really weird case. Also interleaving updates and draws during the same frame is not a very good idea in terms of performance so i don't think we should be promoting that case and i would even try to disallow it if there was a way to do so in c++.

I really don't like the idea of exposing this API. This objects are supposed to know when they've changed and update themselves. exposing them breaks encapsulation and if the internals of the object change there's a good chance that programs using that api will break without a compiler error which is a problem really hard to debug. 

The hasChanged method is just used so the path knows if it needs to update it's internal vbo so i don't see what use case there would be for that outside of the class

c-mendoza:
Fair enough. I’ll rethink my approach. We can close this issue then.

c-mendoza:
Fair enough. I’ll rethink my approach.

PR #5942, opened 2018-03-30
0.10.0 prerelease feedback

danomatika:
Some small fixes while checking out the OF 0.10.0 prerelease.


danomatika:
I added some small ofxOsc updates after double checking things.

Also, should ofxOsc follow the pattern set with the new Settings structs? Currently, the `ofxOscSenderSetting`s and `ofxOscReceiverSettings` are declared globally. Should they be moved into their respective classes and become `ofxOscSender::Settings` and `ofxOscReceiver::Settings`? I think now is the time to change this before the release.

arturoc:
Thanks Dan, I think it's fine to leave them as their own classes, we used ofClass::Settings a couple of times but in certain cases we can't use it cause some interdependencies issues so it's probably better to move everything to ofClassSettings or ofModuleSettings. Once/If we have namespaces we can think of turning these into namespace::Settings if that makes sense but i think by now it's safer to have ofClassSettings since that seems to work in every case.

I'll merge this as soon as all the tests pass

danomatika:
Ok. It would certainly make wrapping the important structs easier without the nesting, although I did find that very recent versions of SWIG finally support nested structs & classes, although with some caveats.

Let me get through updating the ofxLua bindings and some tests before merging. I may find a few more things before I'm through.

arturoc:
ok, great

danomatika:
Using my [check-for-src-pattern script](https://gist.github.com/danomatika/9173240), I'm seeing "struct Settings" in :

* ofMaterial
* ofShader
* ofFbo
* ofTrueTypeFont
* ofAppEGLWindow

danomatika:
Some more feedback:

* ofSerial & ofArduino are insoncistent in the their types for the buffer lengths, ie. some functions use int instead of size_t. I think this should be consistent using size_t

ofSerial:
~~~
long readBytes(unsigned char * buffer, size_t length);
~~~
ofArduino:
~~~
void serialRead(Firmata_Serial_Ports port, int maxBytesToRead);
~~~

* ofArduino seems to be inconsistent in how it accepts binary data ie. signed/unsigned char vectors etc. For instance

~~~
void sendOneWireWrite(int pin, std::vector<unsigned char> devices, std::vector<unsigned char> data);
~~~
takes a char vector while
~~~
void sendI2CWriteRequest(char slaveAddress, unsigned char * bytes, int numOfBytes, int reg = -1);
void sendI2CWriteRequest(char slaveAddress, const char * bytes, int numOfBytes, int reg = -1);
void sendI2CWriteRequest(char slaveAddress, char * bytes, int numOfBytes, int reg = -1);
void sendI2CWriteRequest(char slaveAddress, std::vector<char> bytes, int reg = -1);
~~~
is overloaded to accept multiple types, none of which are `std::vector<unsigned char>`.

I think the overload is fine but the type of data should at least match overall between functions in the same API so you don't have to be casting back and forth using the same ofArduino object.

Further, this API differs from that of ofSerial which doesn't seem to use vectors at all:
~~~
long readBytes(unsigned char * buffer, size_t length);
long readBytes(char * buffer, size_t length);
long readBytes(ofBuffer & buffer, size_t length);
~~~

For instance, maybe both classes could work like ofSerial and pass ofBuffer objects around as base binary buffers and provide overloads to raw buffer pointers for those who want to handle things at a lower level. 

* ofArduino struct and enum naming uses snake case instead of camel case as with the rest of the OF API ie. Firmata_Serial_Data instead of ofArduinoFirmataSerialData or ofArudino::FirmataSerialData. Was it like this before? I suppose I don't know as Ive not used this API before.

* The ofBaseTypes.h header has been split up but ofBaseTypes.cpp is still there. Why not split up the implementation as well?

* I don't see ofMathConstants.h in the OF lib Xcode source tree.

* Do we need templated ofToString redefinitions in: ofImage.h, ofPixels.h, ofGraphics.h? I've commented them out and it builds fine:
~~~
template<typename T>
std::string ofToString(const T & v);
~~~

* ofUtils.h: ofTime, ofTrim* functions, & ofUTF8* functions needs doc commenting

* There is an inconsistent naming between "GL" or "Gl" in API, mostly in ofGLUtils.h, ie. ofGetGlFormat() versus getGLRenderer(), etc. Examples:
~~~
ofGetGlInternalFormat
ofGetGlInternalFormatName
ofGetGlTypeFromInternal
ofGetGlType
ofGetGlFormat
~~~

* Prefer nested or global enums? I suppose that depends on the usage. ofRectangle's various global enums versus ofEasyCam's nested TransformType, ofSoundDevice apis, BoxPrimitive sides, TouchEventArgs. The nested types make sense in their own context, I wonder if ofRetcangle do as well?

* Why does ofFboBeginMode need to be a class enum? A regular enum can be cast to a short. Also, maybe this could also be a nested class as fitting the previous point.

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -28,19 +28,19 @@ ofBuffer::ofBuffer()
 }
 
 //--------------------------------------------------
-ofBuffer::ofBuffer(const char * _buffer, std::size_t size)
-:buffer(_buffer,_buffer+size)
+ofBuffer::ofBuffer(const char * _buffer, std::size_t _size)
```
tgfrerer: I think we might need to be careful with naming variables with leading underscores, as, according to the C++ Standard, 17.4.3.1.2 Global Names [lib.global.names], paragraph 1:

> Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.

see also mozilla's guidelines for variable naming
https://developer.mozilla.org/en-US/docs/Mozilla/Using_CXX_in_Mozilla_code

I assume the leading underscores were inserted for consistency with the old code, but it's probably safer to remove the leading underscores alltogether.

danomatika: Removing would be fine as well. I was going for consistency as it make it easier to handle the buffer conversion with SWIG when the arguments are named the same.

arturoc:
> Why does ofFboBeginMode need to be a class enum?

I think it's much nicer than ultralong names but for consistency sake it should be changed to be a non class enum and the names should be consistent with other enums like: 

```
OF_FBO_BEGIN_MODE_NO_DEFAULTS
OF_FBO_BEGIN_MODE_PERSPECTIVE
OF_FBO_BEGIN_MODE_MATRIX_FLIP
```

Again, if you want to send a PR i think this can be merged before 0.10

libs/openFrameworks/utils/ofFileUtils.cpp
danomatika: "_buffer" & "_size" now "buffer" & "size"

danomatika:
>ofBasetTypes.h was split to make compilation faster so keeping the cpp as one file works better than splitting it. we can now include the base types more modularly but the couple of things that need to be in the cpp are compiled in only one unit

I get that, but I was looking for the implementation and didn't find the function right away. I'm used to the header have a matching .cpp file. I figured it was just an oversight when splitting the headers and not a design decision.

danomatika:
>I think moving to global by now would be a good idea if only for consistency sake since some can't be nested. ofFbo::Settings has been there for ages so we can't just remove them but we need to deprecate them. And i would also deprecate ofEGLWindow::Settings as well since it's not the first version in which it appears
>
>Do you want to send a PR?

That's too large of a change for me to take responsibility for. I just think there should be a concerted effort at some point to consolidate the various enum naming and placement and stick to that moving forward.

danomatika:
Ok, I ended up tackling the Settings structs and moved them all into the global namespace.

I also moved the ofTrueTypeFont direction class enum into a regular global enum. This was mainly because there seem to be issues with forward declaring enum classes and it was the only enum class left at this point anyway. Now it matches the general API, although the naming could possibly be OF_TTF_DIRECTION_*.

danomatika:
I also changed the ofFboBeginMode enum class into a simple ofFboMode enum with OF_FBOMODE_* naming. I feel this better matches it's usage and the *Begin is unnecessary. It could possible be called ofFboOptions, ofFboRenderOptions, but I think ofFboMode could suffice for now.

danomatika:
Last, I deprecated the ofGetGl* functions in favor of ofGetGL* counterparts.

Ok, I think I'm done for now. I tried to structure the commits for easier cherry-picking and you can edit this PR as well.

danomatika:
Is the long term plan to drop the ofVec* classes in favor of the glm versions?

I ask as I notice many of the functions which use to take ofPoints, now take glm::vec3 or glm::vec2 arguments. The ofVec to glm operators work great in C++, but oftentimes these kinds of operators do not translate into a scripting language.

It would great if there was a way to keep an overload to ofPoint instead of using the C++-only pass type operator. Otherwise I will have to do more manual workarounds. Not impossible but just checking before I spend any time on them.

arturoc:
thanks! about glm/ofVec the idea is that ofVec will remain there for some time for compatibility with old projects perhaps be moved back into an addon at some point in the future but the core will only use glm from now on. also addons and examples should be ported to use glm at some point.

ofDefaultVec is there only to be used by the classes that hold collections of glm::vec/ofVec cause for those there's no way to have an automatic conversion so we used a macro on ofConstants to use ofVec instead which should make old projects 100% compatible but it's there only for compatibility with old projects.

Having extra functions for ofVec could work but i'm not sure is the best idea since we would have to addd a ton of code + remove it at some point whenever we remove ofVec. 

Can't you just port your code to use glm? also the memory layout is compatible so from a scripting language i can imagine that you can pass whatever that is 2, 3 or 4 floats and it would work with ofVec or glm without problem

danomatika:
> about glm/ofVec the idea is that ofVec will remain there for some time for compatibility with old projects perhaps be moved back into an addon at some point in the future but the core will only use glm from now on. also addons and examples should be ported to use glm at some point.

Sure, that makes sense.

>ofDefaultVec is there only to be used by the classes that hold collections of glm::vec/ofVec cause for those there's no way to have an automatic conversion so we used a macro on ofConstants to use ofVec instead which should make old projects 100% compatible but it's there only for compatibility with old projects.

Ok

>Having extra functions for ofVec could work but i'm not sure is the best idea since we would have to addd a ton of code + remove it at some point whenever we remove ofVec.

Yeah, it would be a pain when the plan is to move forward with glm. I'd almost think it would make sense to consider moving to namespaces and typdefing the gym types to of::vec2, of::vec3, etc t I suppose that's another discussion.

>Can't you just port your code to use glm? also the memory layout is compatible so from a scripting language i can imagine that you can pass whatever that is 2, 3 or 4 floats and it would work with ofVec or glm without problem

Part of working with scripting language bindings is always having to find workarounds for things that work in C/C++ that don't translate directly. Of course, the C/C++ APIs can be written in ways to facilitate this, hence many of my suggestions.

In this case, of course, it's really a architectural decision toward the future, so I wouldn't want to fight that. SWIG has several methods to help coerce one type to another but to Lua C++ classes just show up as user data pointers, so it doesn't care. I just have to wrangle to conversion in the interfaces.

I was thinking the two types offer different functions though. Does ofVec3 and glm::vec3 have the same API as well as data layout?

PR #5947, opened 2018-04-04
Converted ofxCvBlob points from ofPoint to ofDefaultVec3.

prisonerjohn:
I'm assuming we're phasing out `ofPoint`...

In any case, I encountered an issue with the following code, which this change fixes:
```
ofxCvBlob blob = contourFinder.blobs[i];
ofPolyline polyline = ofPolyline(blob.pts);
```


arturoc:
thanks!

Issue #5948, opened 2018-04-04
F4, F5, F6, F7 keys mapped to same OF_KEY as LEFT, UP, RIGHT, DOWN


prisonerjohn:
If we look at [ofEvents.h](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/events/ofEvents.h#L69-L90), the result of the bitwise OR for the F4, F5, F6, F7, keys is identical to the arrow keys, which makes it impossible to differentiate them in a key event handler.

For example, if we look at F5:
```
   3808 = 1110 1110 0000
OR    5 = 0000 0000 0101
------------------------
   3813 = 1110 1110 0101
```
and UP
```
   3808 = 1110 1110 0000
OR  101 = 0000 0110 0101
------------------------
   3813 = 1110 1110 0101
```
we end up with the same value for `OF_KEY_F5` and `OF_KEY_UP`!

Not sure what the solution is here, maybe change the value of `OF_KEY_MODIFIER` or the bitwise OR operator. 
Or maybe since we're not using GLUT anymore we can just use the [GLFW key macros](http://www.glfw.org/docs/latest/group__keys.html) directly?

arturoc:
the problem with the glfw macros is that some of them clash with characters. The main problem is we use the actual character to check for a key like:

```
if (key == 'a'){
...
}
```
we need to find numbers that are not used for those keys, probably we don't need to use OF_KEY_MODIFIER but we need to figure out codes that are empty in the unicode spec or that are before the symbols you can type.

prisonerjohn:
It looks like there's a nice chunk of 48 unused codepoints from 0870 - 089F. See https://unicode-table.com/en/#syriac-supplement

If it's just a matter of replacing those `OF_KEY_xxx` values, I can go ahead and do that and submit a PR if it appears to work correctly.

prisonerjohn:
Fixed in PR #5956

PR #5956, opened 2018-04-06
Changed values for non-visible keys to use unused Unicode points.

prisonerjohn:
This resolves conflicts with keys that had the same value (like OF_KEY_F5 and OF_KEY_UP).
See #5948 for a better description.


arturoc:
it seems this breaks the glut window. probably we just need to add a switch case there to convert from the original key code to whatever we use.

Also i've just found out that there's some ranges in unicode reserved for this kind o private uses which will never get used by codepoints:

https://www.unicode.org/faq/private_use.html

prisonerjohn:
I've just replaced the code points to use these private use ones you found.

As for GLUT, I wasn't aware we were even still supporting that :) Which platform should I use to test on? I can't find the lib for VS or Mac.

PR #5959, opened 2018-04-08
Bugfix windowing examples

### LINKED ISSUE #5958: multiwindow examples do not compile
As height and width gone protected in ofWindowSettings,  need to use setSize() instead.

oxillo:
fixes #5958.
Replaced settings.width = w; settings.height =h by settings.setSize( w, h )


arturoc:
thanks!

PR #5977, opened 2018-04-18
Bugfix networkTcpClientExample

hiroMTB:
send message when user hit return key.
Otherwise it send strange strings like

input 
1111\n

output
1 11 111 1111

### DIFF
--- a/addons/ofxAndroid/src/ofxAndroidUtils.h
+++ b/addons/ofxAndroid/src/ofxAndroidUtils.h
@@ -59,7 +59,6 @@ bool ofxJavaCallBoolMethod(jobject object, jclass classID, std::string methodNam
 bool ofxJavaCallBoolMethod(jobject object, jclass classID, std::string methodName, std::string methodSignature, ...);
 bool ofxJavaCallBoolMethod(jobject object, std::string className, std::string methodName, std::string methodSignature, ...);
 
-void ofxAndroidPauseApp();
 void ofxAndroidAlertBox(std::string msg);
 int ofxAndroidProgressBox(std::string msg);
 void ofxAndroidDismissProgressBox(int id);
--- a/addons/ofxAndroid/src/ofxAndroidUtils.h
+++ b/addons/ofxAndroid/src/ofxAndroidUtils.h
@@ -59,6 +59,7 @@ bool ofxJavaCallBoolMethod(jobject object, jclass classID, std::string methodNam
 bool ofxJavaCallBoolMethod(jobject object, jclass classID, std::string methodName, std::string methodSignature, ...);
 bool ofxJavaCallBoolMethod(jobject object, std::string className, std::string methodName, std::string methodSignature, ...);
 
+void ofxAndroidPauseApp();
 void ofxAndroidAlertBox(std::string msg);
 int ofxAndroidProgressBox(std::string msg);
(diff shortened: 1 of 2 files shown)


arturoc:
this seems to be working fine for me in linux, in any case this was like this before the difference is not so much in the example but in a retry in the tcp client to solve an issue with long messages: https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxNetwork/src/ofxTCPClient.cpp#L245

arturoc:
this is the commit that changed it: https://github.com/openframeworks/openFrameworks/commit/8e01052ec278bb52ecc26d511dd5da1c6f0ea31c

hiroMTB:
Thanks! 
I noticed this only happen in following condition.

1. start server and client -> both works OK
2. kill server and restart
3. The port is already occupied so server returns error (couldn't bind to port)
4. kill and restart server again
5. type 111 on client and server respond like screen shot.
ss

I thought this could be because server does not shut down properly. 
I put following code on server app, but it does not solve this problem.
`void ofApp::exit(){
    for(unsigned int i = 0; i < (unsigned int)TCP.getLastID(); i++){
        if( !TCP.isClientConnected(i) )continue;
        TCP.disconnectClient(i);
    }
    TCP.close();
}
`
Anyway I will check your link and investigate tmrw.

hiroMTB:
I found this in oF forum.
It solves a re-connection problem but still has a problem like a screenshot above.
https://forum.openframeworks.cc/t/how-to-resolve-quot-ofxnetwork-socket-address-already-in-use-quot/7284/7

hiroMTB:
It's same before and after https://github.com/openframeworks/openFrameworks/commit/8e01052ec278bb52ecc26d511dd5da1c6f0ea31c
Both version has trouble reconnecting server. 
Here is error.

> [ error ] ofxNetwork: ~/openFrameworks/addons/ofxNetwork/src/ofxTCPManager.cpp: 159 EADDRINUSE: the socket address of the given addr is already in use
> [ error ] ofxTCPServer: setup(): couldn't bind to port 11999

Looks like here is two problems.
A. Most of the time, server can not re-connect after server app restart
B. Even when server success to reconnect, it sends strange character like a screenshot above.

Problem A and B happen on both of before & after https://github.com/openframeworks/openFrameworks/commit/8e01052ec278bb52ecc26d511dd5da1c6f0ea31c

hiroMTB:
Problem A can be solved by the code from forum.
But not problem B.
Can you reproduce this reconnection error on Linux?

arturoc:
no, and i think it's only osx where it happens. there's some tests in the test/addons/network* folders that test this among other things and they fail randomly in osx. it's been a while since they don't though

arturoc:
mmh actually yes, i just tried killing the server and i get the same error, i was trying to kill the client only. let me take a look

arturoc:
So the problem was that we weren't resetting the delimiter after the example reconnected so the server and client where using different delimiters, one [TCP] and the other \n

The busy socket a little while after the application is killed is to be expected but I don't think we should reuse by default.

We should have a setup method that specified this things, both the delimiter and the reuse as well as other settings through a Settings struct. And deprecate the individual setters which are kind of error prone

Anyway going to close this for now and we can look into adding a struct settings for the network classes for next release

hiroMTB:
Thanks!

PR #5982, opened 2018-04-23
Fix examples

roymacdonald:
Fixed examples that were not compiling on macos.
This fixes are:
- events/allEventsExamples
	OF_KEY_MODIFIER not found.
- gl/gpuParticleSystemExample
	in main.cpp, ofGLWindowSettings::width is a protected member. switching to use ofGLWindowSettings::setSize(...) instead.
- graphics/polylineBlobsExample
	problems because ofPolyline_<ofPoint> is being used instead of ofPolyline
- video/osxVideoRecorderExample
	- not working on macos >10.11

-general
	ofPoint should typedef ofDefaultVec3 instead of ofVec3f


examples/events/allEventsExample/src/ofApp.cpp
```diff
@@ -38,134 +38,91 @@ void ofApp::keyPressed  (int key){
...
+    }else if( key == OF_KEY_RIGHT_CONTROL){
+        eventString += "RIGHT CONTROL";
+    }else if( key == OF_KEY_LEFT_ALT){
+        eventString += "LEFT ALT";
+    }else if( key == OF_KEY_RIGHT_ALT){
+        eventString += "RIGHT ALT";
+    }else if( key == OF_KEY_RIGHT_SUPER){
+        eventString += "RIGHT SUPER";
+    }else{
+        eventString += (char)key;
```
arturoc: Can you use `ofUTF8Append(eventString, key)` here instead of plain +

examples/events/allEventsExample/src/ofApp.cpp
```diff
@@ -38,134 +38,91 @@ void ofApp::keyPressed  (int key){
...
```
arturoc: Can you use ofUTF8Append(eventString, key) here instead of plain +

examples/video/osxVideoRecorderExample/src/ofApp.cpp
```diff
@@ -1,13 +1,8 @@
 #include "ofApp.h"
```
arturoc: Can you just delete this full example? it doesn't really work in any platform anymore so it's just confusing to have it at all

libs/openFrameworks/types/ofParameter.h
```diff
@@ -3,6 +3,9 @@
 #include "ofEvents.h"
 #include "ofConstants.h"
 #include "ofPoint.h"
+#include "ofVec2f.h"
+#include "ofVec3f.h"
+#include "ofVec4f.h"
```
arturoc: Why do we need this? I'm trying to include as few headers as possible and don't remember this was needed here?

libs/openFrameworks/types/ofPoint.h
```diff
@@ -1,9 +1,9 @@
...
 
-#include "ofVec3f.h"
+#include "ofMathConstants.h"
 
 /// \file 
 /// ofPoint is a typedef (alias) of ofVec3f
 
 /// \brief Look at ofVec3f for documentation
-typedef ofVec3f ofPoint;
+typedef ofDefaultVec3 ofPoint;
```
arturoc: not so sure about this, i left it like it was on purpose so it's not up to date and people start to use glm instead. Does it help with something in particular? otherwise it shouldn't change and much less in the middle of the release candidate process.

examples/video/osxVideoRecorderExample/src/ofApp.cpp
roymacdonald: Sure, it is much easier.

libs/openFrameworks/types/ofParameter.h
roymacdonald: there are some template specializations that need those, in the following lines.
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/types/ofParameter.h#L366-L393

I changed the includes in ofPoint.h which were indirectly including ofVec*f, so I thought that it was better to include this directly in this file rather than having it as a side effect of another include

libs/openFrameworks/types/ofPoint.h
roymacdonald: I see. Maybe it should have some kind of deprecation warning?
there are lots of parts, both in the core and in the examples, that use ofPoint. Not sure if this helps or not but I feel it is more correct having ofDefaultVec3, although a much better way would be to have ofPoint removed from everywhere and having this just for backwards compatibility. I can do all the changes from ofPoint to glm::vec3. should I do?

examples/events/allEventsExample/src/ofApp.cpp
roymacdonald: I made as less changes as possible with this. I see no problem on using `ofUTF8Append`, although I think that this example should be named otherwise and making use of `void ofApp::keyReleased( ofKeyEventArgs & key );` instead of `void ofApp::keyReleased( int key );`
whould make much more sense and it will introduce the use of the newer features of `ofKeyEventArgs`.

examples/events/allEventsExample/src/ofApp.cpp
roymacdonald: do you mean this for all the `eventString +=` in there or just for `eventString += (char)key;`?

libs/openFrameworks/types/ofPoint.h
arturoc: As far as i see though ofPoint is only used in the core for compatibility reasons in a couple of places in ofParameter but nowhere else? there's a coouple of coments still mentioning ofPoint if you want to change those that's fine but other than that i don't want to change any code except for fixing bugs at this moment. Can you rollback this change to ofVec3?

We can discuss what's the best way going forward after the release but right now we shouldn't be changing anything unless it's a bug or stops something else from working

roymacdonald: I just reverted it.
Right it is used only on ofParameter and several of comments, although it is still used on several examples.
best

arturoc: Yes examples and addons still have to be updated to use glm

roymacdonald: I'll go through those and update the PR.
I will make a different PR for the addons.

roymacdonald: @arturoc Just updated all the mac examples to use glm.

arturoc:
Let me know if this is ready to merge.

arturoc:
sorry just realized that you've ported some examples to use glm. what do you mean by the mac examples? it woud be weird to release some examples using ofVec and others glm. perhaps we can separate this PR in 2, one that fixes the bugs and one that changes the examples. I would merge the one that fixes right away and the other right after 0.10 that way we have time to fix the missing examples but having some examples randomly use ofVec and others glm sounds confusing

roymacdonald:
@arturoc mac examples should be osx. there are a few android and ios examples that have ofVec that I can fix too. I will make a different PR with the changes to glm. I was trying to make it easier, sorry if it became the opposite.

roymacdonald:
@arturoc now this pr just contains the bug fixes

arturoc:
thanks!

PR #5985, opened 2018-04-25
fix context WifiManager in ofAndroidLib

kashimAstro:
https://forum.openframeworks.cc/t/0-10-0-release-candidate-3/29348/45?u=kashim


arturoc:
thanks!

PR #5988, opened 2018-04-26
exclude tvOS examples from (linux) build script

edovino:


danoli3:
Oh nice, yeah whoops must have missed this :D

Thanks

PR #5992, opened 2018-05-01
modifier key bitmask fix.

bakercp:
`0x16` is not a helpful (unique) bitmask when the others are `0x1`, `0x2`, `0x4`, etc.  It causes various other keys to be marked as triggered when the `OF_KEY_SUPER` or `OF_KEY_COMMAND` key is pressed.  Also, make the `hasModifiers` const correct for lots of reasons.


arturoc:
The problem with 0x8 is that it's also the backspace. I think what i tried to do here is 0x10 (16 in decimal), but added 0x16 in hex. can you try that?

PR #6002, opened 2018-05-04
use gold linker in linux

arturoc:
makes link times much faster on big projects


arturoc:
it looks like ld.gold is faster though

with a relatively big project after compiling the full project and touching 1 file recompiling takes:

- ld: 28s
- lld: 14s
- ld.gold: 12s

hamoid:
Nice! Massive improvement :)

hamoid:
With existing projects, is it enough to pull the latest OF changes to benefit from this? Or do we need to use the project generator?

arturoc:
@hamoid, yes the qtcreator project has most of it's code in a common folder in the core no need to recreate the project

PR #6037, opened 2018-05-24
Update main.cpp

thejanzimmermann:
ofGLFWWindowSettings size param fixed


bakercp:
This should be added to the next patch release.  I'm @arturoc do all examples get compiled?

bakercp:
This should be merged as it doesn't compile otherwise.

Issue #6016, opened 2018-05-11
ofParameterGroup:remove() issue


eduardfrigola:
I think I found a bug in the `remove` function of `ofParameterGroup`.
When removing parameters of the group with `ofParameterGroup::remove(const string &name)`. The `parametersIndex` (the map that stores the position of each parameter in the `parameters` vector) is not updated when a parameter is removed.

I've tested this solution and works for me, don't know if is the best way to do it:
```cpp
void ofParameterGroup::remove(const string &name){
    if(!contains(name)){
       return;
    }
    int paramIndex = obj->parametersIndex[name];
    obj->parameters.erase(obj->parameters.begin() + obj->parametersIndex[name]);
    obj->parametersIndex.erase(name);
    std::for_each(obj->parametersIndex.begin(), obj->parametersIndex.end(), [&](pair<const string, size_t>& i){
       if(i.second > paramIndex){
           i.second--;
       }
    });
}
```

roymacdonald:
Hi, I think that it might be safer to rebuild the map and have a private function for that, although it can be slower than what you are suggesting.
something like 
```
void ofParameterGroup::rebuildParameterIndexMap(){
   obj->parametersIndex.clear();
   for(size_t i =0; i < obj->parameters.size(); i++){
       obj->parametersIndex[obj->parameters[i]->getName()] = i;
   }
}
```
probably @arturoc  or @bakercp can have a better insight.
cheers

eduardfrigola:
Yes, something like this seems better, and most of the times when using this function speed won't be crucial, but if you have a long list of parameters, and in the case you are only removing the last one, there is no need to rebuild the map because you know for sure the indexs won't change.
My thought was that I only want to edit the parameters that were after the parameter I've removed.
Also I think a map is safer enough to not use the clear method, or am I wrong?
```cpp
void ofParameterGroup::rebuildParameterIndexMap(int beginParameter){
   for(size_t i = beginParameter; i < obj->parameters.size(); i++){
       obj->parametersIndex[obj->parameters[i]->getName()] = i;
   }
}
```
And with beginParameter as 0, you have the same functionality as your function, besides the clear command.

Thanks!

roymacdonald:
Hi,
I see your point and agree with you.  
Maybe it is a good idea to place a Pull Request. can you?
best

eduardfrigola:
Newbie in Pull Requests but I think I can succeed in that.
[...]

PR #6043, opened 2018-05-26
Bugfix ofParameterGroup::remove

eduardfrigola:
Hi!
As discussed in #6016 this pull requests implements a possible solution for the problems described in the issue.


arturoc:
thanks, i fixed this and some other issue with remove in the last commit, fully regenerating the map didn't seem necesary and there was another issue that this didn't fix so i just implemented it in another way

eduardfrigola:
Hi Arturo!
Thanks for taking care of this, I've seen your commit and seems that you are not checking the escaped name of the parameter. When we discussed this in #6016 seemed the right thing to do. What do you think about that?

PR #6078, opened 2018-07-12
fix: 'CLAMP': identifier not found

wasawi:
Not sure if this has to merge into master or some other branch.


arturoc:
thanks, i've fixed it instead by using ofClamp since we've been speaking about removing those macros in the long term

Issue #6025, opened 2018-05-18
ofNode: wrong orientation shows up when it has parent with scaling


yusuketomoto:
Hi there,
I found that ofNode orientation is still buggy while I played with cameraParentingExample. To break it down, the strange behavior happens when an ofNode has parent, scale is applied to the parent, setGlobalOrientation is called. I confirmed that getting parent’s orientation in ofNode::setGlobalOrientation isn't correct on that condition. You can see it by just building example and press 'v' and then 'p' 3~4 times.

Alongside, I also found that ofNode::getGlobalOrientation can be implemented in more efficient way. (3x~ faster)

I can make PRs above. The first PR makes cameraParentingExample work as the same as of_0.9.8 does. Can you tell me what branch would be appropriate? Thanks.

arturoc:
can you send a PR for this? the default branch right now is patch-release that should be ok

thanks!

PR #6093, opened 2018-07-31
Bugfix ofnode orientation with scale

yusuketomoto:
The 1st one in https://github.com/openframeworks/openFrameworks/issues/6025.


arturoc:
thanks so much for adding tests!

PR #6100, opened 2018-08-05
fix `ofSerialize` constness in ofJson.h (for patch-release)

2bbb:
this is same as #6096.
so, i had mistake about target of PR...


2bbb:
sorry, i was noticed to mistake origin of branch.

PR #6104, opened 2018-08-09
Xcode 9.4 Default Travis

danoli3:
Keep Travis Using Latest Default Xcode (also Travis has updated most to latest Xcode).

https://blog.travis-ci.com/2018-07-19-xcode9-4-default-announce


arturoc:
wonder if we should just remove the version though since it seems that then travis defaults to the latest

PR #6107, opened 2018-08-10
remove line break from ofGetVersionInfo() end

dimitre:
remove line break from ofGetVersionInfo() so I can make this kind of comparsion
```c++
if (ofGetVersionInfo() == "0.10.1-stable") {
	largura += 4;
}
```
instead of
```c++
if (ofGetVersionInfo() == "0.10.1-stable\n") {
	largura += 4;
}

```


arturoc:
thanks!

Issue #6066, opened 2018-07-03
Wrong spacing of ofTrueTypeFont


hiroMTB:
ofTrueTypeFont seems like handling letterSpacing in wrong way.
This is problematic when we set big letterSpacing value like more than 2.0.
Here is comparison between oF(top half) and Adobe Illustrator(bottom half).
Please click an image below and check the difference.

![2018-07-03 at 9 09 34 pm](https://user-images.githubusercontent.com/193280/42239944-e99228ce-7f05-11e8-87a3-296541ffda38.png)

Here is test code of oF.
```
#include "ofApp.h"

void ofApp::setup(){
    ofBackground(255);
    ofTrueTypeFont::setGlobalDpi(72);
    font.load("frabk.ttf", 20, true, true);;
    font.setLetterSpacing(4);
}

void ofApp::draw(){
    ofSetColor(0);
    font.drawString("HELLO I AM a BIG LETTERSPACING STRING!!", 20, 250);
    font.drawString("Hello I am a big letterspacing string!!", 20, 330);
}
```

chriship:
Reading here: https://www.freetype.org/freetype2/docs/tutorial/step2.html 

It seems fine with monospaced fonts. I presume it's because they all have the same 'advance' value. Oddly, I don't see any mention of 'bearingX' being used in ofTrueTypeFont.cpp.

It looks to me like Illustrator is using the same 'bearingX' and/or 'advance' value for each character but when I analyse various fonts they have different values for these.

hiroMTB:
Nice idea providing compatibility!

hiroMTB:
thanks!

PR #6109, opened 2018-08-16
Fix ttf font bounding box calculations

### LINKED ISSUE #6106: ofTrueTypeFont::getStringBoundingBox width returning different than previous versions
using ofw 0.10.0 master (patch-release)
It seems to be missing exactly 4 pixels, independent of the string size. (comparing with 0.9.8 or 0.10.0 release)
screen shot 2018-08-10 at 1 23 06 pm
Thanks

XCode 9.2 / OS X 10.12

tgfrerer:
`letterSpacing` scales the default spacing between glyphs, as such it needs to default to `1.f`.

This means, letterSpacing of 1.0 means neutral letter spacing, glyph is placed in default position, respecting the glyph's advance value.

  * If letterSpacing is 0.75, this means that spacing between letters is reduced by 0.25 * the *whitespace width* of the current font.

  *If letterSpacing is 1.25, this means that an additional space of 0.25 * the *whitespace width* is added between letters.

+ Since proportional fonts have per-glyph advance values, we use a constant width as *whitespace width*.

+ additionally, per spec, we scale the default *whitespace width* by the font's property `spaceSize`

Note that to calculate the default whitespace width for a font we don't use the advance value for the literal whitespace (' ') glyph, but, as the docs in the comments specify, we take the whitespace width as the advance of the font's glyph 'p'.

Fixes #6106

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -1054,9 +1054,9 @@ void ofTrueTypeFont::iterateString(const string & str, float x, float y, bool vF
 			} else if (c == '\t') {
 				if (settings.direction == OF_TTF_LEFT_TO_RIGHT){
 					f(c, pos);
-					pos.x += (getGlyphProperties('p').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
+					pos.x += (getGlyphProperties(' ').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
 				} else{
-					pos.x += (getGlyphProperties('p').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
+					pos.x += (getGlyphProperties(' ').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
 					f(c, pos);
 				}
 				prevC = c;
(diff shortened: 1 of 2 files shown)


arturoc:
where does it say that we are using 'p'? that was a really old thing that was fixed years ago but perhaps there's some comment still mentioning that?

arturoc:
I think the problem is that i changed the algorithm to calculate a different letterSpacing in iterateString only thinking that it was used everywhere else but the bb calculation and getTexture do the calculation again so it needs to be changed there as well

My changes were to fix: https://github.com/openframeworks/openFrameworks/issues/6066

were a proportional increase in letter spacing introduced a really weird spacing. instead i'm using an additional space of whatever letterSpacing indicates which makes the spacing look nicer with bigger values

tgfrerer:
Yes, I have just tested the case described in #6106 using this PR and I'm seeing it still being fixed - here's a screenshot of @hiroMTB 's code running this PR:

screenshot from 2018-08-17 15-53-09

(Which looks very much like the Illustrator screenshot they included).

I got 'p' as the base width for whitespace width calculation from the documentation of the `getSpaceSize` method (https://github.com/openframeworks/openFrameworks/blob/patch-release/libs/openFrameworks/graphics/ofTrueTypeFont.h#L293) The documentation for `setSpaceSize` repeats it.

I believe we use the width of the glyph `p` since it reliably returns a value, and the width of the whitespace character (` `) of a font is sometimes given as 0...

arturoc:
Ok, but i think we used p years because we created a table of characters that didn't contained space cause it wasn't in the texture atlas but i fixed that at some point. i think it's safe to use space. I'll change the docs as well since that's not the case anymore. Can you change the calls to getAdvance('p') to getAdvance(' ') instead

tgfrerer:
Done =)

PR #6114, opened 2018-08-23
Bugfix VS build script

hiroMTB:
Two fixes of VS build scripts.
- devenv ignores %PLATFORM% parameter. It compiles against win32 even %PLATFORM% is x64.
- msbuild ignores %ACTION% parameter. It can only excute "build" action even %ACTION% is clean.

One feature
- msbuild has /maxcpucount option to enable parallel compile.

Excute msbuild /? command for detailed available option

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -3,9 +3,9 @@
 
 //-------------------------------
 #define OF_VERSION_MAJOR 0
-#define OF_VERSION_MINOR 10
+#define OF_VERSION_MINOR 11
 #define OF_VERSION_PATCH 0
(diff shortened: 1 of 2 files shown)


arturoc:
The problem is that the branch is started from master instead of patch-release so it has the differences between master and that branch which is right now the default while we work on patch releases

Since this only affects dev scripts you can just send the PR to master if you prefer instead of starting again from patch-release

hiroMTB:
Which is comfortable to maintain for you and all?

arturoc:
i really don't care, i can easily bring this commit later to any of the branches if needed

hiroMTB:
Sent new PR against master. https://github.com/openframeworks/openFrameworks/pull/6121

PR #6128, opened 2018-09-18
Issue when  saving jpg

cyrstem:
changed from auto pixelFormat = OF_PIXELS_BGRA;
to  auto pixelFormat = OF_PIXELS_RGBA;
don/t know if  ALPHA channel should stay but this fix  an issue saving images  with grabScreen

### DIFF
--- a/libs/openFrameworks/gl/ofGLRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLRenderer.cpp
@@ -1914,7 +1914,7 @@ void ofGLRenderer::saveScreen(int x, int y, int w, int h, ofPixels & pixels){
 		y = sh - y;
 		y -= h; // top, bottom issues
 	}
-	auto pixelFormat = OF_PIXELS_BGRA;
+	auto pixelFormat = OF_PIXELS_RGBA;
 	pixels.allocate(w, h, pixelFormat);
 	auto glFormat = ofGetGLFormat(pixels);
 


roymacdonald:
hi. can you provide some more details about the issues that you are having (visual artifacts, errors, logs, etc)?

cyrstem:
the Wrong color configuration were saved when using ofImage  like this `imgpreview.save("preview/screenshot"+ofToString(ofRandom(0,1000),0)+".jpg",OF_IMAGE_QUALITY_HIGH);jpeg`  tracked back to this post 
https://forum.openframeworks.cc/t/wrong-color-saved-when-using-ofimage-save-jpeg-after-using-ofimage-grabscreen-within-v0-10-0/30240

same issue  hope this helps

roymacdonald:
I've checked it on macos 10.12.6, the latest from openFrameworks master and xcode8.3.2 and this problem happens, but it is only when saving to jpegs, so this must be a problem with the library encoding to JPEG or something like that. 
I'll dig a bit a and see what I can find.
cheers

cyrstem:
Im on MacOSX 10.10.5  with Xcode  6.3.2 and OF10 hopes this gives your more input

arturoc:
It would be great to know what's going on with this, some tests would help knowing if this is the correct solution.

could you post how saving an image looks with and without this fix in jpeg, png and bmp?

also a similar fix is probably needed in the programmable renderer so if you could test that it would be really useful but we can test it afterwards too

cyrstem:
this is what happend then and in my new install on linux it happens to the capture is up  and the app running is the one down 
the  the screen shows white light on the object and the capture shows  other color 
Screenshot from 2019-04-03 18-42-43

PR #6131, opened 2018-09-21
Fix compilation on Xcode 10

kkirby:
The architecture i386 is no longer supported on Xcode 10, so I've removed it.
The framework `QuickTime` is no longer available on Xcode 10, so I've removed it.

I have no idea how this PR will work on previous versions of Xcode or if maintaining support for i386 is a priority. If we need to have support for QuickTime and/or i386 then I would guess that we we'll need to modify the project generator to have separate xcconfig files for x86_64 and i386.


ofZach:
thanks for looking into this !   fixing for Xcode 10 is pretty important as it's now the default in app store -- cc @ofTheo

arturoc:
I'm going to merge this since several people are complaining in the forum that they can't compile anymore

PR #6137, opened 2018-09-27
Add PUT request to ofURLFileLoader

chriship:
Add the ability to do a PUT request with ofHttpRequest


arturoc:
you've based this PR of off master and sent it to patch-release, that's why all this extra commits from other people appear in the PR. Since it's a new feature can you send it against master instead?

thanks!

chriship:
Oops! I've changed the base branch to master - does that work?

arturoc:
I'm not sure it's really necessary but we reset the method on every request so the same way we set GET and POST to 0 when doing PUT. PUT should be reset to 0 when doing POST and GET. can you add that?

danoli3 (approved):
Yes

danoli3:
Rebased / Merged and added @arturoc 's PUT to 0 change.

PR #6138, opened 2018-09-27
ofCamera with ortho enabled to respect a lensOffset

m1keall1son:
As is stands, setting an off-axis viewport or a lens offset on an ofCamera will only affect the projection matrix if the camera is _not_ ortho enabled.  I can't think of a case why one would want this behavior.

This PR makes a small change to ofCamera that shifts the ortho viewport by the lensOffset via matrix translation exactly how the perspective counterpart does it.

The use case I have is a multi-computer rendering of the same app in which i need a mix of perspective and orthographic views originating at the same eye point.  In perspective mode, setting the viewports with the `offAxisViewport` method is perfect, but it fails to do anything when I enable ortho.  This PR makes both modes behave in the same way.

If orthogonal and perspective matrices are to be generated by the same camera object (as opposed to two different types with different interfaces) it seems to me they should be affected by all their parameters in as similar a way as possible.

### DIFF
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -151,7 +151,7 @@ glm::mat4 ofCamera::getProjectionMatrix(const ofRectangle & viewport) const {
 	const_cast<ofCamera*>(this)->calcClipPlanes(viewport);
 
 	if(isOrtho) {
-		return glm::ortho(
+		return glm::translate(glm::mat4(), {-lensOffset.x, -lensOffset.y, 0.f}) * glm::ortho(
 			- viewport.width/2,
 			+ viewport.width/2,
 			- viewport.height/2,


roymacdonald:
Hi,
I understand your point. 
The lens offset in the perspective model makes sense because it is simulating how a real pinhole camera would work. Yet, the ortho camera is a completely artificial model, thus it has no such thing as a lens.
I am not sure if what you are proposing would be more confusing or more intuitive. 
If you have a camera in ortho mode cant you achieve the same thing simply by moving along the camera's local x and y angles ? What happens when the camera is rotated? does it still behave as expected?

arturoc:
looks good to me, if we allow to set an ortho camera through ofCamera all the methods should work in orthographic as well. if a method doesn't do anything or the model is so different we shouldn't allow to set a orthographic perspective through ofCamera at all. The rotation of the camera shouldn't be a problem as far as i see.

roymacdonald:
I was being a bit more strict about the theory of the camera model, but yes, from the usability point if view it make a lot of sense. Maybe updating the `set/getLensOffset(...)` method's inline documentation would be a good idea (it still even says that it uses a ofVec2f as an argument instead of the glm::vec2 it actually uses!) Stating that it will work on perspective and ortho mode would be nice too.
best

arturoc:
thanks!

PR #6140, opened 2018-09-30
add ofSystemConfirmDialog for confirming actions

s-ol:
This PR supersedes #3228 (ofSystemChoiceDialog).

I have another cleanup commit ready (b90df914359e7eccde1de5dbb600278d1bf77538) but didn't attach it to this branch so the diff is clean for now.

Status:

|                      | implemented | tested |
| ----------------- | ---| -----|
| Linux (GTK) | yes | yes |
| Windows      | yes | no |
| OS X            | yes | yes |
| Android        | no | - |

Please help with testing!

### DIFF
--- a/libs/openFrameworks/utils/ofSystemUtils.cpp
+++ b/libs/openFrameworks/utils/ofSystemUtils.cpp
@@ -156,6 +156,30 @@ gboolean file_dialog_gtk(gpointer userdata){
 	return G_SOURCE_REMOVE;
 }
 
+struct ConfirmDialogData{
+	string message;
+	bool yesno;
+	bool done;
+	bool result;
+	std::condition_variable condition;
+	std::mutex mutex;
+};
+
+gboolean confirm_dialog_gtk(gpointer userdata){
+	ConfirmDialogData * dialogData = (ConfirmDialogData*)userdata;
+	GtkWidget* dialog = gtk_message_dialog_new (nullptr, (GtkDialogFlags) 0, GTK_MESSAGE_INFO, dialogData->yesno ? GTK_BUTTONS_YES_NO : GTK_BUTTONS_OK_CANCEL, "%s", dialogData->message.c_str());
+	gtk_widget_grab_focus(gtk_dialog_get_widget_for_response(GTK_DIALOG(dialog), dialogData->yesno ? GTK_RESPONSE_YES : GTK_RESPONSE_OK));
+	gint response = gtk_dialog_run (GTK_DIALOG (dialog));
+	gtk_widget_destroy (dialog);
+	dialogData->mutex.lock();
+	dialogData->result = response == GTK_RESPONSE_OK || response == GTK_RESPONSE_YES;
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/utils/ofSystemUtils.cpp
```diff
@@ -601,6 +625,54 @@ LRESULT CALLBACK WndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
...
+		if (bYesNo) {
+			[alert addButtonWithTitle:@"Yes"];
+			[alert addButtonWithTitle:@"No"];
+		} else {
+			[alert addButtonWithTitle:@"OK"];
+			[alert addButtonWithTitle:@"Cancel"];
+		}
+		[alert setMessageText:[NSString stringWithCString:message.c_str()
+												 encoding:NSUTF8StringEncoding]];
+		NSInteger returnCode = [alert runModal];
```
roymacdonald: restoreAppWindowFocus();
needs to be added after this line so the ofApp regains focus.

roymacdonald:
I have just tested this in macos 10.12.6 and xcode 8.3.2 and it works. but the following warning is shown in the console:`
2018-10-02 12:12:02.858818-0300 ofSystemConfirmDialogDebug[48373:912566] [Layout] Detected missing constraints for <NSTextField: 0x100c18f90>.  It cannot be placed because there are not enough constraints to fully define the size and origin. Add the missing constraints, or set translatesAutoresizingMaskIntoConstraints=YES and constraints will be generated for you. If this view is laid out manually on macOS 10.12 and later, you may choose to not call [super layout] from your override. Set a breakpoint on DETECTED_MISSING_CONSTRAINTS to debug. This error will only be logged once.`
Which is also shown when using `ofSystemTextBoxDialog(...)`. By googling  found that it seems to be a problem related to the os version I am using. I will check out this a bit more. 

Also, the OF app window does not get back the focus once the button is pressed so you need to add 
`restoreAppWindowFocus(); ` after running modal. [Check here](https://github.com/openframeworks/openFrameworks/pull/6140/files#r221994381)

That's for now

s-ol:
@roymacdonald thank you for the feedback, I will fix the focus issue asap.

I don't think the warning you mentioned is related since the addition to the code doesn't instantiate an `NSTextField`.

roymacdonald:
Hi, right. Maybe the text that is displayed is inside a NSTextField but it seems to be a harmless warning and it is the same that shows when using `ofSystemTextBoxDialog`. So it works for macos, although I would test it on the newer macos versions.
cheers

s-ol:
CI build failures seem unrelated. Updated OP to reflect mac being tested.

@arturoc feedback on this?

arturoc:
Hey thanks for this, it seems really useful. Additions like this have to work in all platforms or at least in all desktop platforms. If you can take a look at windows we could merge it but meanwhile adding it would create a function that doesn't do anything in that platform

Issue #6139, opened 2018-09-30
mojave + xcode 10 issues


ofZach:
I'm looking at OF on Mojave and Xcode 10 which is kind of painful now -- 

some notes (cc @ofTheo) 

a) 32 bit needs to be removed (this has already been done in master...) 

b) on mojave apps need permission set as a plist for camera, microphone, and any other thing we might need to do that requires permissions (ala iPhone).  I feel like the easiest is we just add these to the template plist so that they are already set. 

c) there's something not right about the overall OF sub project, at the moment you can't open more than one OF project at the same time and build them (before you used to get an integrity warning but now it seems there is more of an error so you actually just can't build more than one project at once...  it's super weird) 

d) openframeworks library always gets rebuilt for each project :(  

(maybe looking at this issue which this reminded me of -- https://github.com/openframeworks/openFrameworks/issues/5895) 

I'm looking into b/c/d at the moment...

ofZach:
update -- if I put projects in "legacy" build system they compile fine (meaning, no redundant compiling of OF, can have more than one OF project open at once and compile it, etc) 

<img width="589" alt="screen shot 2018-09-30 at 4 54 29 pm" src="https://user-images.githubusercontent.com/142897/46262621-c65d1c80-c4d1-11e8-937f-75917f4f7c96.png">

ofZach:
also one issue I see with the main OF project in Xcode 10 is that it doesn't seem to build release lib for OF, I had to add another scheme for release (one scheme is for debug, the other for release)

ofZach:
can confirm also that Xcode 9.4.1 works well on mojave (no repeat compiling of OF per project, no issues with multiple projects open) but we still have issues with the plist / permissions so things like video grabber example doesn't work yet. 

also if you add permissions to the plist 

<img width="1158" alt="screen shot 2018-09-30 at 5 32 07 pm" src="https://user-images.githubusercontent.com/142897/46262991-46d24c00-c4d7-11e8-9656-87e9db5286b9.png">

it seems that this plist doesn't doesn't work when you run the app from within xcode (9.4.1) but it does when you run outside of Xcode 

<img width="944" alt="screen shot 2018-09-30 at 5 38 05 pm" src="https://user-images.githubusercontent.com/142897/46263023-aa5c7980-c4d7-11e8-8665-e195d6c846b0.png">

I think this behavior is better on Xcode 10 but I'm not sure...

arturoc:
[...]

PR #6142, opened 2018-10-03
fixes xcode template to not have OF as a dependcy but as a pre-build …

ofTheo:
…script. related to #6139


ofZach:
just checked this -- on xcode 10 (using new build system) it fails: 

"Command PhaseScriptExecution failed with a nonzero exit code"

full build errors here: 

https://pastebin.com/n0YaEZTE

I will look at xcodebuild a bit closer and try to see what's happening

ofZach:
oh sorry this was because I need to grab the libraries w/ the github repo...  my bad....

ofZach:
can confirm that xcode 10 works but I'm seeing some strange behavior w/ OF apps on xcode 10 with the new build system -- for some reason, objects constructors are not getting called.  for example when you do 

ofImage img;

or 

ofVideoGrabber grabber;

in the h file, the constructor isn't called (but other functions, when you call them, are called).  When you use the legacy build system everything works as expected.   

@ofTheo wondering if you can check this with xcode 9 and the new build system?    I'll keep poking around.

ofZach:
a little weirder (and maybe xcode 10 / new build related) if you change ofApp.h it doesn't force recompile of ofApp.cpp...   something is kind of janky, researching...

ofZach:
In the meantime, I'm going to look at adding a "project.xcworkspace" via project generator that forces legacy build system, which seems to behave normally 

screen shot 2018-10-06 at 9 32 29 am

need to check if this is ok on older xcodes etc

ofTheo:
Whoa, the constructor thing is super weird. I can check on 10.13. Are there different compiler / clang versions you can use in the build settings with the new build system?

Issue #6133, opened 2018-09-24
ofxGui scroll is active when it shouldn't be


samubence:
You can change the value of a parameter e.g ofxSlider using the mouse scroll... which is a great feature.
But it still happens when the group is minimised and the cursor is not over the element. (it is very weird that in this case the slider only goes up and much faster than is was maximised)
The same happens when we don't display the panel, scroll still active.
This is a minimal code to demonstrate the behaviour. 

I'm using of_0.10.0 on macOS, xcode 10.0

```
#include "ofMain.h"
#include "ofxGui.h"

class ofApp : public ofBaseApp
{
public:
    void setup()
    {
        gui.setDefaultWidth(400);
        gui.setup();
        gui.setPosition(0, 0);
        group.setName("GROUP");
        group.add(param.set("param", 0, 0, 255));
        gui.add(group);
        displayGui = true;
    };
    
    void draw()
    {
        ofBackground(param, 0, 0);
        
        if (displayGui)
        {
            gui.draw();
        }
    };
    
    void keyPressed(int key)
    {
        displayGui = !displayGui;
    }
    
    ofxPanel gui;
    ofParameterGroup group;
    ofParameter<float> param;
    bool displayGui;
};

int main( )
{
	ofSetupOpenGL(400,100,OF_WINDOW);
	ofRunApp(new ofApp());
}
```

roymacdonald:
I can see this happenig also. It is really annoying. It is related to the way that the scroll event is handled which is different to the rest of the mouse events. I am looking at this in order to fix it.
cheers

roymacdonald:
just made a PR with the fix. Thanks for pointing it out

PR #6144, opened 2018-10-04
Fix ofx gui slider scrolling

roymacdonald:
Fix for this issue https://github.com/openframeworks/openFrameworks/issues/6133

The scrolling was still being handled when the gui was not being drawn both in ofxSlider and ofxInputField. 

Tested using the code provided in the mentioned issue.

Also, 
there is some redundant code in ofxSlider.cpp and ofxInputField.cpp. Both have a function called `toRange(...)` which is identical in both. When the scroll callback is called this function  is called. Can't we move this function to the ofxBaseGui for instance? Also, the code used in the scroll callback in both ofxSlider and ofxInputField is quite similar, maybe moving this to ofxBaseGui might be better too.


roymacdonald:
anything on this?

arturoc:
thanks! going to merge it into patch-release too so it's in 0.10.1

PR #6150, opened 2018-10-11
Update customEGLWindowSettingsExample main.cpp

jvcleave:
example was no longer working. These changes worked for me but unsure if this is the proper way


arturoc:
thanks!

jvcleave:
@arturoc just realized that the example was using updated syntax from https://github.com/openframeworks/openFrameworks/commit/59fcc11ea0c08d6c673c109f81cfd4460104f47c

sorry for the confusion but this will probably break the new release

Issue #4945, opened 2016-02-24
ofSystemTextBoxDialog: Capturing "Cancel" with pre-populated values


sqrtof5:
When using ofSystemTextBoxDialog with a pre-populated input field, there is no way to know that the "Cancel" option has been chosen if the pre-populated value is a valid answer.

Typical use is something like:

```
string userInputText = ofSystemTextBoxDialog("Question?", "Pre-populated valid answer, not identifiable placeholder");
//check userInputText value
```

The issue is that in this simple case, `userInputText` will hold the pre-populated answer even if Cancel is selected (by implementation in ofSystemUtils.cpp).

I wonder if that could be addressed, maybe by returning an empty string?
E.g. (OSX):

```
text = ( returnCode == NSAlertFirstButtonReturn ) ? [[label stringValue] UTF8String] : "";
```

Or maybe there's another way to handle these (admittedly edge) cases?

ofTheo:
The empty string approach seems like a good way. 
It should probably only be the repopulated text if the person hits the okay button. 
 
We could also switch the return type to a custom type which contains the string and also the button that was hit, but this might be an overly complex way to handle it.

PR #6155, opened 2018-10-18
Make ofSystemTextBoxDialog return empty string on cancel.

ShadowMitia:
See issue #4945.

I've made changes for macOS as suggested in the thread and adapted it for Linux and Windows.

I've only been able to test on Linux right now, I believe it should work fine on macOS, but I'm really unsure on the windows side.

I haven't looked at the parts of the code that deal with android and emscripten.


arturoc:
thanks!

PR #6160, opened 2018-10-27
Revert "Update customEGLWindowSettingsExample main.cpp"

jvcleave:
just realized that the example was using updated syntax from 59fcc11


arturoc:
:) thanks!

Issue #6168, opened 2018-11-01
msys2 binary missing dependency


sphaero:
I’m using appveyor to build an OF binary for windows. This used to work fine however I’m now running into an issue where my final binary is not working due to missing libraw-16.dll.

the log clearly shows libraw (19) being installed through pacman:

```
downloading mingw-w64-i686-libraw-0.19.0-1-any.pkg.tar.xz...
```

I’ve echoed ntldd -R to show what the dependencies are. It depends on libraw-16.dll as a dependency of freeimage but it is not found:

```
    libfreeimage-3.dll => C:\msys64\mingw32\bin\libfreeimage-3.dll (0x00ce0000)
	 libHalf-2_2.dll => C:\msys64\mingw32\bin\libHalf-2_2.dll (0x00760000)
	 libIex-2_2.dll => C:\msys64\mingw32\bin\libIex-2_2.dll (0x000d0000)
	 libIlmImf-2_2.dll => C:\msys64\mingw32\bin\libIlmImf-2_2.dll (0x01750000)
	  libIlmThread-2_2.dll => C:\msys64\mingw32\bin\libIlmThread-2_2.dll (0x000d0000)
	  libImath-2_2.dll => C:\msys64\mingw32\bin\libImath-2_2.dll (0x00760000)
	 libjpeg-8.dll => C:\msys64\mingw32\bin\libjpeg-8.dll (0x00dc0000)
	 libjxrglue.dll => C:\msys64\mingw32\bin\libjxrglue.dll (0x00760000)
	  libjpegxr.dll => C:\msys64\mingw32\bin\libjpegxr.dll (0x00790000)
	 libopenjp2-7.dll => C:\msys64\mingw32\bin\libopenjp2-7.dll (0x00760000)
	 libraw-16.dll => not found
	 libtiff-5.dll => C:\msys64\mingw32\bin\libtiff-5.dll (0x00760000)
```
Also posted on the forum: https://forum.openframeworks.cc/t/appveyor-build-installs-libraw-0-19-but-binary-depends-on-0-16/30916

Possibly an upstream bug in msys2?

sphaero:
btw, I don't see this problem on my own windows build machine with msys2. It has mingw-w64-i686-freeimage Version 3.18.0-1. While on appveyor it has version  mingw-w64-i686-freeimage-3.17.0-4

arturoc:
are you running pacman -Su at the beginning to update the packages?

sphaero:
[...]

PR #6170, opened 2018-11-07
fix ssl dependencies

sphaero:
fixes: #6168 

This does not update msys2 runtime and pacman which could be needed if this fails


sphaero:
ok it still fails, let me try something else to see if we can find some more info

arturoc:
thanks so much!

PR #6174, opened 2018-11-11
fix for xcode template file to allow PG to add frameworks

### LINKED ISSUE #6172: XCode 9.2 can't open project after update 7b87f643f
XCode 9.2 can't open project after update 7b87f643f
Project xxx.xcodeproj cannot be opened because the project file cannot be parsed.

ofTheo:
fix for xcode template file to allow PG to add frameworks and fix broken projects. 
closes #6172

### DIFF
--- a/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj
+++ b/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj
@@ -202,7 +202,9 @@ fi
 				<key>COPY_PHASE_STRIP</key>
 				<string>YES</string>
 				<key>FRAMEWORK_SEARCH_PATHS</key>
-				<string>$(inherited)</string>
+				<array>
+                    <string>$(inherited)</string>
+                </array>
 				<key>GCC_GENERATE_DEBUGGING_SYMBOLS</key>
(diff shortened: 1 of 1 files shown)


roymacdonald:
Hi, now after this update PG fails to create working projects for xcode 8.3.2 that use frameworks as it is the case of ofxSyphon. In the past I made a fix that was properly adding the frameworks but now xcode simply complains that it can not parse the `project.pbxproj` file. Until which version of Xcode is going to be supported?

ofTheo:
hmm @roymacdonald that is the issue this PR should of fixed. 
could you try the nightly zips and see if that works for you?

ofTheo:
@roymacdonald let me know if you see the issue with: 
https://openframeworks.cc/ci_server/versions/nightly/of_v20181113_osx_nightly.zip

I just tried it on Xcode 9 which was having the same error you mentioned before with ofxSyphon and it works fine for me.

roymacdonald:
Hey @ofTheo I tried of_v20181113_osx_release and indeed it works.
Now I forced the pull from the master branch and it also works. Xcode.....
So, ignore all this.
all the best

PR #6176, opened 2018-11-12
support for allocating GL_TEXTURE_3D and GL_TEXTURE_2D_ARRAY

kasparsj:
added support for allocating GL_TEXTURE_3D and GL_TEXTURE_2D_ARRAY textures, to be able to pass into shader as sampler2DArray

covers a small portion of #2622

### DIFF
--- a/libs/openFrameworks/gl/ofTexture.cpp
+++ b/libs/openFrameworks/gl/ofTexture.cpp
@@ -414,47 +414,47 @@ void ofTexture::allocate(const ofTextureData & textureData){
 
 void ofTexture::allocate(const ofTextureData & textureData, int glFormat, int pixelType){
 #ifndef TARGET_OPENGLES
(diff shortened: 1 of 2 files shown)


arturoc:
This looks great, the only problem i see is that there's some things that won't work on OpenGL ES. I've pushed the latest changes from master to your branch since there where some problems that were making some tests fail.  Once the tests finish you can see there which platforms are failing. I can imagine the errors will be similar and once you fix one it'll fix everything else.

IT's probably just a matter of adding a couple of `#ifdef TARGET_OPENGLES` in the new functionality

kasparsj:
Thanks @arturoc. Should be fixed now. I think the failing tests are the same, that are failing on the master branch.

arturoc:
Looks good! I'll wait until we merge the fixes for android to test this there as well and then merge it. Thanks!

Jonathhhan:
Texture3D in core OF would be great :) And I guess it works with GL ES 3...

PR #6179, opened 2018-11-15
ofxGui scaled for retina displays + group header enable/disable feature 

roymacdonald:
I've added a feature to ofxGuiGroup to disable its header. I've found this feature quite useful to unclog the gui and for making dropdowns.
This feature does not alter the normal functioning of ofxGui, just adds a harmless new feature.
Hope this can get merged right away.
cheers

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -55,10 +55,10 @@ void ofxGuiSetDefaultEventsPriority(ofEventOrder eventsPriority){
 	ofxBaseGui::setDefaultEventsPriority(eventsPriority);
 }
 void ofxGuiEnableHiResDisplay(){
-	ofxBaseGui::enableRetina();
+	ofxBaseGui::enableHiDpi();
 }
 void ofxGuiDisableHiResDisplay(){
-	ofxBaseGui::disableRetina();
+	ofxBaseGui::disableHiDpi();
 }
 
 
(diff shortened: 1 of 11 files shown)


roymacdonald:
I just added to this commit an option for scaling the whole gui when using a hi DPI display like the macbook pro's retina display. When it is enabled the gui looks a lot better. I had to add a few changes so the text is rendered properly and scaled accordingly.
Additionaly I made some slight changes on how the mouse callbacks are handled in ofxGuiGroup, so now if the group is not being drawn nothing is passed to its children elements.
Also added a fix that avoids a nasty bug when having more than one ofxPanel being rendered, which used to jump under certain circumstances.

Hope this can get merged. It is not really disruptive but adds and fixes some useful things.

arturoc:
@roymacdonald this is not passing the tests in several platforms also can you remove any references to retina? it's very platform specific, perhaps change it with sethidpi or allow to set a dpi factor or something?

roymacdonald:
Hi @arturoc 
Ok. I found the problem. it was a mousepressed that I added that was not returning. I fixed that.

I renamed retina for hidpi.

There are no platform dependent calls in what I made. it simply scales everything and it also scales the textmesh accordingly. There are hidpi screens in all platforms so it can be useful on all.

There's also the following function I made to remove a lot of redundant and hard coded maths there were. Just in case you wonder
https://github.com/roymacdonald/openFrameworks/blob/feature-ofxGuiAdditions/addons/ofxGui/src/ofxBaseGui.cpp#L204-L215

roymacdonald:
@arturoc now it is passing the tests. Have you been able to try out these changes/improvements I made?
Cheers

PR #6181, opened 2018-11-16
Bugfix sound buffer resampler

dfeeuh:
Second attempt at this, with the correct branch and hopefully useful comments.
Hi all
Last year I worked with @roymacdonald Roy McDonald's ofSoundObjects and discovered a problem when running his example-soundPlayerObject:
roymacdonald/ofxSoundObjects#20

The example loops an audio file and has a speed control. I generate a short stereo tone wave file with a different frequency for each channel but an exact number of wavelengths. When scrubbing the speed there were lots of clicks and sometimes it would crash.
I eventually found this issue here and managed to fix it. There are still problems for very short files (<512 samples) but this is much more robust.
Would love to see this in as I'm looking forward to getting involved.
David Fee

### DIFF
--- a/libs/openFrameworks/sound/ofSoundBuffer.cpp
+++ b/libs/openFrameworks/sound/ofSoundBuffer.cpp
@@ -336,6 +336,7 @@ void ofSoundBuffer::linearResampleTo(ofSoundBuffer &outBuffer, std::size_t fromF
 		return;
 	}
 	
+    // Define the last output sample in terms of frames
 	std::size_t end = fromFrame+numFrames*speed;
 	double position, remainder;
 	std::size_t intPosition, n = 0;
@@ -343,17 +344,23 @@ void ofSoundBuffer::linearResampleTo(ofSoundBuffer &outBuffer, std::size_t fromF
 	std::size_t copySize = inChannels*sizeof(float);
 	std::size_t to, idx;
 	
+    // Since end is defined in terms of frames 
 	if(end<inFrames - 1){
+        // end point is within the buffer
 		to = numFrames;
 	}else{
+        // end point is beyond the end of the buffer
 		to = ceil(float(inFrames-1-fromFrame)/speed) - 1;
 	}
 		
+    // Iterate through each channel. 
+    // This is much faster than iterating through each channel for every frame (sample).
 	for (std::size_t j=0; j<inChannels; j++)
 	{
 		position = fromFrame;
 		n = j;
 		
+        // Iterate through each frame
 		while (n < inChannels*to)
 		{
 			intPosition = position;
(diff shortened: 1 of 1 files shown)


dfeeuh:
[audiowavs.zip](https://github.com/openframeworks/openFrameworks/files/2590293/audiowavs.zip)
Handy test files. Use with the example-soundPlayerObject and listen for the clicks on the original code.

dfeeuh:
@roymacdonald any idea how long these pull requests take?
David

roymacdonald:
hi @dfeeuh 
Patience! ;)
It can take a while. specially a PR as yours that has a lot of changes and needs to be thoroughly reviewed. 
I think that the best idea is to use an external library to handle audio sample rate conversion. There are a few around that I need to test.

arturoc:
This looks good to me. There seems to be a merge conflict now, can you take a look and fix it?

Adding a test that shows the problem (failed before, works now) in the tests folder would be really useful. It should be a test that numerically tests the algorithm not that plays a sound since this is run on a ci server.

If it's too hard to test this no worries just fixing the merge conflict should be fine

dfeeuh:
Hi @arturoc, thanks for the feedback. 
I haven't had a chance to look at this in so long that I've forgotten what I was doing. Anyway, I'd like to try and resolve the conflicts and build the test. My first step in building a test is just to do a visual plot of buffers and inspection, then I'll move to doing it numerically. I was doing this using @roymacdonald 's sound player example but I'd like to avoid using any more addons.
Here's my question - I'm trying to plot the output buffer by registering an ofSoundStream audioOut callback and copy the contents to an intermediate to plot, but the output buffers are always 0. Am I missing something about how this should work?
David

roymacdonald:
Hi, the audioOut call back is expecting you to pass to it something, it is the data that is going to be sent to the speakers. If you want to get the input, use the audioIn callback instead, in which the passed buffer has the audio data from the sound interface input (microphone, line in, etc)

dfeeuh:
Ah, sorry let me clarify: I'm playing from file so it is the final output buffer that I do need. I remember you had some way of doing this in your audio objects examples. AudioIn doesn't work as far as I can tell.
David

roymacdonald:
@dfeeuh 
The sound objects use a pull through architechure, which means that it uses the audioOut callback as a starting point and if there is an attached (connected) object as input it will call its audioOut function, and so on until it reaches the first (source) object of the chain. 
In any case, you shouldn't need ofxSoundObjects to test this update. 
What do you mean by "AudioIn doesn't work"? 
From my side I would prefer to have some external library dealing with resampling as there are a lot of different algorithms for doing it, which I think there is no reason for trying to re code.

roymacdonald:
Hey @dfeeuh  Thanks for this update. nice to have those tests. I've run them, all are passing and it looks good. 
Although, can you update your code so the rest of the class matches the current master?

I've been going through the resampling issue and I think that an external library might be a better choice, as there are a lot of algorithms for different qualities and probably optimized performance.

I've tried [libsamplerate](bugfix-soundBuffer-resampler) and it works nicely. 

Have you benchmarked somehow your updates angainst the previous implementation or other libraries?

dfeeuh:
Thanks Roy
I'm actually now uncovering a few more bugs so will still tinker with those. I haven't had a look at any external resample libraries. For some reason I assumed that openframeworks would maintain its own. Worth looking into then.
By the way, is there an easy way to bring my fork up-to-date with Master? I'm sure I could look it up.
David

roymacdonald:
Hi, I see. How are you testing your changes? do you have an app for that?
You can merge your branch with the master, so assuming you use git in the terminal it would be something like, if you are currently on your branch:
Assuming that your local master branch is OF's master and it is up to date and that you are currently on your branch just call `git merge master` which will probably complain because of the merge conflicts and it will modify the conflicting files with the parts that are making problems. You need to choose what you leave there, save the file and finish the merge. The CLI somehow guides you.

dfeeuh:
All done (I think), though I did end up doing a rewrite. This may defeat the purpose but see what you think...
I've got a small app that plays and displays a waveform for resampling. Attached. Might be useful. 
[testAudioResampler.zip](https://github.com/openframeworks/openFrameworks/files/3806090/testAudioResampler.zip)

PR #6189, opened 2018-11-25
[RFC] Add scripts to provide debian packages

madprog:
These packages are created by the scripts:
- openframeworks-dev
- openframeworks-vectorgraphics-dev

Instructions to build these debian packages:
$ cd scripts/linux
$ ./buildDebianPackages.sh

This script shows how to use the installed libraries to build an example project:
```
#!/bin/bash
set -xe
dpkg -i openframeworks-dev_0.10.0_amd64.deb
dpkg -i openframeworks-vectorgraphics-dev_0.10.0_amd64.deb

curl -O https://raw.githubusercontent.com/openframeworks/openFrameworks/master/examples/graphics/vectorGraphicsExample/src/main.cpp
curl -O https://raw.githubusercontent.com/openframeworks/openFrameworks/master/examples/graphics/vectorGraphicsExample/src/ofApp.h
curl -O https://raw.githubusercontent.com/openframeworks/openFrameworks/master/examples/graphics/vectorGraphicsExample/src/ofApp.cpp

cat > Makefile <<EOF
CXXFLAGS=\\
         \`pkg-config --cflags openframeworks\` \\
         \`pkg-config --cflags openframeworks-vectorgraphics\`

LDFLAGS=\\
         \`pkg-config --libs openframeworks\` \\
         \`pkg-config --libs openframeworks-vectorgraphics\`

vectorGraphicsExample: main.o ofApp.o
	\$(CXX) -o \$@ \$^ \$(LDFLAGS)
EOF

make vectorGraphicsExample
```

This is my first try to write a debian package, so I would be grateful for any advice if I was breaking any guideline or good practice.

The next steps for this PR would be:
- adding other packages for the other addons
- adding support for shared objects
- adding scripts for non-dev packages (shared object binaries only, that could be used as dependencies for other projects)
- splitting the `libopenFrameworks.so` into several `libopenFrameworks-3d.so`, `libopenFrameworks-graphics.so`, etc. (so that users can depend on smaller dependencies when they only need a subpart of openFrameworks)


arturoc:
This is really useful and something that we've been looking into for a while but never managed to make fully work.

An issue i see

- External libraries see, to be included in the packages. As far as i know that is not very good practice. Some of this libraries are already in the debian package repositories in most recent versions, like glm or poco so they should be used from there. For libraries that are not in the repos we could have packages for each library but including them in the OF package or addons packages doesn't seem very correct.

madprog:
@arturoc Thank you for your feedback! Yes, you are right, the external libraries should definitively be tagged as dependencies, and yes, we should create packages for the dependencies missing from the standard repos.
However, as we would take the ones available in the Debian's repositories, do you know if a different version of those libraries would still work with OF?

arturoc:
There's three cases:

- In the case of libraries we use from repos they should be compatible no matter the version. 
- in the case of libraries that we include but are also in the repos, it's usually cause the ones in the repos can be too old to be compatible so we include them ourselves to avoid problems. In those cases though recent distros usually have compatible versions already, like is the case with poco or glm but we would need to analyze case by case. I think apart from those two there's nothing  else that we use and is in the debian repos though. In any case if any library needed to be included we would need a separate package not to include it's files with the corresponding OF package
- In the case of libraries that we include but are not on the repos we should create source packages for those libraries as well not directly include the files with OF and much less include the binaries as happens here in some packages

PR #6194, opened 2018-12-14
Add useful ofParameter constructor

hiroMTB:
Right now we are able to initialize ofParameter like below in ofApp.h.
```
ofParameter<int>  fps{"FPS", 0, 0, 120};
ofParameter<bool> bFullscreen{"Fullscreen", false};
ofParameterGroup  grp{"general settings", fps, bFullscreen};
```
The first parameter (e.g. "FPS") is the name of parameter then initial value, min and max.
This is very useful because we can write in header file and reduce code in ofApp.cpp file.
I'm using this often in my project since I found this syntax in Entropy project(for example [here](https://github.com/Entropy/Entropy/blob/master/Projects/EntropyRender/src/entropy/render/PostEffects.h)), 

It would be nicer if we could initialize "serializable" variable as well in same manner. With this PR we can write this way and reduce .setSerializable(false) in ofApp.cpp to avoid saving parameter to xml (or json).

```
ofParameter<int>  fps{"FPS", 0, 0, 120, false};
ofParameter<bool> bFullscreen{"Fullscreen", false, false};
ofParameterGroup  grp{"general settings", fps, bFullscreen};
```

*Same to ofReadOnlyParameter


hiroMTB:
This PR is closed because of wrong target branch, please refer to this one instead https://github.com/openframeworks/openFrameworks/pull/6195

PR #6195, opened 2018-12-14
Add useful ofParameter constructor

hiroMTB:
Right now we are able to initialize ofParameter like below in ofApp.h.
```
ofParameter<int>  fps{"FPS", 0, 0, 120};
ofParameter<bool> bFullscreen{"Fullscreen", false};
ofParameterGroup  grp{"general settings", fps, bFullscreen};
```
The first parameter (e.g. "FPS") is the name of parameter then initial value, min and max.
This is very useful because we can write in header file and reduce code in ofApp.cpp file.
I'm using this often in my project since I found this syntax in Entropy project(for example [here](https://github.com/Entropy/Entropy/blob/master/Projects/EntropyRender/src/entropy/render/PostEffects.h)), 

It would be nicer if we could initialize "serializable" variable as well in same manner. With this PR we can write this way and reduce .setSerializable(false) in ofApp.cpp to avoid saving parameter to xml (or json).

```
ofParameter<int>  fps{"FPS", 0, 0, 120, false};
ofParameter<bool> bFullscreen{"Fullscreen", false, false};
ofParameterGroup  grp{"general settings", fps, bFullscreen};
```

*Same to ofReadOnlyParameter

### DIFF
--- a/libs/openFrameworks/types/ofParameter.h
+++ b/libs/openFrameworks/types/ofParameter.h
@@ -484,6 +484,7 @@ public:
 	ofParameter(const ParameterType & v);
 	ofParameter(const std::string& name, const ParameterType & v);
 	ofParameter(const std::string& name, const ParameterType & v, const ParameterType & min, const ParameterType & max);
+    ofParameter(const std::string& name, const ParameterType & v, const ParameterType & min, const ParameterType & max, bool serializable);
 
 	const ParameterType & get() const;
 	const ParameterType * operator->() const;
@@ -616,6 +617,14 @@ private:
 		,bInNotify(false)
 		,serializable(true){}
 
+        Value(std::string name, ParameterType v, ParameterType min, ParameterType max, bool serializable)
+        :name(name)
+        ,value(v)
+        ,min(min)
+        ,max(max)
+        ,bInNotify(false)
+        ,serializable(serializable){}
+        
 		std::string name;
 		ParameterType value;
 		ParameterType min, max;
@@ -661,6 +670,10 @@ ofParameter<ParameterType>::ofParameter(const std::string& name, const Parameter
 :obj(std::make_shared<Value>(name, v, min, max))
 ,setMethod(std::bind(&ofParameter<ParameterType>::eventsSetValue, this, std::placeholders::_1)){}
 
+template<typename ParameterType>
+ofParameter<ParameterType>::ofParameter(const std::string& name, const ParameterType & v, const ParameterType & min, const ParameterType & max, bool serializable)
+:obj(std::make_shared<Value>(name, v, min, max, serializable))
+,setMethod(std::bind(&ofParameter<ParameterType>::eventsSetValue, this, std::placeholders::_1)){}
 
 template<typename ParameterType>
 inline ofParameter<ParameterType> & ofParameter<ParameterType>::operator=(const ofParameter<ParameterType> & v){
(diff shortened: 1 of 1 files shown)


arturoc:
The problem with this is that it adds a bool parameter to a constructor which already has a lot of parameters. bool parameters are really hard to understand since a true/false in the 4th position of the call could be anything

Perhaps we could add an enum with OF_SERIALIZABLE / OF_NOT_SERIALIZABLE values or something similar that replaces that parameter so your example would look something like;

```cpp
ofParameter<int>  fps{"FPS", 0, 0, 120, OF_NOT_SERIALIZABLE};
ofParameter<bool> bFullscreen{"Fullscreen", false, OF_NOT_SERIALIZABLE};
ofParameterGroup  grp{"general settings", fps, bFullscreen};
```

which is much clearer

hiroMTB:
Totally agree. I will work on it.
Shall we also consider initializing bInNotify as well?

arturoc:
what is binNotify?

hiroMTB:
Or we could also have ofPrameterSettings as a struct? But it could be almost same with existing  Value class

hiroMTB:
bInNotify is another ofParameter variable to be set false in default. I guess switch on/off calling ofNotifyEvent when variable changed. Will check later when I'm back on my desk.

arturoc:
oh i see, perhaps we can leave that one out or perhaps we could have binary flags like:

```cpp
enum ofParameterFlags{
    OF_PARAM_NOT_SERIALIZABLE = 1 << 0,
    OF_PARAM_DONT_NOTIFY          = 1 << 1,
};


then you can do:
```
ofParameter<bool> bFullscreen{"Fullscreen", false, OF_PARAM_NOT_SERIALIZABLE | OF_PARAM_DONT_NOTIFY};
```

**Comment by eduardfrigola** [comment]
Hi!

I like the idea to have more information inside ofParameter, and to be easy to modify it. I agree with @hiroMTB about having other types than boolean it's needed. Also @arturoc did use also a fourth argument in here: https://github.com/openframeworks/openFrameworks/pull/5598 and that conflicts with it.

I think that going to a more descriptive approach will be simpler, and less confusing. For example what memo did with [ofxMSAControlFreak](https://github.com/memo/ofxMSAControlFreak/blob/master/src/ofxMSAControlFreakTutorial.h)
the only thing that has to be changed is that the "set" methods have to return the parameter itself, useful also when setting the parameter inside a group.

```cpp
ofParameterGroup parameters;
ofParameter<float> floatParam;
// Proposed method
parameters->add(floatParam("Float", 0, 0, 1, false));

// New method
parameters->add(floatParam("Float", 0, 0, 1).setSerializable(false));
```

Also if setSerializable (and all other setters) returned the parameter itself (don't know if that is a cpp aberration), you could do:
```cpp
parameters->add(floatParam("Float", 0, 0, 1).setSerializable(false).setNotify(false).setStep(1.5).setScale(ofParameterScale::Logarithmic));
```


And about future development of ofParameter... I don't know if it's the best place to discus it, but I would like to be able to, for example, declare a simple class, that inherits from ofAbstractParameterFlags or ofAbstractParameterInfo, with custom arguments, and can be stored inside the ofParameter itself. This way you can have the associated information about ofParameter inside the ofParameterGroup, and can benefit from passing only the parameterGroup for GUI related stuff. OfParameter will keeps simple and extendable.

Eduard.

PR #6196, opened 2018-12-14
Fix for OF version number in download libs script

NickHardeman:


scripts/dev/download_libs.sh
```diff
@@ -12,7 +12,7 @@ cat << EOF
 
     Options:
 
-    -v, --version VERSION       OF version to download the libraries for. Defaults to master
+    -v, --version VERSION       OF version to download the libraries for. Defaults to current release. (i.e. 0.10.0)
```
arturoc: thanks!  can you change this back to master? that's really what it defaults too unless you specify a version

ofTheo:
Closing as I think this is already in ( file changes show no difference ).

PR #6201, opened 2018-12-24
Bugfix ofx tcp manager

chriship:
On Windows, when a TCP client disconnects from a TCP server and then reconnects, the `ioctlsocket` method still needs to be called. If it is not then blocking is automatically set to true.

To test:

1. Change code to print out the frame rate of the networkTcpClientExample project. 
2. Run networkTcpServerExample app and networkTcpClientExample app. 
3. Quit networkTcpServerExample app
4. Run networkTcpServerExample app again

You will notice the framerate of networkTcpClientExample drops to ~9fps. This is because the socket has been set with blocking = true and the server is sending messages every ~100ms.

If blocking is set as false then currently the SetNonBlocking method will return before `ioctlsocket` is called and the socket will be set with blocking on.


arturoc:
thanks!

PR #6203, opened 2018-12-25
Fix ofx gui events propagation

roymacdonald:
fixed `ofxGuiGroup` events propagation.
When using an `ofEasyCam` and an `ofxPanel` or `ofxGuiGroup` object,if moving the camera by dragging the mouse and the mouse was released over the gui, then when trying to change something in the gui would also change the `ofEasyCam` object. 
This issue got reported [here](https://forum.openframeworks.cc/t/clicks-on-ofxgui-panel-params-pass-through-to-main-window/31368/3) 

Additionally, I changed all the instances of `ofPoint` and `ofVec*f` into glm.

### DIFF
--- a/addons/ofxGui/src/ofxGuiGroup.cpp
+++ b/addons/ofxGui/src/ofxGuiGroup.cpp
@@ -233,8 +233,6 @@ void ofxGuiGroup::add(ofParameter <ofFloatColor> & parameter){
 void ofxGuiGroup::clear(){
 	collection.clear();
 	parameters.clear();
-//	there is no need for the following line as sizeChangedCB will handle it anyways, and change b.height.
-//	b.height = (bHeaderEnabled?headerRect.height:0) + spacing + spacingNextElement;
 	sizeChangedCB();
 }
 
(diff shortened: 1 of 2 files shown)


arturoc:
can you separate this in two PR's is really difficult to see what belongs to solving the issue and what to porting to glm. there's even some color setting that doesn't seem related to either of those.

roymacdonald:
@arturoc sure, but that's why I made 2 commits in this PR. Just thought it would be easier to review. If    you look at the commits you can see what's done in each case.
The color thing was just that ofxPanel was not drawing it's background. I should have added that in a different PR.
I will update this PR asap.
cheers

roymacdonald:
ok. so I made this into a single commit, only with the changes made for the gui events propagation

roymacdonald:
This is the other PR https://github.com/openframeworks/openFrameworks/pull/6207

roymacdonald:
Hi, the CI gave some errrors which seem to be unrelated to my PR. Can these be re run? 
cheers

addons/ofxGui/src/ofxGuiGroup.cpp
```diff
@@ -233,8 +233,8 @@ void ofxGuiGroup::add(ofParameter <ofFloatColor> & parameter){
 void ofxGuiGroup::clear(){
 	collection.clear();
 	parameters.clear();
-	
-	b.height = (bHeaderEnabled?headerRect.height:0) + spacing + spacingNextElement;
+//	there is no need for the following line as sizeChangedCB will handle it anyways, and change b.height.
```
arturoc: can you just remove this comment?

roymacdonald: Removed.
cheers

arturoc:
thanks!

PR #6210, opened 2019-01-09
Fixes multitouch confusion on iOS

### LINKED ISSUE #6052: 0.10 iOS multitouch doesn't work
touch.id is always 0, and touch.numTouches is always 1

sourya-sen:
Multitouch was never broken on iOS - it had to be enabled from the main.mm file. What made it confusing was that `settings.enableMultiTouch` was not present in the template to begin with, so just added it! Makes life slightly easier :)

### DIFF
--- a/scripts/templates/ios/src/main.mm
+++ b/scripts/templates/ios/src/main.mm
@@ -1,7 +1,7 @@
 #include "ofApp.h"
 
 int main() {
-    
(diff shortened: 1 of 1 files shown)


sourya-sen:
Fixes #6052

arturoc:
thanks!

Issue #6214, opened 2019-01-15
Detect wired & bluetooth keyboard/mouse attached/detached in RPi


namjeongho:
OF cannot detect keyboard or mouse attached/detached after setup process in RPI.
It detect them only once in setup process at program starting.
So if a usb or bluetooth keyboard/mouse is attached after setup process, they cannot detected.

PR #6215, opened 2019-01-16
ofAppEGLWindow: Add control of udev input device attached/detached

namjeongho:
#6214 
Platform : Raspberry Pi
Change class: ofAppEGLWindow
- Control udev input device attached/detached
- Process input event from all connected device at any time during execution
- Unfiy keyboard and mouse event handler. Because bluetooth keyboard with touchpad can be assign to a single dev file like "/dev/input/event0" and both keyboard and mouse event occur from it.


arturoc:
thanks!

PR #6217, opened 2019-01-19
fix buffer overflow in linear interpolation of audio buffer with loop

shenberg:
write pointer is incremented twice, removed the unnecessary increment


arturoc:
thanks!

Issue #6221, opened 2019-01-31
phong.frag: v_transformedNormal and eye should be re-normalized


andylomas:
There appears to be shading bugs caused by the v_transformedNormal and eye not being re-normalized in libs/openFrameworks/gl/shaders/phong.frag

In phong.frag:

1. The raw interpollated value for v_transformedNormal is handed to the light shaders. I think the normal should be re-normalized to make sure it is truly a normalized value. Adding a line to re-nomalize it before the light loop appears to significantly reduce shading artefacts, particularly in specular highlights, if you’ve got vertex normals that are at quite different angles to each other (e.g. with a low poly count sphere).

2. In the light calculations the calculation of the halfVector between the light ray and eye ray appear to use a normalized version of the light ray direction but the eye direction appears to be un-normalized. This appears to be causing some effects such as biasing the halfRay, and hence the position of specular highlight, towards the eye-ray direction if the eye is a long way from the surface.

The suggestion is to normalize v_transformedNormal before it is handed to the lights, as well as normalizing the eye vector in the functions for the point and spot light types.

arturoc:
Thanks!  will take a look when i have a moment

PR #6225, opened 2019-02-01
phong.frag: v_transformedNormal and eye should be re-normalized (#6221)

andylomas:
Proposed bugfix for: #6221 phong.frag: v_transformedNormal and eye should be re-normalized

Changes to libs/openFrameworks/gl/shaders/phong.frag

line 68, function pointLight(...): vec3 eye normalized so that that eye direction is a normalized vector. This means that calculation of the halfVector (in line 82) should now be correct. Previously 'VP' was normalized but 'eye' was un-normalized, which resulted in incorrect calculation of the halfVector, and hence incorrect placement of specular highlights.

line 127, function spotLight(...): vec3 eye normalized so that that eye direction is a normalized vector. This means that calculation of the halfVector (in line 142) should now be correct. Previously 'VP' was normalized but 'eye' was un-normalized, which resulted in incorrect calculation of the halfVector, and hence incorrect placement of specular highlights.

lines 222-233, function main(void): vec3 transformedNormal calculated by normalizing the value of attribute v_transformedNormal. Interpollation of attributes during rasterization means that v_transformedNormal is not necessarily a unit vector, which can cause significant shading artefacts. Using the normalized value (transformedNormal instead of v_transformedNormal) for the light shaders appears to significantly improve the quality of specular highlight when there is smooth interpollation of the normals.

Code tested on Windows 10.

### DIFF
--- a/libs/openFrameworks/gl/shaders/phong.frag
+++ b/libs/openFrameworks/gl/shaders/phong.frag
@@ -65,7 +65,7 @@ static const string fragmentShader = R"(
         float d;            // distance from surface to light source
         vec3  VP;           // direction from surface to light position
         vec3  halfVector;   // direction of maximum highlights
-        vec3 eye = -v_eyePosition;
+        vec3 eye = -normalize(v_eyePosition);
 
         // Compute vector from surface to light position
         VP = vec3 (light.position.xyz) - ecPosition3;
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

PR #6239, opened 2019-02-08
AVFoundationPlayer - set speed, frame/position when beeing ready to play

i-n-g-o:
at the moment AVFoundation player does not allow setting frame/position and speed before the player is ready. isReady is handled internally only - users do not have a way to know when the movie is ready to play.

proposing these changes so frame/position and speed can be set before the video is ready to play.


i-n-g-o:
pinging @openframeworks/video

ofTheo:
This looks good to me. 
Thanks @i-n-g-o 

Merging! :)

PR #6244, opened 2019-02-23
bugfix #2903 Addon data copy

### LINKED ISSUE #2903: ADDON_DATA in addon_config.mk has no effect at all
I'm implementing a small framework within oF, for creating, displaying, and handling visual effects. I want people to be able to develop their own effects to plug into the system, so therefore I want each of them to be contained within a single folder.

To make this happen, I need to be able to copy data (shaders, fonts, etc) from the relevant folders into the bin/data folder of projects that depend on my addon. I set `ADDON_DATA = roses` in my addon_config.mk to accomplish this. It should cause something like `cp -rf $(PROJECT_ADDONS_DATA) bin/data` to be executed towards the end of the make process. (See lines 136, 154, 339 in compile.project.mk. However, when I get to those lines, `$(PROJECT_ADDONS_DATA)` is never defined. (I have checked by attempting to echo it).

It is defined (successfully) on line 113 in config.addons.mk - I have made sure of this by adding these lines towards the end of the file:

```
$(info ---PROJECT_ADDONS_DATA---)
$(foreach v, $(PROJECT_ADDONS_DATA),$(info $(v)))
```

Any help in figuring out, through this jungle of makefiles, why the value isn't defined where it is expected to be, would be greatly helpful.

Currently I'm suspecting that recursive calls to make don't affect each others' environments, and I'm not sure how to solve this. It' surprising to me that a bug like this has never been encountered before (at least it doesn't seem like it from all my searching.)

([Cross posted from of forum](http://forum.openframeworks.cc/t/addon-data-in-addon-config-mk-not-working/15202))

oxillo:
Fix #2903 .
Tested on MSYS2 with a single addon with  2 data files in data folder

```
ADDON_DATA = data/test.txt
ADDON_DATA += data/test1.txt
=> Works copy test.txt and test1.txt in bin/data 👍 
```

```
ADDON_DATA = data/ 
=> Works copy data directory in bin/data 👍 
```

```
ADDON_DATA = dat* 
=> Works copy data directory in bin/data 👍 
```
```
ADDON_DATA = data/test?.txt 
=> Doesn't work 👎 
```


arturoc:
Thanks!

PR #6248, opened 2019-02-27
ofxiOS: fix FBO for GLK view

gumilastik:
In the current version of the addon does not work fbo on ios. Here is the issue:
https://forum.openframeworks.cc/t/fbo-doesnt-work-on-ios/31174


arturoc:
@danoli3 can you take a look?

danoli3:
Looks great! 

https://developer.apple.com/documentation/glkit/glkview/1615593-binddrawable?language=objc

I'll also add it to the other GLKView

arturoc:
thanks!!

Issue #6202, opened 2018-12-25
cairo.h not found Arch Linux


DarkStar1997:
I downloaded the nightly build for openframeworks from 
[linux64gcc6_nightly](https://openframeworks.cc/ci_server/versions/nightly/of_v20181222_linux64gcc6_nightly.tar.gz)
The compilation of 3d/cameralensOffsetExample produces the following output:
```
[rohan@rohan-pc cameraLensOffsetExample]$ make -j8
HOST_OS=Linux
checking pkg-config libraries:   cairo zlib gstreamer-app-1.0 gstreamer-1.0 gstreamer-video-1.0 gstreamer-base-1.0 libudev freetype2 fontconfig sndfile openal libcurl glfw3 rtaudio libpulse-simple alsa gl glu glew gtk+-3.0 libmpg123 
with PKG_CONFIG_LIBDIR=
Compiling OF library for Release
make[1]: Entering directory '/home/rohan/Downloads/of_v20181222_linux64gcc6_release/libs/openFrameworksCompiled/project'
HOST_OS=Linux
checking pkg-config libraries:   cairo zlib gstreamer-app-1.0 gstreamer-1.0 gstreamer-video-1.0 gstreamer-base-1.0 libudev freetype2 fontconfig sndfile openal libcurl glfw3 rtaudio libpulse-simple alsa gl glu glew gtk+-3.0 libmpg123 
with PKG_CONFIG_LIBDIR=
HOST_OS=Linux
checking pkg-config libraries:   cairo zlib gstreamer-app-1.0 gstreamer-1.0 gstreamer-video-1.0 gstreamer-base-1.0 libudev freetype2 fontconfig sndfile openal libcurl glfw3 rtaudio libpulse-simple alsa gl glu glew gtk+-3.0 libmpg123 
with PKG_CONFIG_LIBDIR=
HOST_OS=Linux
checking pkg-config libraries:   cairo zlib gstreamer-app-1.0 gstreamer-1.0 gstreamer-video-1.0 gstreamer-base-1.0 libudev freetype2 fontconfig sndfile openal libcurl glfw3 rtaudio libpulse-simple alsa gl glu glew gtk+-3.0 libmpg123 
with PKG_CONFIG_LIBDIR=
Done!
make[1]: Leaving directory '/home/rohan/Downloads/of_v20181222_linux64gcc6_release/libs/openFrameworksCompiled/project'


Compiling cameraLensOffsetExample for Release
[...]

PR #6252, opened 2019-03-04
Msys2 opencv4

oxillo:
Update ofxOpenCv addon to use openCv 4 if available.
Some methods (especially HaarDetectObejct converted to their C++ equivalent)
Should correct #6202


arturoc:
Thanks so much for looking into this, it's really useful.

Couple of things.

- You've added to the PR your vs code config, can you remove it?
- The way you are doing the addons config is only compatible with the makefiles but won't work with the PG or the qtcreator wizard. What i did recently for another addon is to add an exception in the install_dependencies script that detects the version of the library and modifies addon_config.mk to change the file with whatever is needed. in this case changing opencv -> opencv4

oxillo:
Cleaned up VS code config; 
Added the ofxOpenCv addon_config.mk modification in msys2 install_dependencies.
I was also thinking of having a script in the addon so people upgrading their system can upgrade the script.

arturoc:
i've just compiled opencv 4 for the platforms that use local libraries, can you change the script to instead fix it if you are using and old version? we could also just not have any fix and assume people need to have an up to date install of msys

arturoc:
going to merge this since msys is working fine now and keep working on the other platforms on a different PR.

Thanks!

PR #6256, opened 2019-03-15
Opencv4: fix remaining platforms

arturoc:


pzoltowski:
would be great to update to opencv 4.1 - I guess there shouldn't be any breaking changes but many bug fixes from 4.0.1

PR #6265, opened 2019-03-25
Fix easy cam auto distance

roymacdonald:
when calling `setPosition(..)` on an `ofEasyCam` object during setup, by default it will no go to this position as it will get over riden by the auto distance function of it. Now, once calling `setPosition(..) ` during setup, auto distance is disabled and the camera will get positioned where it needs to be.

### DIFF
--- a/libs/openFrameworks/3d/ofEasyCam.cpp
+++ b/libs/openFrameworks/3d/ofEasyCam.cpp
@@ -70,6 +70,9 @@ void ofEasyCam::reset(){
 	rot = {0,0,0};
 	translate = {0,0,0};
 
+	if(bAutoDistance){
+		bDistanceSet = false;
+	}
 	bApplyInertia = false;
 	currentTransformType = TRANSFORM_NONE;
 }
@@ -506,4 +509,10 @@ bool ofEasyCam:: hasInteraction(TransformType type, int mouseButton, int key){
 //----------------------------------------
 void ofEasyCam::removeAllInteractions(){
 	interactions.clear();
+}
+//----------------------------------------
+void ofEasyCam::onPositionChanged(){
+	if(!bDistanceSet && bAutoDistance){
+		bDistanceSet = true;
+	}
 }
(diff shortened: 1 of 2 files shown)


arturoc:
mmh, isn't there anyway to do this without adding virtual to those methods in ofNode?

If not i'd prefer to leave it as it is and just look into not using inheritance for ofNode and just publish whatever methods make sense for ofEasyCam at some point. That would need a lot of refactoring cause other classes are inheriting from ofNode though

roymacdonald:
Hi,
it could be done by overriding `ofNode::onPositionChanged() ` so it disables the auto distance.
What's the problem with adding `virtual` to those functions in `ofNode`?
Not using inheritance would need some massive refactoring. What would be the benefits of such? I understand the disadvantages that inheritance can have but, Is it worth (in terms of the amount of work needed to get rid of it versus the benefits it will bring)?

arturoc:
I think in the long term we should, even if only as a pedagogical thing so it doens't invite people to use it themselves but also cause it would make the code cleaner and avoid things like this and remove inconsistent methods like having scale on a light for example

Adding virtual to those functions completely changes the way ofNode can be used and if added to any method it should be added to all of them not only to a couple of them randomly.

I think by now if you can do it using onPositionChanged that would be the best

roymacdonald:
I see. well it makes a lot of sense. 
Just out of curiosity, do you think that removing inheritance on all parts of OF is the way to go or should it only be in the parts that give conflicts, like ofNode? There are other parts, like the renderers, where inheritances seems to be a useful thing. Am I right?

I will  make the changes to use `onPositionChanged()`and PR
cheers

arturoc:
The renderers don't really use inheritance, there are some base pure abstract classes, all named as ofBase* but those don't have any implementations and more importantly any data, the way that works is more akin to interfaces in other languages and are there only to provide polymorphism but the classes that implement them are not really inheriting anything from them. 

ofNode is, I think, the only place where OF uses inheritance as in a full class inheriting from another

roymacdonald:
Thanks for the explanation. I would have thought that having pure abstract base classes with no data was inheritance too. 
Cheers

roymacdonald:
@arturoc I just made the change to use `onPositionChanged()`

arturoc:
thanks!

PR #6275, opened 2019-04-08
Fix url for issue tracker

meltar:
Use the correct absolute path instead of a relative path in two links. The malformed path went to a 404 page instead of the issue tracker.

Existing url:
`https://github.com/openframeworks/openFrameworks/blob/master/github.com/openframeworks/openFrameworks/issues`
Fixed url:
`https://github.com/openframeworks/openFrameworks/issues`


arturoc:
thanks!

PR #6281, opened 2019-04-22
fix: Fix the indent

gaocegege:
PTAL


bakercp:
This seems like a pretty clean fix @arturoc .

arturoc:
we actually use tabs not spaces its pretty irregular by now but better not to merge things like this that introduce more errors

bakercp:
Well, it is quite inconsistent so let's just move forward with this merge.

PR #6282, opened 2019-04-23
fix setSpaceSize

crecord:
fix setSpaceSize to be independent of setLetterSpacing and work regardless of whether setLetterSpacing is at default value. More details on the forum: https://forum.openframeworks.cc/t/oftruetypefont-setletterspacing-setspacesize-problem/32327

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -1054,12 +1054,16 @@ void ofTrueTypeFont::iterateString(const string & str, float x, float y, bool vF
 			} else if (c == '\t') {
 				if (settings.direction == OF_TTF_LEFT_TO_RIGHT){
 					f(c, pos);
-					pos.x += (getGlyphProperties(' ').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
+					pos.x += getGlyphProperties(' ').advance * letterSpacing * TAB_WIDTH * directionX;
 				} else{
-					pos.x += (getGlyphProperties(' ').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
+					pos.x += getGlyphProperties(' ').advance * letterSpacing * TAB_WIDTH * directionX;
 					f(c, pos);
 				}
 				prevC = c;
+			} else if(c == ' '){
+				pos.x += getGlyphProperties(' ').advance * spaceSize * TAB_WIDTH * directionX;
+                		f(c, pos);
+               			prevC = c;
 			} else if(isValidGlyph(c)) {
 				const auto & props = getGlyphProperties(c);
 				if(prevC>0){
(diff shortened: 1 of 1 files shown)


tgfrerer:
Looks good ;) I'm mentioning the following issues/prs for added context:  #6109 #6106
The fact that `spacesize` keeps returning might point to it being a bit non-intuitive to use. Maybe we should get rid of `spacesize` alltogether?

arturoc:
as @crecord points out in the forum, it might be useful in somce cases i think the problem might be just the naming is a bit ambiguous? perhaps we could change it to setSpaceCharacterSize or something like that?

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
```diff
@@ -1054,12 +1054,16 @@ void ofTrueTypeFont::iterateString(const string & str, float x, float y, bool vF
...
-					pos.x += (getGlyphProperties(' ').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
+					pos.x += getGlyphProperties(' ').advance * letterSpacing * TAB_WIDTH * directionX;
 				} else{
-					pos.x += (getGlyphProperties(' ').advance * spaceSize * letterSpacing) * TAB_WIDTH * directionX;
+					pos.x += getGlyphProperties(' ').advance * letterSpacing * TAB_WIDTH * directionX;
 					f(c, pos);
 				}
 				prevC = c;
+			} else if(c == ' '){
+				pos.x += getGlyphProperties(' ').advance * spaceSize * TAB_WIDTH * directionX;
```
arturoc: I think TAB_WITH is not correct here? this is only to set the number of spaces per tab so this would add 4 spaces rather than 1 no?

Also i think we should just use spaceSize if it's set instead of advance * spaceSize otherwise it should be callsed space size factor instead of space size to avoid confusions

tgfrerer: No i think it's correct - this branch deals with handling the '\t' tab character, which we expand to 4 (or however many we specify in `TAB_WIDTH`) spaces, unless I'm mistaken.

crecord: Looking at it again. It looks like TAB_WIDTH is unnecessary because the if statement above it should handle that case.  
`pos.x += getGlyphProperties(' ').advance * spaceSize * directionX;`

The only issue with removing getGlyphProperties(' ').advance is that then the default spacing is too tight.

`pos.x += spaceSize * directionX;`

It only includes a letter space. 
Screen Shot 2019-04-23 at 2 15 53 PM

So I'd think it would be best to leave it in.

tgfrerer: yes of course - i was looking at lines 1057, and 1059, sorry. the `TAB_WIDTH` in 1064 looks indeed misplaced.

crecord:
Personally, I think that the naming and function on setSpaceSize makes sense. It just needs to have the expected behavior of changing just the space sizing, be independent of letter spacing, and return the correct bounding box.

I am going to test and see if the bounding box and getStringTexture still works. I'll probably need to push another commit because I only fixed it in that one function!

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
arturoc: @crecord, yes i'm not saying we should remove the advance for the space just that if the space size is set then and only then instead of using the advance we should use the space size.

We can keep track with a flag (`bool spaceSizeSet`) if that value has been set or something like that

Not sure if that makes more sense but if not then that function should be called something like setSpaceSizeFactor or something similar

crecord: I see, I think the factor approach would be preferable because then the spacing would scale with the text rather than needing to set it for a new text size.  Isn't letterSpacing also a factor, but that isn't indicated in the name? Seems like that should be consistent in the naming of both whichever direction you want to go.

crecord:
While I was sorting out the set spacing size issues I came across a couple unrelated problems with getStringTexture.

1. The Y positioning was wonky 
Screen Shot 2019-04-23 at 3 50 31 PM

2. Tab was showing up as a strange character.  
Screen Shot 2019-04-23 at 3 53 55 PM

Both of the above are fixed.

crecord:
Also I made changes to getFirstGlyphPosForTexture, but I'm honestly not sure what how it is useful, so wasn't sure how to thoroughly test it.

arturoc:
That's great, thanks!  getFirstGlyphPosForTexture is weird but it can be used to align several string textures I believe it should be fine.

PR #6285, opened 2019-04-27
armv6: remove failing cd at end of install.sh

lucasrangit:
The `cd $ROOT/raspbian/usr/lib/gcc/arm-linux-gnueabihf/4.9` fails because the GCC version in `rpi_toolchain_gcc6.tar.bz2` is 6.3.0 and not 4.9.

Simply remove it because it is not needed and never did anything (no call to fix relative links after `cd`).

This change is against the stable branch because that's what I use. However, it should be applied to master as well.

Signed-off-by: Lucas Rangit MAGASWERAN <lucasrangit@gmail.com>


dimitre:
Hello Lucas, thanks for the PR. I think this is not needed anymore, scripts were updated to use gcc installed via apt recently.
PRs are very welcome, for the next ones use openframeworks:master instead of stable. Stable is only used for releases.
Thank you

PR #6286, opened 2019-05-02
Added missing ROOT variable to fedora install_dependencies.sh

gaborpapp:
The script uses ROOT when modifies the ofxOpenCv addon. Without it the script fails.


arturoc:
thanks!

PR #6287, opened 2019-05-04
ofxOscSender.setup() fix

bltzr:
as mentioned here: https://forum.openframeworks.cc/t/ofxoscsender-falls-back-on-localhost/32442/2
and discussed there: https://github.com/bltzr/openFrameworks/commit/93ccf317e4377dd754655fb91956a3efe0c43b76 with @danomatika 

when providing an unavailable host to ofxOscSender.setup() it was falling back to localhost, which prevented from knowing that the connection to the desired host failed.

those 2 commits fix that, and post an error message when the endpoint cannot be reached


arturoc:
thanks!

PR #6290, opened 2019-05-16
adding pugi parsing Options to ofXml

dasoe:
pugi offers parsing options like "pugi::parse_default | pugi::parse_comments" in case we want comments not to be stripped etc. Added a variable and a setter to be able to use them in ofXml.
Not sure if I did in accordance to your coding guidelines, but anyway it's just a proposal. Would be great to have for me. 

Would be used like this:
`XML.setParsingOptions(pugi::parse_default | pugi::parse_comments);
XML.load(pathToXML);`

**Thanks for your awesome work!**
oe

### DIFF
--- a/libs/openFrameworks/utils/ofXml.cpp
+++ b/libs/openFrameworks/utils/ofXml.cpp
@@ -6,6 +6,7 @@ using namespace std;
 ofXml::ofXml()
 :doc(new pugi::xml_document){
 	xml = doc->root();
+	parsing_options = pugi::parse_default;
 }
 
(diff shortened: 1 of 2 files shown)


bakercp:
This seems like a clean commit.  Any reason not to merge @arturoc ?

arturoc:
having an int as parameter is dificult to understand what that parameter means. would be better if this was an enum

bakercp:
@dasoe Can you reframe this as an enumeration for the purposes of the ofXml class? ofXml is trying to wrap and simplify more complex features like bit-field operators.

PR #6291, opened 2019-05-16
Improvements in life cycle of Android Activity.

ivorne:
It now preserves EGL context when minimizing and restoring the app Activity, it also has a proper screenshot in Recent Apps list. Support for multiwindowing without restart and pausing/resuming rendering from onStop/onStart callbacks instead of onPause/onResume callbacks.

Detailed information about what my initial problem was and about the whole solution is on stackexchange: https://gamedev.stackexchange.com/questions/171908/how-to-preserve-egl-context-when-pausing-android-ndk-app.


ofZach:
I am very excited about this PR as I have been dealing with and solving some of these issues on my own.   It's great to see the work you've done.

one question I had about this PR is if "pause_gameplay" makes sense or if we can rename it something less "game" centric for apps that aren't games.  For example, I am working on an AR application and I've had issues when sharing videos since it pauses the app briefly.

ivorne:
It's nice to see some feedback.

If you are going to try this out, either combined with your changes or just as it is, please let me know how it worked. An independent verification of the PR would be valuable.

This commit changes the behavior of Activity.onPause and Activity.onResume callbacks - that could  help with the problem with brief pausing in your AR application.

I have no problem renaming the "pause_gameplay" callback to something else if we find a better name. It is to be used in interactive applications with a progression system. This means that the application expects continuous user input and **lack of input** could cause change of state of the application that is nontrivial for the user to revert. The pause_gameplay event marks point in time from which we should not expect continous user input and we should pause all processes that would interpret lack of user input as **intentional** lack of input.

The explanation is getting a bit abstract. The advantage of the name "pause_gameplay" is that it is fairly intuitive. But maybe "pause_interactivity", "pause_input_interrupted", "pause_interactive_progression" or something like that might work.

arturoc:
What is the corresponding event in the android API for onPauseGamePlay? I don't see anything like that there, only the events we had until now

ivorne:
Android has onStart/onStop and onResume/onPause. OnStart/onStop informs about Activity going to backgroud, when it's not visible and should not render or update animations (this corresponds to OF pause and resume events). OnResume/onPause is called when input focus is lost or things like that. Pause_gameplay event would be notified when Activity.onPause is called.

The problem is that OF already uses different naming than Android. But the Android naming is fairly confusing anyway.

arturoc:
I'm not sure i understand exactly what's the problem. So why don't we just use onPause? We already have it, no? Why do we need a new event that isn't part of the official specification?

ivorne:
The official Android specification is that onStop is called when Activity goes to background (not visible on screen) and onPause is called when input focus is lost.

We have three options:

1. Use the same naming as Android. Disadvantage: it would change the meaning of ofxAndroidEvents().pause event. Currently it means that Activity went to background, after the change it would mean that input focus was lost.

2. Extend the naming that OF currently uses - ofxAndroidEvents().pause will still mean that app went to background and we will create new event (eg. pause_gameplay) that will tell us that input focus was lost.

3. Keep the naming that OF uses - don't handle the event which tells us that input focus was lost (Activity.onPause). Gameplay will be paused in the pause event (when app goes to background).

arturoc:
I think 1 would be the best even if we changed the current behaviour slightly. Having something that behaves differently than android's original spcification would be really confusing (as confusing as that already is) and adding a new event that doesn't exist in android even more i think

We are about to do a major release so it would be the moment to do a breaking change like this. In any case aren't we calling OF onPause on Android's onPause right now? If i understand it correctly internally we are then doing some calls to persist the GL state where possible but if we move that to onStop it shouldn't break any user code as far as i understand, no?

libs/openFrameworks/app/ofWindowSettings.h
```diff
@@ -119,13 +119,14 @@ class ofGLWindowSettings: public ofWindowSettings{
 class ofGLESWindowSettings: public ofWindowSettings{
 public:
 	ofGLESWindowSettings()
-	:glesVersion(1){}
+	:glesVersion(1), preserveContextOnPause(true){}
```
arturoc: Just saw this. This class, ofGLESWindowSettings is used in other platforms and is a generic settings class for anything that uses GLES so this shouldn't be here.

If we want to add a new setting that is android specific it should go in an ofxAndroidWindowSettings class that extends this one.

We do something similar with ofGLWindowSettings and ofGLFWWindowSettings if you want to take a look

ivorne:
Good points, I agree. Thinking about it, we don't need to wait for major release if we decide to go this way. It will change the meaning of the OF events but it will not change the actual behavior (which is currently incorrect).

libs/openFrameworks/app/ofWindowSettings.h
ivorne: Right, I looked at the *WindowSettings implementation and got scared by those wild casts in copy constructors :-). I'll change it.

arturoc: I know I should probably replace them with an abstract class that every settings of the same class implement instead of using inheritance but if you can just copy that style that should be ok by now

ivorne:
So I incorporated both of those changes into the PR.

1. The events are now 'start', 'stop', 'resume', 'pause', and their meaning is the same as in Android.

2. I created ofxAndroidWindowSettings which interits ofGLESWindowSettings and I moved the 'preserveContextOnPause' parameter into it, so that it now only affects Android.

libs/openFrameworks/app/ofWindowSettings.h
ivorne: Done. It is in e8a0d92799a22505b1e4aba45f81edfe44b2dfa4.

ivorne:
So what now? Any interest in merging this PR?

arturoc:
can you merge current master against your branch?

ivorne:
Right, will do.

arturoc:
thanks!

PR #6292, opened 2019-05-19
Add handling of double precision position data in PLY model loader

CharStiles:
If users try to load a PLY with double precision position data instead of float data, OF will act like its floating point precision, instead of silently loading the model incorrectly.
This was discussed in my forum post [here](https://forum.openframeworks.cc/t/solved-double-precision-ply-file/32506 ) It was suggested I make a PR so here we are. :) thanks.


arturoc:
That's great thanks!

Issue #6294, opened 2019-05-20
ofPath::tessellator initialization order fiasco


roymacdonald:
Hi, I've seen several times apps crashing randomly (ie. you run it once and it works, then run it again without even recompiling and it will crash, then run again and it works fine), where always the crash points to `ofPath::tessellator`. I think that the problem is due to the initialization order fiasco, which is properly addressed on lots of other parts of the OF core but not in this case.
The one issue that raises my doubt here, rather than just simply implementing the fix and making a PR, is that these `static` variable is also, sometimes, a `thread_local` variable. 
As far as I can tell the implementation for the the `thread_local` and the regular `static` variables is almost the same, but I might be wrong. Can you guys enlighten me please? @arturoc @bakercp 

 I think that we should implement a getter function for the tessellator like this.

```
static ofTessellator & getTessellator(){
#if defined(TARGET_EMSCRIPTEN)
	static std::unique_ptr<ofTessellator> i =  make_unique<ofTessellator>();
	return *i;
#elif HAS_TLS
	static thread_local std::unique_ptr<ofTessellator> i =  make_unique<ofTessellator>();
	return *i;
#else
	return tessellator; // non static instance defined in ofPath.h
#endif
}
```
Would this be correct? Please let me know what you think. cheers

arturoc:
I doubt this is related to static initialization order since the tesselator is never used until draw or getTesselation are called explictly, never on any constructor so unless you are calling yourself those on any constructor called from a static initialization it shouldn't be a problem.

Also the initialization order is not random and definetly wont't change between runs without recompiling and in most cases even after recompiling.

The tesselator sometimes crashes due to malformed triangles (three points on the same line for example) so i think that might be the problem. If you are running the tesselator on a randomly initialized shape it would make the crashes random. I haven't checked in a while if there's a new version of the tess2 library but if there is we might want to update it. Other than that or trying to detect malformed triangles (which might make tesselation much slower) I'm not sure there's any other solution

roymacdonald:
I see. Thanks for answering.
This happens when using ofxGui actually. So the geometry is not random. I think it happens when instantiating an ofxFloatSlider, for example, as it usually points to the to it.
<img width="1400" alt="Screen Shot 2019-05-21 at 09 53 47" src="https://user-images.githubusercontent.com/974878/58101948-7b9e2b00-7bae-11e9-9256-4bc3393ec0cf.png">

[...]

PR #6297, opened 2019-05-22
update method for drawing slider bars

tgfrerer:
Instead of calling `ofPath` tessellation on each draw call for sliders, slider knows how to generate mesh for bar rectangles itself. The mesh is cached, and updated **iff** the slider value changes.

Rect mesh caching works similarly to the slider's internal caching of text meshes.

You might notice a slight performance improvement on GUIs with multiple sliders with this PR.

As an additional check/optimisation, slider fill rectangles with either width or height < 1 pixels will also not be drawn (they would not be visible anyway).

By *not* using `ofPath` (and by implication the tessellator) to draw the slider we avoid having to ask the tessellator to tessellate extremely small paths, in which case it had a tendency to throw its hands up in
despair or disgust (hard to tell from a stack trace) and cause crashes (see issue #6294).

Fixes issue #6294.

### DIFF
--- a/addons/ofxGui/src/ofxGuiUtils.h
+++ b/addons/ofxGui/src/ofxGuiUtils.h
@@ -1,0 +1,84 @@
+#pragma once
+
+#include "ofColor.h"
+#include "ofRectangle.h"
+#include "ofVboMesh.h"
+
+namespace ofxGui {
+
+/*
+ * Internal helper to generate and cache rectangle meshes for ofxGui.
+ */
+class RectMesh {
+	ofColor     mColorFill = {};
+	ofRectangle mRect      = {};
+	ofVboMesh   mMesh      = {};
+	bool        isDirty    = true;
+	bool        mHasMesh   = false;
+
+  public:
+	void clear() {
+		mColorFill = {};
+		mRect      = {};
+		mMesh.clear();
+		mHasMesh = false;
+		isDirty  = false;
+	}
+
+	void draw() {
+
+		if ( isDirty && mHasMesh ) {
+			mMesh.clear();
+			mHasMesh = false;
+			isDirty  = false;
+		}
+
+		if ( mRect.width < 1.f || mRect.height < 1.f ) {
+			// We will not draw a mesh for rectangles
+			// which are smaller than one pixel for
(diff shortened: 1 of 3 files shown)


arturoc:
:) thanks! this looks actually really useful for other gui elements since pretty much everything is a rectangle in ofxGui and should be much faster to generate a mesh than tessellate a path. Can you move it outside of oxfSlider into utils or something like that? Perhaps just it's own file. No need to adapt other controls but we can start using it whenever we modify other classes

Also can you remove the allocation of the vectors on every change? given how small they are it's probably insignificant but better to avoid memory allocations. Just add the vertex info directly in the mesh.

roymacdonald:
Ohh thanks @tgfrerer ! :D
I will be able to look at this in depth next week and hopefully adapt it to the rest of ofxGui. I'll keep you posted. 
Cheers!

tgfrerer:
@arturoc I've updated the PR as to your suggestions. I've extracted the rectangle draw code into `ofxGuiUtils.h`, a separate header file which I have created in `ofxGUI`, where I felt it most likely belonged. 

I've placed the helper into the `ofxGui` namespace so that it doesn't appear in the general namespace, and to indicate that it belongs to `ofxGui`.

I've also added a dirty flag and checks for it so that the helper becomes easier to use - with this, calling `clear()` is not mandatory anymore which i hope makes its use a bit more intuitive.

arturoc:
Great thanks, can you remove the namespace though? we don't have namespaces anywhere else in the official addons so it would be weird to use it for this one class.

Also, and it probably doesn't make a big difference so feel free to leave it as it is. But it seems that most (all?) of the flags could just be replaced with a check to see if the mesh is empty which would make it even easier to follow the state? So just clear the mesh when something changes and then on draw if the mesh is empty regenerate it before draw which would avoid the boolean flags.

As i said apart from the namespace the other changes don't make such a big difference so feel free to leave it like it is if you prefer

arturoc:
Sorry I meant ofxGui::RectShape -> ofxGuiRectShape otherwise it would easily lead to clashes with some other library

arturoc:
thanks!

PR #6303, opened 2019-05-29
ofxSvg improved compatibility

sebleedelisle:
Hi team! 

Some fixes for ofxSVG here. I hope they're helpful, and let me know your thoughts / feedback etc. 

- added support for the SVG "use" command, a common feature that allows multiple instances of the same graphic elements. This is commonly found in SVGs exported with vvvv and Flash. 

- fixed an issue where line weights less than 1 would not render. 

- fixed an issue where elements marked "display:none" would still render. 

- separated the parsing from the loading for increased flexibility (my use case was parsing SVG files being sent over a socket connection from vvvv). 

It works by expanding out the original SVG text before passing it into lib_svgtiny so there is somewhat of a performance hit - so I've added the ability to disable it if required by using the ofxSVG::setImprovedCompatibilityMode(bool mode) method. It also required a minor addition to ofXML as outlined below. 

ofXML - provided access to the underlying pugi::xml_node method "getParent()"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,7 +149,7 @@ script:
   - ccache -z
   - ccache -s
   - if [ "$TARGET" = "linux64" ] || [ "$TARGET" = "linuxarmv6l" ] || [ "$TARGET" = "linuxarmv7l" ]; then
-      wget http://mirrors.kernel.org/ubuntu/pool/main/c/ccache/ccache_3.3.4-1_amd64.deb;
+      wget https://mirror.fccn.pt/repos/pub/ubuntu/pool/main/c/ccache/ccache_3.3.4-1_amd64.deb;
       sudo dpkg -i ccache_3.3.4-1_amd64.deb;
     fi
   - if [ "$TARGET" = "osx" ] && [ "$OPT" = "makefiles" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,8 +149,8 @@ script:
   - ccache -z
   - ccache -s
   - if [ "$TARGET" = "linux64" ] || [ "$TARGET" = "linuxarmv6l" ] || [ "$TARGET" = "linuxarmv7l" ]; then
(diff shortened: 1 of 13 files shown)


arturoc:
have you meassured how bad is the performance loss? if it's not too much i think we can just remove this new method to disable the full parser

In any case instead of having a flag that sets a state it would be better to have an additional method like loadLimitedSvg() or something like that

sebleedelisle:
Hey Arturo! Performance will vary hugely depending on the size of the SVG file, and the amount of fixing it needs. For the tiger image in the SVG example, it actually seemed to load a bit faster than before (which I don't actually understand!). So maybe we don't need the fallback? Happy to remove it and re-commit if you're happy. :)

arturoc:
mmh not sure what to do then, have you seen any case where loading a file would take way longer than before?

sebleedelisle:
After more testing I can confirm that there is no measurable performance degradation for SVG files that do not need fixing. So it only takes longer if the files _do_ need fixing, in which case they wouldn't have worked at all before, and the performance decrease is marginal in most SVGs... so I think we're good. 

Let me know if you're happy to take out the flag and I'll do another pull request.

arturoc:
Sure, if you can remove the flag I think it0s ready to merge, no need for another PR just push the new commit to your branch

sebleedelisle:
Ah great, new version pushed... I also added some more documentation to the file. Does this automatically get pushed out to the ofSite or do we need to do that separately?

arturoc:
documentation only updates when we do a major release which will happen soon and it happens automatically so nothing else to do.

there's a problem right now with some of the tests, nothing related to this PR just a general problem, i'll look into it and merge it as soon as that is solved

thanks!

sebleedelisle:
Great thanks Arturo!

arturoc:
thanks!

PR #6304, opened 2019-06-06
Fix makefile errors when MAKEFILE_DEBUG is defined.

bakercp:
Prevent "commands commence before first target" error when `MAKEFILE_DEBUG` is defined.

There were stray tabs under the MAKEFILE_DEBUG sections.


arturoc:
thanks!

PR #6307, opened 2019-06-06
Fix Makefile Verbosity level

bakercp:
This continues to place extra output behind the `MAKEFILE_DEBUG` guard to decrease chatter during the build and processes.

Also, ignores printing other errors during the build process (e.g. `cat .compiler_flags` when `.compiler_flags` doesn't exist.)


arturoc:
thanks!

PR #6309, opened 2019-06-13
Implement ofMesh .save() of triangle strips

hamoid:
`ofMesh.inl` contains a comment like this:

    //TODO: add index generation for other OF_PRIMITIVE cases

This PR implements one of those missing cases (`OF_PRIMITIVE_TRIANGLE_STRIP`)


arturoc:
thanks!

PR #6312, opened 2019-06-16
ofSoundBuffer::copy from was not being set properly

roymacdonald:
there was a small type that was making `ofSoundBuffer::copyFrom` not to set the samplerate to the value passed as argument.
ofSoundBuffer's variable for the sample rate is called `samplerate`while the argument passed to the copyFrom function was called `sampleRate`. The function was setting the sample rate using `samplerate` thus it would never actually set it to the argument passed. This PR fixes this issue.


arturoc:
can we use a _ in the parameters or something to avoid this problem in the future, it's really subtle and easy that someone modifying the file would do the same mistake

roymacdonald:
Sure. Just added what you suggested.

PR #6314, opened 2019-06-17
Included missing header in ofImage.h required for instantination of of ofImage_::clone()

thesstefan:
Included missing header in ofImage.h required for instantination of ofImage_::clone().


arturoc:
thanks!

PR #6318, opened 2019-06-25
Update the android sdk version to 23.

lewisou:
Update the Android SDK version to 23.

### DIFF
--- a/libs/openFrameworksCompiled/project/android/build.gradle
+++ b/libs/openFrameworksCompiled/project/android/build.gradle
@@ -13,12 +13,12 @@ model {
         compileSdkVersion 25
 
         defaultConfig.with {
-            minSdkVersion.apiLevel 19
+            minSdkVersion.apiLevel 23
             targetSdkVersion.apiLevel 25
         }
 
         ndk {
-            platformVersion = "19"
+            platformVersion = "23"
             moduleName "openFrameworksLib"
             toolchain  = buildToolchain()
             stl        = compilerStl()
--- a/libs/openFrameworksCompiled/project/android/build.gradle
+++ b/libs/openFrameworksCompiled/project/android/build.gradle
@@ -13,12 +13,12 @@ model {
         compileSdkVersion 25
 
         defaultConfig.with {
-            minSdkVersion.apiLevel 23
+            minSdkVersion.apiLevel 19
             targetSdkVersion.apiLevel 25
         }
 
         ndk {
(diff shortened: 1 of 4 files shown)


danoli3:
All libs need this as well.

``` In Android Q, the platform will warn users the first time they run any app that targets a platform version lower than Android 6.0 (API level 23). If the app requires the user to grant permissions, the user is also given an opportunity to adjust the app's permissions before the app is allowed to run for the first time.

Due to Google Play's target API requirements, a user would only see these warnings if they run an app that has not been updated recently. For apps that are distributed through other stores, similar target API requirements are also being introduced in 2019. For more information about these requirements, see Expanding target API level requirements in 2019. ```

arturoc:
@danoli3 the libraries were compiled for android 23 here: https://github.com/openframeworks/apothecary/pull/134

@lewisou I thought you were going to compile the libraries for android 21 instead of 23. In any case i'm not opposed to update to 23 but It's not very clear to me how compatibility is affected by this. Will OF apps keep working on devices with android versions older than android 6? Does it matter?

Do you have any idea @danoli3? cc @danzeeeman @HalfdanJ

arturoc:
I've just recompiled apothecary against ndk 21 which seems to work and modified this PR to use 21 for the ndk and keep 19 as the minimal version for the sdk since that should be unrelated

if tests pass i'll merge this later

arturoc:
thanks!

danoli3:
Target SDK is very much similar to iOS Target SDK vs Minimum SDK so it is backwards compatible.

That being said, the current Target SDK Minimum is 23 for the moment, however within 3 months this moves to Target SDK 28 for all new Android Apps and 26 for all existing Android Apps on the Store.

> When you upload an APK, it needs to meet Google Play’s target API level requirements. Starting August 1, 2019, Google Play requires that new apps target at least Android 9.0 (API level 28), and that app updates target Android 9.0 from November 1, 2019. Until these dates, new apps and app updates must target at least Android 8.0 (API level 26).
https://developer.android.com/distribute/best-practices/develop/target-sdk

Might be worth just pushing everything forward. I'll be looking into this as well more for a oF Android App pretty soon, be jumping back on oF so I can help out

cerupcat:
@danoli3 Did you update to newer NDK? I'm looking to use NDK platform version 26, but not sure if it's as simple as changing the 'platformVersion' to '26'. It doesn't appear so from my early tests.

PR #6323, opened 2019-06-28
Correctly parse escaped linker flags in addons for QTCreator.

bakercp:
Currently when dealing with complex addon_config.mk files, QTCreator was failing in situations like this:

```make
	ADDON_LDFLAGS  += -Wl,--start-group
	ADDON_LDFLAGS  += /opt/intel/mkl/lib/intel64/libmkl_intel_lp64.a
	ADDON_LDFLAGS  += /opt/intel/mkl/lib/intel64/libmkl_sequential.a
	ADDON_LDFLAGS  += /opt/intel/mkl/lib/intel64/libmkl_core.a
	ADDON_LDFLAGS  += -Wl,--end-group
	ADDON_LDFLAGS  += -lpthread -lm -ldl
```

While makefile works fine, QTCreator/QBS failed when linker flags were already escaped (-Wl, or -Xlinker,).

According to the documentation, linker flags should not be escaped.

https://doc.qt.io/qbs/qml-qbsmodules-cpp.html#linkerFlags-prop

So, this PR fixes that for addons.


arturoc:
thanks!

PR #6325, opened 2019-07-04
Rpi4

eshkrab:
some attempts at making oF work on raspberry pi 4, as per discussion [here](https://forum.openframeworks.cc/t/rpi-4-raspbian-buster-openframeworks/32856/13). Everything compiles (as TARGET_LINUX, not TARGET_RASPBERRY_PI) but currently fails at runtime with a segfault because unable to init GLFW

### DIFF
--- a/addons/ofxOpenCv/addon_config.mk
+++ b/addons/ofxOpenCv/addon_config.mk
@@ -61,7 +61,7 @@ common:
 	# ADDON_LIBS_EXCLUDE =
 
 linux64:
-    ADDON_PKG_CONFIG_LIBRARIES = opencv harfbuzz
+    ADDON_PKG_CONFIG_LIBRARIES = opencv4 harfbuzz
 	ADDON_LIBS_EXCLUDE = libs/opencv/%
 	ADDON_INCLUDES_EXCLUDE = libs/opencv
 	ADDON_INCLUDES_EXCLUDE += libs/opencv/%
--- a/addons/ofxOpenCv/addon_config.mk
+++ b/addons/ofxOpenCv/addon_config.mk
@@ -61,7 +61,7 @@ common:
 	# ADDON_LIBS_EXCLUDE =
 
 linux64:
-    ADDON_PKG_CONFIG_LIBRARIES = opencv4 harfbuzz
+    ADDON_PKG_CONFIG_LIBRARIES = opencv harfbuzz
 	ADDON_LIBS_EXCLUDE = libs/opencv/%
 	ADDON_INCLUDES_EXCLUDE = libs/opencv
 	ADDON_INCLUDES_EXCLUDE += libs/opencv/%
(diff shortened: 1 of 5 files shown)


addons/ofxOpenCv/addon_config.mk
```diff
@@ -61,7 +61,7 @@ common:
 	# ADDON_LIBS_EXCLUDE =
 
 linux64:
-    ADDON_PKG_CONFIG_LIBRARIES = opencv4 harfbuzz
+    ADDON_PKG_CONFIG_LIBRARIES = opencv harfbuzz
```
arturoc: this should be kept as opencv4 you might have commited it after the install_dependencies script changes it cause opencv4 is not yet avaliable on the pi.

reverting it to opencv4 should fix some of the issues in other platforms

libs/openFrameworks/app/ofMainLoop.cpp
```diff
@@ -51,13 +53,15 @@ shared_ptr<ofAppBaseWindow> ofMainLoop::createWindow(const ofWindowSettings & se
...
+	#elif defined(TARGET_RASPBERRY_PI) && defined(TARGET_OPENGLES)
 	shared_ptr<ofAppEGLWindow> window = std::make_shared<ofAppEGLWindow>();
+	#elif defined(TARGET_RASPBERRY_PI) && defined(TARGET_OPENGL)
+	shared_ptr<ofAppGLFWWindow> window = std::make_shared<ofAppGLFWWindow>();
 	#elif defined(TARGET_EMSCRIPTEN)
 	shared_ptr<ofxAppEmscriptenWindow> window = std::make_shared<ofxAppEmscriptenWindow>();
 	#elif defined(TARGET_OPENGLES)
 	shared_ptr<ofAppGLFWWindow> window = std::make_shared<ofAppGLFWWindow>();
-	#else
+	#elif defined(TARGET_LINUX)
```
arturoc: this should just be an else, otherwise windows and osx are not considered

libs/openFrameworks/utils/ofConstants.h
```diff
@@ -188,14 +188,18 @@ enum ofTargetPlatform{
 
 	#include <unistd.h>
 
+  #include "bcm_host.h"
+  #undef countof
 	#ifdef TARGET_LINUX_ARM
-		#ifdef TARGET_RASPBERRY_PI
+		// #ifdef TARGET_RASPBERRY_PI
```
arturoc: this should be kept as it was and the include for bcm_host above removed, since it only belongs to the PI not to general linux. Also the define countof. if it's needed at all should go here not in the general linux section

addons/ofxOpenCv/addon_config.mk
eshkrab: ooh, I was wondering how that file got changed without me touching it

libs/openFrameworks/utils/ofConstants.h
eshkrab: so If I'm trying to compile with TARGET_LINUX, what'd be the right way to stick that countof undefine?

ofTheo:
@eshkrab any luck getting this to work? 
Could you maybe share the glfw error you are seeing? 

Thanks!

ofTheo:
closing as #6436 solves this. 
Thanks though!

PR #6331, opened 2019-07-07
Upgrade MSYS2 packages in appveyor (including GCC)

### LINKED ISSUE #6330: MSYS2 CI broken
Currently (July 2019), Continuous Integration in Appveyor for MSYS2 fails when trying to run the tests. 
Compilation works fine.

oxillo:
Fix #6330.
Version of GCC included in appveyor image is older than the one used to build openframeworks.
This results in an error when loading the Dlls at runtime.
install_dependencies.sh upgraded to use the MINGW_PACKAGE_PREFIX env variable defined by MSYS environment instead of the custom made mingw-w64-$arch. That should make install_denpendencies compatible with MINGW64 with no changes.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,11 +3,14 @@ version: 1.0.{build}
 environment:
   global:
     APPVEYOR_OS_NAME: windows
+    CHERE_INVOKING: 1
+    MSYS2_PATH: c:\msys64
   matrix:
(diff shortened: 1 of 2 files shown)


arturoc:
sorry for the long wait, have been wanting to look into this myself but didn't have the time lately, thanks!

PR #6335, opened 2019-07-10
A few different fixes

elliotwoods:
- be able to manually force an update for `localTransformMatrix` on `ofNode` (often this isn't desireable, but since this was previously supported, ofNode's internals aren't perfect for all situations, and it is entirely compatible with ofNode internals, please consider returning support for this function)
- add an operator for `(glm::mat4) * (glm::vec3)`. This is such a common operation. Again, this was available before and I believe this makes sense to be universal rather than implemented per use-space.
- `size` on ofThreadChannel - we've talked about this before. Can we please please keep this? I need to know the size!! And it is there in our reference implementation.


ofTheo:
Hey @elliotwoods - the second two features both make sense to me. 

I don't think it would hurt to add size() access to the queue even if it is occasionally off. 
For others the ThreadChannel size discussion is here: #4809

For the first one - I can't see how you could change the local matrix ofNode without it being updated?
Maybe I am missing something?

Issue #6320, opened 2019-06-26
[REQUEST] better background color when generating new projects


lewislepton:
I think that the generated background color could be better by being darker when generating a new project.

I think for learners, when creating shapes with no colors etc at first, it can be hard to see shapes right from the start. you can see that the circle is not perfect, but with a darker background, this makes it easier to see. the same with 3d objects as well. its easier on the eyes & you can see properly.
even more so if your shapes are smaller on screen

I set the background on grayscale to `60` & it feels right, because its not too dark. but just think its nicer for people starting out to have a project that generates something a bit darker & makes it clearer for people to see shapes.

a better experience in total ;)

I know we can set this with `ofBackground()` or `ofSetBackgroundColor()`. but for folks starting out, they may not know that. so are stuck with this brighter & harder to see way until they figure the solution out

2d shape
<img width="1136" alt="60177721-85502980-9811-11e9-8841-bbeaa03f5e45" src="https://user-images.githubusercontent.com/11299322/60190983-76767080-982b-11e9-8997-e71add7ae275.png">
<img width="1136" alt="60177720-85502980-9811-11e9-8dd7-0c772979805a" src="https://user-images.githubusercontent.com/11299322/60190994-7c6c5180-982b-11e9-8743-bbf9f9cd6e45.png">

3d shape
<img width="1136" alt="60177732-8f722800-9811-11e9-9122-26f79cd65c0a" src="https://user-images.githubusercontent.com/11299322/60190999-81c99c00-982b-11e9-82ca-3919e5c58ae2.png">
<img width="1136" alt="60177733-8f722800-9811-11e9-8f29-6b7045b783d2" src="https://user-images.githubusercontent.com/11299322/60191016-87bf7d00-982b-11e9-87dc-47a29ea2356e.png">

small shapes
<img width="1136" alt="60177958-24752100-9812-11e9-96b6-9fb3a01689ad" src="https://user-images.githubusercontent.com/11299322/60191029-8db55e00-982b-11e9-98b0-bf343ae5605a.png">
<img width="1136" alt="60177955-24752100-9812-11e9-89e4-3fd4ca205cdc" src="https://user-images.githubusercontent.com/11299322/60191036-90b04e80-982b-11e9-8bf1-427084467561.png">

ofZach:
I support for this small and helpful change -- darker grey still feels neutral and helps white things pop.  We default with white for draw color since beginners can sometimes get tripped up with drawing images and colors tinting the draw.

[...]

PR #6339, opened 2019-07-16
making the default background color 60

lewislepton:
related to what i brought up in issues #6320

tested & working. though double check if needed


arturoc:
thanks!

PR #6344, opened 2019-07-21
Add default values for ofPath::Command members.

bakercp:
Explicitly set default values for `ofPath::Command`.

Setting these makes serialization and other customization a bit easier.


arturoc:
thanks!

Issue #5863, opened 2018-01-29
fontShapesExample not working


schmidh:
Dear OF Community,

in the **_fontShapesExample_** the calls to **_getOutline_** are not working. So the loops to access the individual points are never run:

["// we can also access the individual points"](https://github.com/openframeworks/openFrameworks/blob/master/examples/graphics/fontShapesExample/src/ofApp.cpp#L61)

Any ideas?

Kind regards!

schmidh:
Dear OF community,

think I found the smelly line of code without knowing why. (Because I do know too little of the inner workings of OF.)

I commented out following line in the method `makeContoursForCharacter(FT_Face face)` of `ofTrueTypeFont` and then **fontShapesExample** works again:


[charOutlines.setPolyWindingMode(OF_POLY_WINDING_NONZERO)](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofTrueTypeFont.cpp#L174)

Maybe someone could have a look at the code at this point who has more knowledge than me.

Kind regards!

ofTheo:
can confirm this! 
thank you for finding the fix, will PR against this issue.

arturoc:
mmh sorry, there was another problem with the way we where using the contour path where we never set the line width to one so the first contour for th char was never drawn either. That's fixed now but not really this issue.

I'm going to reopen it but i think the winding mode is the correct one for fonts so it shouldn't be changed

arturoc:
oh, ok this was really the same problem. when we took the font path we were never setting the width to 1 so it wasn't generating a contour at all. it wasn't related to the winding. not sure why changing the winding fixes it though.

arturoc:
btw, theo, i didn't saw your comment until after committing the fix, i had been fixing something related to this some days ago and had the fixes for the example in my computer but had never committed them

schmidh:
Thanks a lot!

ofTheo:
abs, no worries @arturoc :) 
awesome that its fixed!  thanks @schmidh  for reporting this!

Issue #5792, opened 2017-10-31
don't strip gitignore/.gitignore from script/templates when building releases


Potato95:
so i'm getting into using Git with my projects. In the OF book it says that there is a gitignore file in the file path scripts/templates/gitignore/.gitignore. But when i go to that folder there is only a config file?? am i missing something? Cheers!!

bakercp:
It's there.  https://github.com/openframeworks/openFrameworks/blob/master/scripts/templates/gitignore/.gitignore

Perhaps, since the file starts with a "dot" (e.g. `.gitignore`) and is therefore considered "hidden" on many file systems you simply can't see it with your current Finder / Explorer / File Browser settings?

d-egan:
Hmm, I think this should be reopened.  
It's not a problem if you clone from the repository but "of_v20170714_linux64gcc6_nightly.tar.gz" doesn't have the `.gitignore`
Checked with `ls -a` `ls .*`

danzeeeman:
Yeah the releases strip git. This is not a bug.

d-egan:
But it's a template `.gitignore` for importing into your own project. right? So it's not used for maintaining the openframeworks repository and should be exempt from the strip.  
Without this `.gitignore`, the gitignore template is useless for everyone who uses the releases.

bakercp:
@d-egan I understand you and am reopening this.

This needs to be fixed:

https://github.com/openframeworks/openFrameworks/blob/b8f124f59ac20ae7651fc1ffa64d203362255f96/scripts/dev/create_package.sh#L477

bakercp:
I haven't tested it extensively yet, but I just submitted a PR #5824

bakercp:
Hi @d-egan I think e got it resolved for future releases. Thanks for your persistence!  Closing this.

PR #6346, opened 2019-07-27
Update from original (#1)

### LINKED ISSUE #2449: ofxiOSKeyboard need fix y poision?
ofxiOSKeyboard.mm

inside init() and updateOrientation() there is _y = _h need to fix to _y = _yOriginal ?

### LINKED ISSUE #4596: ES2 glsl precision
ive been working on a iOS ES2 project a have been experiencing a lot of triangles z-flighting.
figured out that the issue was with the glsl precision settings, currently set to `precision mediump float;`
by settings it to `precision highp float;`, this fixed all the z-fighting issues.

the two lines of code which define the ES2 glsl precision are here,
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp#L1826
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp#L1832

the only way i can increase the precision to high without hacking at the OF core is to write my own custom shader... but this won't be obvious for most users and is a bit of a hassle.

so wondering if there is a way of programmatically setting the ES2 precision in ofGLProgrammableRenderer?

### LINKED ISSUE #4942: Deprecate ofTTFCharacter (typedef for ofPath)
in ofTrueTypeFont we have a data type called "ofTTFCharacter" which is just a typedef of ofPath: 

https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofTrueTypeFont.h#L28

I suggest we deprecate this in favor of just using ofPath.   I don't think this class name gives us much benefit, has a somewhat different naming convention (TTF instead TrueTypeFont like the main class name) and on a personal note I find myself always having to look this up each time I use, which is usually a good sign it's not as intuitive as it could be.

### LINKED ISSUE #5859: ofTessellator Error
Dear OF community,

just trying to run example _**fontShapesExample**_ from master branch of OF on GitHub. Unfortunately I get following error message:

**[ error ] ofTessellator: performTessellation(): mesh polygon tessellation failed, winding mode 0**

Could also be a consequential error from _**ofTrueTypeFont**_ loading a font. Not sure. Has anybody else seen this error?

Kind regards!

### LINKED ISSUE #5861: windowMode not being passed on to ofAppGLFWWindow

### LINKED ISSUE #5865: Mac OF_TTF_SANS changed
On `TARGET_OSX`, `OF_TTF_SANS` corresponds to:

```c++
fontname = "Helvetica Neue";
fontID = 4;
```

But the way this font looks has changed at some point (either 10.12 or 10.13 I believe).

Here are some variations, from fontId 6 to 0:
 
fontid 6
fontid 5
fontid 4
fontid 3
fontid 2
fontid 1
fontid 0

I suggest changing it to fontId 0 (the final image).

### LINKED ISSUE #5904: OF_FULLSCREEN and OF_GAME_MODE loooks buggy
Hey, I'm trying to draw an FBO in fullscreen but it is not working as expecting. I'm on Linux(ubuntu) and I'm using the last OF master.

This is the code of the main.cpp file

```cpp
#include "ofMain.h"
#include "ofApp.h"

int main( ){
    settings.setGLVersion(3, 2);
//  settings.setSize(1480, 920);
    ofCreateWindow(settings);
    ofRunApp(new ofApp());
}
```

And this is the ofApp.cpp
```cpp
void ofApp::setup(){
    // finalFbo was declared in the header file as ofFbo
    ofSetVerticalSync(true);
    //FBO
    cout << "FULLSCREEN" << endl;
    cout << ofGetWidth() << endl; // this print 1024
    finalFbo.allocate(ofGetWidth(), ofGetHeight(), GL_RGBA);
    finalFbo.begin();
    ofClear(0, 0, 0, 0);
    finalFbo.end();
}

void ofApp::draw(){
    // fullfill the FBO
    finalFbo.begin();
    ofClear(0, 0, 0, 255);
    myClass.draw();
    finalFbo.end();
    // draw the FBO
    ofSetColor(255);
    finalFbo.draw(0, 0);
}
```

The result is that the window goes fullscreen, but not the content.
full1

While trying to fix this, I've substitute OF_FULLSCREEN with OF_GAME_MODE
```cpp
settings.windowMode = OF_GAME_MODE;
```
When the setupt function is called, this line
```cpp
 cout << ofGetWidth() << endl;
```
prints 1850

And the screen is still cut
gamemode

In both case, resolution looks good and the window appears without decorations.

As suggested by @hamoid in the forum, I have tried to use ofGetScreenWidth, but the the result does not change. Just the background color is different.

OF_FULLSCREEN
screenshot from 2018-02-24 10-49-26

OF_GAME_MODE
gamemode

cuinjune:
* ofQTKitVideoGrabber: fix includes

* ofAppEGLBasewindow: fix includes

* fixing include optimization for osx mk, ios and linux arm

* fixing ofAppEGLWindows includes

* osx: don't use basename, seems to crash

* ofxiOS: fix includes optimizations for tvOS

* osx: trying to fix crash on tests script with basename

* trying to fix osx basename crashing by upgrading coreutils

* ofAppNoWindow: init events pointer

* Fixes for includes optimizations on windows

* unit tests: fixes for includes optimizations

* fix vs project for includes optimizations

* fixes for msys2 and vs includes optimizations

* removemissing binary lib from repo

* ofxUnitTests fixes for appveyor after includes optimizations

* ofxUnitTests: fixes for includes optimizations

* linux run_tests: don't use gdb to run examples by default

* fix network tests on vs

* testing qbs linux builds

* fixing qbs ci tests

* qbs ci tests: cache linuxbrew

* linux ci qbs: don't try to install qbs if it still exists

* travis: fix linux qbs

* qbs ci: testing cache

* Fixes .gitignore template stripping #5792

* fixes for qbs ci tests

* fixing qbs ci tests

* fix travis for qbs cache

* fixing qbs builds

* fixing qbs ci tests

* fixing qbs ci tests

* fix qbs ci tests

* fixing qbs ci cache

* fixing qbs ci builds

* fixing qbs ci builds

* fixing qbs builds

* fixing qbs builds

* fix qbs builds

* comment out qbs builds by now

* [Docs] - openframewroks/openframeworks - README.md - minor tweaks (spelling, grammar, syntax, writing)

[Docs] - openframewroks/openframeworks - README.md - minor tweaks (spelling, grammar, syntax, writing)

* add missing ofBaseApp.cpp file into xcode prj

* Added make_preferred() to paths where necessary.

* travis: fixing binary upload

* fixing travis upload libs in osx

* fixing binaries upload

* fixing binaries uploads

* fixing binaries upload on ci tests

* add ccache to ci tests to make them faster

* adding ccache to ci tests

* fixing ccache in ci tests for linux arm

* fixing ccache for linux arm

* fixing ccache in ci tests for osx makefiles

* makefiles: don't set CC, CXX.. if already set in env var

* [Docs] - openFrameworks/openFrameworks - README.md - minor tweaks - changing small details in markdown language

[Docs] - openFrameworks/openFrameworks - README.md - minor tweaks - changing small details in markdown language

* fixing ccache for linux arm, osx, ios and tvos

* fixing ccache for armv7

* fix osx ccache and add parallel builds for run tests

* fixing ccache for ios, tvos and checking armv7

* testing ccache on armv7

* fixing ccache for osx, ios, tvos and linux armv7

* fixing osx ccache

* fix ccache config for armv7 and android

* fixing osx ccache

* fix travis config wrongly commited from older version

* fix ccache for osx

* refactor ccache config for osx based targets to ccache.sh script

* fixing osx based targets ccache

* adding ccache to emscripten

* fix ccache in osx

* adding gradle cache for android

* fixing ccache for osx

* fix ccache for emscripten

* fixing ci cache for android and emscripten

* fix makefiles

* fixing ccache and adding to android

* makefiles: fix includes for local freetype

* fixing android builds after applying ccache

* travis: show cache status at the end

* fix cache for armv7

* ccache won't work for android with the experimental gradle plugin

* makefiles: bring back recursive search paths

* fix ccache builds for emscripten ios and tvos

* exec permissions for ios and tvos install scripts

* fixing ccache script for ios

* fix a problem in linux arm7 that happens intermitently

* ccache fix for osx

* fix ci tests for armv7

* add ccache to msys2

* appveyor: trying to enable ccache for msys2

* fixing ccache for msys2

* fixing msys2 ccache

* fixing appveyor

* emscripten: trying to install latest ccache from .deb

* fixing msys2 ccache and output stats at end

* ccache: correct settings for msys2

* testing newest ccache for emscripten

* fixing ccache for msys2

* fixing msys2 ccache

* fixing msys2 ccache

* fix msys2 ccache

* fixing ccache for msys2

* msys2 run_tests: -j2

* appveyor: fixing msys2 ccache

* Update .appveyor.yml

* reset ccache stats before compiling

* trying to cache objs for emscripten instead of using ccache

* appveyor: debuging ccache

* debugging ccache on appveyor

* fixing ccache

* testing emscripten cache

* fixing emscripten cache

* debuging emscripten ci cache

* fixing appveyor ccache

* fixing ccache for emscripten

* fixing emscripten cache

* appveyor: fixing cache msys2

* fixing emscripten ccache

* travis: run ccache -s at beginning too to check that cache was downloaded correctly

* travis: install latest ccache for linux

* fix travis config syntax

* trigger new build

* fixing emscripten ci

* Fix emscripten install

* emscripten: trying to update emscripten to check if we can use ccache

* trying to fix ccache on emscripten

* disable ccache for emscripten by now since it's not working

* add back ofBaseTypes which includes each modules base types to not break so much code

* fix download_libs scripts to work with new packages bundles

* upload_of_lib.sh: fix upload for osx

* linuxarmv7 fixing ci tests

* armv7 ci: quiet wget

* fixing armv7 ci

* fixing linuxarmv7 ci tests

* upload_lib: fix uploading library after tests on osx

* fix linuxarmv7 ci install script

* fixing linuxarmv7 ci install script

* linuxarmv7 ci tests: fix age detectin for cached image

* fixing linux armv7 ci tests cache age detection on install

* fixes for linux armv7 ci tests install

better checks for cache age and use absolute paths

* ci upload_of_lib: fix android final binary upload

* Fixing hideCursor() for Fullscreen on Linux

This should behave well on all other platforms as well.
On Linux the cursor does not disappear in fullscreen, guessing this has to do with how GLFW_CURSOR_HIDDEN is handled internally, in window mode it work fine but in fullscreen GLFW_CURSOR_DISABLED is necessary.

* ofPath: don't tesselate empty paths + fix tessellation cache logic.

Closes #5859

* create_package: output dots while compiling PG

* create_package: download linux lib from ci server instead of compiling full OF when building PG

* create_package: Compile PG in debug  [ci skip]

* create_package: debug problem when compiling PG

* create_package: detect if we are on server (very naive) and log to logs or console [skip ci]

* create_package: remove latest OF library before downloading it

* nightly builds: don't include commit hash in download page

it can be incorrect cause we are downloading the library from the
server

* nightlybuilds: bring back the commit hash [skip ci]

we only download the OF library to compile the PG but the commit hash
is the correct one

* create_package: android doesn't need paths.make anymore [skip ci]

* create_packages: create packages for vs 2015 & 2017

* create_package: remove reference to eclipse docs

* linuxarmv7 ci: echo dots while downloading so it doesn't timeout

* create_package: fix download_libs parameters [skip ci]

* linuxarmv7 ci tests: fix install phase, bring back wget logs to avoid timeout

* linux64 -> linux64gcc4

* travis: use gcc4 suffix for download_libs

* addons ci: download gcc4 package for linux [skip ci]

* ofxPoco: addon_config remove extra quotes that where breaking the addon

[skip ci]

* nightlybuilds: send mail on completion

* nightly builds: add mail address for reports

* ofAppGLFWWindow: better handling of window mode

Fixes #5861
Also fixes a problem where setting GAME_MODE using a window settings
struct would change the screen resolution to 1024x768 when the size
wasn't set.

To fix this we are making width and height in window settings protected
which will break some code but the fix is pretty straightforward, just use
setSize instead of directly assigning values to width and height

* ofAppGlutWindow: fix window settings after changing width/height to protected

* ofAppRunner.h: fixes for windowSettings width/height protected

* EGL window: changes for widnow settings width/height protected

* ofAppEmscriptenWindow: fixes for window settings width/height as protected

* fontShapesExample: get character as fill and contour.

Fixes: #5863

* fonstShapesExample: use contour to draw font point by point

* remove deprecated ofTtfCharacter and just use ofPath instead

Closes #4942

* fix for OS X OF_TTF_SANS default on 10.13 closes #5865

* soundBufferExample: use a sound stream object instead of global

* ofNodeExample: fix starting pov

* fix download_libs.ps1 script

+ script was deleting temporary libs folder "32", "64" repeatedly
meaning only the libs from the very last .zip slice of libs would
survive the cull.

Since the temporary folders "32/64" get deleted automatically as they
are part of the "cull list" when preparing the libs directory for
upgrade this additional delete routine is not needed in the first place.

+ also added an echo to instruct users on how to install libs
for visual studio 2017, instead of the default, visual studio 2015.

+ also added an echo telling users that that the script finished
successfully.

* set vs2017 as default platform

+ platform libraries for vs2017 will be installed if script called
without paramters.

+ echo message on how to override this default behaviour in case users
want to install legacy vs2015 libraries.

* remove http request from regex example

* scale image

* remove obsolete example

* cleanup

* flip plane primitive tex coordinates

* flip sphere primitive tex coordinates

* flip ico sphere primitive tex coordinates

* flip cylinder primitive tex coordinates

* flip cone primitive tex coordinates

* flip box primitive tex coordinates

* update 3d primitive example

* fix ofCamera-drawFrustum

+ fixes bug when drawing ofCamera Frustum which appeared recently.
+ simplifies code drastically
+ adds comments focusing on the "how" rather the "what".

* fix ortho projection matrix calc

+ ortho matrix must not use viewport offset for its own calculations,
only viewport dimensions.

This is in line for how projection is calculated for perspective
matrices, where only the aspect ratio of the viewport is used.

* modernise advanced3DExample

+ remove redundant sub-class of ofCamera, OrthoCam
+ clarify some variable names
+ clean up camera assignments
+ mention built-in frustum draw method in code comment

* fix save screen

Use pbo only in programmable renderer to avoid problems with intel
Draw image on image saver example

* imageSaveExample: remove images from data folder

* create_package: don't include transformFeedbackExample in osx or linux arm

* PG latest commit

* apothecary latest commit

* remove reference to OrthoCam.h

+ removes a leftover reference to deleted header file

* fix box primitive winding order

aligns box primitive triangle winding order so that it is consistent
with the winding order of other primitives, that is, when viewed through
an ofCamera, and GL_CULL_FACE is enabled with glFrontFace(GL_CW) it will
look correct, the front faces visible, the back faces invisible.

All other oF primitives appear to follow this convention.

* ofImage: add return value to save image

* ofImage: fix save image returning false after some correct cases

* Fix for SSL certificate issue when load a URL

* Add unimplemented methods to ofTexture

* Remove unused methods

* ofAppGLFWWindow: the size reported on setup was wrong.

Fixes #5904

* ofAppGLFWWindow: calculate pixel scale even in fullscreen

but don't try to position the window

* ofAppGLFWWindow: don't reposition window if in fullscren or gamemode

* Add gradle wrapper to android studio templates

* gradlew in templates didn't have execution permissions

* Fixes #2449 ofxiOSKeyboard position

* Fixes #4596 - Medium P to High P for Shader Floating Point Precision

* Fixes Update and Draw being called before Setup


arturoc:
Hey, no idea what this is, the title doesn't explain anything and it's conflicting on a ton of files. I'm going to close it but feel free to keep commenting on this issue

cuinjune:
Sorry It was my mistake. Please ignore this.

Issue #6345, opened 2019-07-24
Can't hear sound using ofSoundStream on Emscripten (Minimal Example Included)


cuinjune:
I found out `ofSoundStream` based examples don't work properly on Emscripten.
Here' my minimal example code that works on macOS but doesn't work on Emscripten.

**`ofApp.h`**:

```
#pragma once

#include "ofMain.h"

class ofApp : public ofBaseApp{
    public:
    void setup();
    void update();
    void draw();
    
    void keyPressed(int key);
    void keyReleased(int key);
    void mouseMoved(int x, int y);
    void mouseDragged(int x, int y, int button);
    void mousePressed(int x, int y, int button);
    void mouseReleased(int x, int y, int button);
    void mouseEntered(int x, int y);
    void mouseExited(int x, int y);
    void windowResized(int w, int h);
    void dragEvent(ofDragInfo dragInfo);
    void gotMessage(ofMessage msg);
    
    void audioOut(ofSoundBuffer & buffer); //only this is added
};
```
**`ofApp.cpp`**: (only relevant methods)

```
#include "ofApp.h"

//--------------------------------------------------------------
void ofApp::setup(){
    
    ofSoundStreamSettings settings;
    settings.numInputChannels = 0;
    settings.numOutputChannels = 2;
    settings.sampleRate = 44100;
    settings.bufferSize = 512;
    settings.setOutListener(this);
    ofSoundStreamSetup(settings);
}

//--------------------------------------------------------------
void ofApp::audioOut(ofSoundBuffer & buffer)
{
    for (size_t i = 0; i < buffer.getNumFrames(); ++i)
    {
        buffer[i*buffer.getNumChannels()    ] = ofRandom(0, 1) * 0.1;
        buffer[i*buffer.getNumChannels() + 1] = ofRandom(0, 1) * 0.1;
    }
}
```
Result: When I run it, it says `Exception thrown, see JavaScript console`.
<img width="1431" alt="3e4b709e073a165056de62f19eb2338a3dd23972" src="https://user-images.githubusercontent.com/24947986/61776416-edf0ed00-ae35-11e9-90d0-1ef06d7448c5.png">

In the Javascript console, I get the following error message: 

> Uncaught TypeError: Runtime.dynCall is not a function at ScriptProcessorNode.stream.onaudioprocess

[...]

PR #6347, opened 2019-07-27
Fix Uncaught TypeError: Runtime.dynCall

cuinjune:
The Runtime object has been removed for quite some time at this point. 

So `Runtime.dynCall` should be changed to just `dynCall` in `library_html5audio.js` file.

Before the fix, it generated the following error: (No sound)

> Uncaught TypeError: Runtime.dynCall is not a function at ScriptProcessorNode.stream.onaudioprocess

After the fix, I could properly run examples that use `ofSoundStream` on a browser through Emscripten.

Relevant issue: https://github.com/openframeworks/openFrameworks/issues/6345#issue-472141591


arturoc:
thanks!

PR #6355, opened 2019-08-18
Bugfix - last commit hash written before build successful.

ofTheo:
This isn't the cause of the current nightly build errors, but it hides the actual cause. 

Because lasthash.txt is written before the build is successful it means it's not trying to do the build each night even though the actual packages are older than the last commit.  

Moving it to the end of the script will mean that our log file will show the real error messages preventing the build.


arturoc:
thanks, please send PRs from your own repo otherwise tests are triggered twice and extra branches are created in the OF one that i need to delete

ofTheo:
ah - yeah! thought I had.  
Will do :)

PR #6357, opened 2019-08-19
bugfix changes the create_package script to call executable generated by DebugNoOF

ofTheo:
Project generator seems to be compiling fine now after installing poco. 
This seems to be the last step to getting the nightlies to build again. 

compiling the PG with DebugNoOF produces bin/projectGenerator_debug but a few lines down the script calls bin/projectGenerator

### DIFF
--- a/scripts/dev/create_package.sh
+++ b/scripts/dev/create_package.sh
@@ -189,7 +189,7 @@ function createProjectFiles {
         else
             pg_platform=$pkg_platform
         fi
-        ${main_ofroot}/apps/projectGenerator/commandLine/bin/projectGenerator --recursive -p${pg_platform} -o$pkg_ofroot $pkg_ofroot/examples > /dev/null
+        ${main_ofroot}/apps/projectGenerator/commandLine/bin/projectGenerator_debug --recursive -p${pg_platform} -o$pkg_ofroot $pkg_ofroot/examples > /dev/null
     elif [ "$pkg_platform" == "linuxarmv6l" ] || [ "$pkg_platform" == "linuxarmv7l" ]; then
         for example_group in $pkg_ofroot/examples/*; do
             for example in $example_group/*; do


arturoc:
hey theo, i find pretty strange that we've been using a wrong version of the PG for years, can you check when this change was introduced? also it seems likely that using a debug version of the PG to generate all the project files for all the examples for every platform might take for ever so i'm not sure it's the way to go. it was surely me who did that change cause the DebugNoOF target is not standard but don't remember exactly why i chose to use debug

ofTheo:
Yeah. I was trying to figure out why/how it stopped working and its really strange. 

The DebugNoOF has been that way for 2 years now: https://github.com/openframeworks/openFrameworks/blame/master/scripts/dev/create_package.sh#L177 

And since that time the script has called projectGenerator not projectGenerator_debug. 

My first guess was that DebugNoOF wasn't actually producing a binary with _debug at the end. 
So it just worked because the debug command line binaries had the same name as the release ones. 

But looking here: https://github.com/openframeworks/openFrameworks/blame/master/libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk 

It has been putting _debug at the end for a long time now. 

I looked back at the commits between March 24 and now and nothing in the commits seemed like it would relate to the bin name changing.  

This was close to that time - but I don't think it would affect anything. 
https://github.com/openframeworks/openFrameworks/commit/9baea23a9a635270ff4e08282dfdcb4b62175e31#diff-c6952bc7cc674bb2a15ff505b4683336 

I'll dig in a bit more. Nothing really obvious stood out though. 
The nightlies are buildings again now with this fix so at least its working. 

Def could be better to have it in release though - this felt like the safer change at least for now, wasn't sure if there was a reason for the DebugNoOF so didn't want to touch that :)

arturoc:
mmh yeah that's strange in any case it seems it's building fine now and it's only taken 45 minutes so it's not that terrible, let's leave it like this, i think i used debug cause building all of OF took really long but we already build it for addons tests and store it somewhere so the script is downloading the compiled library and using it i believe. thanks for looking into it!

PR #6361, opened 2019-08-26
small improovement for getResampledBySpacing

jonasfehr:
To avoid bigger gaps at the end of the resampled polyline, adding almost a full spacing will help adding a rather small last space than a big one. The last point is replaced anyway.


jonasfehr:
A more elegant solution, indeed. It just  needs <= in the for-loop to work if the spacing is an accurate fit, as if you use getResampledByCount.



    float f=0;
    for(f=0; f<=totalLength; f += spacing) {
        poly.lineTo(getPointAtLength(f));
    }
    
    if(!isClosed()) {
        if( f != totalLength ){
            poly.lineTo(points.back());
        }
        poly.setClosed(false);
    } else {
        poly.setClosed(true);
    }

_________________

www.jonasfehr.ch

Furesøvej 161
2830 Virum

mail@jonasfehr.ch
+45 52 777 432

PR #6366, opened 2019-09-05
fixes ofPolyline to prepare oF for glm 0.9.9.5

npisanti:
as we discussed in https://github.com/openframeworks/apothecary/pull/136

### DIFF
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -151,7 +151,7 @@ glm::mat4 ofCamera::getProjectionMatrix(const ofRectangle & viewport) const {
 	const_cast<ofCamera*>(this)->calcClipPlanes(viewport);
 
 	if(isOrtho) {
-		return glm::translate(glm::mat4(), {-lensOffset.x, -lensOffset.y, 0.f}) * glm::ortho(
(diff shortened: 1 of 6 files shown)


ofTheo:
Hi @npisanti - I think there are still 3 glm::lerp 's left in the file as far as I can tell. 
I am guessing those will have the same compilation error.

arturoc:
the problem is this is a templated class, so nothing gets really compiled until it's used. Perhaps it would be useful to include a test in the tests folder that uses as many methods as possible from a polyline that way we can know from the ci tests if it's really working. also if something changes in future versions of glm or someone changes ofPolyline somehow

oxillo:
There are also problems with the glm::mat constructors which I understand do not initialize values anymore.
@npisanti : I've just submitted a PR to your branch with the fixes (including missing lerp as mentionned by @ofTheo). I've managed to compile all examples under MSYS2
Can you merge it ?

npisanti:
yes, i just changed the things needed for compiling, i substituted glm::lerp only in those points as I thought that as it was compiling ALL the class not just some methods

just merged your PR

bakercp:
Is this ready to merge @arturoc ?

arturoc:
thanks!

PR #6367, opened 2019-09-06
Feature SSBO

thesstefan:
Hey! I've recently wanted to pass an array of variable size to a shader. After some research I found out that the only way of doing that is by using an SSBO so I coded a quick implementation for it. 

(The change to ofBufferObject was made because I couldn't figure out how to set the target to GL_SHADER_STORAGE_BUFFER without having to call bind on an array buffer first)

### DIFF
--- a/libs/openFrameworks/gl/ofBufferObject.cpp
+++ b/libs/openFrameworks/gl/ofBufferObject.cpp
@@ -6,10 +6,10 @@
 
 using namespace std;
 
-ofBufferObject::Data::Data()
+ofBufferObject::Data::Data(GLenum target = GL_ARRAY_BUFFER)
 :id(0)
 ,size(0)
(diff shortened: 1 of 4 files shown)


arturoc:
Thanks, there's already an example of how to use SSBOs directly with ofBufferObject.https://github.com/openframeworks/openFrameworks/tree/master/examples/gl/computeShaderParticlesExample  it's simply a matter of binding them with the right target so not sure we need to have an extra class for this the same way we don't have one for uniform buffers or texture buffers or other specific types of buffers.

This is also a feature that doesn't exist in most of the platforms we support so it's failing most of the tests.

I'm going to leave this open for a while in case others have a different opinion but i would vote to not include it and just leave ofBufferObject as the universal way to use all kinds of buffers

Issue #6372, opened 2019-09-25
Run scripts cannot handle spaces in folder names


fred-dev:
The scripts for downloading libs fail if there are spaces in the fodler names. I have a use case where I cannot control the folder names, It would be great if the scripts could handle spaces.

bakercp:
This goes for apothecary too.

ofTheo:
Are you able to tell which lines this fails on? 
I took a quick look and didn't see anything that stood out.

fred-dev:
For download_libs.sh for OSX and IOS I get an error on  line 141. I don't get further than that.

kkshmz:
Hi, I found that double quoting works quite nicely on OSX, I can do pull request if you want to.
cd "$SCRIPT_DIR"  on line 141

with that being said, I don't know how this would work on other platforms as you could potentially pipe a sed arguement to escape the spaces. 
`SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd | sed 's/ /\\ /g' )"`

arturoc:
yes adding quotes should solve the problem in every platform in any case the script is tested in the CI servers so if you want to send a PR we'll see if at least it keeps working fine on every platform

fred-dev:
Double quotes were added in the scripts/dev/download_libs.sh script, but for this to work on OSX they also need to be added to the scripts/ios/download_libs.sh and scripts/osx/download_libs.sh scripts.

danomatika:
Yes. Path variables should always be escaped in the apothecary scripts. Hell, for shell scripts in general.

PR #6374, opened 2019-09-30
add double quotes to allow for folder names with spaces

kkshmz:
Pull request for 
#6372


kkshmz:
@arturoc Hey, it seems to be working most of the time, is there a way I could debug the one that didn't pass? (linuxarmv7)

arturoc:
That it's not related to your change which surely works on linux arm7 if it works on other linus platforms. it just failed cause of a timeout downloading some packages, i've just restarted that job and hopefully it'll work this time. thankS1

kkshmz:
Thanks for the reply, seems like the IL server is the only one down, but should work if we tried later.

kkshmz:
IL Server now seems to be up, when you have time!

PR #6394, opened 2019-10-13
bugfix. fixes error with Xcode build due to depracted function error.

ofTheo:
This is an error I saw in the nightly. 
Just switching the deprecated function to the replacement function name.


ofTheo:
Hmm don't merge just yet. 
Looks like pugixml is older for the linux64 ci build in Travis and missing the newer api function which is in 1.9 ( what should be installed with apt-get ) .

arturoc:
you'll need to add an ifdef to check for the pugixml version so it compiles on the server. the ci machines have relatively old versions of ubuntu so the versions of the libraries that come with it are old as well. that way also we keep compatibility with older versions of the library if someone is using an old linux version

ofTheo:
Ahh okay that makes sense. 
Thanks!! I'll add that check.

PR #6398, opened 2019-10-18
fixes glm::quat to use good default values. 

### LINKED ISSUE #6397: glm 0.9.9.5 merge errors
Since we merged in glm 0.9.9.5 looks like there are a few different glm related errors. 
- I think this linux64 one might be related to this not being defined: GLM_HAS_TEMPLATE_ALIASES, https://travis-ci.org/openframeworks/openFrameworks/jobs/599075106 
- MacOS and Windows VS/MSYS seems to be an issue with the orientation failing in the tests `tests\math\ofnoderegressiontests\src\main.cpp: 136` and `tests\math\ofnoderegressiontests\src\main.cpp: 142` 

```
[  error ] value1: 0, 0, 0, 1
[  error ] value2: 0, 0, 0, 0
[  error ] 	orientation failed 
```

From what I have read: "(0,0,0,0) is not a valid rotation quaternion, as they are required to have length 1." 

The 0,0,0,0 value is from glm::quat() - see: https://github.com/openframeworks/openFrameworks/blob/master/tests/math/ofNodeRegressionTests/src/main.cpp#L136 
	
`ofxTest(aprox_eq(n3.getOrientationQuat(), glm::quat()), "\torientation");`

So maybe glm::quat() switched from 0,0,0,1 to 0,0,0,0 in 0.9.9.5 and thats why we see the failing tests? 

Hmm. See: https://github.com/g-truc/glm/issues/417 
It could be that this is now defined as default hence the issue: GLM_FORCE_NO_CTOR_INIT 

Actually looking at the 0.9.9.5 header it looks like we are missing this #define 

```
#if defined(GLM_FORCE_CTOR_INIT) && GLM_HAS_INITIALIZER_LISTS
#	define GLM_CONFIG_CTOR_INIT GLM_CTOR_INITIALIZER_LIST
#elif defined(GLM_FORCE_CTOR_INIT) && !GLM_HAS_INITIALIZER_LISTS
#	define GLM_CONFIG_CTOR_INIT GLM_CTOR_INITIALISATION
#else
#	define GLM_CONFIG_CTOR_INIT GLM_CTOR_INIT_DISABLE
#endif
```

If `GLM_CONFIG_CTOR_INIT == GLM_CTOR_INIT_DISABLE`
Then the quat is not set to 0,0,0,1 

```
#	if GLM_CONFIG_DEFAULTED_FUNCTIONS == GLM_DISABLE
		template<typename T, qualifier Q>
		GLM_FUNC_QUALIFIER GLM_CONSTEXPR qua<T, Q>::qua()
#			if GLM_CONFIG_CTOR_INIT != GLM_CTOR_INIT_DISABLE
			: x(0), y(0), z(0), w(1)
#			endif
		{}
```

So... to end a very long issue. 

**Solution?**
- I think forcing #define for GLM_FORCE_CTOR_INIT or making sure it doesn't end up as GLM_CTOR_INIT_DISABLE should fix the test issues.  this seems related: https://github.com/g-truc/glm/issues/735 
- And defining GLM_HAS_TEMPLATE_ALIASES should fix linux64.

ofTheo:
The tests were the only part of OF using the glm::quat() with no arguments so fairly simple to fix. 
Partially closes #6397

### DIFF
--- a/tests/math/ofNodeRegressionTests/src/main.cpp
+++ b/tests/math/ofNodeRegressionTests/src/main.cpp
@@ -133,13 +133,13 @@ public:
 			ofxTest(aprox_eq(n3.getLocalTransformMatrix() * zero, n2.getGlobalTransformMatrix() * zero), "\tmatrices");
 			n3.setParent(n2, true);
 			ofxTest(aprox_eq(n3.getPosition(), glm::vec3(0,0,0)), "\tposition");
-			ofxTest(aprox_eq(n3.getOrientationQuat(), glm::quat()), "\torientation");
+			ofxTest(aprox_eq(n3.getOrientationQuat(), glm::quat(0,0,0,1)), "\torientation");
 			ofxTest(aprox_eq(n3.getScale(), glm::vec3(1,1,1)), "\tscale");
 			ofxTest(aprox_eq(n3.getGlobalTransformMatrix() * zero, n2.getGlobalTransformMatrix() * zero), "\tmatrices");
 			n3.setGlobalPosition(n3.getGlobalPosition());
 			n3.setGlobalOrientation(n3.getGlobalOrientation());
 			ofxTest(aprox_eq(n3.getPosition(), glm::vec3(0,0,0)), "\tposition");
-			ofxTest(aprox_eq(n3.getOrientationQuat(), glm::quat()), "\torientation");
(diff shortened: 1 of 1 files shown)


ofTheo:
ugh - this went from OF instead of my repo.

PR #6400, opened 2019-10-18
ofPolyline::removeVertex(...)

### LINKED ISSUE #6375: removeVertex() method is missing in the ofPolyline class
It seems strange that _ofPolyline_ has an _addVertex()_ method, but not a _removeVertex()_. This method [has been already defined](https://github.com/openframeworks/openFrameworks/blob/b34f9a936badb4b9667daaa90d16e719fab8eb03/libs/openFrameworks/3d/ofMesh.inl#L351) in _ofMesh_ and does exactly what one would expect:

`vertices.erase(vertices.begin() + index);`

This could be especially useful for beginners who want to remove a vertex without dealing directly with iterators and vectors.

bakercp:
Resolves #6375


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 
 CORE
 ----
+
```
arturoc: No need to add changelog, i'm adding tags when i merge PRs so we can autogenerate it later. No need to remove it from this one either :) just so you know for next time

libs/openFrameworks/graphics/ofPolyline.inl
```diff
@@ -93,6 +93,19 @@ void ofPolyline_<T>::insertVertex(float x, float y, float z, int index) {
 }
 
 
+//----------------------------------------------------------
+template<class T>
+void ofPolyline_<T>::removeVertex(int index) {
+	if(index >= points.size()){
```
arturoc: this kind of bounds check only happens in c++ in debug mode, not sure we should have it

bakercp: I just copied this pattern from ofMesh.  I'd prefer to remove it.  Should we keep it for consistency?

arturoc: oh, ok, letr's leave it then although it's weird

PR #6401, opened 2019-10-18
Bugfix fix glm default

ofTheo:
didn't realize constructor was w, x, y, z. ;) 
this should help with #6397


ofTheo:
so this is still failing on linux64 even with qualifier.hpp which has the template defines. 

a fairly straightforward fix is to not use any of the templated types. 
we are only using them in a few places.  

`tvec3<T, P>` becomes `vec<3,T, P>` 
`tmat3x3<T,P>` becomes mat<3,3,T,P>`

etc

PR #6405, opened 2019-10-20
[tiny fix] brush up ofAVFoundationGrabber

2bbb:
* add init for grabber was not init.
* replace substitute by nil for unused allocation.


ofTheo:
Thanks!

PR #6410, opened 2019-10-26
ofxSVG : use libxml2 system-package on MSYS2

oxillo:


oxillo:
already merged with #6413

PR #6413, opened 2019-10-30
Mingw64 support

oxillo:
use OpenAL for sound support.
FMOD still used on mingw32.


arturoc:
thanks! it's great to have this merged finally!

Issue #6419, opened 2019-10-31
glitch in ofxGuiColorPicker


arturoc:
Sometimes the color picker draws the saturation slider with the last vertex extending outside of the control

roymacdonald:
@arturoc does this happen under any special condition?

roymacdonald:
sent PR with fix

arturoc:
I've seen this happening after drawing two images, actually a video and an image and then a gui with an slider and a color picker. all of them coming from parameters

roymacdonald:
I was looking at this issue and I still don't get a consistent behavior from it. 
It seems that some points of the mesh have garbage in the z value of its position. This might be related to the GLM changes done recently. Just setting to zero the z value at the ofRectangle-s constructor seems to work.
Does this make sense to you @arturoc ?

This might be also relate to [this](https://github.com/openframeworks/openFrameworks/issues/6435)
I'll test this a bit more before trying a fix.

arturoc:
yes that sounds very likely. i also was getting this randomly so an uninitialized variable sounds very possible

roymacdonald:
I still don't get to a conclusive test with this issue, although I think that we should initialize ofRectangle's position to 0 in order to avoid any other issue with it. What do you think?

tgfrerer:
@roymacdonald can you point me to a test case which still fails?

roymacdonald:
Lol @tgfrerer  you posted that PR just when I was saying that it should be done.

PR #6426, opened 2019-11-02
fixed ofxColorPicker glitch

roymacdonald:
fixes https://github.com/openframeworks/openFrameworks/issues/6419
@arturoc


arturoc:
this doesn't make sense the default mode is already OF_PRIMITIVE_TRIANGLES. going to reopen but let me know if i'm missing something

roymacdonald:
Nah, I got something odd too with the sliders when right clicking. But probably related to the same thing.

PR #6429, opened 2019-11-04
ofxGui: Updating events to allow multiwindow functioning

roymacdonald:
So currently if you try to use an ofxGui on a multiwindow app and draw it on a window other than the main window, you need to setup the gui before the main app setup is run, which is quite cumbersome.
Besides this, if you have an app and want to open a new window and draw in it an already setup gui the mouse events will not work. You need to disable and reenable mouse events while the window on which you want to draw the gui is the currently renderer window. 

So, the fix for this is quite simple and effective. It takes the same idea from the ofEasyCam events.
 
It stores a pointer to the ofEvents() and it checks against it on every draw call so if it changed it will reregister the mouse events.

This allows:
*  to avoid [this nasty line ](https://github.com/openframeworks/openFrameworks/blob/master/examples/windowing/multiWindowOneAppExample/src/main.cpp#L22)
* automatically switch to different windows. 
* even render the same gui on different windows at the same time while having mouse interactions on both.
@ofTheo @arturoc @bakercp @ofZach

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -138,8 +138,19 @@ void ofxBaseGui::unregisterMouseEvents(){
 	ofUnregisterMouseEvents(this, defaultEventsPriority);
 	bRegisteredForMouseEvents = false;
 }
-
+void ofxBaseGui::setEvents(ofCoreEvents & _events){
+	if(&_events != events){
+		bool wasMouseInputEnabled = bRegisteredForMouseEvents;// || !events;
+		unregisterMouseEvents();
+		events = &_events;
+		if (wasMouseInputEnabled) {
+			// note: this will set bMouseInputEnabled to true as a side-effect.
+			registerMouseEvents();
+		}
+	}
+}
 void ofxBaseGui::draw(){
+	setEvents(ofEvents());
 	if(needsRedraw){
 		generateDraw();
 		needsRedraw = false;
--- a/addons/ofxGui/src/ofxBaseGui.h
+++ b/addons/ofxGui/src/ofxBaseGui.h
@@ -71,8 +71,8 @@ class ofxBaseGui {
 		static void loadFont(const ofTrueTypeFontSettings & fontSettings);
 		static void setUseTTF(bool bUseTTF);
 
-		void registerMouseEvents();
-		void unregisterMouseEvents();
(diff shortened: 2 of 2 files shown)


moebiussurfing:
sorry if it's not related to this but:

I noticed that when changing window mode WINDOW/FULL SCREEN with  ofSetFullscreen(MODE),
ofxGui does not receive the mouse coordinates fine when switching from full-screen to window mode.
(Window goes like 20 pixels above the top border in the vertical y axis. The mouse pointer is like displaced vertically 20 pixels below.  In full-screen mode, it works.)

I need to click the window top title bar to resize again and then the mouse interacts fine with ofxGui again.

I tried to remove and add again mouse listeners without success.

This is on macOS, using only one window, and up to date repository.

roymacdonald:
Hi @moebiussurfing 
I think this has to do with a change in the coordinate system rather than an mouse events problem.
As for what you say, those 20 pixels on the y axis seem to be there to compensate the window's top bar.
I am not sure if there is a specific window event for when it changes to/from fullscreen. Although I guess that the window size event will work. So, register to this event so when it triggers then you just flag for a re-render. I'll try out this and let you know.

roymacdonald:
@moebiussurfing can you open a new issue with this.
As well, I tested it on macosx 10.12, xcode 8.3.2 and the latest OF code, and it runs fine. Can you please share the code you are using in which this error happens.
best

moebiussurfing:
ok @roymacdonald , I'll recheck my code. nevermind about it. thanks

I am using a dual monitor setup if this could mean something...
also, the ofxPanels are into some classes, not in ofApp. And the screen mode switching is called into ofApp.

Not sure what you mean about 'change in the coordinate system', but drawing/ofTranslate push/pop looks fine...

roymacdonald:
@moebiussurfing the dual monitor probably plays some role on this. Do you see this happening on a specific monitor(main or secondary)?

roymacdonald:
by cahnge in the coordinate system I mean that there could be some sort of change as in fullscreen there is not top bar, and the gui drawing is cached and it wont change until you interact with it. It might nnot be the case. I was just guessing. Can you share some kind of test code?

ofTheo:
this looks good to me! 
could you submit this to patch-release so we can include it in 0.11.1 ?

roymacdonald:
sure.

PR #6434, opened 2019-11-05
Bugfix nightly node

ofTheo:
See discussion also in: 
https://github.com/openframeworks/projectGenerator/issues/225

### DIFF
--- a/scripts/dev/create_package.sh
+++ b/scripts/dev/create_package.sh
@@ -395,6 +395,8 @@ function createPackage {
     if [ "$pkg_platform" = "vs2015" ] || [ "$pkg_platform" = "vs2017" ]; then
 		cd ${pkg_ofroot}/apps/projectGenerator/frontend
 		npm install > /dev/null
+        npm install -g n
+        n 6.11.0
 		npm run build:vs > /dev/null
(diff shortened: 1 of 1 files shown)


ofTheo:
hold off on merging. going to try this on the server itself to avoid pushing a bunch of commits.

ofTheo:
Okay so the issue is this:

On the CI server: 
/usr/bin/node is 4.8.7 ( bad / breaks builds) 
/usr/local/bin/node is 6.11.0 ( good / works fine ) 

In the user environment /usr/local/bin/node nvm etc is used. 
In crontab /usr/bin/node is used. 

This is why running nightly_builds.sh from the command line works fine but it breaks in cron. 

Adding this to the crontab *should* fix the issue as it will default to the /usr/local/bin like the user environment. 

`PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin `

Other possible fixes include updating /usr/bin/node on the server. 
Or specifying /usr/local/bin/nvm in create_package.sh but that makes it more platform specific. 

Hoping setting the path in the crontab fixes it all. 
Fun stuff! ;)

PR #6439, opened 2019-11-10
Fixed OpenGL ES2 shader example 08

n1ckfg:
OpenGL ES2 needs texture2DLod() instead of texture2D() in the vertex shader


arturoc:
thanks!

PR #6444, opened 2019-11-12
add default member initialiser for position

### LINKED ISSUE #6419: glitch in ofxGuiColorPicker

tgfrerer:
If `ofRectangle`s field `position.z` was not explicitly set, this would mean that garbage memory most likely remained stored in place of the .z coordinate of a rectangle's position.

Whilst this didn't affect any 2D operations, as soon as the rectangle in question was going to be drawn, its .z position would be fed straight into an ofMesh, and then a vertex shader, and all sorts of mayhem could happen.

Test this with `guiFromParametersExample` when opening the `ofxColorPicker` GUI element after selecting a colour...

By adding a default member initialiser to `.position`, the .z coordinate will receive default 0, which is what any 2D graphical element would be expected to be initialised with.

Fixes #6419

### DIFF
--- a/libs/openFrameworks/types/ofRectangle.h
+++ b/libs/openFrameworks/types/ofRectangle.h
@@ -923,7 +923,7 @@ public:
     ///
     /// \warning The z-component of this position is preserved and can be used
     /// but all ofRectangle operations will ignore the z-component.
-	glm::vec3 position;
+	glm::vec3 position{};
 
     /// \brief The x position of the ofRectangle.
     float& x;


bakercp:
Only comment would be -- this syntax seems inconsistent with other recent changes. Many of the other glm default initialization updates have been initialized in the constructor e.g.

```
ofRectangle::ofRectangle():
    position(0)
{
}
```

tgfrerer:
@bakercp I see - when initialising in the constructors, there is a small chance however, that one or the other *specialised* constructor might get missed if we're not super careful, because each such `glm::*` field must be initialised in all constructors (and specialised constructors) explicitly.

If we, on the other hand, initialise using such a default member initialiser, this covers all current constructors and all possible future constructors.

bakercp:
Agreed, we should probably be doing it this way across the board. My
comment was more about consistency.

ofTheo:
Thanks @tgfrerer ! 
I am sort of surprised this works though.

How does adding the member initializer zero out the values?

This is where the constructor values get zeroed, but in this case we don't have that define enabled: https://github.com/g-truc/glm/blob/master/glm/detail/type_vec3.inl#L12

ofTheo:
Also the more I think about it the more I think we should define the default constructor variable. 
`GLM_FORCE_CTOR_INIT` 

Already the transition from ofVec to glm is quite hard.  
Now with glm types having garbage values unless initialized seems even less friendly to me. 
( eg: look how hard its been to track down the bugs in the core )

I know glm is trying to have parity with glsl but I don't think we should make things more difficult for users or for large projects which are working with the assumption of good initialization of glm types. 

I think  its fine to fix these issues in the core, but I think we should enable `GLM_FORCE_CTOR_INIT` before release and maybe make some notes that in the future GLM may remove this define.   

I would lobby pretty hard to not have them drop it. 
Who wants to have to remember what an identity quat is anyway? :)

arturoc:
i might be missing something but i don't think there's any difference between adding the {} which would call the default constructor and not having it which should also call the default constructor no?

@ofTheo yes let's add that macro once everything is working, my main worry is that it might go away at some point and it would make people introduce errors that are really hard to catch. do you know if this is an experimental extension or is already stabilized?

tgfrerer:
@ofTheo it works because empty braces intialization is a shorthand in cpp for the most type-appropriate zero initialisation: https://en.cppreference.com/w/cpp/language/zero_initialization.

arturoc:
oh i see, i thought {} would call the default constructor.

ofTheo:
@arturoc @tgfrerer - Ha that's new for me too! Thanks for the explanation. 

@arturoc I'll check about that. I think it is not even an extension just part of the configuration of the core of glm. But I'll do some more research before submitting a PR.

ofTheo:
Merging - AppVeyor error doesn't look related. Thanks @tgfrerer

PR #6455, opened 2019-11-15
fix tabs and spaces with rpi makefiles. fix permissions for install_dependcies.sh

ofTheo:
Okay this should be good now. 
Everything new is using tabs ( as with the existing make files ) except `$(info ` calls which need spaces. 

@bakercp @arturoc if this looks good feel free to merge.


arturoc:
thanks!

PR #6456, opened 2019-11-15
bugfix fixes fullscreen issues on RPI. closes #6445

### LINKED ISSUE #6445: RPi with GLFW won't set correct size coming out of fullscreen.
Once #6325 is merged newer RPi devices can use GLFW. 

One of the only issues I have noticed is that when toggling fullscreen GLFW continues to use the fullscreen window dimensions when returning the screen width and height ( and frame buffer width and height ). This means that even if the window is back at 1024x768 all of the OF rendering calls think it is at 1920x1080

It seems related to this call. 
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/app/ofAppGLFWWindow.cpp#L798 

Immediately after the above call windowW and windowH are the correct non fullscreen size, but as the window is resizing back to the non-fullscreen the window dimensions never get set to these values, but they stay at the fullscreen values. 

Also all the resize_cb calls are showing the fullscreen sizes. 

I think this is because it takes a few frames to get out of fullscreen and somehow the fullscreen values are getting set for the window shape. I think also it is a bit of a case of X battling GLFW. 

Posting the details here incase someone who is more familiar with Linux / X has dealt with this. 

A fix could be to store the desired window size as we do on other platforms and then use glfwSetWindowSize to restore the window size.   Hoping there is something more elegant though.

ofTheo:
this fixes #6445 

I tried a few different methods first but this is needed due to how many frames it takes the window to restore on the RPI. 

Tested toggling fullscreen from both an app started in windowed and fullscreen mode.

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -312,6 +312,11 @@ void ofAppGLFWWindow::setup(const ofGLFWWindowSettings & _settings){
 
 	windowW = settings.getWidth();
 	windowH = settings.getHeight();
+	
+#ifdef TARGET_RASPBERRY_PI
+	windowRect.width = windowW;
+	windowRect.height = windowH;
+#endif 
 
 	glfwMakeContextCurrent(windowP);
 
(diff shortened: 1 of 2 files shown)


ofTheo:
I don't think this is the most perfect solution, but it gets fullscreen working well for all the cases I've tried and only affects the RPI. 

I think after the release I'll look into some other approaches, maybe there is a way to make X block on changing the window so the resize_cb happens at the right time?  

Happy if anyone has thoughts / suggestions though. 
Def not super familiar with the X calls being used.

arturoc:
looks good. glfw window is a mess by now though, full of this kinds of hacks for each platform just using booleans to check when something is happening. at some point we should turn it into some kind of state machine otherwise it's super difficult to maintain.

PR #6458, opened 2019-11-16
bugfix fixes dylibs not working in qtcreator if consoleApplication is true. 

### LINKED ISSUE #6420: qtcreator projects fail to link

ofTheo:


arturoc:
#6420 is happening in linux so i don't think this will fix it. i'll try to take a look next week

ofTheo:
Oh ha! 
Well found this macOS issue as a result so that's good :) 
I'll leave the linux stuff to you then.

PR #6460, opened 2019-11-19
[[Need discussion]] makes ofRectangle `memcpy`able and slim.

2bbb:
* make slim from 40byte per object to 20byte per object.
* remove virutual from destructor.

## Pros:
This PR makes `ofRectangle ` to `memcpy`, `fwrite` and other low level memory operation friendly.
In addition, this makes less memory usage.
(`float &` use 4/8byte at 32/64bit env. but this PR makes these into `union` doesn't need memory.)

## Cons:
But, this PR removes `virtual` from destructor.
This makes what memory leak may occur in some code that inherit from ofRectangle and be use with polymorphism.

### DIFF
--- a/libs/openFrameworks/types/ofRectangle.cpp
+++ b/libs/openFrameworks/types/ofRectangle.cpp
@@ -9,6 +9,12 @@ using namespace std;
 ofRectangle::ofRectangle() : position(), width(), height() {
 }
 
+ofRectangle::ofRectangle(const ofRectangle& rect)
+: position(rect.position)
+, width(rect.width)
+, height(rect.height)
+{}
+
 ofRectangle::ofRectangle(float px, float py, float w, float h) {
 	set(px,py,w,h);
 }
@@ -774,6 +780,13 @@ ofRectangle ofRectangle::mapClamp(const ofRectangle & coeff) const {
        );
 }
     
+//----------------------------------------------------------
+ofRectangle& ofRectangle::operator = (const ofRectangle& rect) {
+    position = rect.position;
+    width = rect.width;
+    height = rect.height;
+}
+
 
 //----------------------------------------------------------
 ofRectangle ofRectangle::operator + (const glm::vec3 & point){
--- a/libs/openFrameworks/types/ofRectangle.cpp
+++ b/libs/openFrameworks/types/ofRectangle.cpp
@@ -785,6 +785,7 @@ ofRectangle& ofRectangle::operator = (const ofRectangle& rect) {
     position = rect.position;
     width = rect.width;
     height = rect.height;
+    return *this;
 }
 
 
--- a/libs/openFrameworks/types/ofRectangle.cpp
+++ b/libs/openFrameworks/types/ofRectangle.cpp
@@ -6,8 +6,7 @@
 using namespace std;
 
 //----------------------------------------------------------
-ofRectangle::ofRectangle() {
-    set(0,0,0,0);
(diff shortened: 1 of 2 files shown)


arturoc:
looks great but as you say removing the virtual destructor could introduce problems for people inheriting from ofRectangle, which is probably not super common but pretty sure some people are doing it. since it's such a subtle error that won't fail to compile or anything i'm unsure of what to do...

there's also some problems compiling on linux 64

2bbb:
@arturoc 
Exactly what you say. It doesn't make error.

I know 2 addons use inheritance from ofRectangle.
https://github.com/roymacdonald/ofxScaleDragRect
https://github.com/nariakiiwatani/ofxMeshShape

I don't think these suppose to use with polymorphic way, because I think most codes don't use ofRectangle with (shared) pointer and reference.
Maybe, better way is these use simply has-a relation (or private inheritance) and provide features for cast from/to ofRectangle.

Two ways I think:

One solution is making ofRectangle to `final` class solve it safely.
But this is extreme way to solve. it will break backward compatiblity.

Other one is don't mind for above considerations, only notify on release note for version what this is included.

Anyway, I'll try to fix compile error on linux64.

ofTheo:
Adding in @bakercp and @NickHardeman for comment.

ofTheo:
ps: I am guessing not too many people use the virtual destructor of ofRectangle to release memory. ( just a guess though ) 

If that is the only impact it might not be too bad.

bakercp:
Inheriting from `ofRectangle` is somewhat common, as the virtual destructor has signaled to developers that it is ok and expected to do so. Whether or not that was a good design move in the past is certainly debatable, but I think we should be careful not to change `ofRectangle` fundamentally without careful consideration.  Here are some examples of this use pattern:

https://github.com/search?q=%22public+ofRectangle%22&type=Code

I would suggest that we instead work through a discussion of what this idea rectangle structure would contain.  For instance, there has been much discussion over the years of if -- and why -- we need a 'z' component in `ofRectangle` when many of the methods explicitly ignore it.

So, perhaps it would be better to talk about a simpler ofRectangle2D or a similar approach?

roymacdonald:
I tend to use make classes that inherit from `ofRectangle,` as it is so handy, and the removal of the virtual destructor would be bad for me. Yet, why does it need to be removed?
 
I agree with @bakercp as we should be careful, as well as the discussion about the 'z' component of it. At leas I've always felt that it is unnecessary to have the position as a vec3.
If I need a rectangle like thing on a 3D space I would go for an `ofPlanePrimitive` rather than the ofRectangle, although I understand that other people might not.
Maybe having a templated class for which the template is the position type?

NickHardeman:
I also tend to make classes that extend `ofRectangle` so removing it would cause some issues with some of my current code.  I agree with @bakercp and @roymacdonald that the z position component seems unnecessary, but removing it could cause issues for some people. Seems like most of the functions inside ofRectangle convert from `vec3` to `vec2`.

2bbb:
@bakercp @roymacdonald @NickHardeman 

Thanks for comments.

I read some code what @bakercp point out.
And I think many of class inherited from `ofRectangle`  is not assume with polymorphic usage by `ofRectangle*` and `ofRectangle&`.
However I understand what virtual destructor is sensitive topic.
I'll add `virtual` for destructor again.

So, I believe replacing `x` and `y` from `float &` to `union` makes it better performance for memory and processing.

And I agree to z position is unnecessary. but it will give more huge impact. so, I don't talk about it here.

PR #6468, opened 2019-11-21
bugfix. fixes errors if legacy vector math is enabled. closes #6463

### LINKED ISSUE #6463: OF_USE_LEGACY_VECTOR_MATH not working
when OF_USE_LEGACY_VECTOR_MATH is set to 1, i cant compile and get several Shell Script Invocation Error Group like:

`'reference' is a protected member of 'std::__1::__deque_base<ofVec3f, std::__1::allocator<ofVec3f> '`

on macOs 10.14.6 and XCode 11.1 
oF 10.1_osx_release and of_v20191118_osx_release nightly build.

ofTheo:
Wasn't as bad as it initially seemed. 
Just needed the non forward decelerations in ofPolyline. 

Closes #6463

### DIFF
--- a/libs/openFrameworks/graphics/ofPolyline.h
+++ b/libs/openFrameworks/graphics/ofPolyline.h
@@ -7,6 +7,10 @@
 #include "glm/fwd.hpp"
 #include <deque>
 
+#include "ofVec2f.h"
+#include "ofVec3f.h"
+#include "ofVec4f.h"
+
 /// \file
 /// ofPolyLine allows you to combine multiple points into a single vector data
 /// object that can be drawn to the screen, manipulated point by point, and


arturoc:
perhaps this could be ifdef'd with OF_USE_LEGACY_VECTOR_MATH? instead of always including it?

also it would be really useful to run the tests with this macro defined, the script that runs the tests could perhaps uncomment OF_USE_LEGACY_VECTOR_MATH in ofConstants and re-run the tests or we could have other targets for the main platforms (windows, macos and linux) that run the tests with that flag enabled.

that way when someone changes something if that's introducing errors in this version it would be tested. Otherwise there's no way to catch these errors at the moment they are introduced.

if you can look into that it would be great otherwise feel free to merge it as it is

thanks!

ofTheo:
I might leave the idea for another PR, but that seems like a good idea!
The only issue is it ends up in quite a long recompile for the core, so I wonder if we could do it on just one platform which should catch most of these errors. 

ie: run linux64 tests once regular and once with OF_USE_LEGACY_VECTOR_MATH set to 1. 
I think we might even be able to do that just from the command line env vars so it could be quite easy.  I'll take a look though.

arturoc:
most of the tests run in parallel so if we do it in separate jobs it would be fine

Issue #6467, opened 2019-11-20
ofParameter<void>::trigger() does not notify parents.


eduardfrigola:
When listening to parameterChangedE() event from an ofParameterGroup, ofParameter<void> triggers does not get caught by this event.

Is there a reason for the trigger() method not implementing notifying events to parents? similar to  what ofParameter::eventsSetValue(...) does:
https://github.com/openframeworks/openFrameworks/blob/8d8c8ff5280f7140d5bf2cdfda08ad90f552383d/libs/openFrameworks/types/ofParameter.h#L745-L763

I can send a PR implementing this if needed.

arturoc:
can you explain a use case for this?  trigger doesn't change the state of the parameter, as `ofParameter<void>` doesn't have any, so notifying a parameter changed event didn't seem necessary but we could add it if there's a use case where it would make sense

eduardfrigola:
I'm using it to build a gui (`ofParameter<void>` are buttons) via scripting language, where I can add/delete parameters on the fly. The communication is only by ofParameter event system (all of them inside an `ofParameterGroup`), and I want to catch all events inside the group and call a function inside the script. Now I have implemented a new listener for every ofParameter<void> that gets created, that i have to keep track and delete if necessary. But seems like a workaround.

When I added `ofParameter<void>` functionality I didn't thought parameterChanged event would not be notified. I get your point that the name parameterChanged does not apply for `ofParameter<void>`, that is true. As there is no "some event inside the group happened" event, I use `parameterChangedE` for that, and that confused me, as it does not get `ofParameter<void>` events. For me it is better to include it than not, but maybe my use case is not for what is intended.

Thanks

arturoc:
ok, i don't think adding it would be problematic for any case so if you want to send a PR? we'll probably merge it after the 0.11 release since we are already mostly fixing bugs.

eduardfrigola:
Should we close this?

arturoc:
yeah thanks!

PR #6471, opened 2019-11-25
ofParameter<void>::trigger() notifies parents

eduardfrigola:
Implements what is discussed in: #6467


arturoc:
thanks!

eduardfrigola:
Sorry to bring back this old issue, but I see two problems in here.
As it is a 0.11.1 milestone issue, it should have been requested against patch-release and not into master... I don't know how to fix this... Can this be fixed? Does this have to be removed from the 0.11.1 milestone?
Another problem is that the code in this pull request introduced an issue #6602, and i think that should be resolved to be able to mark the milestone as completed.

Sorry about those mistakes, let me now what I can do to address them!

Issue #6274, opened 2019-04-08
transformFeedbackExample doesn't work on OSX


neilmendoza:
Runs fine but blank canvas on OSX 10.13.6.

roymacdonald:
Hi, it will  not work on macos, because of its openGL implementation. It was mentioned somewhere which I cant recall where, but I think it was @arturoc who did.

arturoc:
supposedly this should work in osx cause transform feedback is supported but for some reason it doesn't so if anyone wants to investigate what it's going on that would be really useful

roymacdonald:
here it is, https://blog.openframeworks.cc/post/173258830179/opengl-new-features.
@arturoc There you say it will not work on osx. Has anything changed that you say that now it should run. Something related to mojave? If such, I am not running Mojave (and will not) so I cant test any further.
cheers

arturoc:
yes that was wrong. I assumed that osx just didn't support it but transform feedback is old, supported since 3.0: https://www.khronos.org/opengl/wiki/Transform_Feedback  

Which makes me thing that it might just be that you need to use the programmable renderer to make it work

roymacdonald:
Ha. Thanks for pointing it out.
The example at least is running the programable renderer. On my machine it does not work either, macos 10.12.6

I'll try yo dig about this on the next days.

best

eduardfrigola:
Hi, I found a way to make transform feedback work on macOS.
I've tested it in both macOS 10.13.6, and win10, with latest master.
Without changing any code on the transform feedback example (using default gl 3.3).

While reviewing the full implementation of transform feedback, some tutorials point out that is "preferable" to cut gl pipeline just after vertex or geometry shader (http://ogldev.atspace.co.uk/www/tutorial28/tutorial28.html).

Seeing that in OF's implementation that didn't happen I implemented that and seems to fix it.
Just calling `glEnable(GL_RASTERIZER_DISCARD)` right before `glBeginTransformFeedback`, and then `glDisable(GL_RASTERIZER_DISCARD)` (to return to the previous state) right after `glEndTransformFeedback`.

It seems plausible that when apple did their implementation they required this state to be activated.

If you want to try it: https://github.com/PlaymodesStudio/openFrameworks/tree/bugfix-transform_feedback_macos

What do you think?

roymacdonald:
[...]

PR #6473, opened 2019-11-26
Fix for transform feedback on macOS

eduardfrigola:
as discused in #6274


eduardfrigola:
I tested it with Ubuntu 18.10 and it worked exactly as without the change. Can I help with some more tests?

arturoc:
can you send this to the patch-release branch instead of master?

PR #6474, opened 2019-11-26
Include how-to-update section

hamoid:
Also minor headline formatting to make sections more obvious.

### DIFF
--- a/INSTALL_FROM_GITHUB.md
+++ b/INSTALL_FROM_GITHUB.md
@@ -1,4 +1,4 @@
-## How to compile openFrameworks from github
+# How to compile openFrameworks from github
 
 Follow these instructions only if you want to submit a pull request or if you want to always use the most recent version of openFrameworks. In other cases you will not need to clone the entire openFrameworks repository.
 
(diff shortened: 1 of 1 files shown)


roymacdonald:
Hi, Thanks for this.
If anyone is using the github version I assumes that the user has fork and changed the remotes to point to its own fork, which is what you would do in order to make pull requests.
So, I would add to check
```
git remote -v
```
and see that which is the remote that is pointing to the openFrameworks github repo, and pull on that. For example, if I call this command I get the following:
```
origin	git@github.com:roymacdonald/openFrameworks.git (fetch)
origin	git@github.com:roymacdonald/openFrameworks.git (push)
upstream	https://github.com/openframeworks/openFrameworks.git (fetch)
upstream	https://github.com/openframeworks/openFrameworks.git (push)
```
Thus, it is the `upstream` remote which is pointing to OF's github repo, and it is from which I need to pull. In order to do so I need to call
```
git pull upstream master
```
The download dependencies is not absolutely mandatory, it is only needed when a dependency has been updated, which does not happen often, so I would add that this should be done if the local repo has not been updated in a long time.

hamoid:
Good point @roymacdonald , I'll update it.

hamoid:
Better now @roymacdonald ?

roymacdonald:
Hi. sure, sorry to bother, but I  would explicitly add how to add the other remote as it is super easy.
```
git remote add upstream git@github.com:openframeworks/openFrameworks.git
``` 
in which `upstream` can be any name that is not already used for a remote, thus listed on the `git remote -v command`

to update the project generator it is better to call `git submodule update --recursive`

hamoid:
Updated.

roymacdonald:
Great! looks good to me.

roymacdonald:
thanks

INSTALL_FROM_GITHUB.md
```diff
@@ -43,7 +43,46 @@ Once the success message appears you are done.
...
+
+It is possible that you only have one or the other. If you are missing the official remote you can add it like this:
+
+    git remote add upstream git@github.com:openframeworks/openFrameworks.git
+
+Note that `upstream` is a name you choose, and it's how you will refer to the official openframeworks repository. Use that same name in the following command to update your local copy of openFrameworks:
+
+```bash
+git pull upstream master
+git submodule update --recursive
```
ofTheo: I ran into this recently. 
If you haven't pulled in the submodules you have to do this first: 
`git submodule update --init --recursive`

hamoid: Added :)

arturoc:
thanks!

PR #6478, opened 2019-11-29
Changelog for 0.11.0 

ofTheo:
For #6412


ofTheo:
okay - this should be good to go.

PR #6481, opened 2019-11-30
add features about JSONlike formats.

2bbb:
implemented thin wrappers of feature of nlohmann::json about `BSON`, `CBOR`, `MessagePack`, `UBJSON`.

### DIFF
--- a/libs/openFrameworks/utils/ofJson.h
+++ b/libs/openFrameworks/utils/ofJson.h
@@ -112,3 +112,202 @@ inline void ofDeserialize(const ofJson & json, ofAbstractParameter & parameter){
 		}
 	}
 }
+
+namespace of {
+    namespace json_like_utils {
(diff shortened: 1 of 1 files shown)


bakercp:
Thanks for this @2bbb. This is a sensible and well-written PR.  That said, I'm not convinced we need to add it as OF is moving to slim down / modularize the core moving forward.  

At the same time, I do use these functions in my own work and it does make the exception-less behavior more familiar -- but I'd suggest we move these thin wrappers to a future well-namespaced ofJson addon rather than add to the core.  

Additionally, json.hpp makes heavy use of exceptions which we largely don't handle at this point, so I think we need to decide how much we want to try to hide exceptions in any future addon.

2bbb:
@bakercp 
ok. i close this PR now.
and i will wait for ofJson addon.

PR #6482, opened 2019-12-01
update submodules to latest for master/stable

ofTheo:


ofTheo:
@arturoc @bakercp - not sure if we need to do this on master branch or just in the release tag?
 
figured we would merge master into stable now that 0.11.0 is out, but wanted to have this commit in master first.   

If we don't need it feel free to close. 
Otherwise we can bring it in and update stable from master.

PR #6485, opened 2019-12-04
Ensures depth buffer for ofxAppEmscriptenWindow. Closes #6484

### LINKED ISSUE #6484: emscripten - weird issues with 3D meshes / geometry
With latest master / 0.11.0 I am seeing meshes with what seems like inverted normals or depth testing failing. 

assimpExample: 
image

3dPrimitiveExample:
image

however 2D shapes seem fine:
image

- UPDATE - 
I just tried disabling depth testing on an Xcode build of assimpExample and it looks the same.
So I am guessing depth testing is failing somehow on emscripten. 
image

ofTheo:
closes #6484

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -118,7 +118,7 @@ void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
             }
 
         }else{
-            ofLogNotice() << " depth is successfully set to " << depthPreference[i] << endl;
+            // Got a good configuration. Stop searching. 
             break;
         }
     }
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -65,7 +65,7 @@ void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
 		   EGL_GREEN_SIZE, EGL_DONT_CARE,
 		   EGL_BLUE_SIZE, EGL_DONT_CARE,
 		   EGL_ALPHA_SIZE, EGL_DONT_CARE,
-		   EGL_DEPTH_SIZE, 16, // So we can have depth testing. 16 / 24 or 32bit 
+		   EGL_DEPTH_SIZE, 32,
 		   EGL_STENCIL_SIZE, EGL_DONT_CARE,
 		   EGL_SAMPLE_BUFFERS, EGL_DONT_CARE,
 		   EGL_NONE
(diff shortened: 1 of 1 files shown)


arturoc:
that's weird, this has worked in the past but i guess some browsers might default to no depth buffer. does this happen in a browser in particular or in any?

In any case 16bit could be really low depending on the scene. perhaps the best would be to check the largest option iterating through every option returned by:

```
eglGetConfigs(display, NULL, 0, &numConfigs)
```

Fixing this values has been problematic in the past in other platforms that use EGL like android and rpi.

ofTheo:
Yeah, I was wondering if it might have issues hard coding it to a value. 

I currently see the issue with both the latest Chrome and Safari on 10.14. 
It could be just a macOS thing where the system provides the option for a config with no depth buffer, or it could be to do with the engines Chrome/Safari are using GL. 

I'll try the approach you suggested and see what it lists out. 
Thanks!

ofTheo:
Hmm - so I tried using eglGetConfigs and it always returns 1 for chrome and safari on macOS. 
Tried a few other approaches and couldn't get back a list of different configs. 

I tried another approach though. 
You can give higher number values for EGL_DEPTH_SIZE and it will select the highest configuration that it can find.  

So for example on my system, passing:
`EGL_DEPTH_SIZE, 64` or 
`EGL_DEPTH_SIZE, 32` 

Gives me a depth size of 24 which must be the max for my system. 
So I think `EGL_DEPTH_SIZE, 32` should default to the highest resolution for any system. 

I'll add a check just incase to use EGL_DONT_CARE if the config fails for any reason.

arturoc:
this looks really hacky, is there any documentation for this api where it says that it works like that? i don't remember that fixing a value meant that value or less.

ofTheo:
yeah, the whole thing is a little strange. 
I think the idea is that if a value doesn't match, the smallest possible value to what you specify is returned. ( http://directx.com/2014/06/egl-understanding-eglchooseconfig-then-ignoring-it/ ) 

The default for EGL_DEPTH_SIZE though is 0. ( https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglChooseConfig.xhtml ) 
So that is what will probably be returned for most platforms with `EGL_DEPTH_SIZE, EGL_DONT_CARE` which is what we have now. 

On macOS no matter what I pass to `eglGetConfigs` and `eglChooseConfig` I always get back 1. 
It seems like macOS is not interested in giving a range of options for which to pick the best match. 

If I set `EGL_DEPTH_SIZE` to anything **except for** `EGL_DONT_CARE` I get 24 for the depth size. Otherwise I get 0 and no depth buffer. 

So probably the safest approach is to default to 24 ( seem standard and same as in ofAppEGLWindow.cpp ), if `eglChooseConfig` fails try 16 and then `EGL_DONT_CARE`. 

That should cover all the bases.

ofTheo:
This should be safe now for macOS and any other platforms.

addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
```diff
@@ -89,14 +101,27 @@ void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
...
 
-	ofLogNotice("ofxAppEmscriptenWindow") << "Got " << numConfigs << " display configs";
-
-	// Choose config
-	if ( !eglChooseConfig(display, attribList, &config, 1, &numConfigs) ){
-		ofLogError() << "couldn't choose display";
-		return;
-	}
+        }else{
+            ofLogNotice() << " depth is successfully set to " << depthPreference[i] << endl;
```
arturoc: Can you remove this notice or at least set it to verbose?

ofTheo: Sure. Just pushed!

PR #6486, opened 2019-12-04
add msys2 64bit builds to nightly and release scripts.

ofTheo:
This should work for nightly builds and releases. 
@oxillo maybe you could test the generated package and let us know if it works okay for you?

( msys2 64bit build ) 
http://ci.openframeworks.cc/versions/v0.11.0/of_v0.11.0_msys264_release.zip

### DIFF
--- a/scripts/dev/nightlybuilds.sh
+++ b/scripts/dev/nightlybuilds.sh
@@ -53,6 +53,7 @@ fi
 ./create_package.sh linux64 $lastversion master gcc5
 ./create_package.sh linux64 $lastversion master gcc6
 ./create_package.sh msys2 $lastversion master
+./create_package.sh msys2 $lastversion master 64
 ./create_package.sh vs2017 $lastversion master
 ./create_package.sh ios $lastversion master
 ./create_package.sh osx $lastversion master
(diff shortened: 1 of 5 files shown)


ofTheo:
Just tested the 64bit msys2 package above it works great. 
So I think this should be good to merge. I can add a link to that zip on the downloads page.

arturoc:
the tests for the new platform are not passing, no?

ofTheo:
yeah, the test for msys2 64bit isn't passing. 

I think its been like this for a while - not relating to this PR though and it compiles fine for me on Desktop so I am guessing its a CI configuration issue?  

Maybe @oxillo, could you take a look?

oxillo:
I'll look into it during the week-end.

oxillo:
Need to have openframeworks/apothecary#155 accepted

ofTheo:
@arturoc do you think you could you re-run the AppVeyor CI now that the apothecary changes have been merged? Thanks!!

arturoc:
done, also, if you add yourself to the openframeworks/appveyor team you should have admin access to appveyor

ofTheo:
Ahh awesome! 
Thanks 🙂

scripts/dev/nightlybuilds.sh
```diff
@@ -52,7 +52,8 @@ fi
 ./create_package.sh linux64 $lastversion master gcc4
 ./create_package.sh linux64 $lastversion master gcc5
 ./create_package.sh linux64 $lastversion master gcc6
```
ofTheo: @arturoc do you think here we should try and programmatically pull the default branch? So during patch-release phases it is targeting patch-release instead of master?  

```
git remote show origin | grep "HEAD branch" | cut -d ":" -f 2
``` 

This outputs `patch-release` currently.

arturoc: yes that would be great but not sure there's a way to do that without using the github api or something like that. i usually just change it manually while we are on a patch release cycle

arturoc:
going to merge this as it is by now, thanks!

PR #6492, opened 2019-12-05
Forward declare OpenAL structs to allow for api change. closes #6490

### LINKED ISSUE #6490: ofOpenALSoundPlayer conflicts with latest OpenAL
See: https://forum.openframeworks.cc/t/compilation-failing-due-to-confliting-definition-in-openal/33927/2 

Due to OpenAL 1.2 ( released less than a week ago ) changing some struct typedefs. 
This should be an easy fix.

ofTheo:
Don't merge yet - just want to check if this approach passes the CI tests. 
Addresses #6490

### DIFF
--- a/libs/openFrameworks/sound/ofOpenALSoundPlayer.cpp
+++ b/libs/openFrameworks/sound/ofOpenALSoundPlayer.cpp
@@ -9,14 +9,6 @@
 #include "ofEvents.h"
 #include <sndfile.h>
 
-#if defined (TARGET_OF_IOS) || defined (TARGET_OSX)
-#include <OpenAL/al.h>
-#include <OpenAL/alc.h>
-#else
-#include <AL/al.h>
-#include <AL/alc.h>
-#endif
-
 #ifdef OF_USING_MPG123
 #include <mpg123.h>
 #endif
--- a/libs/openFrameworks/sound/ofOpenALSoundPlayer.h
+++ b/libs/openFrameworks/sound/ofOpenALSoundPlayer.h
@@ -10,9 +10,9 @@
 
 typedef unsigned int ALuint;
 /** Opaque device handle */
-typedef struct ALCdevice_struct ALCdevice;
+struct ALCdevice;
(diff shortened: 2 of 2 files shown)


ofTheo:
yeah, doesn't work and I understand why. 
not sure if there is a clean solution aside from matching what is in alc.h in the header. or including alc.h in the header instead of the cpp.  might just try that and see how the tests do...

ofTheo:
As far as the tests go this looks good. 
The includes are still ifdef'd inside of the `#ifdef OF_SOUND_PLAYER_OPENAL` and this avoids having duplicate typedef's in two places.

arturoc:
I moved all the includes for external libraries inside cpps to make compiling and autocomplete faster. otherwise you end up with this files being included everywhere. can we check the openal version and do one of the forward declaration types or the other based on that?

ofTheo:
I totally get the reasoning, I just couldn't figure out a better solution :) 

The problem with detection is it would need to be in the ofOpenALSoundPlayer.h header ( as that's where the typedefs are done ) and to detect the OpenAL version we would need to include the OpenAL headers anyway which already would solve the issue :) 

The other option would be to just typedef the newer syntax, but then everyone would need to be on the latest OpenAL.

ofTheo:
So this issue came up again on the [forum](https://forum.openframeworks.cc/t/compileof-sh-fails-on-arch-linux-gcc9/33963). 
I think the safer option would be to merge this PR as is. 

Another approach would be to forward declare the new typedefs the same way as it was done before, but this could mean that systems with slightly older OpenAL versions ( by a matter of days ) would be broken. 

So I think probably good to merge as is. 
I looked at the average compile times on the CI over the last PRs and this change doesn't seem to be significantly slower than normal times. ( the last few are faster, but then other slightly older PRs are slower ).

arturoc:
going to close this in favor of https://github.com/openframeworks/openFrameworks/pull/6498 which doesn't expose headers globally

PR #6494, opened 2019-12-06
Add libgconf-2-4 to debian/install_dependencies.sh

### LINKED ISSUE #6493: `projectGenerator` requires gconf2, but it is not installed (Debian)
The script `scripts/linux/debian/install_dependencies.sh` doesn't install the `libgconf-2-4` package required by `projectGenerator`, resulting in the following error when trying to execute the generator:

    error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory

This script should install this dependency to allow the `projectGenerator` to work.

Tested with openFrameworks 0.11.0, on Debian 10.2 (buster).

galexite:
This commit adds libgconf-2-4 to the `apt install` command in the
`install_dependencies.sh` script for Debian in
`scripts/linux/debian`.

Tested on Debian 10.2 (buster).

Closes #6493.

### DIFF
--- a/libs/openFrameworks/utils/ofJson.h
+++ b/libs/openFrameworks/utils/ofJson.h
@@ -36,10 +36,10 @@ inline bool ofSaveJson(const std::filesystem::path& filename, const ofJson & jso
 	try{
 		jsonFile << json;
 	}catch(std::exception & e){
-		ofLogError("ofLoadJson") << "Error saving json to " << filename.string() << ": " << e.what();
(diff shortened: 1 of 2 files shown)


galexite:
Rebasing on master.

ofTheo:
Hmm - I think this should have been sent to patch-release. 
But we'll need to do a PR for other linux platforms so I can do that to patch-release.

PR #6506, opened 2019-12-12
added warning suppression for glm warnings coming from gcc 8.

### LINKED ISSUE #6496: MSYS2 64bits : many GLM warnings
When compiling, many warnings emitted :

```
N:/openFrameworks/libs/glm/include/glm/gtc/packing.inl:506:37: warning: 'void* memcpy(void*, const void*, size_t)' writing to an object of type 'glm::i16vec4' {aka 'struct glm::vec<4, short int, glm::packed_highp>'} with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
  506 |   memcpy(&Unpack, &v, sizeof(Unpack));
      |                                     ^
```

Undefining GLM_FORCE_CTOR_INIT in ofConstants.h solve the problem.
Seems like a GLM issue g-truc/glm#854

ofTheo:
Added warning suppression for glm warnings coming from gcc 8. 
Closes #6496 

We could remove the warning suppression once its fixed in glm's upstream.

### DIFF
--- a/libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
+++ b/libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk
@@ -135,6 +135,7 @@ PLATFORM_REQUIRED_ADDONS =
 # >= 4.9.x c++14
 # other compilers c++11 by now
 ifeq ($(CXX),g++)
+    GCC_MAJOR_GTEQ_8 := $(shell expr `gcc -dumpversion | cut -f1 -d.` \>= 7)
 	GCC_MAJOR_EQ_4 := $(shell expr `gcc -dumpversion | cut -f1 -d.` \= 4)
 	GCC_MAJOR_GT_4 := $(shell expr `gcc -dumpversion | cut -f1 -d.` \> 4)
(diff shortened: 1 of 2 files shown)


arturoc:
I don't think we should be suppressing warnings, this is probably happening on arch linux which gets new versions the day after they come out, by the time this hits other distributions glm will surely have released a fix.

going to close it but let me know if you think otherwise

ofTheo:
@arturoc, according to @oxillo its currently the reason msys2 isn't passing tests on the CI. 
Thats the main urgency - otherwise it is just a bit annoying :) 

Also on Rpi I see these warnings too, which might slow down compile times a bit. 
I feel like its okay to add this flag as its a new type of warning that isn't present on GCC < 8. 

No real harm done from suppressing it and when GLM has a fix ( which might still be a while as it doesn't technically break anything ), we can remove the flag.

arturoc:
we could temporarily patch the makefile in that specific test but i'm reluctant to add hacks here and there to solve problems in other external libraries, later it's difficult to remember where they were and makes the codebase more difficult to maintain

PR #6508, opened 2019-12-12
Add latomic linker for legacy Rpi builds in Stretch and newer

ofTheo:


arturoc:
does this flag fail on older versions? it would be easy to test since the CI server is still on stretch i think. in that case it would be simpler to just add `-latomic` always and not introduce special cases for each version in the makefile

PR #6512, opened 2019-12-22
[WIP] Setting c++ std from project

### LINKED ISSUE #5445: std=c++14 flag being passed to PLATFORM_CFLAGS (android)
The default android project makefile (config.default.android.mk) passes -std=c++14 to PLATFORM_CFLAGS, resulting in the following compile error for c files:  

`error: invalid argument '-std=c++14' not allowed with 'C/ObjC'`

Removing the flag from PLATFORM_CFLAGS and adding to PLATFORM_CXXFLAGS fixes the issue. 

Will offer pull request shortly.

### LINKED ISSUE #5520: Add mechanism to explicitly set c++ version in config.make
Currently the c++ version (e.g. c++11, c++14) used by a project is determined by `./libs/openFrameworksCompiled/project/makefileCommon/config.linux.common.mk` (on Linux). Provide a way such that the version can be overridden by a project in `config.make`.

This is useful for projects needing to be compiled on different platforms whose c++ version support differs. By being able to be explicit about a version a project can set a version that would be supported across all targeted platforms.

[Related forum thread: How can I use c++11 instead of c++14](https://forum.openframeworks.cc/t/how-can-i-use-c-11-instead-of-c-14/26331/4)

oxillo:
This PR adds the ability to set the C++ standard to use at project level (PROJECT_CXX_STD) or at platform level (PLATFORM_CXX_STD).
If none is defined, it tries to set the standard from the compiler version (GCC only) or set to c++11.

@arturoc @ofTheo : can I have your feedback on it before going further ? Do you prefer it to go to the master branch ?

Fix #5520 
Fix #5445

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -257,7 +257,7 @@ void ofxAppEmscriptenWindow::setWindowPosition(int x, int y){
 }
 
 void ofxAppEmscriptenWindow::setWindowShape(int w, int h){
-	emscripten_set_canvas_size(w,h);
+    emscripten_set_canvas_element_size(NULL,w,h);
 }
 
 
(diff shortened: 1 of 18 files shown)


arturoc:
this would be really useful, it seems you are bringing a lot of stuff from master in the PR but as you mentioned it would be better if you sent this to master instead since it's adding a new feature

thanks!!

PR #6517, opened 2020-01-08
Fix ofGstVideoPlayer frame by frame

roymacdonald:
I am using macos 10.14.6, xcode 11.2.1 and the current OF master.
When trying to use ofGstVideoPlayer  and calling `setFrameByFrame(true)` on it,  I get a wrongly decoded image. See below.
Screen Shot 2020-01-08 at 15 46 43

And when I do not call `setFrameByFrame(true)` I get the movie decoded properly, as seen in the following image
Screen Shot 2020-01-08 at 15 49 37

I don't have any previous experience with GStreamer, and it seems massive. Out of looking at some of its examples and the `ofGstVideoPlayer`'s code I found out that the code in the update function was being done wrong, so I replaced with some of the code used in `GstFlowReturn ofGstVideoUtils::process_sample(shared_ptr<GstSample> sample)` which is the function that properly decoded the movie when not using the frame by frame feature. 
Copying and pasting code is a super bad idea and a function should be used instead on both instances, but as I do not know if what I changed is correct from the GStreamer point of view, I opted for leaving it like this for know. As well I am not sure how this change would affect other platforms.

### DIFF
--- a/libs/openFrameworks/video/ofGstUtils.cpp
+++ b/libs/openFrameworks/video/ofGstUtils.cpp
@@ -742,7 +742,7 @@ bool ofGstUtils::gstHandleMessage(GstBus * bus, GstMessage * msg){
 	return true;
 }
 
-GstElement 	* ofGstUtils::getPipeline() const{
+	GstElement 	* ofGstUtils::getPipeline() const{
 	return gstPipeline;
 }
 
@@ -853,6 +853,16 @@ ofTexture * ofGstVideoUtils::getTexture(){
 #endif
 }
 
+	static GstVideoInfo getVideoInfo(GstSample * sample){
+		GstCaps *caps = gst_sample_get_caps(sample);
+		GstVideoInfo vinfo;
+		if(caps){
+			gst_video_info_from_caps (&vinfo, caps);
+		}else{
+			ofLogError() << "couldn't get sample caps";
+		}
+		return vinfo;
+	}
 void ofGstVideoUtils::update(){
 	if (isLoaded()){
 		if(!isFrameByFrame()){
@@ -877,7 +887,7 @@ void ofGstVideoUtils::update(){
 			ofLogError() << "frame by frame doesn't work any more in 0.10";
 #else
 			GstBuffer * buffer;
-			GstSample * sample;
+			GstSample* sample;
 
 			//get the buffer from appsink
 			if(isPaused()){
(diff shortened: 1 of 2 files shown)


roymacdonald:
Just in case, might be useful for testing

ofApp.h

```cpp
#pragma once

#include "ofMain.h"
#include "ofGstVideoPlayer.h"

	public:
		void setup();
		void update();
		void draw();
		void keyReleased(int key);

		
	ofVideoPlayer player;
	shared_ptr<ofGstVideoPlayer> gst;
};

```

ofApp.cpp

```cpp
#include "ofApp.h"
#include "ofGstVideoPlayer.h"

void ofApp::setup(){
	ofSetLogLevel(OF_LOG_VERBOSE);
	auto r =  ofSystemLoadDialog();
	if(r.bSuccess){
		
	
		gst = make_shared<ofGstVideoPlayer>();
//		gst->setFrameByFrame(true);
		
		player.setPlayer(gst);
		player.load(r.getPath());

		
		
		guint major;
		guint minor;
		guint micro;
		guint nano;
		

		
		gst_version	(&major, &minor, &micro, &nano);
		
		
		
		cout << "GST Version major " << major << endl;
		cout << "GST Version minor " << minor << endl;
		cout << "GST Version micro " << micro << endl;
		cout << "GST Version nano " << nano << endl;
		
		
//		cout << (string)gst_version_string () << endl;
		
//		player.play();
	}else{
		ofExit();
	}
}

void ofApp::update(){
	if(!gst->isFrameByFrame())
		player.update();
}

void ofApp::draw(){
	 
	ofRectangle r (0,0,player.getWidth(), player.getHeight());
	r.scaleTo(ofRectangle(0,0,ofGetWidth(), ofGetHeight()));
	player.draw(r);
	stringstream ss;
	ss << "is Frame by Frame : " << boolalpha << gst->isFrameByFrame() <<endl;
	ss << "Current Frame : " << player.getCurrentFrame() << endl;
	ss << "Total Frames : " << player.getTotalNumFrames() <<endl;
	ss << "is Paused "  << player.isPaused() << endl;
	ss << "is Playing "  << player.isPlaying() << endl;
	
	ofDrawBitmapStringHighlight(ss.str(), 30,30);
}


void ofApp::keyReleased(int key){
	
	if(key == OF_KEY_LEFT){
		player.previousFrame();
		if(gst->isFrameByFrame()) player.update();
	}else if(key == OF_KEY_RIGHT){
		player.nextFrame();
		if(gst->isFrameByFrame()) player.update();
	}else if(key == ' ' ){
		if(!player.isPaused()){
			player.setPaused(true);
		}else{
			player.play();
		}
	}else if(key == 'f' ){
		if(gst->isFrameByFrame()){
			gst->setFrameByFrame(false);
//			player.play();
		}else{
			gst->setFrameByFrame(true);
//			player.setPaused(true);
		}
	}
}

```

roymacdonald:
@arturoc can you check this? I am using this fix for an ongoing project and has been working fine.

PR #6521, opened 2020-01-16
add libgconf-2-4 to install dependicies for linux. closes #6493

### LINKED ISSUE #6493: `projectGenerator` requires gconf2, but it is not installed (Debian)

ofTheo:


arturoc:
not sure this is correct for every distribution, archlinux for example usually doesn't add the lib prefix to packages and depending on what's the problem we might need the -dev package in debian based distros or the -devel one in fedora. what's the issue this is related to?

ofTheo:
Ahh good to know! 
I can check what the correct naming is for the different distros and revise. 

See #6493 - I think you mentioned we needed this for all linux distros.

sigmike:
Arch Linux doesn't provide a package for gconf anymore (probably because it's obsolete[[1](https://askubuntu.com/questions/249887/gconf-dconf-gsettings-and-the-relationship-between-them)]). There's an AUR package that works: https://aur.archlinux.org/packages/gconf.

dimitre:
Hello @ofTheo recently in Ubuntu 24.04 install script was failing because of libgconf-2-4
Maybe we can close this one

PR #6526, opened 2020-01-24
Added clarification of how drawBox works

rhodso:
After using oF in a recent project, I was trying to draw a box, when I
encountered some weird behaviour. So I added this to clarify how drawBox
works for future use


roymacdonald:
I would rather say that the box's origin is in the center of the box as the parameters passed use the word origin.

PR #6536, opened 2020-02-11
emscripten  DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR

Jonathhhan:
Hi, 
I hope I did the pull request the right way...
You need to set the flag DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR to 0 to make Open Frameworks work with the latest fastcomp Emscripten (1.39.6).
I also added some additional flags but they are mainly for debugging...
https://github.com/Jonathhhan/openFrameworks/blob/patch-release/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk
I am not able to change the template.html so that it is possible to use the flag DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1 which would be even nicer.
Here is the topic from the forum: https://forum.openframeworks.cc/t/latest-emscripten-doesnt-work-on-examples/30463/54

### DIFF
--- a/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk
+++ b/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk
@@ -1,3 +1,34 @@
+
+Skip to content
+Pull requests
+Issues
+Marketplace
+Explore
+@Jonathhhan
+Learn Git and GitHub without any code!
+
+Using the Hello World guide, you’ll start a branch, write comments, and open a pull request.
+openframeworks /
+openFrameworks
+
+541
+7.5k
+
+    2.4k
+
+Code
+Issues 820
+Pull requests 89
+Actions
+Projects 0
+Wiki
+Security
+Insights
+openFrameworks/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk
+Arturo Castro emscripten: fix problems with latest versions c070afc on 26 Jun 2019
+@arturoc
+@frauzufall
+284 lines (245 sloc) 13.2 KB
 ################################################################################
 # CONFIGURE CORE PLATFORM MAKEFILE
 #   This file has linux common rules for all the platforms (x86_64, i386,armv6l
@@ -64,7 +95,7 @@ PLATFORM_REQUIRED_ADDONS = ofxEmscripten
 ################################################################################
 
 # Code Generation Option Flags (http://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html)
-PLATFORM_CXXFLAGS = -Wall -std=c++14 -Wno-warn-absolute-paths     
+PLATFORM_CFLAGS = -Wall -std=c++14 -Wno-warn-absolute-paths
 
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


arturoc:
the errors seem not related so i'll merge it, but can you send it to the patch-release branch instead?

Also can you remove the extra flags? I don't think we need them or at least not for release if they are only for debugging. if you think they are useful perhaps you can send a second PR with those only for the debug builds explaining what they do 

Thanks!

Jonathhhan:
Alright, now I only added `-s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=0`
To make it work with Ofelia I also had to change `PLATFORM_CFLAGS = -Wall -std=c++14 -Wno-warn-absolute-paths` to `PLATFORM_CXXFLAGS = -Wall -std=c++14 -Wno-warn-absolute-paths`, for compiling c++ and c. But I think that is not necessary if compiling only Open Frameworks to Emscripten. Here is the config file that I use for Ofelia Emscripten: https://github.com/Jonathhhan/ofxOfelia/blob/master/scripts/Emscripten/updateOF/config.emscripten.default.mk
Here is a description of the flags: https://github.com/emscripten-core/emscripten/blob/master/src/settings.js
I try to describe the flags that i use with with my own words (please correct me if I am wrong, I was just trying to find something that works):
WASM=1 compiles to webassembly.
ALLOW_MEMORY_GROWTH=1 allows emscripten to resize the memory.
TOTAL_MEMORY=1920MB max available memory.
ASSERTIONS=2 prints error messages.
SAFE_HEAP=1 prints an error message if there is a memory leak.
ERROR_ON_UNDEFINED_SYMBOLS=0 allows undefind symbols (i think popen was the problem).
DEMANGLE_SUPPORT=1 prints symbols in a human readable way.

arturoc:
This is now failing in emscripten as well. it seems that you've added some raw text at the beginning and the end of the makefile. if you can remove that?

Jonathhhan:
How could that happen? Anyway, I hope now it is right.

arturoc:
that looks great thanks!

arturoc:
about the rest of the flags, i think WASM=1 is the default no? the rest seem useful except ERROR_ON_UNDEFINED_SYMBOLS=0 which would allow for undefined symbols that will just generate an exception later at runtime. if there's some undefined symbol we need to fix it not ignore it

can you send another PR with those? this time to master since these are new features not a bug fix

Jonathhhan:
Yes, you are right, it seems WASM=1 is the default. The undefined symbol is indeed popen, but I am not sure if that is caused from OF or Ofelia. And it seems not possible to resolve popen, so for now I need to use ERROR_ON_UNDEFINED_SYMBOLS=0, but would be nice if that is not needed:
https://stackoverflow.com/questions/55861324/how-to-make-emscripten-to-link-popen
"My point is popen is undefined because is impossible to implement UNIX process in a browser context so you simply can not solve the problem. It is much better to leave it undefined than to define a non-functional code. I agree that Emscripten documentation is not great and is very confusing but it is not hard to guess that why popen is not defined considering the browser's and WemAssembly engine's capabilities. "

arturoc:
the popen problem comes from openframeworks. from ofSystem specifically, it makes sense that this doesn't exist in emscripten and we just need to ifdef that function so it doesn't exist in empscripten. if you want to send a PR with that change or just open an issue explaining the problem that would be super useful

PR #6539, opened 2020-02-13
ofxGui: Updating events to allow multiwindow functioning

roymacdonald:
@ofTheo

### DIFF
--- a/addons/ofxGui/src/ofxBaseGui.cpp
+++ b/addons/ofxGui/src/ofxBaseGui.cpp
@@ -138,8 +138,19 @@ void ofxBaseGui::unregisterMouseEvents(){
 	ofUnregisterMouseEvents(this, defaultEventsPriority);
 	bRegisteredForMouseEvents = false;
 }
-
+void ofxBaseGui::setEvents(ofCoreEvents & _events){
+	if(&_events != events){
+		bool wasMouseInputEnabled = bRegisteredForMouseEvents;// || !events;
+		unregisterMouseEvents();
+		events = &_events;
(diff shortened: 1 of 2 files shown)


arturoc:
this seems kind of hacky the recommended way to use several windows is to have a different App class per window which solves this issue. 

In any case i'm going to revert this and merge it into master instead so we can test it before having it on an official release

roymacdonald:
Hi @arturoc  
Sure it can seem hacky but I find it way much less elegant and prone to error having to do [this](https://github.com/openframeworks/openFrameworks/blob/master/examples/windowing/multiWindowOneAppExample/src/main.cpp#L22)  in order to allow the Gui to capture the events from the desired window. 

Having the same gui on two different windows might seem like a stretch as the events listeners would be changing on each frame, but on a more common situation it would happen much less times, and will reduce the pain of figuring out why a gui is not working on a different window.
Last but not least. It is more or less the same way that `ofEasyCam` has for acquiring the mouse events.
Best

arturoc:
as i said, if you use a different application for each window then you don't need to do [this](https://github.com/openframeworks/openFrameworks/blob/master/examples/windowing/multiWindowOneAppExample/src/main.cpp#L22)

in any case i see how allowing to use the gui like this if possible is an advantage but somehow also makes it easier to use a model of using multiwindow that is problematic in other areas so even if this solves the problem for guis, other problems will appear once someone has a more complex application if people use this model instead of one app per window

roymacdonald:
I have not used multiwindowing a lot, mostly having guis on a different window but I see what you mean. Lets see how other users testing goes..

PR #6540, opened 2020-02-14
config.emscripten.default.mk Emscripten flags proposal

Jonathhhan:
Coming from this PR https://github.com/openframeworks/openFrameworks/pull/6536

I suggest that these additional flags perhaps could be useful:
`--bind` enables embind.
https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html?highlight=embind
`--profiling-funcs` preserve function names in profiling.
https://emscripten.org/docs/tools_reference/emcc.html
`DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=0`
Arturo: "Yes it seems they’ve changed the way to handle events so changing the template would allow it to work without enabling this flag. that would be of course better but if you are not sure on how to update the template, sending a PR with this flag and opening an issue explaining this would be also super useful".
`ALLOW_MEMORY_GROWTH=1` allows emscripten to resize the memory.
`TOTAL_MEMORY=1GB` max available memory (Firefox does not work above 1GB).
`ASSERTIONS=2` prints error messages.
`SAFE_HEAP=1` prints an error message if there is a memory leak.
`DEMANGLE_SUPPORT=1` prints symbols in a human readable way.

`--profiling-funcs`, `ASSERTIONS=2`, `SAFE_HEAP=1`, and `DEMANGLE_SUPPORT=1` are only useful for debugging (but helped me a lot).

Here is a complete list of the flags: https://github.com/emscripten-core/emscripten/blob/master/src/settings.js

I also changed `PLATFORM_CFLAGS = -Wall -std=c++14 -Wno-warn-absolute-paths` to `PLATFORM_CXXFLAGS = -Wall -std=c++14 -Wno-warn-absolute-paths`, for compiling c++ and c. 
I am not sure if that is useful for Open Frameworks, but I need it for compiling ofxOfelia with OF.

I also use `ERROR_ON_UNDEFINED_SYMBOLS=0` because of "popen", but I do not suggest it.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,9 @@
 version: 1.0.{build}
 
+platform:
+  - x86
+  - x64
+
 environment:
   global:
     APPVEYOR_OS_NAME: windows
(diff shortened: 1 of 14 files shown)


danoli3:
What happened here @oxillo app veyor was never fixed after the Cache turned on / Matrix changed...

Issue #6547, opened 2020-02-19
Emscripten: Webaudio crashes if heap resizes (solved)


Jonathhhan:
I opened an issue on the Github page from Emscripten, because Webaudio is crashing when heap  resizes. Apparently it is more an issue with html5audio which is part of ofxEmscripten, so I post it here....
https://github.com/emscripten-core/emscripten/issues/10488

"Ah, then I'd guess this is a bug in

https://github.com/openframeworks/openFrameworks/blob/patch-release/addons/ofxEmscripten/libs/html5audio/lib/emscripten/library_html5audio.js

in Open Frameworks. Specifically, it probably keeps a view alive to the old data. My guess is it's in html5audio_stream_create, where inbufferArray is captured once, then referred to from later callbacks, at which time it may be invalid. If that's it, the fix is to not create a subarray once and reuse it, but create one in each callback."

Jonathhhan:
Thanks kripken, your guess fixed the issue.
I just put moved those two lines of code to the callback:
```
var inbufferArray = Module.HEAPF32.subarray(inbuffer>>2,(inbuffer>>2)+bufferSize*inputChannels);
		        var outbufferArray = Module.HEAPF32.subarray(outbuffer>>2,(outbuffer>>2)+bufferSize*outputChannels);
```


```
html5audio_stream_create: function(context_id, bufferSize, inputChannels, outputChannels, inbuffer, outbuffer, callback, userData){
		var stream = AUDIO.contexts[context_id].createScriptProcessor(bufferSize,inputChannels,outputChannels);
		var id = AUDIO.lastStreamID++;

		stream.onaudioprocess = function(event){
                        var inbufferArray = Module.HEAPF32.subarray(inbuffer>>2,(inbuffer>>2)+bufferSize*inputChannels);
		        var outbufferArray = Module.HEAPF32.subarray(outbuffer>>2,(outbuffer>>2)+bufferSize*outputChannels);
			var i,j,c;
			if(inputChannels>0){
				for(c=0;c<inputChannels;++c){
					var inChannel = event.inputBuffer.getChannelData(c);
					for(i=0,j=c;i<bufferSize;++i,j+=inputChannels){
						inbufferArray[j] = inChannel[i];
					}
				}
			}

			dynCall('viiii',callback, [bufferSize,inputChannels,outputChannels,userData]);

			if(outputChannels>0){
				for(c=0;c<outputChannels;++c){
[...]

PR #6549, opened 2020-02-22
fix "Webaudio crashes if heap resizes"

Jonathhhan:
It does not crash if I put those two lines to the call back, so that the arrayBuffer is created every time a new sound file is loaded:

```
var inbufferArray = Module.HEAPF32.subarray(inbuffer>>2,(inbuffer>>2)+bufferSize*inputChannels);
var outbufferArray = Module.HEAPF32.subarray(outbuffer>>2,(outbuffer>>2)+bufferSize*outputChannels);
```
https://github.com/Jonathhhan/openFrameworks/blob/patch-release/addons/ofxEmscripten/libs/html5audio/lib/emscripten/library_html5audio.js
https://github.com/openframeworks/openFrameworks/issues/6547


bakercp:
There are quite a few changes in this PR and it is difficult to tell what you are proposing at a glance.  Can you clean up and submit a new PR?

Jonathhhan:
I am sorry. I need to check out how the whole PR thing works. Until then it is perhaps more helpful if I just open an issue instead of a PR (if it is helpful at all). Just wanted to mention that it does not crash for me anymore with the change from above...

bakercp:
Definitely appreciate it -- an issue would probably make more sense.

Jonathhhan:
I already opened an issue before the PR, but then I thought the right way to submit would be a PR. Next time I know it better: https://github.com/openframeworks/openFrameworks/issues/6547

PR #6558, opened 2020-03-23
adding pugi parsing Options to ofXml (OF enum version)

dasoe:
Hi all,

this is a re-opened pull request after change requests from your side.
pugi offers parsing options like "pugi::parse_default | pugi::parse_comments" in case one wants comments not to be stripped etc. Added a variable, a setter and (as requested) OF-like enum decleration to be able to use them in ofXml. 

Would be used like this:
XML.setParsingOptions( OF_PARSE_DEFAULT | OF_PARSE_COMMENTS ); XML.load(pathToXML);

Hope I got this enum-thing done appropriately?

**Thanks for your awesome work!**
oe

### DIFF
--- a/libs/openFrameworks/utils/ofXml.cpp
+++ b/libs/openFrameworks/utils/ofXml.cpp
@@ -6,6 +6,7 @@ using namespace std;
 ofXml::ofXml()
 :doc(new pugi::xml_document){
 	xml = doc->root();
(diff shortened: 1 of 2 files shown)


dasoe:
Hi all, 

I think this pull request is good (I add it manually to each OF download), change request from your side is impelmented also. 
Is there something I can do to further support it? 

have a good day!
oe

Issue #6145, opened 2018-10-04
Possible memory leak in ofxGui


roymacdonald:
I just noticed that in ofxGuiGroup all the `add(ofParameter<T> & parameter)` functions call new as you can see [here](https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxGui/src/ofxGuiGroup.cpp#L178-L229), thus creating the gui object needed. but these newly created objects never get deleted even after the `ofGuiGroup` is destroyed. In a lot of cases this might not be a problem as a lot of times the the `ofxGuiGroup` or `ofxPanel` object exists for the whole runtime of the app, but in scenarios where the gui is created and destroyed dynamically this is a problem.

The following code is a proof of this.
````cpp
#include "ofMain.h"
#include "ofxGui.h"

class ofApp : public ofBaseApp
{
public:
	//----------------------------------------------------------------
	void setup(){		
		gui = new ofxPanel();
		gui->setup();
		gui->add(param.set("param", 0, 0, 255));
		sliderPtr = &(gui->getFloatSlider("param"));
		
	};
	//----------------------------------------------------------------
	void draw(){
		
		ofBackground(param, 0, 0);
		
		if (displayGui){
			if(gui){
				gui->draw();
			}else if (sliderPtr){
				sliderPtr->draw();
			}
		}
		
		stringstream ss;
		
		if(sliderPtr){
			ss << "sliderPtr  " << sliderPtr << endl;	
		}else{
			ss << "sliderPtr  is null" << endl;
		}
		if(gui){
			ss << "gui  " << gui << endl;	
		}else{
			ss << "gui is null" << endl;
		}
		ofDrawBitmapStringHighlight(ss.str(), 20, ofGetHeight() - 50);
		
	};
	//----------------------------------------------------------------
	void keyPressed(int key)
	{
		if(key == ' '){
			displayGui = !displayGui;
		}else if(key == 'd'){
			delete gui;
			gui = nullptr;
		}
	}
	//----------------------------------------------------------------
	ofxPanel* gui= nullptr;
	ofParameter<float> param;
	ofxFloatSlider* sliderPtr = nullptr;
	bool displayGui = true;
};

int main( )
{
	ofSetupOpenGL(400,200,OF_WINDOW);
	ofRunApp(new ofApp());
}

````

Maybe moving all the pointers used into smart pointer can be a good idea. What do you think @arturoc  @bakercp @ofZach ?

arturoc:
[...]

PR #6570, opened 2020-04-13
Fix ofxGui memory leak

roymacdonald:
ofxGui: Fix possible memory leak, because keyword "new" was being used but not deleted. Now using only unique_ptr
Related to:
#6145
and partially addressed in commit
b961214
(It was my fault not adding this fix to that commit)


roymacdonald:
@arturoc  @ofTheo ? This is kinda hot fix. as I said it was my fault not adding it in the already mentioned and merged commit.

ofTheo:
@roymacdonald this looks good to me. 👍
will merge later today. Thanks

ofTheo:
@roymacdonald this looks good to me. 👍
will merge later today. Thanks

PR #6566, opened 2020-04-10
Merge make43 msys2ccache

oxillo:
- remove use of ccache in MSYS2 appveryor build (fix #6565)
- fix "missing separator" error with make 4.3 (fix #6535)

Both fix in a single PR to have build pass on Appveyor 
Replace #6564

PR #6572, opened 2020-04-15
MSYS2 : (re)enable ccache in appveyor

### LINKED ISSUE #6565: MSYS2 : build stalls in  appveyor
Lately MSYS2 stalls when building in Appveyor. Build works fine locally.

Disabling CCache during Appveyor build solves the problem.

oxillo:
Fix #6565 - use mingw(32|64) version of ccache instead of msys version.

Revert disabling of ccache in #6566 

Remove duplicate use of setting ccache (by setting CXX in ccache.sh) and by using USE_CCACHE and setting PLATFORM_CXX in msys2 makefile. makefile solution (with USE_CCACHE) has been retained.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,18 +5,19 @@ environment:
     APPVEYOR_OS_NAME: windows
     CHERE_INVOKING: 1
     MSYS2_PATH: c:\msys64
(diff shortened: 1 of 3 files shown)


arturoc:
thanks!

Issue #6576, opened 2020-04-22
ofParameterGroup contains(string name) if parameter changed name.


eduardfrigola:
Hi everyone.
Playing with ofParameterGroups I found an issue with ofParameterGroup::contains(...)
https://github.com/openframeworks/openFrameworks/blob/5453d35ae88022d2c09f47d3beb9a57c1e87a65d/libs/openFrameworks/types/ofParameterGroup.cpp#L422-L424 it also applies in all places where [`std::map<std::string,std::size_t> parametersIndex;`](https://github.com/openframeworks/openFrameworks/blob/5453d35ae88022d2c09f47d3beb9a57c1e87a65d/libs/openFrameworks/types/ofParameter.h#L269`) is used.

The issue pops up when you change the name of a parameter that is already inside the parameterGroup. The name changes of the parameter changes, but the name in parametersIndex does, not and so we cannot find our parameter inside the parameterGroup. Because the parameterIndex map is only created when adding parameters, and only modified when removing, the only workaround for this is to remove the element from the group, changing the name, and adding it again. With that changing the order of the elements in the group.

As ofParameter has a public method of `::setName()`, this case should be considered right?

As there is no event fired by the ofParameter when the name is changed, I think the best solution is to get rid of `parametersIndex` and search the parameters vector directly so we get the updated name every time we make the request.

Eduard

roymacdonald:
Hey @eduardfrigola 
It makes sense to me to remove anything that is duplicated data, as it is in this case.
Although, I guess that `parametersIndex` was added to improve performance. So, what about testing how performance is afected by removing `parametersIndex` ? Otherwise, having an ofEvent triggered when the name changes can be another option, that can be useful on some other scenarios.

eduardfrigola:
Hi @roymacdonald thanks for answering.
I made test with bundled OF unittests for ofParameter and didn't notice any notable difference... But the test is with small amount of parameters inside the group, so I will make a test adding a ton of parameters.
Your proposal of a change name event will be a better option, for example, for detecting if the group already has a parameter with that name on it, as it does when you `add()` a parameter.

Maybe handling name changes once added to parameterGroup it is not a thing to be allowed, since probably you passed that group to somewhere else (ofxGui) and modifying that causes some errors. Have you encountered something similar?

roymacdonald:
Hi, let me know how it goes with those tests. 
[...]

PR #6577, opened 2020-04-22
[ofParameterGroup] removed parameterIndex

eduardfrigola:
This pull request is for addressing the issue described on #6576.

This removes the usage of parameterIndex as does not keep track of the changes in names of ofParameters inside the group.

I have also used the already implemented methods to avoid code duplication. (used `contains()`or `getPosition()`).

Worried about performance with this change, I run the unitTest for parameter and found that the change in time is negligible. (About 300 microseconds for both current implementation and this branch).

I can edit the [parameter test](https://github.com/openframeworks/openFrameworks/tree/patch-release/tests/types/parameters) to show the fix if that is needed.

Eduard

### DIFF
--- a/libs/openFrameworks/types/ofParameter.h
+++ b/libs/openFrameworks/types/ofParameter.h
@@ -266,7 +266,6 @@ private:
 
 		void notifyParameterChanged(ofAbstractParameter & param);
 
-		std::map<std::string,std::size_t> parametersIndex;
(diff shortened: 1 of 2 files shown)


eduardfrigola:
Better fix for the same bug: https://github.com/openframeworks/openFrameworks/pull/6614

PR #6578, opened 2020-04-23
Storing render surface stack on fbo.begin() to restore at fbo.end().

sadmb:
By this change, you can use nested fbo:
fbo.begin();
fbo.begin();
fbo.end();
fbo.end();

### DIFF
--- a/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
@@ -1293,6 +1293,11 @@ void ofGLProgrammableRenderer::unbind(const ofShader & shader){
 void ofGLProgrammableRenderer::begin(const ofFbo & fbo, ofFboMode mode){
 	pushView();
     pushStyle();
+    ofBaseDraws* currentRenderSurface = const_cast<ofBaseDraws*>(matrixStack.getCurrentRenderSurface());
+    if(currentRenderSurface)
+    {
+        renderSurfaceStack.push_back(currentRenderSurface);
+    }
     if(mode & OF_FBOMODE_MATRIXFLIP){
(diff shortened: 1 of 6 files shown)


arturoc:
thanks for the contribution, this has been discussed before though and was considered that it doesn't add any functionality that is really useful since there can only be one FBO bound at any given time and adds certain complexity to the FBO class so we decided against including it.

do you have any specific use case? usually you would do 

```
fbo1.begin();
fbo1.end()
fbo2.begin();
fbo2.end();
fbo1.begin();
```

if at the end of using fbo1 and then fbo2 you want to then use fbo1 again

sadmb:
@arturoc 
Thanks for your reply. :)
Well, then maybe I will make subclass for my project.

I usually use layer manager system to make some app.
It uses fbo in manager class for combining layers like:

```
Class LayerManager {
  vector<ofPtr<LayerBase>> layers;

  void update()
  {
    for(auto& layer: layers)
    {
      layer->update();
    }
    fbo.begin();
    for(auto& layer: layers)
    {
      layer->draw();
    }
    fbo.end();
  }

  void draw()
  {
    fbo.draw()
  }
}

Class LayerBase {
  LayerBase(string name, ofBlendMode blendmode) {}
  virtual ~LayerBase() {}
  virtual void update() = 0;
  virtual void draw() = 0;
}
```

then I make each layer class inherits layer base class like 

```
Class Layer1 : public LayerBase
{
public:
  Layer1() : LayerBase("Layer1", OF_BLENDMODE_ALPHA) {}
  ~Layer1() {}
  void update() {}
  void draw() {}
}
```

Sometimes I need to use fbo inside those layers with changing viewport.

Other than this example, there're some addons using fbo expect you to write inside, and sometimes you may want to use fbo inside those addons so I thought it is better to store render surfaces. :)

PR #6586, opened 2020-05-14
Fix ofLineSegmentIntersection test between line and rectangle.

sebleedelisle:
There is currently an edge case where ofLineSegmentIntersection incorrectly returns false if the line's end points sit directly on the edges of the rectangle. This commit should fix that.

Example code to prove this : 
```
	ofRectangle rect(0,0,800,800);
	glm::vec2 p1(400,0);
	glm::vec2 p2(400,800);
	glm::vec2 p3(0,400);
	glm::vec2 p4(800,400);

	ofLogNotice((rect.intersects(p1, p2)) ? "TRUE" : "FALSE");
	ofLogNotice((rect.intersects(p3, p4)) ? "TRUE" : "FALSE");
```

Both tests should return true, previously they returned false. 

Hope this helps!

### DIFF
--- a/libs/openFrameworks/math/ofMath.h
+++ b/libs/openFrameworks/math/ofMath.h
@@ -441,13 +441,13 @@ bool ofLineSegmentIntersection(const vectype& line1Start, const vectype& line1En
 	compareB = diffLB.x*line2Start.y - diffLB.y*line2Start.x;
 	if (
 		(
-			( ( diffLA.x*line2Start.y - diffLA.y*line2Start.x ) < compareA ) ^
-			( ( diffLA.x*line2End.y - diffLA.y*line2End.x ) < compareA )
+			( ( diffLA.x*line2Start.y - diffLA.y*line2Start.x ) <= compareA ) ^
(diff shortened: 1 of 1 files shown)


bakercp:
Is this related?

https://github.com/openframeworks/openFrameworks/issues/4871

bakercp:
There are some referenced discussions in #4871, e.g. https://github.com/openframeworks/openFrameworks/pull/4872

sebleedelisle:
Oh sorry, I did do a search on issues but I didn't see those! 

My pull request fixes the issue referenced in #4871, I'm pretty sure it was a logical issue rather than a floating point one.

roymacdonald:
Did this got fixed? can we close this PR?

sebleedelisle:
It has not been fixed from what I can tell (in the master branch at least) - it's a pretty simple fix, change the '<' symbols to '<=' as in the commit in the PR. Thanks!

roymacdonald:
sounds like a reasonable merge to me. what do you think @ofTheo @bakercp  ?

PR #6590, opened 2020-05-30
Fixed title not being shown when calling ofSystemLoadDialog in macos

roymacdonald:
Macos only: Minor change that properly shows the title of the NSOpenPanel when calling ofSystemLoadDialog


roymacdonald:
@ofZach @ofTheo @arturoc @bakercp  can we merge this one. It is a really small fix, which should not affect any other and it would be actually quite useful.

PR #6593, opened 2020-06-11
Raspberry Pi blur shader example bugfix

pierrep:
This fixes the examples/shader/09_gaussianBlurFilter app under the Raspberry Pi using GLES2.

The RPI uses normalised texture coordinates in shaders (as all GLES2 shaders do) and this was not being taken into account. There was also this following line that wasn't present in the other desktop shaders and seemed redundant, so I took it out:
`color /= 25.0;`


pierrep:
@ofTheo any chance you could have a look at this? It's been languishing and it's really a simple fix. It's also been broken for forever on RPI. Also, #6749 is a recent PR with critical implications for RPI and Linux...also been broken since forever. Thanks!

PR #6604, opened 2020-07-31
Fix flatNormals()

michaelt919:
flatNormals() doesn't work because indices gets reset by clear(), after which the indices are needed.  The simple fix is to make a local copy of indices.


arturoc:
thanks!

PR #6608, opened 2020-08-07
Fix regex when ofShader int define is set to 0.

prisonerjohn:
Regex did not work when #define was originally set to `0` (because of the hex format). This is now fixed.
Also made this block of code use tabs like the rest of the file.


arturoc:
Thanks!

PR #6625, opened 2020-10-27
split PLATFORM_CFLAGS into PLATFORM_CXXFLAGS to avoid compile errors on .c files

armadillu:
Similar to this: https://github.com/openframeworks/openFrameworks/issues/5445 

When your project contains .c files, the of makefile currently passes the -std=c++11 to the compiler, which throws an error.
This wasn't a problem when using GCC, but it is when trying to use on CLANG.

Using the same strategy as mentioned on the issue above fixed the problem.


arturoc:
thanks!

Issue #6627, opened 2020-11-04
download_libs.sh Android On Windows64 Fails: rsync: command not found


danoli3:
### rsync: command not found
Error Bash:
> 2020-11-05 01:26:29 (994 KB/s) - 'openFrameworksLibs_master_androidx86.tar.bz2' saved [68716649/68716649]
> 
> Removing old libraries
> Uncompressing libraries android from openFrameworksLibs_master_androidarmv7.tar.bz2
> Uncompressing libraries android from openFrameworksLibs_master_androidarm64.tar.bz2
> Uncompressing libraries android from openFrameworksLibs_master_androidx86.tar.bz2
> Copying opencv to ofxOpenCv
> Removing old opencv libraries
> ../dev/download_libs.sh: line 244: rsync: command not found
> Error on or near line 244; exiting with status 1
> 

### Proposed Solution

```
if ! command -v rsync &> /dev/null
        then      
            cp -a ${addonslibs[i]}/ ../addons/${addons[i]}/libs/${addonslibs[i]}     
        else
            rsync -a ${addonslibs[i]}/ ../addons/${addons[i]}/libs/${addonslibs[i]}
fi
```

danoli3:
Fixed

PR #6628, opened 2020-11-05
Fix: download_libs.sh Android On Windows64 Fails: rsync

danoli3:
Fixes: #6627


danoli3:
Lol it's failing on Appveyor trying to download rsync for windows...

PR #6629, opened 2020-11-05
remove flags

natxopedreira:
removed march and mtune flags from config.linux.common.mk and config.msys2.default.mk

related to this post
[https://forum.openframeworks.cc/t/linux-problem-please-help/36509](https://forum.openframeworks.cc/t/linux-problem-please-help/36509)


arturoc:
thanks!

PR #6635, opened 2020-11-18
Fix nightly builds to use the current default branch

ofTheo:
grabs the current default branch ( normally master but currently patch-release ) so nightly builds target what we are currently working on.

will need to merge this to master too, so the nightly script ( which just checks master ) is updated.


arturoc:
Not sure about this, i think it'll give problems wihen we switch branches. Right now the repo in the server is using mater for example so this won't be picked up until we merge this change into master?

In any case the patch branches should be merged into master often so keeping nightlies using master should be fine i think. If you want to see some change in next nightly just merge patch into master.

Anyway i think this could be useful but seems like it could also be problematic. 

+1 to changing the mail from mine to of though

ofTheo:
@arturoc yeah I think this makes sense if we can keep master up to date with patch-release. I was thinking if they weren't kept in sync this would be useful.  

I think there might be a way to setup the repo to auto merge patch-release commits to master, that could be a nice solution so we don't need to keep doing PRs to keep master up to date during patch-release.

PR #6636, opened 2020-11-19
update travis.yml to fix linux build

armadillu:
Remove the "exclude gcc" section of the travis.yml file to bring back auto ci linux builds


arturoc:
Thanks! Do you know when that was disabled?

armadillu:
a few months ago, they must have changed things on Travis. I have a lot of addons on CI and I realized all the linux builds were just not running and took a few tests to figure out why...

Issue #6542, opened 2020-02-14
archlinux install_dependencies.sh missing ROOT variable


themancalledjakob:
**Steps to reproduce:**
- run install_dependencies.sh

**Expected result:**
- it should install dependencies without relying on additional environment variables, which are not handled by the script.

**Actual result:**
- the script implies a `ROOT` variable to be present, containing the parent directory of the script itself. If this is not present, the script fails to update openCV, because it can't find the right addons directory.

terminal output:
```
Updating ofxOpenCV to use openCV4
sed: can't read /addons/ofxOpenCv/addon_config.mk: No such file or directory
sed: can't read /addons/ofxOpenCv/addon_config.mk: No such file or directory
```

**Potential fix:**
```
--- a/scripts/linux/archlinux/install_dependencies.sh
+++ b/scripts/linux/archlinux/install_dependencies.sh
@@ -18,8 +18,12 @@ if [ $exit_code != 0 ]; then
        exit $exit_code
 fi

+# get the root directory of this script
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
 # Update addon_config.mk files to use OpenCV 3 or 4 depending on what's installed
 addons_dir="$(readlink -f "$ROOT/../../../addons")"
```

System details:
openFrameworks v0.11.0
5.4.18-1-MANJARO

ImtiazKabir:
I am struggling with the same problem. I tried with the nightly builds too. But apparently it is still broken.
Can anyone help me with this problem? I am using manjaro

themancalledjakob:
hey,

the issue is, that the script uses a "$ROOT" variable. This variable seems to be empty for me, and apparently for you as well.
you can either use the fix that I proposed above to create the variable, or just hardcode "addons_dir" in line 22 to the addons directory in your openFrameworks installation to fix it.
For example: 
`addons_dir = /absolute/path/to/openFrameworks/addons`
that should do it.

ImtiazKabir:
Thank u so much! I changd the directory and it worked!

themancalledjakob:
You can also put `ROOT=$(cd $(dirname $0); pwd -P)` somewhere above in the script.
In scripts for ubuntu, etc you'll find that line, they just forgot to put it in here (and also some other install scripts). It is basically just finding the directory in which the script is in.
If it's gonna be fixed ever, then that is how it'll be done.

[...]

Issue #6631, opened 2020-11-08
OF_ROOT does not exist


ImtiazKabir:
Hey,
I recently downloaded OF and installed the dependencies. Then I compiled OF. There was no problem so far.
When I tried to compile polygon example, it said
`Makefile:13: /home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk: No such file or directory
make: *** No rule to make target '/home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk'.  Stop.`
I noticed that the variable `OF_ROOT` in the makefile, does not exist. So where did the project Makefile go?!
Also, `projectGenerator` is not recognised as a command. So I must have missed something.
I previously used OF in Ubuntu 18.04 and didnt face any problem.
I am currently using arch linux (manjaro)

ofTheo:
Hi - I think this is the same issue as #6650 and is fixed in latest master/patch-release. 
Can you try the nightly builds at the bottom of the download page and see if this is still an issue?  
https://openframeworks.cc/download/

themancalledjakob:
Hey Theo,

I think what you reference to is the OF_ROOT in the make system.
These scripts are run independently in bash, just as a script. Therefore the OF_ROOT variable is missing if it not set in the very same script.
Afaik this is not fixed yet.

PR #6638, opened 2020-11-26
set ROOT environment variable

themancalledjakob:
code has been tested on Manjaro 20.2
assumed fix would work for all other distros as well, since it's bash

[edit] whoops, I referenced the wrong issue! sorry about that
#6542

### DIFF
--- a/scripts/linux/archlinux/install_dependencies.sh
+++ b/scripts/linux/archlinux/install_dependencies.sh
@@ -10,6 +10,8 @@ if [ $EUID != 0 ]; then
    exit 1
 fi
 
+ROOT=$(cd $(dirname $0); pwd -P)
+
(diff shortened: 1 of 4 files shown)


ofTheo:
as mentioned in #6631 I think this is fixed. 
apologies!!

themancalledjakob:
Hey @ofTheo ,

I think what you reference to is the OF_ROOT in the make system.
These install_dendencies.sh-scripts are run independently in bash though, just as a script. Therefore the ROOT variable is missing if it not set in the very same script.
Afaik this is not fixed yet in master or patch-release. Just checked.

If I do anything wrong with the pull request, please let me know. I am happy to fix it. :)

ofTheo:
@themancalledjakob 

Apologies again :) 
You are correct. Sorry!!  

Would it be possible to PR this against patch-release and not master? 
That way it can be fixed in the next patch-release. 0.11.1 

Thanks!
Theo

Issue #5490, opened 2017-03-14
Licensing Issues | ofSoundPlayer for FMOD to upgrade to FMOD 5 from FMOD 4.4 (Ex) 


TheGrimOfCrazy:
FMOD Ex has stopped development and its license was not as friendly. I contacted them to ask about the use of FMOD Ex (FMOD 4.4) and they requested that openFrameworks upgrade due to licensing issues. There is currently no way to obtain an actual license for FMOD Ex.

FMOD 5's new license allows use for free as long as the development budget is under $500k and licensing after that is fairly cheap when compared to the budget. FMOD Ex can not be licensed anymore and is only used by people who are current licensees.

PR #6639, opened 2020-11-28
bugfix replace deprecated fmod system fft call with future compatible calls

ofTheo:
This allows us to update the fmodex.dylib to support Big Sur without getting linker errors. 
Also compatible with current fmodex.dylib.


ofTheo:
don't merge this yet - latest fmod with Apple Silicon support doesn't have these calls either 🤦‍♂️

ofTheo:
Okay this is now working with the newer FMOD api ( which is needed for Apple Silicon and Big Sur ). 
The new approach requires us to do our own binning as the new api doesn't allow us to specify the number of bands so there might be some small changes in the FFT results. Testing against the old approach it actually seems a little better in terms of distribution and we have more options in terms of nBands requested. 
 
NOTE: We'll need to update headers and libs for all platforms even though the api changes are minimal. 

Some references which were helpful: 
https://www.parallelcube.com/2018/03/10/frequency-spectrum-using-fmod-and-ue4/ 

FMOD reference for transitioning from older api to newer: 
https://www.fmod.com/resources/documentation-api?version=2.0&page=white-papers-transitioning-from-fmodex.html#fmod-studio-core-api-versus-fmod-ex-core-api

ofTheo:
I have all the new fmod platforms libs up here:
http://ci.openframeworks.cc/libs/fmod/ 

I am thinking to add a fmod.sh formula for apothecary and then update the OF codebase to use `fmod` instead of `fmodex` as all the libs are no longer called fmodex ( since the last 9-10 years ). I think this will be a bit clearer than using the old fmodex naming when referring to the newer non fmodex api ( 99% is the same just the few changes in this PR ).  

That way fmodex.sh still works as it currently does in apothecary and all the old links / urls will remain for the older fmodex libs. 

Before I go down this path I would love some thoughts on this.  

The other option would be to rename the fmod libs / folders to be the same as the old one but would sort of mask the library naming and the fact that its a more recent version.

arturoc:
yes that sounds like the way to go.

oxillo:
Shouldn't we move fmod to an external addon ? It's licencing is very restrictive (you need a license as soon as you use it for money-making activity including internal use in a company). see also #5490.
If we don't go the addon way, I would suggest that 
 - [ ] Of is compiled __without__ FMOD support
 - [ ] FMOD is only available throudh a dedicated compiler option
 - [ ] Licence.md is updated to highlight these license issues/risks.

ofTheo:
@oxillo Hey - I think that makes absolute sense for the future roadmap. Either way as this is a patch release to get OF working on macOS again I think we should keep this in for now.

danoli3:
Yeah I always have straight away removed fmod for this reason from any
commercial macOS projects. Would agree with this path to prevent issues

roymacdonald:
Hi! I second @oxillo We should get fmod out of the way and simply leave as an optional thing.

danoli3:
100% had to patch it out for some AbletonLink tests I was doing on macOS
yesterday. Lets rip it out asap

PR #6643, opened 2020-12-14
Initial aarch64 build

hfiguiere:
- built on RaspberryPi OS 64-bits


dimitre:
Hello @hfiguiere we now have a working target aarch64.
Can you please check to see if there is something important from here missing in master?

PR #6652, opened 2021-01-02
add support for Objective C and Objective C++ source in addon Make builds

bangnoise:
This PR extends existing support for .m and .mm source files (Objective C and Objective C++) to addons built using Make.

Support already exists for these files in user projects built using Make, and the same files in user projects and addons built using Xcode.


ofTheo:
@bangnoise thanks for this! 
could you push a commit to your branch ( white space or something ) so it can trigger the new CI tests?

PR #6664, opened 2021-01-23
msys2: Make OpenAL the default soundplayer

oxillo:
Change ofConstants.h to use OpenAL as default soundplayer if nothing is set in makefiles.
To use old FMOD, set PLATFORM_DEFINES += OF_SOUND_PLAYER_FMOD in Makefile


danoli3:
Amazing!

ofTheo:
thanks @oxillo !

PR #6666, opened 2021-01-26
Bugfix downloader

ofTheo:
The nightly build script was broken by #5848 because $2 ( for silent args ) wasn't been checked if set before testings its value. 
I tried a few different solutions that felt like they should of worked but didn't. 
These changes fixes the downloader function.


danoli3:
Ah damn, thanks for fixing!

PR #6672, opened 2021-01-27
WIP : make OF works when installed in a directory with spaces

oxillo:
Currently MSYS2 is compiling with all examples and tests (including a new one with space in its name!) with OF being installed in 'open Frameworks' instead of 'openFrameworks'


ofTheo:
@oxillo just switched the target to master instead of patch-release

PR #6673, opened 2021-01-28
core: fix GLM warnings 

### LINKED ISSUE #6496: MSYS2 64bits : many GLM warnings

oxillo:
Fix #6496 fix class-memaccess warning on unused glm types

### DIFF
--- a/libs/openFrameworks/math/ofVectorMath.h
+++ b/libs/openFrameworks/math/ofVectorMath.h
@@ -18,7 +18,6 @@
 #include "glm/trigonometric.hpp"
 #include "glm/exponential.hpp"
 #include "glm/vector_relational.hpp"
-#include "glm/ext.hpp"
 
 #include "glm/gtc/constants.hpp"
 #include "glm/gtc/matrix_transform.hpp"


ofTheo:
Huh. I guess we are not using it in the core as all the test/checks pass. 
It [also shows a warning saying](https://github.com/g-truc/glm/blob/master/glm/ext.hpp#L13) not to include it :) 

Maybe some people might be using it in their projects but I think its probably good not to include at the OF level.

action-script:
GLM EXT-GTC are not being used on the core either the examples but this removal has broken some projects I had.

For people with issues compiling GLM/GTC functions:
It requires manual header inclusion.
`#include "glm/gtc/random.hpp"`

It is confusing that some GLM functions are part of the OF documentation, but are not being actually included on code.
Should it be clarified on the OF documentation?

Issue #6457, opened 2019-11-16
ofxNetwork : wrongly reported errors


oxillo:
ofxNetwork may report errors that should be ignored.
This happens when there are construct like : 

```cpp
int err= ofxNetworkCheckError();
if ( err != OFXNETWORK_ERROR(NOTSOCK) ){
	// Ignore this error
}
``` 
As ofxNetworkCheckError() gets the error and logs it, all the errors will be reported even if they should be **ignored**.

This explains the error seen here:
```
[  error ] ofxNetwork: C:/projects/openframeworks/addons/ofxNetwork/src/ofxTCPManager.cpp: 200 ENOTSOCK: socket argument is not a socket
```

https://ci.appveyor.com/project/arturoc/openframeworks/builds/28847653/job/0duai87vuou8ctv7#L1358

_Originally posted by @ofTheo in https://github.com/openframeworks/openFrameworks/issues/6412#issuecomment-553731166_

## Proposal
 - Deprecate ofxNetworkCheckError()
- Add `int ofxNetworkGetLastError()` to return the last error
- Add void ofxNetworkLogError(int err, const char* file, int line) to log an error that happened on `line` line in file `file`
- Add ofxNetworkLogLastError() that is a shortcut for `ofxNetworkLogError(ofxNetworkGetLastError(), __FILE__, __LINE__)`

P-v-K:
Should that read "all the errors will be _**ignored**_ even if they should be reported."?

oxillo:
It's more "all the errors will be reported even if they should be **ignored**". I'll edit the post to make it clearer.

ofTheo:
@oxillo so would ofxNetworkGetLastError() return no error for errors that can be safely ignored? 
By the way I have also seen this issue occasionally on the CI for macOS makefile tests.

ofTheo:
This is one of the last issues left before we can release 0.11.1. 
I get the sense that it could be quite a lot of changes. Do you think we should push this to 0.11.2 / 0.12 ?

P-v-K:
I have no opinion on pushing or fixing, as I just ran into it once and it hasn't been impacting my current work. Maybe it could be mentioned in the documentation to help people be more likely to figure out what's (not) happening if/when they do run into it? Because once you know, and find the lines of code mentioned, you could hack it to do what you want? Just a thought, trying to be helpful. Again, I don't have enough knowledge/experience with it to have a strong opinion.

oxillo:
[...]

PR #6675, opened 2021-01-29
Fix ofx network

oxillo:
As proposed in #6457 

-   Deprecate ofxNetworkCheckError()
-   Add int ofxNetworkGetLastError() to return the last error
-   Add void ofxNetworkLogError(int err, const char* file, int line) to log an error that happened on line line in file file
-   Add ofxNetworkLogLastError() that is a shortcut for ofxNetworkLogError(ofxNetworkGetLastError(), __FILE__, __LINE__)

### DIFF
--- a/addons/ofxNetwork/src/ofxNetworkUtils.cpp
+++ b/addons/ofxNetwork/src/ofxNetworkUtils.cpp
@@ -3,13 +3,18 @@
 #include "ofUtils.h"
 
 
-
-int ofxNetworkCheckErrno(const char* file, int line) {
+//-----------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


ofTheo:
@oxillo thanks for this!! 
To be totally honest, I am not sure I 100% understand the fix but the logic and approach seems sounds. 

I'll give it a day to let @arturoc and maybe @bakercp to chime in. 

But if there is no other feedback we can merge it tomorrow.  It does pass the tests and I was seeing similar issues on macOS with errors in the TCP tests before which I don't see now. 

Thanks again!

PR #6678, opened 2021-01-30
ofXML fixes

### LINKED ISSUE #6111: ofXml float comma contradiction
Hello,
There is a contradiction between `setAttribute()` and `getFloatValue()`.
When you input a float like this:
```
xmlObj.setAttribute("FloatName", 2.5);
// or even
xmlObj.setAttribute("FloatName", 2.5f);
```
the XML file will contain "2.5" but 
`xmlObj.getAttribute("FloatName").getFloatValue();`
gives 2.
If you manually write 2,5 instead, then 
`xmlObj.getAttribute("FloatName").getFloatValue();`
will give 2.5 correctly.

I guess it is cool to read commas, but point should be taken in account, at least to be coherent.
Thanks very much.

### LINKED ISSUE #6676: ofXml : getChildren(name) returns children with incorrect name
In a document like
```xml
<graphic>
  <line/>
  <rectangle/>
  <circle/>
  <circle/>
  <triangle/>
  <circle/>
  <ellipse/>
</graphic>
```
I would expect that `graphic.getChildren("circle")` would just return an iterator on the 3 circles. 
The returned iterator iterates over the children between the first circle and the last circle; so, in this case, the `<triangle/>` is iterated.

### LINKED ISSUE #6677: ofXml: select_single_node() is deprecated in pugiXml

oxillo:
fix #6677 : pugiXML select_single_node deprecation
fix #6676 : getChildren(name) not returning only children with name
fix #6111 : Always process floats in C locale (with '.' as decimal separator) independently of current system locale (XML floats are always with '.')
Log error when something goes wrong when loading or saving a document.

### DIFF
--- a/libs/openFrameworks/utils/ofXml.cpp
+++ b/libs/openFrameworks/utils/ofXml.cpp
@@ -201,7 +201,7 @@ bool ofXml::removeAttribute(ofXml::Attribute && attr){
 
 ofXml ofXml::findFirst(const std::string & path) const{
 	try{
-		return ofXml(doc, this->xml.select_single_node(path.c_str()).node());
+		return ofXml(doc, this->xml.select_node(path.c_str()).node());
 	}catch(pugi::xpath_exception & e){
 		return ofXml();
 	}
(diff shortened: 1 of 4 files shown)


ofTheo:
@oxillo if you want to switch the ofLogNotice for ofLogVerbose I can merge this in. 
Thanks!

ofTheo (approved):
thanks!!

PR #6682, opened 2021-02-09
Android fix Music Player and SoundEffects

danoli3:
Fixes: Bug on Music Player (streaming / looped) audio from ofSoundPlayer in Android Sound Wrapper.
Problem occurs when setting music / checking if playing / seekTo on track that is not yet seen as "isPlaying" by the underlying system and then creates a MediaPlayer exception / bugs out. Android 9/11 tested. 
Fix guards the player a bit more with bool. Force release on Destructor to and I think android needs it

Added Attributes and new Build Constructor for MediaPlayer / Pool (required api change / future )

Fixes: SoundPool for Sound Effects played through ofSoundPlayer (no stream / no loop / multiplay) 
Corruption occurring when loading many sounds to pool, each one would corrupt the pool. Fixed with Build constructor and attributes and guards


danoli3:
Todo: Verify all API's are guarded

PR #6693, opened 2021-02-26
fix regular expression example

thomasgeissl:
hello
the regex example was crashing on osx, m1.
i removed the closing parenthesis. i guess adding an opening one would work as well. not sure this example requires a capture group.

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::regex_error: The parser did not consume the entire regular expression.
terminating with uncaught exception of type std::__1::regex_error: The parser did not consume the entire regular expression.
```


ofTheo:
thanks!

PR #6695, opened 2021-03-02
Feature: Android MSAA 

danoli3:
Adding MSAA to Native Android Window System OpenGLES 2.0 Devices

Tested on a suite of 12 varying android devices from 5 to 11 with no performance impact. 
Super Low Tier Devices do not have the ability to render MSAA so falls back to configs without feature

Works by finding and choosing Android Config from the Android Graphics Chooser and defaulting to MSAA4 then MSAA 2 then No MSAA.

Also patches and fixes Config Chooser where would fail now has a few fallback passes

### DIFF
--- a/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidWindow.java
+++ b/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidWindow.java
@@ -15,13 +15,14 @@ import android.view.SurfaceHolder;
 
 class OFEGLConfigChooser implements GLSurfaceView.EGLConfigChooser {
 	
-    public OFEGLConfigChooser(int r, int g, int b, int a, int depth, int stencil) {
+    public OFEGLConfigChooser(int r, int g, int b, int a, int depth, int stencil, int samples) {
         mRedSize = r;
         mGreenSize = g;
         mBlueSize = b;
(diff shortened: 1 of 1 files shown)


addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroidWindow.java
```diff
@@ -249,13 +337,13 @@ public OFGLSurfaceView(Context context,AttributeSet attributes) {
...
 //        We don't need this callback that much anyways, the renderer does not call render callbacks when gl context is invalid, so the OFAndroidWindow.onSurfaceCreated callback should be enought for us to make things work.
-/*
+
     @Override
 	public void surfaceDestroyed(SurfaceHolder holder) {
     	super.surfaceDestroyed(holder);
 		OFAndroid.onSurfaceDestroyed();
 	}
-*/
+
```
ofTheo, line 346: Just wanted to check it was intentional to uncomment surfaceDestroyed callback?
Considering the note above.  

If it is needed, then this PR looks good to go to me.

danoli3: Actually this is another Fix where Shader's are killed and never re-loaded with this commented out

PR #6699, opened 2021-03-03
Android Fix PS4/PS5 Controller KeyCodes to oF KeyPressed/KeyReleased

danoli3:
Basically current KeyCode parsing in OFAndroid layer parses keys into Unicode or kills them into 0: 

Problem: Controller Keycode: 22 becomes 0 passed back to **keyPressed** - Key should be 22 here:
Pasted Graphic
Pasted Graphic 1

Source of Problem:
image

Fix: Exclusion for Range >0 to < 256
PS5 ranges from 20-> 106

Android Supports Game Controllers for Accessibility and Games and these fall into the < 256 int range so using an exclusion for this range in the OFAndroid Layer that was zeroing out the KeyPressed / KeyReleased events for Hardware Controllers

### DIFF
--- a/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroid.java
+++ b/addons/ofxAndroid/ofAndroidLib/src/cc/openframeworks/OFAndroid.java
@@ -858,8 +858,10 @@ public class OFAndroid {
            		return false;
            	}
         }
-
 		int unicodeChar = event.getUnicodeChar();
+		if(unicodeChar == 0 && keyCode < 256 && keyCode > 0) {
+			unicodeChar = keyCode;
+		}
 		return onKeyDown(keyCode, unicodeChar);
 	}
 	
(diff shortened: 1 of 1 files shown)


ofTheo:
@danoli3 is it not possible to query the keyCode variable instead of relying on key? 
I think some of the PS4/PS5 keycodes overlap with the OF special keys:  

eg: 
```
	OF_KEY_ESC		=	27,
```

I think this is the same as the O on the PS controller. 
Just curious if for controllers the info is getting passed through via event args ?

danoli3:
Yeah keycode event data is stripped in `OfAndroid.java` before getting down to oF layer due to that command listed above... except where the commands are registered back or OF_KEY_CONTROL. This is due to a keyCode security function getUnicodeChar which is stripping any non-unicode or ASCII keys pressed/released as they are not real keyboard keys in that functions mind.

oF C++ event sent to ofxAndroidApp is just the keycode itself:
```
keyPressed(int key);
keyReleased(int key);
  ```

An alternative solution would be to add a GameController Pressed/Release just for this however it speaks the same keycode language. Proposed solution retains existing function and supports the safe keys used by controllers (0-256)

PR #6709, opened 2021-03-14
This change adds support for an OpenGL debug context. 

paulhoux:
When you create your window using the `ofGLFWWindowSettings` structure, you can set `debugContext` to true. Once created, OpenGL debug messages will appear in your console window if the GPU driver detects any potential problems or errors. This greatly helps in finding hard to debug issues.

IMPORTANT: this change has only been tested on Windows 10.

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -41,6 +41,28 @@
 using namespace std;
 
 //-------------------------------------------------------
+#if defined(TARGET_WIN32)
+void __stdcall	debugMessageCallback(GLenum /*source*/, GLenum /*type*/, GLuint /*id*/, GLenum severity, GLsizei /*length*/, const GLchar *message, void *userParam)
+#else
+void			debugMessageCallback(GLenum /*source*/, GLenum /*type*/, GLuint /*id*/, GLenum severity, GLsizei /*length*/, const GLchar *message, void *userParam)
+#endif
+{
+	switch( severity ) {
+		case GL_DEBUG_SEVERITY_HIGH:
+			std::cout << "OpenGL error: " << message << std::endl;
+		break;
+		case GL_DEBUG_SEVERITY_MEDIUM:
+			std::cout << "OpenGL warning: " << message << std::endl;
+		break;
+		case GL_DEBUG_SEVERITY_LOW:
+			std::cout << "OpenGL message: " << message << std::endl;
+		break;
+		case GL_DEBUG_SEVERITY_NOTIFICATION:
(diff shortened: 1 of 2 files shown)


ofTheo:
I think we'd need to get all the checks to pass above for this to get merged. 

Also as macOS is limited to openGL 3.3 and [this function is 4.3 ](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDebugMessageCallback.xhtml) and up we'd probably need some checks to only include this functionality if openGL 4.3 support and up is available. Otherwise I imagine it would cause a crash if called.

paulhoux:
I created this PR in response to your [post](https://forum.openframeworks.cc/t/solved-problem-using-a-fork-of-of-in-combination-with-the-download-libs-script/37418/2) on the forums, in the hope it would be useful.

This functionality may also be available to OpenGL version < 4.3, as long as they support the KHR_debug or ARB_debug_output extension. See: https://www.khronos.org/opengl/wiki/Debug_Output

I am not sure where to go from here. For instance, I do not have a Linux machine to test things on, nor do I know how you prefer to handle OpenGL extensions in OpenFrameworks. But I guess it should be relatively easy to check if there is support for this and then either include or exclude it from the build.

ofTheo:
Hi. 

For PRs we generally need them to work across all platforms OF supports. 
I personally only have regular access to macOS and Windows, but you can use the Github Actions checks above to make sure your PR works for all supported OS's. 

In this case it seems that the linux arm builds don't recognize that function. 

```
/home/runner/work/openFrameworks/openFrameworks/libs/openFrameworks/app/ofAppGLFWWindow.cpp:393:67: error: ‘glDebugMessageCallback’ was not declared in this scope
128
   glDebugMessageCallback( (GLDEBUGPROC)debugMessageCallback, this );
```
See: https://github.com/openframeworks/openFrameworks/pull/6709/checks?check_run_id=2106028450 

Going to add some inline comments that could help this pass the checks.

paulhoux:
Perhaps updating the OpenGL extensions loader (like Glew, or GLAD) for the Linux build is sufficient? Although that would admittedly be quite a big change. In any case, thank you for taking the time to review this PR.

arturoc:
I'd say this might be related to an older version of GLEW being used in the CI server but glDebugMEssageCallback will for sure work on newer versions which are present any newer distribution. I would make the ifdef conditional to just checking if that extension is available using GLEW macros instead of disabling it per platform

danoli3:
Yeah using https://registry.khronos.org/OpenGL/extensions/KHR/KHR_debug.txt
from GLFW should make these extensions for GLFW work on macOS 4.1 OpenGL and ES

PR #6710, opened 2021-03-14
This change fixes an error message at the start of your application…

paulhoux:
…if you call `ofRestoreWorkingDirectoryToDefault();` in your `main` function.

IMPORTANT: only tested on Windows 10.

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -87,7 +87,7 @@ namespace{
 
     //--------------------------------------------------
     std::filesystem::path & defaultWorkingDirectory(){
-            static auto * defaultWorkingDirectory = new std::filesystem::path();
+            static auto * defaultWorkingDirectory = new std::filesystem::path(ofFilePath::getCurrentExeDir());
             return * defaultWorkingDirectory;
     }
 


libs/openFrameworks/utils/ofUtils.cpp
```diff
@@ -87,7 +87,7 @@ namespace{
 
     //--------------------------------------------------
     std::filesystem::path & defaultWorkingDirectory(){
-            static auto * defaultWorkingDirectory = new std::filesystem::path();
```
ofTheo, line 90: What is the difference you get in terms of the path between before and after this change?

Also more of a general question as I didn't write this code. 
Why are we returning a reference and using new? Why not just return std::filesystem::path via copy instead? 

Thanks!!

paulhoux: When calling `defaultWorkingDirectory()` early in your application, the error "[warning] ofUtils: ofToDataPath: error while trying to change back to default working directory" is shown in the output window. It is linked to the use of `ofToDataPath`. This fix will prevent the error and always return a valid path for `ofToDataPath`.

I agree that simply returning a reference to a static variable, instead of a dereferenced pointer, is the better choice. However, I did not want to change the code too much, as I did not know why a pointer was chosen in this case.

ofTheo: Thanks! 
Looking more into the code I get it now. Your change makes sense and I don't think impacts most users anyway.

PR #6711, opened 2021-03-14
This change fixes a const-correctness problem…

paulhoux:
…with `ofFbo::begin(bool)` and also fixes an issue with `ofbo::updateTexture(int)` where the original read buffer wasn't restored properly after blitting.

IMPORTANT: only tested on Windows 10.


ofTheo:
thanks!

PR #6725, opened 2021-04-05
Fixed bug: crash when call ofVAArgsToString("%d", 0);

2bbb:
when call `ofVAArgsToString("%d", 0);`, second argument `0` will be interpreted to null pointer of `va_list`.
and function overload will be resolved to `auto ofVAArgsToString(const char * format, va_list args)` and crush.

this PR will fix this problem.

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -1072,7 +1072,10 @@ string ofVAArgsToString(const char * format, ...){
 	return s;
 }
 
-string ofVAArgsToString(const char * format, va_list args){
+template <typename VAList>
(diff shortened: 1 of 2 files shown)


2bbb:
i made mistake, and fixed it with create new function `std::string ofVAListToString(const char * format, va_list args)`.

implementation of this function is moved from `std::string ofVAArgsToString(const char * format, va_list args)`
and fixed `template <typename VAList> auto std::string ofVAArgsToString(const char * format, VAList args)  -> ...` calls this function in inline.

if this change is not acceptable, i will rename `ofVAListToString` to proper name to internal use, and warp into `priv` namespace.

arturoc:
thanks!

PR #6726, opened 2021-04-05
Update actions

eyt5rfyt:


ofTheo:
@openframeworks/core 

FYI this PR attempted to add a Github Action which would spin up a bunch of crypto mining operations. 
I have blocked the user and reported them to Github, but we might want to disable Github Actions on new yml files. 

See: the last file https://github.com/openframeworks/openFrameworks/pull/6726/files
Xmrig is a cpu/gpu miner

Issue #167, opened 2010-04-06
replace fmod


ofTheo:
can we do it?
we need cross platform:
playback 
streaming
pitch
pan 
multiplay 

for us to be compatible with what exists. 
OpenAL seems like the best choice.
What other options are there?

DomAmato:
dusting this one off. Has there been any ideas about how to address this? the most current master uses fmod for VS 32/64 bit builds

oxillo:
Fmod removed from MSYS2 in PR #4807

oxillo:
Hi,

We are trying to clean out the many issues that we have.
Many versions of OF have been released since this was issued.

Can you help us by checking if this feature is still requested for a next OF version ?

In case of inactivity, we'll close this issue by end of january, 2020.

Thanks

ofTheo:
Reviving one of the oldest issues for 0.12.0 :) 

Hoping we can remove fmod as the default player for macOS and Windows. 

For macOS I am working on an AVAudioEngine player which I think can be a 1:1 swap for ofFmodSoundPlayer. I believe it will also work on iOS / tvOS and be a good update for the iOS / tvOS sound player. 

For Windows I imagine a Media Foundation replacement sound player could provide similar functionality. But I know a bit less about that. 

Would be open if anyone Windows focused wanted to take a stab at it.
@oxillo @DomAmato @paulhoux

danoli3:
In iOS/tvOS and we already not using fmod it’s all AVFoundation

ofTheo:
oh yeah @danoli3 I just meant we could eventually remove the ofxiOSSoundPlayer and have iOS use the AVAudioEngine one which has some nice additional features ( multiplay, setSpeed able to go below 0.5 and above 2.0 etc )

danoli3:
Sounds great!

ofTheo:
@danoli3 @julapy here is the wip sound player.
https://github.com/ofTheo/openFrameworks/blob/AVEngine-SoundPlayer/libs/openFrameworks/sound/ofAVEngineSoundPlayer.m

It has pretty much all the ofFmodSoundPlayer functionality. 
I just need to add multiplay ( easy-ish ) and FFT ( harder ). 

If you want to give it a try you can use the soundPlayerExample and replace the top of ofApp.cpp with:

```
#include "ofAVEngineSoundPlayer.h"

//--------------------------------------------------------------
void ofApp::setup(){
	
[...]

PR #6728, opened 2021-04-06
macOS / iOS sound player. AVAudioEngine - WIP

ofTheo:
Addressing #167 

Features:
- [x] Multiplay
- [x] Pitch/Speed 
- [x] Handles switching audio outputs mid play
- [ ] Global FFT
- [ ] Streaming from disk vs loading to memory? 

ToDo:
- [ ] Add in code in ofConstants.h to allow easy switching from fmod to this player. 
- [ ] Add in iOS interrupt handling ( when a call comes in etc ) 
- [ ] Test on iOS 
- [ ] Look at Global FFT( see @arturoc 's comment here: https://github.com/openframeworks/openFrameworks/issues/167#issuecomment-809208468 ) 

Questions:
- Multiplay works by creating new internal players for each play request.  ofAVEngineSoundPlayer uses an ofEvents().update listener to cleanup the extra multiplay players. This seems more reliable than using a thread ( had some unexpected crashes when using a thread ), but curious if people see any downside to an update event on the main thread, vs spawning a new thread for each player.

### DIFF
--- a/libs/openFrameworks/sound/ofAVEngineSoundPlayer.m
+++ b/libs/openFrameworks/sound/ofAVEngineSoundPlayer.m
@@ -1,0 +1,568 @@
+//
+//  ofAVEngineSoundPlayer.cpp
+//  soundPlayerExample
+//
+//  Created by Theo Watson on 3/24/21.
+//
+
+#include "ofAVEngineSoundPlayer.h"
+#include "ofUtils.h"
+#include "ofMath.h"
+#include "ofLog.h"
+
+#define TARGET_OSX
+
+//REFS: https://github.com/ooper-shlab/AVAEMixerSample-Swift/blob/master/AVAEMixerSample/AudioEngine.m
+// https://developer.apple.com/documentation/avfaudio/avaudioengine
+// https://developer.apple.com/forums/thread/14138
+// https://developer.apple.com/forums/thread/48442
+// https://github.com/garynewby/XYAudioView/blob/master/XYAudioView/BasicAVAudioEngine.m
+// https://github.com/twilio/video-quickstart-ios/blob/master/AudioDeviceExample/AudioDevices/ExampleAVAudioEngineDevice.m
+
+#import <Foundation/Foundation.h>
+#import <AVFoundation/AVFoundation.h>
+
+@interface AVEnginePlayer : NSObject
+
+@property(nonatomic, retain) NSTimer * timer;
+
+- (BOOL)loadWithFile:(NSString*)file;
+- (BOOL)loadWithPath:(NSString*)path;
+- (BOOL)loadWithURL:(NSURL*)url;
+
+- (void)unloadSound;
+
+- (void)play;
+- (void)play: (float)startTime;
+- (void)pause;
+- (void)stop;
+
+- (BOOL)isLoaded;
+- (BOOL)isPlaying;
+
+- (void)volume:(float)value;
+- (float)volume;
+
+- (void)pan:(float)value;
+- (float)pan;
+
+- (void)speed:(float)value;
+- (float)speed;
+
+- (void)loop:(BOOL)value;
+- (BOOL)loop;
+
+- (void)multiPlay:(BOOL)value;
+- (BOOL)multiPlay;
+
+- (void)position:(float)value;
+- (float)position;
+
+- (void)positionMs:(int)value;
+- (int)positionMs;
+
+- (float)positionSeconds;
+- (float)soundDurationSeconds;
+
+@end
+
+
+@interface AVEnginePlayer ()
+
+@property(nonatomic, strong) AVAudioEngine *engine;
+@property(nonatomic, strong) AVAudioMixerNode *mainMixer;
+@property(nonatomic, strong) AVAudioUnitVarispeed *variSpeed;
+@property(nonatomic, strong) AVAudioPlayerNode *soundPlayer;
+@property(nonatomic, strong) AVAudioFile *soundFile;
+@property(nonatomic, assign) bool mShouldLoop;
+@property(nonatomic, assign) int mGaurdCount;
+@property(nonatomic, assign) bool mMultiPlay;
+@property(nonatomic, assign) float mRequestedPositonSeconds;
+@property(nonatomic, assign) AVAudioFramePosition startedSampleOffset;
+
+
+@end
+
+@implementation AVEnginePlayer
+
+@synthesize timer;
+
+// setupSharedSession is to prevent other iOS Classes closing the audio feed, such as AVAssetReader, when reading from disk
+// It is set once on first launch of a AVAudioPlayer and remains as a set property from then on
+- (void) setupSharedSession {
+#ifndef TARGET_OSX
+	static BOOL audioSessionSetup = NO;
+	if(audioSessionSetup) {
+		return;
+	}
+	NSString * playbackCategory = AVAudioSessionCategoryPlayAndRecord;
+#ifdef TARGET_OF_TVOS
+	playbackCategory = AVAudioSessionCategoryPlayback;
+#endif
+	[[AVAudioSession sharedInstance] setCategory:playbackCategory error: nil];
+    AVAudioSession * audioSession = [AVAudioSession sharedInstance];
+    NSError * err = nil;
+    // need to configure set the audio category, and override to it route the audio to the speaker
+    if([audioSession respondsToSelector:@selector(setCategory:withOptions:error:)]) {
+        if(![audioSession setCategory:playbackCategory
+             						  withOptions:AVAudioSessionCategoryOptionMixWithOthers
+                                        error:&err]) { err = nil; }
+    }
+	[[AVAudioSession sharedInstance] setActive: YES error: nil];
+	audioSessionSetup = YES;
+#endif
+}
+
+- (instancetype)init
+{
+    self = [super init];
+    if (self) {
+		[self setupSharedSession];
+		NSError * error = nil;
+		
+		// Engine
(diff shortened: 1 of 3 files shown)


danoli3:
Looking good!!

roymacdonald:
Hi, I have all this working on [ofxSoundObjects](https://github.com/roymacdonald/ofxSoundObjects/). it only on [dr_wav](https://github.com/mackron/dr_libs/blob/master/dr_wav.h), [dr_mp3](https://github.com/mackron/dr_libs/blob/master/dr_mp3.h), [dr_flac](https://github.com/mackron/dr_libs/blob/master/dr_flac.h) and [stb_vorbis](https://github.com/nothings/stb/blob/master/stb_vorbis.c), to load the files. 
Then it uses ofSoundStream to send whatever it is to the sound device. The good thing of this is that the player and its data is not "black boxed" on its way to the sound device. It is already fully cross platform. Having an global FFT object so it can behave exactly as ofSoundPlayer is no problem. I could add such. It also comes with the extra feature of writing audio files to disk.

I think that having an approach as this would allow to have a more unified sound engine, where you could be able to manipulate audio data from a sound file in the same way as you can with a sound input.

The current audio resampling of ofSoundBuffer is not good. I have also implemented using [libsamplerate](https://github.com/libsndfile/libsamplerate) and works really well.

I could add a bunch of optimizations as using SIMD as [ofxPDSP](https://github.com/npisanti/ofxPDSP)  does.

I think that this implementation would be a lot cleaner than having a different one for each platform.

ofTheo:
@roymacdonald couple of questions about ofxSoundObjects 

- Does it support m4a audio files? 
- Does it support multi play? ( ie: playing the same sound multiple times ) 
- Have you checked performance vs the fmod audio player? 

I have to update this PR but I basically have everything except getSpectrum / fft working. 
I might have Obj C peeps like @2bbb and @danoli3 take a look once its ready to make sure I am not leaking memory or doing anything terrible from an Obj C perspective 🙂 but I am curious if ofxSoundObjects or ofxAudiFile could be a good solution too ( as we'll need to make a windows based sound player too to switch away from fmod ).

roymacdonald:
Hi @ofTheo 
>     Does it support m4a audio files?
ofxSoundObjects use ofxAudioFile as the backend for loading files. So far it supports Wav, mp3, FLAC and Ogg Vorbis. But adding a library for m4a does not sound like much of a problem.

> Does it support multi play? ( ie: playing the same sound multiple times )
Yes it does. I have 2 branches, one that is just like the regular ofSoundPLayer and another one which gives you granular control over each playing instance, ie: you can set different speeds or change the playback position of a particular play instance, yet still it defaults to mimic ofSoundPLayer.
> Have you checked performance vs the fmod audio player?
Nope, but I have used it on some projects being able to play 2000+ sounds at the same time with no problem.

I think that it would be reasonable for me to make a special branch before adding into OF because there is useful things than can be extracted from the different branches.

ofxSoundObjects has an example that runs an fft, so adding the getSpectum part to the player is not much of a problem.

And running benchmarks to compare with "native" players should be also done. 

I would be super happy if we merged ofxSoundObjects into the OF core. Let me know what you think so I can work on it to make it mergeable.

BTW, I also have ofxSoundObjects working with ofxAudioUnits and ofxNDI.

ofTheo:
@roymacdonald nice that it doesn't need any libs compiled!
I imagine `aiff` and `m4a` might be fairly important if it was the core audio player. 

```
Yes it does. I have 2 branches, one that is just like the regular ofSoundPLayer and another one which gives you granular control over each playing instance, ie: you can set different speeds or change the playback position of a particular play instance, yet still it defaults to mimic ofSoundPLayer.
```

I'll give it a spin and try it with the soundPlayer example. Which branch would you recommend?

roymacdonald:
> I imagine aiff and m4a might be fairly important if it was the core audio player.
Certainly. I will look at such.

Which branch would you recommend? 
go with the master branch. experimental has more features, but the core of it is pretty much the same. 

Currently you need to set up the player along with the sound stream, etc. but It is quite straight forwards to implement something that behaves exactly as the current ofSoundPLayer. I mean, no need to setup the sound stream, connect to it, etc.

The one thing I think would be very important to update/fix in order to make this all work nicely is to properly implement the ofSoundBuffer resampling, which can be quite tricky and currently does not work nicely. I found out that the best approach is to use a library for such, like libsamplerate.

dimitre:
this is great news @ofTheo and @roymacdonald 
I would love to have global FFT, being able to write audio files to disk and not having fmod

dimitre:
Just wondering here @ofTheo if this could be merged as it is, and improved by PRs later.
I can update Xcode project to list the files.

ofTheo:
@dimitre I think I have a few local updates I need to push, but then might ask @2bbb @danoli3 and @roymacdonald to take a look. The FFT stuff should be doable natively, but after a couple of attempts I might ask people with better knowledge to take a look.

roymacdonald:
Hi @ofTheo i have been extra busy recently but I can take a look at this during the week. I will let you know.

ofTheo:
Just updated this - it still needs FFT. 
If you want to test quickly you can grab the two new `ofAvEngineSoundPlayer` files and then at the top of `soundPlayerExample` add this: 

```
	synth.setPlayer(make_shared<ofAVEngineSoundPlayer>());
	beats.setPlayer(make_shared<ofAVEngineSoundPlayer>());
	vocals.setPlayer(make_shared<ofAVEngineSoundPlayer>());
```

Would be great if someone who has a better idea of `Objective C` and or Sound like @roymacdonald @2bbb or @admsyn could take a look. I feel like its close but needs some tweaks. 

One nice thing it does handle well is switching audio outputs while its playing. 
It pauses the engine, saves the info of all the playing sounds and then restarts everything with the new output hardware / rate etc as it was when the audio was switched ( a lot of other apps just crash when this happens ).

dimitre:
Hey @ofTheo I'll test this one today and made some changes to make it work with ARC, not sure if they are correct.
https://github.com/dimitre/openFrameworks/tree/avengine
Cheers

ofTheo:
Thanks @dimitre !! 
Definitely in need of an objective-c expert here 🙂

libs/openFrameworks/sound/ofAVEngineSoundPlayer.m
```diff
@@ -0,0 +1,568 @@
...
+
+- (void)play;
+- (void)play: (float)startTime;
+- (void)pause;
+- (void)stop;
+
+- (BOOL)isLoaded;
+- (BOOL)isPlaying;
+
+- (void)volume:(float)value;
```
2bbb, line 45: this is a small part, getter/setter method in basic obj-c style is

```
- (void)setVolume:(float)value;
- (float)volume;
```

then we can use property style access like

```
obj.volume = 0.5f; // equivalent to [obj setVolume:0.5f];
float v = obj.volume; // equivalent to float v = [obj volume];
```

libs/openFrameworks/sound/ofAVEngineSoundPlayer.m
```diff
@@ -0,0 +1,709 @@
...
+@property(nonatomic, assign) bool mMultiPlay;
+@property(nonatomic, assign) float mRequestedPositonSeconds;
+@property(nonatomic, assign) AVAudioFramePosition startedSampleOffset;
+
+
+@end
+
+@implementation AVEnginePlayer
+
+@synthesize timer;
```
2bbb, line 95: basically, no need to explicit `@synthesize` if we use timer as `self.timer` (or `_timer` on internally. `_timer` will be defined automatically with `@prorpety` declaration).
if we want to use simply as `timer` then `@synthesize timer` is needed.

dimitre:
One thing I really miss compared to default ofSoundPlayer is the capability of playing audio backwards setting a negative speed. I've noticed the values are clamped but probably varispeed doesn't allow negative speed or maybe it is other property. I've noticed a while ago the same thing in @admsyn ofxAudioUnit.
https://github.com/admsyn/ofxAudioUnit/

Maybe looking at ofxAudioUnit can be helpful for fft too

dimitre:
Heyy I think this should be merged soon, so it is easier for all to make PRs on the top of the existing code.
And there are other changes in the project that should be made to acommodate this:
- [ ] Adding the files to osx XCode template
- [ ] Adding the files to exclusions in certain platforms (if needed)
- [ ] Adding the files to ios XCode template
I can take care of this changes in parallel

danoli3:
Will need Objective-C++ filename of .mm instead of .m 
Giving it a spin

dimitre:
Yeah @danoli3 I've did that in my fork
and some other ARC changes. 
PS: I don't know much about obj-c
https://github.com/dimitre/openFrameworks/tree/avengine

danoli3:
Awesome yeah I've been testing it out tonight! Looks and works great! I'm testing in the ARC branch however just with "-fno-objc-arc" currently on this .mm! I think this is production ready

dimitre:
I would love to see this merged, even if incomplete. I don't think it will interfere with any other OF functionality.
I am available this week to work on template updates to include the new files.

ofTheo:
Thanks! 
There are definitely some things in there that I’m not sure I quite understand correctly. :)

dimitre:
Heyy let's merge this one! 
I'll be using / testing a lot. 
Future work can be done in other PRs :)

ofTheo:
okay sounds good @dimitre ! 
@danoli3 if you have any small fixes you could PR to master.

ofTheo:
reverting this as there is too much to fix in master to get this working. 
the ARC requirements are quite intsensive and there is still a lot of iOS code which needs ifdefs

PR #6732, opened 2021-04-07
loadModel deprecated, using load instead

dimitre:
loadModel now calls load function and shows deprecated message.
I've tested using loadModel and load here, it is OK in my side.
I've seen ident was mixed between tabs and spaces and changed to tabs. 
Let me know if it is not the OF style.
Thanks

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
@@ -13,100 +13,110 @@ ofxAssimpModelLoader::ofxAssimpModelLoader(){
 }
 
 ofxAssimpModelLoader::~ofxAssimpModelLoader(){
-    clear();
+	clear();
+}
+
+// DEPRECATED
+bool ofxAssimpModelLoader::loadModel(string modelName, bool optimize){
+	load(modelName, optimize);
+}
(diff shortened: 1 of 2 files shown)


dimitre:
I've just noticed the errors, missed a return, will fix soon

ofTheo:
@dimitre Thanks!

fyi - if you do the deprecation like this: 
https://github.com/openframeworks/openFrameworks/blob/51bcbf3c6bd727db9b31782d29fb078c9b431fde/libs/openFrameworks/video/ofAVFoundationPlayer.h#L88-L91 then it shows up as a warning in the compiler.

dimitre:
Thanks for the info @ofTheo 
I've did this way in this pull request :
```c++
		OF_DEPRECATED_MSG("Use load instead", bool loadModel(std::string modelName, bool optimize=false));
		OF_DEPRECATED_MSG("Use load instead", bool loadModel(ofBuffer & buffer, bool optimize=false, const char * extension=""));
```
and it shows a warning on XCode. Is this the same thing you mentioned?
I like the more verbose message in your example, I can update that.

ofTheo:
Thanks @dimitre !
If you want to and more verbose deprecated message as above we can merge this.

dimitre:
@ofTheo done!

Issue #6738, opened 2021-04-30
ofGetUnixTime() returns a signed int, 2038 is not so far ahead any more.


armadillu:
Can we change the return type of ofGetUnixTime() into a size_t?
https://en.wikipedia.org/wiki/Year_2038_problem

danoli3:
size_t and we should think about also remaking all the core functions that
use floats to support doubles as well :)

arturoc:
can you send a PR?

armadillu:
PR ready. https://github.com/openframeworks/openFrameworks/pull/6740

2bbb:
is `size_t` better?
i think, it's better to use `uint64_t`.
there is a possibility what `size_t` is `uint32_t` on some environments.

armadillu:
You are probably right, I bet the rpi does that. I'll prepare another PR.

arturoc:
size_t is 32bits on 32bits platforms yes

PR #6740, opened 2021-05-02
ofGetUnixTime() return changed to size_t from uint

armadillu:
As discussed on Issue #6738, make ofGetUnixTime() return size_t to hopefully avoid troubles in 2038.


armadillu:
safer to make this return uint64_t

PR #6741, opened 2021-05-02
Freeimage flags

elliotwoods:
allow the user to send their own flags to freeimage when loading an image through ofLoadImage


elliotwoods:
ah - also there's a fix for colour maths (especially with integer it would overflow)
also the ofThreadChannel size - but i thought that was already merged?

elliotwoods:
yes that was already merged here:
https://github.com/openframeworks/openFrameworks/commit/e06be9a4c61a0bb7298e17bf36f9e4f4954e76c8

ofTheo:
Hey @elliotwoods. Thanks!
Can you remove the non relevant commits from the PR? 

Usually I never work on master locally but just keep it tracked to OF upstream master. 
Then just checkout a new branch from that and add the changes.

ofTheo:
ping @elliotwoods

Issue #6742, opened 2021-05-02
linux testAllExamples.sh fails


themancalledjakob:
discovered a tiny bug in `testAllExamples.sh`
It stumbles over README.md (and eventual qtCreator build-* directories), and then fails.
```
./testAllExamples.sh: line 11: cd: README.md: Not a directory
-----------------------------------------------------------------
building  + README.md
./testAllExamples.sh: line 20: cd: README.md: Not a directory
make: *** No rule to make target 'Debug'.  Stop.
error compiling README.md
```

Since all Examples end with "Example"... would it be an idea to do something like this instead?

```
#!/bin/bash

export LC_ALL=C
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
OF_ROOT=$SCRIPT_DIR/../..

THREADS=""
if [[ $1 == "-j*" ]]; then
  THREADS="$1"
fi

find $OF_ROOT/examples -type d \( -name android -o -name ios -o -name gles \) -prune -false -o -type d -name '*Example' -exec bash -c 'cd {} && make $THREADS && make run' \;

```
Also, why not make it possible to use multiple threads :)

I am never sure, is it a good idea to immediately create a PR for a small fix like this and link it, or is it better to first have it confirmed?

arturoc:
looks good, send a PR please

PR #6743, opened 2021-05-06
fix testAllExamples.sh

themancalledjakob:
testAllExamples.sh stumbled over README.md (and eventual qtCreator build-* directories), and then failed.
This fixes it and adds multithreaded compile

#6742

### DIFF
--- a/scripts/linux/testAllExamples.sh
+++ b/scripts/linux/testAllExamples.sh
@@ -1,40 +1,25 @@
 #!/bin/bash
 
 export LC_ALL=C
-cd ../../examples
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+OF_ROOT=$SCRIPT_DIR/../..
 
-for category in $( ls . )
+# parameters
+THREADS=""
+BUILD_TYPE="Release"
+for arg in "$@"
 do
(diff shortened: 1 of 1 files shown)


cdorsman:
Hi,

Does anyone look at the state of some of these checks?
Two of these are still being queued for at least 527 days.

themancalledjakob:
hmm i don't know how to re-run a check. Since it's changing merely a shell script, I don't think that any of these checks could possibly fail though.

not sure why the PR is not accepted yet, either it wasn't seen, it changed too many things in one go, or possibly someone didn't like the implementation. Also, I don't think I created an issue referencing this PR...

ofTheo:
apologies @themancalledjakob 
part of the issue I see with this is that it removes the checks for the platform specific examples which means you can run this script if you are working with the main repo. 

At this point it might be simpler to close this and do a new PR as some of the checks don't even exist anymore. 

Apologies for the hassle. 
If you want to point to the new PR in this thread, I'll close this one then. 

Thanks!! 🙏

themancalledjakob:
hey theo! :-) really no need to apologize, I am pretty impressed by how fast, clear and friendly you respond keeping in mind the complexity of such a project. thanks for that!
In any case, I will check the current script, if it still has the same issues I'll do a new PR for the fix and a separate one for multithreaded compile options.

themancalledjakob:
@ofTheo  #7159

themancalledjakob:
@ofTheo  #7160 - add multithreaded option

PR #6750, opened 2021-05-29
Fix 6722 of easy cam copy

jaysonh:
Added a copy constructor for ofCamera and ofEasyCam

Not sure if this is useful to other people but used this quite a bit in some of my own projects.

Please note ofEasyCam copy constructor does not copy the listeners, this wasn't necessary for my own projects and seemed like a bad idea.

### DIFF
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -35,8 +35,8 @@ ofCamera::ofCamera(const ofCamera & camera) :
 	vFlip(vFlip),
 	renderer(renderer)
 { 
-
 }
+
 //----------------------------------------
 ofCamera::~ofCamera() {}
 
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -24,6 +24,20 @@ vFlip(false)
 }
 
 //----------------------------------------
+ofCamera::ofCamera(const ofCamera & camera) :
+	isOrtho(camera.isOrtho),
+	fov(camera.fov),
+	nearClip(camera.nearClip),
+	farClip(camera.farClip),
+	lensOffset(lensOffset),
+	forceAspectRatio(forceAspectRatio),
+	aspectRatio(aspectRatio),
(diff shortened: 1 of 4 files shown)


ofTheo:
Looks like this isn't working in the CI. @jaysonh 
https://github.com/openframeworks/openFrameworks/runs/2769932119?check_suite_focus=true#step:6:131 

I wonder if it would  be simpler to remove the constructors and destructors ( as we are not currently using them for anything ) and then let C++ do a default copy constructor?

arturoc:
ofCamera needs a virtual destructor cause otherwise destroying an ofEasyCamera would leak memory. so the copy constructor on ofCamera would make sense. ofEasyCamera can't be copied properly because it uses events so it's better to not allow it to be copied imo

ofTheo:
@arturoc doh, yup I see! 

Yeah, I was hoping to avoid stuff like this: 
https://github.com/openframeworks/openFrameworks/pull/6750/files#diff-5917e4588a8cab432cbfe3dab04f711ad05e1b2eff22889a88723423bf564598R29-R72 which gets pretty hard to stay on top of over time. 

We could probably at least allow the copy constructor for ofCamera and maybe use this approach so we don't have to add variables if they change over time? https://stackoverflow.com/a/35127609

arturoc:
yeah copy constructors are horrible. i didn't remember you can now do that in c++11, let's do that yes.

ofTheo:
Thanks @arturoc - yeah also had no idea that was an option :) 

@jaysonh do you want to update your PR to drop the ofEasyCam constructors/copy constructors/destructors and then implement the copy constructors for ofEasyCam using `default` described here: https://stackoverflow.com/a/35127609 ?

Otherwise I can close this and take a stab at it. 

Thanks!
Theo

jaysonh:
hey guys sorry for the hassle, the ofEasyCam was just a quick fix that I thought might help the original issue opener.

@ofTheo  i'll update the PR with ofEasyCam default constructors in the next couple of days.

arturoc:
The idea was to do this for ofCamera, ofEasyCam probably shouldn't have copy constructors at all since it has events that are problematic to copy

PR #6771, opened 2021-09-28
Bugfix. Fixes broken video renderer for newer codec tools on Windows. closes #6765

### LINKED ISSUE #6765: ofDirectShowPlayer hits "failed to find the video renderer"
When I tried to play `videoPlayerExample` in v0.11.2 with VisualStudio, just got: 

```
failed to find the video renderer
```
and wouldn't update video frame. Here is the line.https://github.com/openframeworks/openFrameworks/blob/7c46c8a9415703c1cf7ce1bb66c91cc97851fa19/libs/openFrameworks/video/ofDirectShowPlayer.cpp#L619

my env:

* Windows 10 Home 21H2 19044.1202
* VisualStudio2019 Community (v142)
* K-Lite Codec Pack 16.4.0

Before half a year I could open any video files on same machine, but I keep almost all system (like Win OS/framework, NVidia firmware, K-Lite codec etc.) up-to-date to their latest, and now I've got the error above. Looking up to docs DirectShow from Microsoft, I found some features are going to be deprecated such as [SampleGrabber](https://docs.microsoft.com/en-us/windows/win32/directshow/using-the-sample-grabber). Yet I couldn't make sure what the exact problem is and how to hot-fix it.

An alternative that works for me is just to use this: [DomAmato/ofxWMFVideoPlayer](https://github.com/DomAmato/ofxWMFVideoPlayer). This plugin can be built with v0.11.2 without errors and can open/play videos.  Downgrading software/apis like VisualStudio might work too, but it's not the one for me.

This issue should be relatable with #5156 .

ofTheo:
Newer installs of K-Lite Codec tools are defaulting to a newer Directshow rendering pipeline, breaking video playback on Windows. This approach tries both the existing and newer approaches and only fails if it can't find either. 

Fixes #6765

Thanks to @nama-gatsuo for reporting and @hanasaan for the fix suggestion!

### DIFF
--- a/libs/openFrameworks/video/ofDirectShowPlayer.cpp
+++ b/libs/openFrameworks/video/ofDirectShowPlayer.cpp
@@ -615,11 +615,16 @@ class DirectShowVideo : public ISampleGrabberCB{
             IPin* pinOut = 0;
 
             hr = m_pGraph->FindFilterByName(L"Video Renderer", &m_pVideoRenderer);
-            if (FAILED(hr)){
-                printf("failed to find the video renderer\n");
-                tearDown();
-                return false;
-            }
+
+			if (FAILED(hr)) {
+				//newer graphs use Video Mixing Renderer 9
+				hr = m_pGraph->FindFilterByName(L"Video Mixing Renderer 9", &m_pVideoRenderer);
+				if (FAILED(hr)) {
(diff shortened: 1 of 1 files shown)


ofTheo:
closing as GH support thinks this PR can't trigger actions. going to do a new PR to see if actions will work again.

PR #6777, opened 2021-10-26
Fix issue in ofSerial during Win32 enumeration

spolette:
This fix a bug occurring if the device's friendly name contains "COM" string.
For instance, if the device's friendly name is "Virtual COM Port (COM15)", then
enumeration will return the following device path : "COM Port (COM15" instead
of "COM15"


spolette:
> Doesn't this break the case where the name is simply COM*

Well, in theory, it could. However, I've never seen the case where windows shows a name with simply "COM*". I've always seen something like "Communication Port (COM1)", or "USB Serial Port (COM12)".
Moreover, the original code was expecting a friendly name which ends with ")", so we can assume that there is a "(" somewhere before the ")", therefore I don't think this change would break anything.

PR #6778, opened 2021-11-02
ofURLFileLoader: don't reuse CURL* handle

### LINKED ISSUE #6776: ofURLFileLoaderImpl crashes curl when used from multiple threads
OF 0.11.0 - Linux + Android (but any platform using curl should see the bug)

When two threads try to download a file, the app is likely to crash (perhaps after eating all the RAM available...).

The crash can be exposed with a little change to `imageLoaderWebExample`:
ask OF to download the online image in sync mode just after the async download has been triggered by key press:
``` C++
	img.clear();
	ofLoadURLAsync("https://openframeworks.cc/about/0.jpg","about");
	ofLoadURL("https://openframeworks.cc/about/0.jpg"); // ADDED LINE
	loading =true;
}
```

The reason is that a CURL handler (which is created by `curl_easy_init()`) can only be used by a single thread  (see: https://curl.se/libcurl/c/threadsafe.html : "You must never share the same handle in multiple threads").

A possible quick fix is to remove the `curl` member from the `ofURLFileLoaderImpl` class, and create it locally at the beginning of the `handleRequest()` function:
``` C++
ofHttpResponse ofURLFileLoaderImpl::handleRequest(const ofHttpRequest & request) {
	std::unique_ptr<CURL, void(*)(CURL*)> curl =
		std::unique_ptr<CURL, void(*)(CURL*)>(curl_easy_init(), curl_easy_cleanup);
(...)
}
```

Ant1r:
fixes #6776

`ofLoadURLAsync` and `ofLoadURL` are called from different threads, however from https://curl.se/libcurl/c/threadsafe.html you can read that: "You must never share the same handle in multiple threads".

So here we fix this by creating a new handle for each request. I think the overhead is acceptable; it could be noticeable when downloading many small files though.

There might be a better way using:
``` C++ 
thread_local std::unique_ptr<CURL, void(*)(CURL*)> curl 
```
but I couldn't figure how to to it properly, sorry...

### DIFF
--- a/libs/openFrameworks/utils/ofURLFileLoader.cpp
+++ b/libs/openFrameworks/utils/ofURLFileLoader.cpp
@@ -51,15 +51,12 @@ private:
 	ofThreadChannel<ofHttpResponse> responses;
 	ofThreadChannel<int> cancelRequestQueue;
 	set<int> cancelledRequests;
-	std::unique_ptr<CURL, void(*)(CURL*)> curl;
 };
 
-ofURLFileLoaderImpl::ofURLFileLoaderImpl()
-:curl(nullptr, nullptr){
+ofURLFileLoaderImpl::ofURLFileLoaderImpl() {
 	if(!curlInited){
 		 curl_global_init(CURL_GLOBAL_ALL);
 	}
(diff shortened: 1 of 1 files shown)


arturoc:
Thanks!

Issue #6585, opened 2020-05-10
Discussion about supporting C++17


2bbb:
Let's discussion about C++17.

2bbb:
https://github.com/openframeworks/openFrameworks/issues/6488

2bbb:
https://forum.openframeworks.cc/t/does-of-not-work-with-c-17/29925

2bbb:
on macOS early 10.14, we can't use C++17 with oF0.11.0 release, with both `OF_USING_STD_FS 0` and `OF_USING_STD_FS 1`.
the biggest cause of this problem is by Apple LLVM/clang.

but essentially, I think it lies under extending `std` namespace adding `filesystem`.
extending `std` is so evil way.

I fixed bug about difference between `std::filesystem` and `std::experimental::filesystem`.
but it doesn't fix problem on macOS early 10.14.

and if try to use C++17 on macOS 10.14, then we need to remove `boost::filesystem` from `std` with `OF_USING_STD_FS 0` or adding so many CPP macros and separating codes.

I want to propose what adding `of` (or other name) namespace and add `boost::filesystem` or `std::experimental::filesystem` or `std::filesystem`.
it still needs some macros to absorb the diff of some APIs, but doesn't need to adding macros.

what do you think?

roymacdonald:
Hey @2bbb I think that this move to support C++17 will eventually happen so we should consider it right now how to address it. I think that it is important that these changes don't alter the way in which of is used in general, as far as this is possible and does not bloat the codebase or needs the use of hacky fixes.
Can you post the code changes that you are proposing?
best!

oxillo:
Do we still need to use boost::filesystem in C++17 (meaning C++17 and OF_USING_STD_FS=0) ?
I would say that most people using C++17 will expect to use std::filesystem.
So I would suggest that OF_USING STD_FS is only applicable to C++14 and below.
This will also make solving #6587 a lot easier...

2bbb:
@roymacdonald 

i will push to my repository asap.

@oxillo 
basically, i agree to what you say.
but, on older macOS (~Mojave) has C++17 option but can't use `std::filesystem::path`...
it still needs option to use `boost::filesystem`.

arturoc:
boost was only used temporarily until there was filesystem support in std. 

The idea is not to keep it but to remove it as soon as we can and stop even shipping the boost libraries. 

[...]

PR #6783, opened 2021-11-14
fix: ambiguous byte conflict with windows header

edmBernard:
A really small PR.

There is conflict with windows header and the definition of byte symbol.


edmBernard:
I just found there is a discussion with other ambiguous issues and the fix at the end seem to be the same https://github.com/openframeworks/openFrameworks/issues/6585
but I don't find PR with it

ofTheo:
Just ran into this. 
Thank you for fixing it!

Issue #6784, opened 2021-11-19
macOs XCode Schemes - Target building unexpected executables


dimitre:
Hello, I'm trying to debug an addon and found some different behaviors on Release and Debug versions, both Silicon or Rosetta, so I've tested all the configurations, erasing the software each time and getting the file information in finder, here are the results:

Target | Result
---|---
Debug | Application (Apple Silicon)
Debug Rosetta | Application (Intel)
Release | Application (Universal)
Release Rosetta | Application (Universal)

In fact I've compared the output of Release and Release Rosetta and they build the same.
```
diff /Volumes/tool/ofw/apps/WerkApps/EmptyTest/bin/Empty.app/Contents/MacOS/Empty /Volumes/tool/ofw/apps/WerkApps/EmptyTest/bin/Empty0.app/Contents/MacOS/Empty 
```
And to double check I've tested both with git latest and 0.11.2
Anybody else can confirm this issue?
Thanks

ofZach:
this isn't addressing your issue, but in case it's helpful one thing I have noticed I've need to add arm64 to "excluded  architectures" to excluded arm linking even on rosetta targets -- when I needed to link for example with an intel only library or code that's kind of intel specific.

dimitre:
Thanks @ofZach for this workaround.
I'll just mention one issue that it is probably related to this
https://github.com/leadedge/ofxNDI/issues/19

dimitre:
I think this is the flag that should be set to Yes, so Release builds are as expected. 
![Screen Shot 2021-11-22 at 10 46 55](https://user-images.githubusercontent.com/58289/142872669-4329ac28-0e78-43aa-99d6-699514576468.png)

dimitre:
Closing here, fixed by the pull request

PR #6785, opened 2021-11-22
Fix XCode release schemes

dimitre:
XCode Release and Release Rosetta both generates Universal identical apps right now.
This pull request keeps Release behavior similar to Debug, compiling Apple Silicon or Intel code depending on the scheme selection.
I think the absence of this key turned the default to NO, so it was building other architectures together causing some issues on addons based in a specific architecture.

More details of this issue here:
https://github.com/openframeworks/openFrameworks/issues/6784#issuecomment-975544869


ofTheo:
Cool - looks good to me! 
Thanks!!

Issue #6534, opened 2020-02-09
config.make PROJECT_CC and PROJECT_CXX not working


dimitre:
I'm having issues when trying to set a custom compiler in config.make
using PROJECT_CXX and PROJECT_CC variables.
after some investigation on openFrameworks makefiles I've noticed both CC and CXX variables are added only when empty using ?= operator.
But in fact they were already set in the beginning of the config.make

it is possible to inspect using this as the custom Makefile
```
$(info PROJECT_CC begin is $(PROJECT_CC))
$(info PROJECT_CXX begin is $(PROJECT_CXX))
$(info CC begin is $(CC))
$(info CXX begin is $(CXX))


ifneq ($(wildcard config.make),)
	include config.make
endif



# make sure the the OF_ROOT location is defined
ifndef OF_ROOT
	OF_ROOT=$(realpath ../../..)
endif

# call the project makefile!

include $(OF_ROOT)/libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk

$(info PROJECT_CC end is $(PROJECT_CC))
$(info PROJECT_CXX end is $(PROJECT_CXX))
$(info CC end is $(CC))
$(info CXX end is $(CXX))

```

I've noticed I had to unset this variables for the per project CC and CXX settings work, like this
```
make -e CC='' CXX=''
```

using
openframeworks git HEAD (1afca9f5925ffd0484405bd015181fa8b02bfa09)
macOS 10.14.6

more info: https://forum.openframeworks.cc/t/building-of-with-clang/15329/13?u=dimitre

dimitre:
I'm now testing on Ubuntu 20.04 and I'm having the same issue there. 
using openFrameworks 0.11.0
I can only make it work with a custom compiler invoking CXX as a parameter like this
```
make -e CXX='zapcc++ '
```

PR #6787, opened 2021-11-22
PROJECT_CC and PROJECT_CXX fix

dimitre:
changes to fix https://github.com/openframeworks/openFrameworks/issues/6534

I've studied the concatenation of make file includes to understand how CXX and CC variables were not being set per project when using PROJECT_CC and PROJECT_CXX
with this two changes it arrives at the end with the right variables set.

To understand better how variables were set I've added this in the end of my test project Makefile
```make
$(info -----------------)
$(info PROJECT_CC final is $(PROJECT_CC))
$(info PROJECT_CXX final is $(PROJECT_CXX))
$(info CC final is $(CC))
$(info CXX final is $(CXX))
$(info -----------------)
```

as the changes are isolated inside the ifdef PROJECT_CC / ifdef PROJECT_CXX it should not affect anything else

### DIFF
--- a/libs/openFrameworksCompiled/project/makefileCommon/config.project.mk
+++ b/libs/openFrameworksCompiled/project/makefileCommon/config.project.mk
@@ -374,7 +374,7 @@ ifdef MAKEFILE_DEBUG
 endif
 
 ifdef PROJECT_CXX
-	CXX ?= $(PROJECT_CXX)
+	CXX = $(PROJECT_CXX)
 endif
 
 ifdef PLATFORM_CXX
@@ -382,7 +382,7 @@ ifdef PLATFORM_CXX
 endif
 
 ifdef PROJECT_CC
-	CC ?= $(PROJECT_CC)
+	CC = $(PROJECT_CC)
 endif
 
 ifdef PLATFORM_CC


ofTheo:
I _think_ this is okay. 
I am guessing we don't want to _not set_ these variables if the `PROJECT_CC` is set and `CC` already has a value. 

The other option would be to append the CC instead of setting it. 
Curious what others thoughts are as this is in the common makefile?

Will merge in a couple of days  if there aren't any objections.

dimitre:
nicee

dimitre:
> I _think_ this is okay. I am guessing we don't want to _not set_ these variables if the `PROJECT_CC` is set and `CC` already has a value.
Exactly! In the few cases one chooses to have a custom CC in a project one explicitly wants to use it :)
It took a while for me to discover it wasn't working, after this change I was able to test Zapcc compiler on linux.

Btw, just sharing a modern and fast linker here:
https://github.com/rui314/mold

arturoc:
Mmh not sure, the way it's now you can do 

```
CC=gcc make
```

to quickly get gcc as C compiler if that wasn't the default. That's pretty standard in makefiles. Also CC is usually not set so setting PROJEC_CC in the make config in the project should work fine.

dimitre:
@arturoc ok so maybe I'm doing it wrong 
I've read in config.make this section
################################################################################
# PROJECT COMPILERS
#   Custom compilers can be set for CC and CXX
#		(default) PROJECT_CXX = (blank)
#		(default) PROJECT_CC = (blank)
#   Note: Leave a leading space when adding list items with the += operator
################################################################################

if we set PROJECT_CXX and PROJECT_CC it doesn't change final CXX and CC values
if we directly set CC and CXX it works, so maybe it is a fact of changing the somment section?

arturoc:
what do you get if you type

```
echo $CC
```

on the terminal?

dimitre:
> what do you get if you type
> 
> ```
> echo $CC
> ```
> 
> on the terminal?

a blank new line. it seems to echo a line feed character. macOs here

arturoc:
Then it might be a bug in the makefiles where it's setting CC before trying to read PROJECT_CC and set it. Ideally it should happen in reverse but otherwise this PR is better than the current state for sure

PR #6788, opened 2021-11-26
ofAVFoundationPlayer.mm Minor tweaks, avoid texture allocation every frame

dimitre:
comment out CVPixelBufferUnlockBaseAddress in every frame. not needed for normal video
Some minor changes to avoid allocating videoTexture every frame.


ofTheo:
Looks good to me! Thanks!!

PR #6791, opened 2021-12-02
ofAVFoundationVideoPlayer - Add autorelease to dispatch_async

dimitre:
I've noticed in XCode the video player create lots of threads and release few. 
This change allows the program to release threads over time.

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -191,6 +191,9 @@ static const void *PlayerRateContext = &ItemStatusContext;
 	}
 	
 	dispatch_async(queue, ^{
+		@autoreleasepool {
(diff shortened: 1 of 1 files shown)


ofTheo:
Cool!! 
Thanks! Did you see the threads getting removed after this change?

dimitre:
yess solved the issue

ofTheo:
Thanks!! 🙂

armadillu:
Wouldn't the  autorelease pool wrap the whole block? Did it not compile doing it that way? This seems very odd...

dimitre:
@armadillu yes it can wrap the entire block.
I just didn't know it was the correct way. Easy to change that.

I'm happy the thread problem is finally gone here. check it out if you have time / still use macOs
it now has a similar thread usage as other video players like this
https://github.com/unixzii/MetalPlayer

Issue #6780, opened 2021-11-11
ofVideoPlayer::setPosition glitch on MacOS


kylemcdonald:
The following code will work for the `fingers.mov` but when I try some other videos, sometimes it fails.

```c++
#pragma once

#include "ofMain.h"

class ofApp : public ofBaseApp{
	public:
    ofVideoPlayer player;
    void setup(){
        player.load("fingers.mov");
        player.play();
    }
    void update(){
        player.update();
    }
    void draw(){
        player.draw(0, 0, ofGetWidth(), ofGetHeight());
    }
    void mousePressed(int x, int y, int button) {
        player.setPosition(ofMap(x, 0, ofGetWidth(), 0, 1, true));
    }
};
```

One easy way to get a video that fails is to upscale fingers:

```
ffmpeg -i fingers.mov -vf scale=3840:-1 fingers-4k.mp4
```

Using that will cause the video playback to pause.

I have other videos with other codecs that will work (e.g. 4K Prores 422 LT) but if I add `player.update()` immediately after `player.setPosition()` then it will crash the app.

For all these videos I can easily seek in the QuickTime app with no latency, so this feels like a problem with the way OF uses the Mac SDK rather than a problem with my codec choice or videos.

2bbb:
@kylemcdonald

hi, kyle!

i tried to research this problem, but this code and fingers-4k.mp4 don't crash app on my envs.
what envs do you use?

FYI, my envs:
* both debug/release build
* oF 0.11.2
* Xcode 12.4
* Catalina 10.15.7
* MacBook Pro (15-inch, 2019)

dimitre:
hello Kyle, I've upscaled fingers with ffmpeg and had intermittent issues decoding too. 
But the same issues appear on Quicktime/Preview also. In the same video positions.
Only VLC is able to play correctly this specific video after upscaling.

Other than that I think there is room for improvement in ofAVFoundationVideoPlayer too

dimitre:
I've tested a lot AVFoundation player and in my usage the playback stops working if I seek a position before the actual position. and resumes after the playhead is equal to the position it was.
Today I found a fix which is commenting out this lines in ofAVFoundationVideoPlayer.m
```
	if (!bStream && (CMTimeCompare(time, videoSampleTime) < 0)) {
		// if jumping back in time
		[self createAssetReaderWithTimeRange:CMTimeRangeMake(time, duration)];
	}
```
And it works perfect here.
[...]

PR #6797, opened 2021-12-14
AVFoundationVideoPlayer fix - set video position to a previous position than the playhead

dimitre:
this particular call of createAssetReaderWithTimeRange is exclusive if we seek a position of the video before the actual playhead.
but it creates a strange situation where videoSampleTime is not correctly updated and we got stuck outside the "while" loop that checks the video buffer for new content.
no video is displayed until the playhead goes to the previous position again.

This fix supress this function here, but  it is going to be called anyway in updateFromAssetReader, but this time videoSampleTime is correctly updated and video is displayed correctly with good performance.

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -1229,10 +1229,10 @@ static const void *PlayerRateContext = &ItemStatusContext;
 	time = CMTimeMaximum(time, kCMTimeZero);
 	time = CMTimeMinimum(time, duration);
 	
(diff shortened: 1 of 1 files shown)


ofTheo:
Hey @dimitre.
I gave this a whirl and it seems to work as expected for me. 
It might make sense to comment out the if statement its in too?

Happy to merge when you are good to go.

dimitre:
Hello @ofTheo great news! I've just commented out the if and we are good to go. 
I think this PR can potentially close https://github.com/openframeworks/openFrameworks/issues/6780

ofTheo:
Thanks @dimitre! 
Hmm some of the build checks are failing, but unrelated to this PR so just going to merge this.

Issue #6816, opened 2021-12-24
warning TARGET_OS_IPHONE_SIMULATOR


dimitre:
There are some warnings regarding macro TARGET_OS_IPHONE_SIMULATOR
```
ofConstants.h:94:67: warning: 'TARGET_OS_IPHONE_SIMULATOR' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
```
From what I've read the valid ones are 
```
    TARGET_OS_SIMULATOR
    TARGET_IPHONE_SIMULATOR
```
and it seems TARGET_IPHONE_SIMULATOR is deprecated, so I think TARGET_OS_SIMULATOR is the best choice.
I can make a PR correcting that if you agree

dimitre:
This PR keeps deprecated define TARGET_IPHONE_SIMULATOR but not the TARGET_OS_IPHONE_SIMULATOR which doesn't work anymore
https://github.com/openframeworks/openFrameworks/pull/6818

PR #6818, opened 2021-12-26
Updating old defines - TARGET_OS_IPHONE_SIMULATOR

dimitre:
This commit changes the old define TARGET_OS_IPHONE_SIMULATOR
to this one : TARGET_IPHONE_SIMULATOR which is now deprecated but still working.
fixes some compiler warnings when using make

This PR closes https://github.com/openframeworks/openFrameworks/issues/6816


danoli3:
Awesome yeah looks good

PR #6819, opened 2021-12-26
int to std::size_t to match ofMatrix4x4

dimitre:
and it fixes some more strict compiler warnings "Implicit conversion loses integer precision:"
And some space to tabs fix.

Let me know if it is best to make different PRs for readability
only changed code was this lines
```c++
	float& operator[]( std::size_t n ){
	float operator[]( std::size_t n ) const {
```

instead of
```c++
	float& operator[]( int n ){
	float operator[]( int n ) const {
```


dimitre:
@ofTheo I'll be merging this one as it was reviewed by @artificiel as good to merge here: https://github.com/openframeworks/openFrameworks/issues/7741

Issue #6814, opened 2021-12-16
Suggestion : ofSystemTextBoxDialog - focus on text input on macOs


dimitre:
It would be great to have the focus on the text input on ofSystemTextBoxDialog.
Today we have to click inside text input after window appears to be able to type

PR #6820, opened 2021-12-27
ofSystemTextBoxDialog - Focus on textfield

dimitre:
this way we can answer without using the mouse to focus the textfield.
it closes https://github.com/openframeworks/openFrameworks/issues/6814


dimitre:
I'll now explain better this PR.
ofSystemTextBoxDialog is a modal window, but "text field" isn't focused, so to be able to type inside now I should click the window, type and press enter. 
With this PR there is no more need to click in the window to focus.

I'm using this a UI project, so I can "command + " click any slider to input manually the values, and soon strings and other values. This way I don't need to worry about programming a text input, handling accentuation, enconding etc.
https://github.com/dimitre/ofxMicroUI

and values can be input without having to click a new window. very immediate.

PR #6821, opened 2022-01-06
ofVideoPlayer setPlayer std::move shared_ptr

dimitre:
so it doesn't increment and decrement use_count.

### DIFF
--- a/libs/openFrameworks/video/ofVideoPlayer.cpp
+++ b/libs/openFrameworks/video/ofVideoPlayer.cpp
@@ -17,7 +17,7 @@ ofVideoPlayer::ofVideoPlayer (){
 
 //---------------------------------------------------------------------------
 void ofVideoPlayer::setPlayer(shared_ptr<ofBaseVideoPlayer> newPlayer){
-	player = newPlayer;
+	player = std::move(newPlayer);
 	setPixelFormat(internalPixelFormat);	//this means that it will try to set the pixel format you have been using before. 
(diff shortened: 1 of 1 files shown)


ofTheo:
Sorry. I think this is probably not desired behavior. 

I had left this open to just double check, but I think we don't want to treat the setPlayer as a unique_ptr which the use of std::move is basically doing. 

If someone has a couple of different `shared_ptr<ofBaseVideoPlayer>` players and want to switch back and forth between them it is quite easy to do by using setPlayer ( we have a project where we do this exact thing ). But with this PR the player they pass in will be NULL after they call setPlayer and they won't be able to use it again. 

I think for the cost of an increment / decrement it is worth keeping that flexibility?

ofTheo:
@tyhenry ahh I see. It is just affecting the extra use_count from the local copy. 

I did a quick sanity check and it looks to be the case.
So then this all looks good ( and I learned something new as a bonus  ) 
Thanks all!

```
	shared_ptr <ofAVFoundationPlayer> tPlayer = make_shared<ofAVFoundationPlayer>();
	
	fingerMovie.setPlayer(tPlayer);
	fingerMovie.load("movies/fingers.mov");
	fingerMovie.setLoopState(OF_LOOP_NORMAL);
	fingerMovie.play();
	
	cout << " tPlayer value is " << tPlayer.get() << " use count is " << tPlayer.use_count() << endl;
```
```
 	tPlayer value is 0x100f2aa88 use count is 2
```

PR #6826, opened 2022-01-10
fixed addTrailingSlash on std filesystem

2bbb:
on std::filesystem (not std::experimental::filesystem), `addTrailingSlash` is broken.
because, `std::filesystem::path("/hoge") / std::filesystem::path("/")` returns `std::filesystem::path("/")`.
so, this bug causes the wrong behavior of `ofToDataPath`.

ref: https://wandbox.org/permlink/QzINUEPwoDVOjYWr

this PR fixes it.

### DIFF
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -1641,12 +1641,17 @@ string ofFilePath::getPathForDirectory(const std::filesystem::path& path){
 	// if a trailing slash is missing from a path, this will clean it up
 	// if it's a windows-style "\" path it will add a "\"
 	// if it's a unix-style "/" path it will add a "/"
-	auto sep = std::filesystem::path("/").make_preferred();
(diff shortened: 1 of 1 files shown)


2bbb:
related?: #6585

2bbb:
related stackoverflow article:
https://stackoverflow.com/questions/48347642/differences-between-c14-stdexperimentalfilesystemv1-and-c17-stdfiles

petewerner:
Hey just want to say this is working for me on macos with c++20. From https://github.com/openframeworks/openFrameworks/issues/6585#issuecomment-990974476 i think getPathForDirectory needs updating too https://github.com/openframeworks/openFrameworks/blob/e71175c606c5b32269687f830a13a596abbb6b97/libs/openFrameworks/utils/ofFileUtils.cpp#L1641

2bbb:
@petewerner 
That's right!!
I will fix and add it.

thanks!

Issue #6793, opened 2021-12-03
ofVideoPlayer simplification, room for improvement


dimitre:
I'm spending some time reading ofVideoPlayer code to understand better how it is structured.
So I'm opening here to ask opinions and share thoughts. 

Some things caught my attention in the process, the first of them is this function
```c++
void ofVideoPlayer::setPlayer(shared_ptr<ofBaseVideoPlayer> newPlayer)
```
which has a shared_ptr as a parameter. and assigns a shared_ptr to another shared_ptr here
```c++
player = newPlayer;
```
I can see they are using two different addresses if I print them
```c++
cout << "setPlayer " << &player << " " << newPlayer << endl;
```

I've noticed there are some checks in different parts of the code which call this function, some call setPixelFormat, and some don't.
```c++
	if( !player ){
		setPlayer(std::make_shared<OF_VID_PLAYER_TYPE>());
		player->setPixelFormat(internalPixelFormat);
	}
```

We can substitute this checks with a function which creates the shared pointer without passing as a function parameter, something like this
```c++

void ofVideoPlayer::checkPlayer(){
	if( !player ){
		player = std::make_shared<OF_VID_PLAYER_TYPE>();
		player->setPixelFormat(internalPixelFormat);
		setPixelFormat(internalPixelFormat);	//this means that it will try to set the pixel format you
	}
}

```

ofTheo:
I would do this check: 

`cout << "setPlayer " << player.get() << " new player is " << newPlayer.get() << endl;`

For me I get the same:
setPlayer 0x10162a778 new player is 0x10162a778

dimitre:
Yes I can understand that. there are more haha.
I don't think 
		player->setPixelFormat(internalPixelFormat);
is needed after
		setPlayer(std::make_shared<OF_VID_PLAYER_TYPE>());

because internally setPlayer calls setPixelFormat which by itself calls player->setPixelFormat

dimitre:
and this line
```
class ofVideoPlayer : public ofBaseVideoPlayer,public ofBaseVideoDraws{
```
can be changed to 
```
class ofVideoPlayer : public ofBaseVideoDraws{
```
as it is, because of double inheritance.

ofTheo:
Ahh, yeah, I think the extra setPixelFormat calls could be removed wherever setPlayer is called. 

I would be more hesitant of messing with the base class stuff though. It is currently working and who knows what changing it might break. :)

dimitre:
[...]

PR #6829, opened 2022-01-10
ofVideoPlayer unneeded inheritance

dimitre:
It was only needed if ofVideoPlayer was the object of playing video itself, 
but I understand now ofVideoPlayer will always have a player object inside with playing capabilities (player)
Player itself is called for ofBaseVideoPlayer properties, so it makes no difference.

Related https://github.com/openframeworks/openFrameworks/issues/6793#issuecomment-985820882


dimitre:
Let's merge this one @ofTheo

ofTheo:
hmm - I am still nervous about this one. 

I could see someone having something like: 

```
//someones custom class 
void setPlayer(ofBaseVideoPlayer * player){
    playerPtr = player; 
    //some custom code here 
};
```

That is designed to allow any class with a ofBaseVideoPlayer. 

eg: 

```
ofVideoPlayer player; 
ofFFMpegPlayer ffmpegPlayer; 

myClass.setPlayer( &player ); //use default 

myClass.setPlayer( &ffmpegPlayer); //use ffmpeg 
```

I feel like we have done this in the past for a project. 
And we also have this approach here too: 
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/video/ofVideoGrabber.h#L49 

So I think it is there to allow maximum flexibility.

dimitre:
ofVideoPlayer object is never a player (maybe only in remote past) it is only a container.
it always hold the player object inside itself that can always be set as in your examples
you can see this here
```c++
void ofVideoPlayer::setPlayer(shared_ptr<ofBaseVideoPlayer> newPlayer){
	player = std::move(newPlayer);
	setPixelFormat(internalPixelFormat);	//this means that it will try to set the pixel format you have been using before. 
											//if the format is not supported ofVideoPlayer's internalPixelFormat will be updated to that of the player's
}
```
and the player object
```
		std::shared_ptr<ofBaseVideoPlayer>		player;
```

ofTheo:
@dimitre oh yeah totally, I understand that. 

It is more that the way it currently is you can store a pointer to ofBaseVideoPlayer and it can refer to both ofVideoPlayer and ofDSVideoPlayer - which can be useful  in some situations ( as you can use implementation players without ofVideoPlayer via the base functions ). 

obviously you wouldn't want to pass in an ofVideoPlayer to setPlayer ( which can happen now ). 

I do think you are correct in your thinking, I am just worried about some people storing the pointers but yeah, it is probably fine to merge.

dimitre:
I can be responsible for reverting this if any issue arises in the near future.

I've tried to check out ofDSVideoPlayer to see if it has other base class in common, like ofBaseVideoDraws but didn't find in any place. is that a public addon?

Other info: binary code size without inheritance is 86% of the object size, so less compile time and memory
```
-rw-r--r--  1 z  staff  38368 Sep 29 19:56 ofVideoPlayer1.o
-rw-r--r--  1 z  staff  44752 Sep 29 19:57 ofVideoPlayer2.o
```

PR #6837, opened 2022-01-19
shared_ptr as const in function parameter to avoid temporary copies

dimitre:
It does't change how this functions are used, only avoid one more internal copy.


ofTheo:
hey @dimitre - maybe avoid doing code formatting with code-change PRs as it makes it harder to differentiate?
Thanks!!

dimitre:
> hey @dimitre - maybe avoid doing code formatting with code-change PRs as it makes it harder to differentiate?
> Thanks!!

sorry it was the intention, I'll close this and review better before submitting

PR #6839, opened 2022-01-19
ofMainLoop::run shared_ptr as const in function parameter to avoid temporary copies

dimitre:


roymacdonald:
should go with https://github.com/openframeworks/openFrameworks/pull/6867

PR #6841, opened 2022-01-20
Suggestion: shallow clone

dimitre:
pro: much quicker install
con: have to unshallow to access other branches if needed.


danoli3:
Yeah this is good advice till we do remove all binaries from the commit tree

PR #6843, opened 2022-01-21
ofxOsc - Excluding a include folder for macOs

dimitre:
so the compiler doesn't search here
```
-I/Volumes/tool/ofw/addons/ofxOsc/libs/oscpack/src/ip/win32
```

### DIFF
--- a/addons/ofxOsc/addon_config.mk
+++ b/addons/ofxOsc/addon_config.mk
@@ -78,7 +78,5 @@ vs:
 	# a specific platform
 	ADDON_SOURCES_EXCLUDE = libs/oscpack/src/ip/posix/%
 	ADDON_DEFINES = OSC_HOST_LITTLE_ENDIAN
-	
-osx:
-	ADDON_INCLUDES_EXCLUDE = libs/oscpack/src/ip/win32/%
+
 	
--- a/addons/ofxOsc/addon_config.mk
+++ b/addons/ofxOsc/addon_config.mk
@@ -63,7 +63,7 @@ common:
 	# when parsing the file system looking for sources exclude this for all or
 	# a specific platform
 	ADDON_SOURCES_EXCLUDE = libs/oscpack/src/ip/win32/%
-
+ 	ADDON_INCLUDES_EXCLUDE = libs/oscpack/src/ip/win32/%
 	# when parsing the file system looking for include paths exclude this for all or
 	# a specific platform
 	# ADDON_INCLUDES_EXCLUDE =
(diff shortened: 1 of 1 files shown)


addons/ofxOsc/addon_config.mk
```diff
@@ -78,3 +78,7 @@ vs:
 	# a specific platform
 	ADDON_SOURCES_EXCLUDE = libs/oscpack/src/ip/posix/%
 	ADDON_DEFINES = OSC_HOST_LITTLE_ENDIAN
+	
+osx:
+	ADDON_INCLUDES_EXCLUDE = libs/oscpack/src/ip/win32/%
+
```
ofTheo, line 84: @dimitre - I think we would need this for all other platforms too.

dimitre: Great, I'll change the PR to a common block

dimitre: done

ofTheo:
@dimitre I am still seeing the path being included on macOS in the build log: `I/Users/runner/work/openFrameworks/openFrameworks/addons/ofxOsc/libs/oscpack/src/ip/win32` 

I am guessing once the exclude is working correctly we will need to add a specific include for msys2 and VS.

dimitre:
@ofTheo are you seeing this using xcode or make? 
I've made two commits after the first because at first I forgot to include the correct line.

dimitre:
@ofTheo, One question: if we include again in specific blocks of msys2 and VS it overrides the ADDON_INCLUDES_EXCLUDE right? thanks

ofTheo:
@dimitre seeing this in the Github Actions CI / Checks for Xcode and make. 
I think its getting included on all platforms. eg: this is linux 
```-I/home/runner/work/openFrameworks/openFrameworks/addons/ofxOsc/libs/oscpack/src/ip/win32```

It might be that the projectGenerator is somehow allowing this through or it is not using that variable at all? 

should be pretty easy to test locally though.

**Comment by dimitre** [comment]
I think I know already maybe the line should not have the %, so it considers the folder itself, like this
```
	ADDON_SOURCES_EXCLUDE = libs/oscpack/src/ip/win32/%
  	ADDON_INCLUDES_EXCLUDE = libs/oscpack/src/ip/win32
```

dimitre:
@ofTheo let me know if this did the trick on latest build.
About the excluisions / inclusions in VS and WIN, how about just keeping the ADDON_INCLUDES_EXCLUDE in platforms we choose, like linux, ios, osx ?

PR #6845, opened 2022-01-23
Fix bugs on tvos

2bbb:
this PR fixes wrong include guard (maybe come from copy/paste?) and about unknown property `glViewController`.


danoli3:
Thanks @2bbb ! Looks good, my mistake :)

PR #6862, opened 2022-01-30
add sanity workaround for xlib macro and openmaxil compile error on raspberrypi bullseye               

yoosamui:
**reference issue:  ofxCv macro error on X11**

raspberry pi on 11 (bullseye) sanity workaround. 
I dint't test on other OS.  However the script part can be extended to check other OS/distros easily.
The script part reads the OpenCV version but not used it except to display the version number. 
In case that an OpenCV version number is relevant can also added easily.

Please review

regards
Juan

### DIFF
--- a/scripts/linux/debian/install_dependencies.sh
+++ b/scripts/linux/debian/install_dependencies.sh
@@ -70,6 +70,28 @@ if [ "$OS_CODENAME" = "7 (wheezy)" ]; then
     sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 1 --force
 fi
 
+if [[ $(uname -a) == *"raspberrypi"* ]] && [[ "$OS_CODENAME" = "11 (bullseye)" ]]; then
+    # sanity for stitching
+    version=/usr/bin/opencv_version
+    if [ -f $version ]; then
+        version=$($version)
+        # check version if necessary.
+        if [ ! -z $version ]; then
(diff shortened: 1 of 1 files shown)


yoosamui:
my guess:
sed don't have write access.

Please cancel this pull request. I would like to create a new one to find out what the real problem is.

regards
juan

ofTheo:
Another solution that could work: 

Either for X11 or for ofxOpenCv do something like this:

In ofxCvConstants.h: 
https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxOpenCv/src/ofxCvConstants.h#L10-L17 

```
#define Status _PrivateCvStatus

#include "opencv2/core/version.hpp"
#if CV_MAJOR_VERSION < 4
    #include "cv.h"
    #define USE_OLD_CV
#else
    #include "opencv2/opencv.hpp"
    #include "opencv2/imgproc/imgproc_c.h"
#endif

#undef Status
```

It is not exactly pretty but neither is X11 typedefing Status in its header file 🙂

yoosamui:
yes, that was my first attempt, i just forgot: #define Status before #undef
definitely your proposal is the better workaround  for the ugly xlib macro.
I tested on my raspberry and it working fine.  

also i got a lot of trouble with the build-server and the sanity workaround i wrote. 
Don't  know exactly way...
I already close PR #6862. 

what do you suggest, will you take care of this, or should i do a new PR with your proposal? 

regards
juan

PR #6867, opened 2022-02-01
ofMainLoop::addWindow, avoid temporary copy

dimitre:


roymacdonald:
> "const" reference (isn't mutable reference) argument can receive temporary(r-value) vars.
> but if we need to make more strict, then we should overload with const Type & and Type &&arg.
This makes sense and it would be a good thing to make this change in the most strict way at once.

2bbb:
I would like to add:

https://codexpert.ro/blog/2014/10/17/c-gems-ref-qualifiers/

```cpp
class X {
    std::vector<int> xs;
public:
    std::vector<int> &get() & { 
        return xs;
    }
    const std::vector<int> &get() const & { 
        return xs;
    }
    std::vector<int> get() && { 
        return std::move(xs);
    }
};
```

This qualitifier and using explicitly move can make more strict memory management.
But, I don't know which part in oF is suitable for this semantics... ☺️

roymacdonald:
@2bbb There are lots of parts where this can be implemented. Mostly add the `get() && ` function. For instance, just from the top of my mind, ofMesh has several getters that could add this. Yet, what does it imply to move a vector? will the data in the vector in the class instance become invalid ?

2bbb:
@roymacdonald 
Yes, instance become invalid when vector are moved.

https://wandbox.org/permlink/7CX8t9bjztLZzE31

in this code, `auto v = std::move(x).get();` is similar to `auto v = std::move(x.vs);`.
`auto ws = std::move(vs);` is similar to `decltype(vs) ws; std::swap(vs, ws);`

so, we need to implement those carefully. because some situation like below code causes bug.

```cpp
void f(X &&v) {
#ifdef __DEBUG__
  std::cout << v.get() << std::endl;
#endif
  auto value = v.get(); // <-- if __DEBUG__ is defined, this value is already invalidated
  ...
}
```

roymacdonald:
> so, we need to implement those carefully. because some situation like below code causes bug.
right. 

So under which circumstances does the  `get() && ` override gets called instead of the other ones?
Where will this be benefical?

ofTheo:
I dug in a bit to the link provided and can't see a huge reason to add this sort of complexity.  ( not the PR, but the ref qualifiers stuff now being discussed ).  

If there is some examples of it closing open issues/bugs or increasing performance 2x 3x etc - would be good to then think about, but we currently have so many open issues might be good to focus on those? 🙂

PR #6868, opened 2022-02-01
Workaround for xlib macro issue - create a clean commit. only lines affected

yoosamui:
Please review

Regards
Juan


ofTheo:
Thanks @yoosamui !

ofTheo:
Hey @yoosamui 
I think this is causing other issues for other platforms. I am going to see if I can find a solution that maybe just fixes it for X11. 

Going to close, but will  make sure the fix is in the next release. Thanks for finding this issue!

PR #6878, opened 2022-02-20
supress warning unused needsResizeCheck - RPI only

dimitre:
RPI only variable, so it goes inside #ifdef macro
```
/Volumes/tool/ofw/libs/openFrameworks/app/ofAppGLFWWindow.h:222:9: warning: private field 'needsResizeCheck' is not used [-Wunused-private-field]
```


dimitre:
one easy here @ofTheo

PR #6880, opened 2022-02-20
ofArduino.h remove unused numEncoders

dimitre:
supress this warning
```
/Volumes/tool/ofw/libs/openFrameworks/communication/ofArduino.h:792:6: warning: private field '_numEncoders' is not used [-Wunused-private-field]

```
I've just commented it out, in the case somebody wants to actually implement


dimitre:
Maybe it is best to remove completely? what do you think @ofTheo  ?

PR #6882, opened 2022-02-21
ofxOpenCV - ofPoint to glm::vec2

dimitre:
This PR supress some of the warnings about the differences of the base class and in ofGraphicsBaseTypes.h and overloaded classes here

```
/Volumes/tool/ofw/addons/ofxOpenCv/src/ofxCvImage.h:97:15: warning: 'ofxCvImage::draw' hides overloaded virtual functions [-Woverloaded-virtual]
        virtual void draw(const ofPoint & point) const;
                     ^
```


dimitre:
This PR supress some of the warnings about the differences of the base class and in ofGraphicsBaseTypes.h and overloaded classes here

ofTheo:
Before merging this can you check that you can still use an ofPoint or an ofVec2f as an argument with this change? 
Thanks!

ofTheo:
Be great if people could add any open issues not listed above to this - either in the comments or by editing this issue.

Issue #6848, opened 2022-01-24
[Proposal] About supporting obj-c ARC default.


2bbb:
Currently, many Objective-C codes are assuming ARC.
and, codes will be made strict under ARC.
(for example, #6845 was found when I trying to make obj-c codes to ARC.)

the problem is compatibility about addons use obj-c.
but, codes with no-ARC (i.e. use retain, release, autorelease, etc... ) will get build errors explicitly on ARC envs.
and those can be fixed with add build flag `-fno-objc-arc` to implementation codes.

this branch ( https://github.com/2bbb/openFrameworks/tree/obj-c-arc ) is what I try to support ARC.
(now updated core codes only. next targets are core addons.)
if this proposal gets approval, I will post this as PR.

I think, now is good timing to support ARC env on updating to v0.12.
What do you think, everyone?

2bbb:
Oh, great. 
maybe no core addons codes need obj-c...?

dimitre:
Do you think it breaks compatibility with other addons like ofxSyphon, ofxNdi, ofxAudioUnit, ofxMidi for ios?

2bbb:
Sadly, yes. maybe.
because these codes use non-ARC methods like `retain`, `release`, `autorelease` etc.

Fortunately(?), it raises an error on ARC env, will not be the cause of memory leaks.
and, users need to add flag `-fno-objc-arc` to implementation files on Xcode, targets setting > build phase > compile source.

roymacdonald:
@2bbb How do you see that moving to ARC will benefit what we currently have?
It is ceirtainly nicer not having to manage memory but if it already works, why fix it?
Besides, I think that the fact that it will be problematic for the addons which use Obj-C can lead to broken addons, etc. 
If using `-fno-objc-arc` would allow using non-ARC code, is it a problem to have it by default in the xcode project?

2bbb:
@roymacdonald 
thanks for comments.
I understand what you say.

First of all, I propose to supporting ARC for maintenancibility.
no ARC is so legacy like using C++03. So many documents are wrote by Swift and Obj-C with ARC.

Problem is what codes written by ARC work on no ARC env.
e.g. someone fixed bug with ARC codes, then it will be compiled and works, but memory will leaks...
e.g. someone added new iOS feature what is documented by ARC, then it works, but memory will leaks...

Second, ARC makes type safe than no ARC envs. (again, I found bugs #6845 when developing ARC support)
[...]

PR #6889, opened 2022-03-11
[W.I.P.] Supporting ARC

2bbb:
this PR makes goal to fix to support all not ARC codes as ARC codes.
please see discussion on: #6848


danoli3 (commented):
Looking great @2bbb ! Great work

Maybe just add this to the prime delegate of each platform 
```
#if !__has_feature(objc_arc)
 #   error need ARC
 #endif
 ```
 
 I'll check it out tomorrow and test vs some projects 
 
 So far looking good, reviewed changes

**Comment by 2bbb** [comment]
@danoli3 

thanks for the comments.

> Maybe just add this to the prime delegate of each platform and not in each obj-c converted but no issues as it's just safeguard

OK.
I will remove from .m/.mm and add ARC check macro in ofConstants.h like:

```
#ifdef TARGET_OSX
	#ifndef __MACOSX_CORE__
		#define __MACOSX_CORE__
	#endif
	#include <unistd.h>
	#include "GL/glew.h"
	#include <ApplicationServices/ApplicationServices.h>

	#if defined(__LITTLE_ENDIAN__)
		#define TARGET_LITTLE_ENDIAN		// intel cpu
	#endif
    
    #ifdef __OBJC__ // added
        #if !__has_feature(objc_arc)
            #error "need ARC"
        #endif
    #endif
#endif

...

#ifdef TARGET_OF_IOS
	#import <OpenGLES/ES1/gl.h>
	#import <OpenGLES/ES1/glext.h>

	#import <OpenGLES/ES2/gl.h>
	#import <OpenGLES/ES2/glext.h>


	#define TARGET_LITTLE_ENDIAN		// arm cpu

    #ifdef __OBJC__ // added
        #if !__has_feature(objc_arc)
            #error "need ARC"
        #endif
    #endif
#endif
```

I will test too, but do you think is this fine?

ofTheo (changes requested):
Could we change: 
`error need ARC` 
to something a bit more descriptive? 
`error Please enable ARC (Automatic Reference Counting) at the project level`

2bbb:
@ofTheo 
I will change this message when same as solving @danoli3 point out.

danoli3:
Looking good, did initial tests. will continue before fully approving

We are now locked to iOS 13 + with inclusion of C++17 filepath just a FYI! No issue with this though!

danoli3:
Might be a good idea to generate a guide to upgrade a project that is not ARC into ARC oF future though, I had some custom AppDelegates and views that forced into the new ARC realms. Will help devs

2bbb:
I updated about @danoli3 and @ofTheo point out about ARC feature checking.

2bbb:
I think what replacing many pairs of an explicitly defined instance variable, `@property` and `@synthesize` to `@property` including implicitly defined instance variable and removing `@synthesize` is better to Maintainance.
because currently, if we need to change type of variable then we need to change two points, but if there is only one `@property` then we need to change once.

I will do this with other PR after this PR is finished.

ofTheo:
Going to leave @danoli3 to okay the final merge of this! 
@danoli3 just make sure to do **Squash and Merge** above with `#changelog #osx #ios #tvos` ( this helps with generating the changelog for releases )

danoli3:
Screen Shot 2022-05-06 at 3 15 22 pm


Looks good! For anyone with issues with old non-arc code that cannot be converted: You can add this to compile flag per src file 

```-fno-objc-arc```

danoli3:
Been testing looks good

danoli3:
Good to merge @ofTheo

ofTheo (approved):

dimitre:
Hello @2bbb I've tried to port ofxSyphon to ARC with no success. 
As I need it for several projects I should change XCode project to disable ARC and comment out a line in ofConstants.h
My suggestion is we can disable ARC enforcing now in ofConstants.h as everything else works great anyway. 
cc: @ofTheo

2bbb:
Hi, @dimitre 

I changed about those on https://github.com/openframeworks/openFrameworks/pull/6889/commits/96f11b878c5eb6dac137a4454f2b2a0f73632feb
but this was wrong.

maybe correct way is what condition macro is wrote in all .m / .mm file, or make one header and include from those files.
at least it was a completely mistake to put them together in `ofConstants.h`...

PR #6891, opened 2022-03-11
ofUtils.cpp warning  [-Wmisleading-indentation]

dimitre:
this PR fixes this warning and note 
```c++
/home/pi/ofw2/libs/openFrameworks/utils/ofUtils.cpp:523:5: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
     if (makeAbsolute && path.is_absolute())
     ^~
/home/pi/ofw2/libs/openFrameworks/utils/ofUtils.cpp:526:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  if (!enableDataPath)
  ^~
```

### DIFF
--- a/libs/openFrameworks/utils/ofUtils.cpp
+++ b/libs/openFrameworks/utils/ofUtils.cpp
@@ -520,11 +520,13 @@ void ofSetDataPathRoot(const std::filesystem::path& newRoot){
 
 //--------------------------------------------------
 string ofToDataPath(const std::filesystem::path & path, bool makeAbsolute){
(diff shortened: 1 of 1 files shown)


2bbb:
as a side note , `ofUtils` and `ofFileUtils` reference each other about `ofToDataPath`.
I want to solve this complexitiy... but I don't know what is better solution...

2bbb:
ah, maybe, moving ofToDataPath from ofUtils to ofFileUtils is solve this complexity?

roymacdonald:
@ofTheo this looks good to merge

PR #6893, opened 2022-03-14
Move functions about data path from ofUtils to ofFileUtils.

2bbb:
This PR make solve little complexity about ofUtils and ofFileUtils references each other.

those functions and private variable are moved from ofUtils to ofFileUtils.
(and remove declaration of `setWorkingDirectoryToDefault` has no implementation...)

```cpp
namespace {
    bool enableDataPath = true;
    string defaultDataPath();
    std::filesystem::path & defaultWorkingDirectory();
    std::filesystem::path & dataPathRoot();
}
void ofEnableDataPath();
void ofDisableDataPath();
std::string ofToDataPath(const std::filesystem::path & path, bool absolute=false);
bool ofRestoreWorkingDirectoryToDefault();
void ofSetDataPathRoot(const std::filesystem::path& root);
```

and those line is moved from `of::priv::initutils()` to new private function `of::priv::initfileutils()`.

```cpp
        defaultWorkingDirectory() = std::filesystem::absolute(std::filesystem::current_path());
```

and add line to call this function after `of::priv::initutils() called` in ofAppRunner.

This is not needed absolutely. but I think this update makes a sense. because, ofToDataPath and other functions are about Files, and these are in ofFileUtils is better.

In some situations, maybe this update will break backwards compatibility when using `ofToDataPath` without including ofFileUtils only including ofUtils and other headers too.
If needed, I will add `#include ofFileUtils` in ofUtils.h for backwards compatibility.

### DIFF
--- a/libs/openFrameworks/app/ofAppRunner.cpp
+++ b/libs/openFrameworks/app/ofAppRunner.cpp
@@ -127,18 +127,19 @@ void ofInit(){
 	signal(SIGABRT, &ofSignalHandler);  // abort signal
 #endif
 
-        of::priv::initutils();
+    of::priv::initutils();
+    of::priv::initfileutils();
 
-	#ifdef WIN32_HIGH_RES_TIMING
-		timeBeginPeriod(1);		// ! experimental, sets high res time
-								// you need to call timeEndPeriod.
-								// if you quit the app other than "esc"
-								// (ie, close the console, kill the process, etc)
-								// at exit wont get called, and the time will
-								// remain high res, that could mess things
-								// up on your system.
-								// info here:http://www.geisswerks.com/ryan/FAQS/timing.html
-	#endif
+#ifdef WIN32_HIGH_RES_TIMING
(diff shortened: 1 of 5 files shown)


roymacdonald:
looks like a reasonable thing and makes sense. I dont see much problem on merging it so far it is just moving code from one file to another, rather than adding/deleting code

ofTheo:
@2bbb just to check, the functionality hasn't changed, just where the implementation is defined?

2bbb:
@ofTheo 

basically, I used Cmd+x and Cmd+v.

but, some points are not only changing location.

* delete declaration `void setWorkingDirectoryToDefault()` from `ofUtils.h`, and this doesn't exist on oF codes now.
* add new private function `of::priv::initfileutils()` and some initialization code move into this from `of::priv::initutils()`.
* add to call `of::priv::initfileutils()` on `ofInit()`

ofTheo:
Thanks @2bbb for the explanation. 
This all looks good from my perspective.! 

Unless there is anything else you need to do, I can merge this.

PR #6900, opened 2022-03-17
change Xcode settings about indentation from space to tab

2bbb:
So, I didn't know oF code style, and it says use hard tab (4).
https://github.com/openframeworks/openFrameworks/wiki/oF-code-style

Because Xcode settings about indentation is space (4).

This PR change it (only oF library project, not include pg-template) to hard tab (4).


danoli3:
Ahhh yes should merge this.

Issue #6892, opened 2022-03-12
[Proposal] change default value of wrapInBundle flag on ofxOscSender::sendMessage to `false`


2bbb:
This proposal is because some of my friends got trouble.

Some osc libraries don't support bundle, and sometimes I am consulted about "can't communicate by osc".

I think what `ofxOscSender::sendMessage` doesn't use bundle default is better.

how do you think about this?

roymacdonald:
To me it sounds like sending in a bundle by default is just adding unnecesary overhead. not sure though.
Do you see any potential issue about backwards compatibility by not sending by bundle by default?

2bbb:
@roymacdonald 
Maybe there is no problem with backwards compatibility. there are libraries are incompatible to Bundle, but I haven't to know the library what receive only bundle.
But, If there is case like the app sending so many messages and if sending with bundle wrapped is effective, this change will make those app slow maybe.

2bbb:
oh, I was misunderstanding.
`sendMessage` with `bWrapBundle = true` is only wrap a single message. I had thought to this flag bundles messages smart.

I become to don't know why this is default behavior...

dimitre:
I didn't know it was using bundle by default. Are you sure? 
If it is I vote to change to single message unless needed.
I've always used different types for the regular single message and bundle, so If I'm sending multiple messages in the same "frame" like loading a preset I can wrap in a bundle, and for syncronize regular data just plain message.
```c++
ofxOscBundle bundle;
ofxOscMessage message;
```

2bbb:
@ofTheo
thanks for the information.
issue #1804 notes a historical reason.

so, the reason for default value of `wrapInBundle = true` is for backward compatibility to the era of default bundles, no flags.
is it correct?

if then, I think now is time to change to `false` default. (I can't ensure it will keep backward compatibility completely, but maybe it is a rare case, I think.)

ofTheo:
@2bbb I agree - it was a weird default to have and in most cases ( I believe ) bundle is less supported than raw messages. So I think it should be safe to switch to false.

2bbb:
🎉

PR #6901, opened 2022-03-17
Switch default value of wrapInBundle from true to false.

2bbb:
related to: #6892


dimitre:
+1 for this PR merge.

PR #6904, opened 2022-03-26
ofxUDPManager : unused variable bytesleft removal

dimitre:
variable is unused, I can see "int total" variable replaced this logic, this cleanup removes the following warning
```
ofw/addons/ofxNetwork/src/ofxUDPManager.cpp:312:6: warning: variable 'bytesleft' set but not used [-Wunused-but-set-variable]
```


roymacdonald:
looks fine to merge @ofTheo

PR #6909, opened 2022-04-04
This ofBaseApp instance is never used

dimitre:
this PR serves to increase readability in the fact the ofBaseApp here is not used in any way. 
it is only used in the original void ofMainLoop::run
```c++
	windowsApps[window] = app;
```
this PR in fact doesn't do anything other than increase readability.
And I think this map can be changed in the future with something else, like a struct.


ofTheo:
think this is not needed. 
the old code is the same - it is not creating a object just setting an unallocated shared_ptr.

dimitre:
Allright thanks!

PR #6916, opened 2022-04-04
ofArduino Cleanup unused code - remove warning

dimitre:
ofArduino.cpp:990:7: warning: variable 'shift' set but not used [-Wunused-but-set-variable]


ofTheo:
Hmm - this one seems to be like it was there for a reason - and might be more of something that needs a fix as opposed to just removing?

dimitre:
Reading the code
it seems to me it was a bit shift operation used before in every operation
when the value was stored here
int pinVal[2] = { pin, val };

and now it is being used in previous code, maybe some previous case didn't use bit shift at all so they decide to manipulate case by case

DomAmato:
Hmm been awhile since I touched this code base but it does look like its just clearing anything in the buffer and those variables were likely just used for debugging and can be removed.

PR #6920, opened 2022-04-05
Removing QTKit

dimitre:
makes sense?
I think it was discontinued in 10.7 in 2010


ofTheo:
Thanks!!
Want to add a few more things to this PR? 

Removing these lines:
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/video/ofVideoGrabber.h#L19-L22 

And: 
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/video/ofVideoPlayer.h#L17-L20

Maybe remove everything in the TARGET_OSX section here except this line:
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofConstants.h#L307 

Same - keep just this line:
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/utils/ofConstants.h#L356

dimitre:
Hey sorry closed automatically when I tried to rename branch.
I'll be submitting a PR removing this lines and Quicktime ones too. 

one question : do you think we remove everything QT or Quicktime from code? I don't mind.
if one wants to use abandonware can always download OF old releases

ofTheo:
oops sorry thought I had replied to this. 
I think it would be good to do QTKit and Quicktime as separate PRs just for the history and testing. 

Maybe doing the above for QTKit as one PR. 
Then Quicktime as another. 

Finally if everything seems to be working fine we can remove the actual .h and .cpp files from git as a later PR.

ofTheo:
I think maybe just QTKit for this PR? 
We could do another one for Quicktime. 

But yeah both are super legacy at this point. 🙂

PR #6930, opened 2022-04-05
Specific diagnostics supressed in some files

dimitre:
diagnostic ignored on ofMatrix4x4 ofFileUtils ofxUDPManager ofxTCPManager ofxTCPServer ofxTCPClient


ofTheo:
This one feels like it could be set at the project level no?

dimitre:
Nice, I didn't know. how that can be made @ofTheo ?
The idea behind this PR was only use this pragma temporarly until we decide how to cast types case by case.

Issue #6929, opened 2022-04-05
Warnings general


dimitre:
I've spent some time to fix warnings, I think they get in the way of you reading important things on screen.
So I've just finished a lot of changes that can help clearing space.
## XCode
No warnings anymore after the following PRs are merged. I think they can be reviewed and merged as one.
cc: @ofTheo @2bbb @danoli3 
https://github.com/openframeworks/openFrameworks/pull/6927
https://github.com/openframeworks/openFrameworks/pull/6924
https://github.com/openframeworks/openFrameworks/pull/6923
https://github.com/openframeworks/openFrameworks/pull/6922
https://github.com/openframeworks/openFrameworks/pull/6921
https://github.com/openframeworks/openFrameworks/pull/6920

## General code
https://github.com/openframeworks/openFrameworks/pull/6919
https://github.com/openframeworks/openFrameworks/pull/6918
https://github.com/openframeworks/openFrameworks/pull/6917
https://github.com/openframeworks/openFrameworks/pull/6916

## Related issues (future)
https://github.com/openframeworks/openFrameworks/issues/6926
https://github.com/openframeworks/openFrameworks/issues/6925

And other thing that can be fixed soon is this (int to size_t convesion):
https://github.com/openframeworks/openFrameworks/issues/6866#issuecomment-1061906274

dimitre:
Implicit conversion loses integer precision: 'std::vector<ofxOscMessage>::size_type' (aka 'unsigned long') to 'int'
https://github.com/openframeworks/openFrameworks/blob/290dbfc23930425024c9456b5c5519611fb1990d/addons/ofxOsc/src/ofxOscBundle.cpp#L46-L54

dimitre:
Implicit conversion loses integer precision: 'unsigned long' to 'osc::osc_bundle_element_size_t' (aka 'int')
https://github.com/openframeworks/openFrameworks/blob/290dbfc23930425024c9456b5c5519611fb1990d/addons/ofxOsc/src/ofxOscSender.cpp#L238

oxillo:
in MSYS2 : warning :  ignoring attributes on template argument

```
E:/openframeworks/libs/openFrameworks/utils/ofUtils.h:603:60: warning: ignoring attributes on template argument 'va_list' {aka 'char*'} [-Wignored-attributes]
  603 |     -> typename std::enable_if<std::is_same<va_list, VAList>::value, std::string>::type
```

oxillo:
[...]

PR #6931, opened 2022-04-06
[W.I.P.] suppress warning about implicit type conversions

2bbb:
related: #6929

### DIFF
--- a/libs/openFrameworks/app/ofAppNoWindow.cpp
+++ b/libs/openFrameworks/app/ofAppNoWindow.cpp
@@ -52,10 +52,10 @@ int kbhit()
 
 int getch()
 {
-    int r;
+    ssize_t r;
     unsigned char c;
     if ((r = read(0, &c, sizeof(c))) < 0) {
-        return r;
+        return (int)r;
     } else {
         return c;
     }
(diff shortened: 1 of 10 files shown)


2bbb:
@dimitre

I try to supress warnings, but I can't decide what is correct way in many situations.
I think, rest of warnings are from gaps between C++ types and OpenGL (or other third-party library) types.

Maybe, we have to do force cast, but is it fundamental solution...?

for example, `ofTexture::allocate`:
* many `ofTexture::allocate` methods delegating to other `allocate` method with size_t arguments like `allocate(pix.getWidth(), pix.getHeight(), ...)`
* but width and height arguments of `ofTexture::allocate` is typed as `int`
* and allocate method that is finally called, `w` and `h` will assinged to `texData.width`, `texData.height` are typed as `float`

I can't decide what is right type to `allocate`.

what do you think...?

dimitre:
I suppose texData is float for a reason, like the tex hack to remove artifacts, 
so maybe ofTexture allocate should use float parameters? What do you think @ofTheo ?

roymacdonald:
What is the problem with implicit conversion other than the annoying warnings?

dimitre:
@roymacdonald I think the warnings are there so we can make decision about control of the types. This specific one already brings the question : should a texture be allocated with float? if yes should ofTexture receive float parameters too ?

Issue #6719, opened 2021-03-25
ofSetFullscreen - issues after entering and exit fullscreen


dimitre:
After enter fullscreen and exit, sometimes aspect ratio is changed and window will only work correctly on fullscreen.
I've noticed this on Big Sur, apple m1, not sure if it is affect other builds too.
this only happens when my main window is created with dimensions exceeding or equaling screen size.
for example in this computer if my window width is 1439 it doesn't happen, if it is 1440 it start to happen.
vertical window size seems to influence but not sure how.
mouse coordinates are still good after the issue.

```c++
void ofApp::draw(){
    ofSetColor(255,0,70);
    ofDrawCircle(300, 300, 300);
}

bool fs = false;
void ofApp::keyPressed(int key){
    if (key == 'f') {
        fs ^=1;
        ofSetFullscreen(fs);
    }
}
```
Before Fullscreen
<img width="1552" alt="Screen Shot 2021-03-25 at 12 04 48" src="https://user-images.githubusercontent.com/58289/112495085-55291800-8d62-11eb-881b-cb43f4139621.png">

After Fullscreen
<img width="1552" alt="Screen Shot 2021-03-25 at 12 04 52" src="https://user-images.githubusercontent.com/58289/112495113-5bb78f80-8d62-11eb-93d7-d9b5c0306791.png">

ofTheo:
Huh. Weird. I just tried on an intel MBP and don't see the issue. Is this with 0.11.1 / 0.11.2 ?

Is your window setup for Default for Resolution in the Display settings of System Preferences?
Or one of the other modes? 

<img width="655" alt="image" src="https://user-images.githubusercontent.com/144000/112504339-eb504480-8d48-11eb-826d-18ea06aca146.png">

dimitre:
not sure if it helps, this display gives this warning when switching fullscreen and back
```
CVCGDisplayLink::setCurrentDisplay didn't find a valid display - falling back to 60Hz
[ error ] ofAppGLFWWindow: 65544: Cocoa: Failed to find service port for display
```

ofTheo:
Hmm. I see that error coming up for other apps that use GLFW with M1 machines. 

Out of curiosity does it go away if you do this in `ofAppGLFWWindow::update()` ?

	if( targetWindowMode == OF_WINDOW){
			 int winW, winH;
			glfwGetWindowSize(windowP, &winW, &winH);
			resize_cb(windowP, currentW, currentH);
	}

[...]

PR #6934, opened 2022-04-08
Update ofAppGLFWWindow.cpp for openframeworks/openFrameworks#6719

petewerner:
Set the window properties after resizing, to fix openframeworks/openFrameworks#6719

- coming out of fullscreen, we call cocoa's setStyleMask, which triggers the glfw framebuffer_resize callback
- then we call setWindowShape()
    - since the glfw resize was already triggered, the glfw internals do not get called, do not see the glfw callbacks trigger
    - i assume because the window size set by the cocoa call equals the window size we want to set, glfw does not do the 2nd resize
    - theres some y axis offset adjustment that is maybe getting skipped 
        - https://github.com/glfw/glfw/blob/master/src/cocoa_window.m#L1038
        - haven't really looked into the guts of glfw
- workaround is to call setWindowShape() first then setStyleMask
- seems to fix the issue macos 12.3/m1 and 11.6/intel

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -930,12 +930,12 @@ void ofAppGLFWWindow::setFullscreen(bool fullscreen){
 			windowRect.height = getWindowSize().y;
 		}
 
+		setWindowShape(windowRect.width, windowRect.height);
+		setWindowTitle(settings.title);
+
 		[NSApp setPresentationOptions:NSApplicationPresentationDefault];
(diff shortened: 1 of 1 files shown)


ofTheo:
Thanks @petewerner! 
Will test on my mac here too. Looks promising!

ofTheo:
Just tested this on my Intel mac and it seems to address the issue. 
@dimitre could you try this on your setup and see if it closes #6719 ?

dimitre:
Yess this correct things and close the issue

ofTheo:
Great!!! 
Thanks @petewerner and @dimitre !!

Issue #6925, opened 2022-04-05
ofAVFoundationGrabber : 'devicesWithMediaType:' is deprecated: 


dimitre:
Just posting here if somebody more experienced in obj-c can jump in this one

https://github.com/openframeworks/openFrameworks/blob/bc294119135e62eeb07b6845a01990820c2ec0e5/libs/openFrameworks/video/ofAVFoundationGrabber.mm#L40-L41

https://github.com/openframeworks/openFrameworks/blob/bc294119135e62eeb07b6845a01990820c2ec0e5/libs/openFrameworks/video/ofAVFoundationGrabber.mm#L245-L247

'devicesWithMediaType:' is deprecated: first deprecated in macOS 10.15 - Use AVCaptureDeviceDiscoverySession instead. [-Wdeprecated-declarations]

cheers!

PR #6935, opened 2022-04-12
alternating deprecated devicesWithMediaType with AVCaptureDeviceDiscoverySession

dimitre:
This PR closes https://github.com/openframeworks/openFrameworks/issues/6925


ofTheo:
Thanks @dimitre going to ask @danoli3 and @2bbb to take a quick look at this one. 
Just want to make sure we aren't missing anything with the change.

2bbb:
`AVCaptureDeviceDiscoverySession` is from 10.15.
Will we remove target before 10.15?
isn't currently deployment target 10.9?

If we support old macos (< 10.15), I recommend to use available macro like:

```
    NSArray<AVCaptureDevice *> *devices;
    if (@available(macOS 10.15, *)) {
        AVCaptureDeviceDiscoverySession *session = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:@[
            AVCaptureDeviceTypeBuiltInWideAngleCamera,
            AVCaptureDeviceTypeExternalUnknown,
        ] mediaType:nil position:AVCaptureDevicePositionUnspecified];
        devices = [session devices];
    } else {
        devices = [AVCaptureDevice devicesWithMediaType:AVMediaTypeVideo];
    }
```

`@available` macro is from Xcode9, if we need to support more older Xcode, then we use other macro with `Availability.h` maybe...

dimitre:
thank you @2bbb great idea! we can make it to a function like getDevices or something.

ofTheo:
Thanks @2bbb 
Def don't need to support Xcode < 9  - so an Xcode 9 onwards solution would be fine! :)

dimitre:
Even with @available it still generates the warnings. It can be merged as it is or if you prefer I can investigate a bit more if we have some kind of preprocessor checks of macOs version

ofTheo:
Ahh cool!. 
Well feel free to **squash and merge** when you are ready.  

I am also seeing a bunch of weird print outs when opening the camera on macOS 12.3.1 

`CMIO_Graph_Helpers_Analytics.mm:36:sendAnalytics Missing key usesFileOutputDelegate`
`CMIO_Graph_Helpers_Analytics.mm:36:sendAnalytics Missing key usesVideoFieldMode` etc

PR #6937, opened 2022-04-13
migrating more settings from pbxproj to here

dimitre:
this is more readable format, I've only brought settings that are equal in the three builds (Debug, Release, Appstore)
once merged I'll make another PR to remove them from pbxproj


dimitre:
Hey @ofTheo what do you think of this one?
the idea behind it is to bring some configurations to this more readable format, so we can finetune them in xcconfig files soon if needed. (ex: GCC_GENERATE_DEBUGGING_SYMBOLS)
For now it doesn't change anything because these settings are already in place in pbxproj

ofTheo:
Looks good to me! 
Might be better  though to add the corresponding changes to pbxproj to this PR so its a clean swap.

dimitre:
Done! I've found two more identical configurations and moved them too.
Feel free to check it again when you have time

PR #6941, opened 2022-04-17
Xcconfig - Moving Release and Debug common to Shared

dimitre:
Comparing all configurations that are the same from Debug and Release I've moved them for shared
This way Debug and Release only have the exclusive configurations for each one.

I'm not sure if this configurations are still used or they are leftovers from the time OF had a separate project than the project. Anyway I think it makes more sense this way


ofTheo:
Thanks!! @dimitre

Issue #6926, opened 2022-04-05
ofFileUtils.h - std::iterator deprecated


dimitre:
This file uses std::iterator which is deprecated on c++17
utf-8 library updated std::iterator to templates, so maybe it is a good reference 
https://github.com/openframeworks/openFrameworks/issues/6905

https://github.com/openframeworks/openFrameworks/blob/bc294119135e62eeb07b6845a01990820c2ec0e5/libs/openFrameworks/utils/ofFileUtils.h#L172-L186

dimitre:
Cheers! Thanks @ofTheo @2bbb !

PR #6945, opened 2022-04-21
Refactoring ofBuffer::Lines/RLines

2bbb:
* remove `std::iterator` with writing type alias directly
* implement `Lines::rbegin` and `Lines::rend` are not implemented (!!)
* add some `const` qualitifiers (and `cbegin`, `cend`, etc...)
* fix logic on RLine:
  * first line (i.e., last line in buffer) was wrong.
* fix edge case on Line/RLine: 
  * empty line was ignored when line break appears at last/first of text (e.g. `"\na\nb\n"`)
    * is this not correct?

related: #6926

### DIFF
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -14,6 +14,8 @@
 	#include <limits.h>        /* PATH_MAX */
 #endif
 
+#include <iterator>
+
 using namespace std;
 
 namespace{
@@ -266,7 +268,7 @@ vector<char>::const_reverse_iterator ofBuffer::rend() const{
 }
 
 //--------------------------------------------------
-ofBuffer::Line::Line(vector<char>::iterator _begin, vector<char>::iterator _end)
+ofBuffer::Line::Line(vector<char>::const_iterator _begin, vector<char>::const_iterator _end)
 	:_current(_begin)
 	,_begin(_begin)
 	,_end(_end){
@@ -332,7 +334,7 @@ bool ofBuffer::Line::empty() const{
 
 
 //--------------------------------------------------
-ofBuffer::RLine::RLine(vector<char>::reverse_iterator _rbegin, vector<char>::reverse_iterator _rend)
+ofBuffer::RLine::RLine(vector<char>::const_reverse_iterator _rbegin, vector<char>::const_reverse_iterator _rend)
 	:_current(_rbegin)
 	,_rbegin(_rbegin)
 	,_rend(_rend){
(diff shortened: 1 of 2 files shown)


2bbb:
@arturoc

I have a question.

The specification of iteration `Lines` is what the last line break will be ignored?
i.e. "A\nB\n" will be `"A", "B"`, not `"A", "B", ""`

Test case is so. Is this what you intended?
https://github.com/openframeworks/openFrameworks/blob/073f17d1bc8cf77407234cabf58fa37ad0b22b53/tests/utils/buffer/src/main.cpp#L165

dimitre:
@2bbb thanks for the PR. I've just checked out and tested. everything working here!

dimitre:
Great one! I've noticed the builds using make are generating an error
```c++
/Users/z/Desktop/src/ofw/libs/openFrameworks/utils/ofFileUtils.cpp:439:20: error: no member named 'make_reverse_iterator' in namespace 'std'
        return RLine(std::make_reverse_iterator(_begin), std::make_reverse_iterator(_begin));
```


I think it can be solved by a template here or bumping gcc version
https://stackoverflow.com/questions/50161210/error-make-reverse-iterator-is-not-a-member-of-std#50162394

ofTheo:
Hmm, I think there are more issues as the tests are failing too. 
Going to revert this @2bbb until the tests / checks pass the CI ( sorry I missed that ) 

```
[ notice ] lines iterator
[ notice ] all lines are correct passed
[  error ] lines iterator correct numLines failed 
[  error ] test_eq(numLines, lines.size())
[  error ] value1: numLines is 18
[  error ] value2: lines.size() is 17
[  error ] /Users/runner/work/openFrameworks/openFrameworks/tests/utils/buffer/src/main.cpp: 176
[ notice ] -------------------
[ notice ] lines reverse iterator
[  error ] all lines are correct failed 
[  error ] /Users/runner/work/openFrameworks/openFrameworks/tests/utils/buffer/src/main.cpp: 212
[  error ] lines iterator correct numLines failed 
[  error ] test_eq(numLines, lines.size())
[  error ] value1: numLines is 18
[  error ] value2: lines.size() is 17
[  error ] /Users/runner/work/openFrameworks/openFrameworks/tests/utils/buffer/src/main.cpp: 213
[  error ] 3/31 tests failed
[ notice ] took 2ms
```

dimitre:
Hey @ofTheo this test failing is related to the second @2bbb post where he asks about appending "\n" so it returns one more line. If the \n is unneeded we can update the test itself too

2bbb:
@dimitre @ofTheo 
Hi. sorry for the late response.

at first, `make_reverse_iterator` is from C++14. if we need backward compatibility, I will rewrite with another way.

----

second, I know the test is failed now. and I want to know is the current test really correct.
I think fine result is the using getLines with iteration is same as result of split by "\n" (or "\r\n").

so, e.g. when content of ofBuffer is `"\n\nabc\ndef\n\n"`, current results of getLines and getRevrseLines are:

```
[notice ] lines: 0 
[notice ] lines: 1 
[notice ] lines: 2 abc
[notice ] lines: 3 def
[notice ] lines: 4 

[notice ] rlines: 0 
[notice ] rlines: 1 def
[notice ] rlines: 2 abc
[notice ] rlines: 3 
[notice ] rlines: 4 
```

I think doesn't this makes confuse?
and, if we want to ignore last empty line, then "\n\n" isn't needed to ignored too? or "\n\n\n"?

thanks

oxillo:
I would really prefer the option of having the last line included i.e.

```
[notice ] lines: 0 
[notice ] lines: 1 
[notice ] lines: 2 abc
[notice ] lines: 3 def
[notice ] lines: 4 
[notice ] lines: 5
```

That would be consistent with the number of lines shown in most text editors (`"\n\nabc\ndef\n\n"` has 6 lines in VS Code, Notepad++ and nano but vim has 5...) .
Also, I would expect that `getLines()` on `"\n\ndef\nabc\n\n"` would return the same as `getReverseLines()` on `"\n\nabc\ndef\n\n"`

PR #6946, opened 2022-04-21
Icon simplification

dimitre:
this PR simplificates the app icon. all the logic is removed from pbxproj except the rsync script
and everything is now concentrated inside the readable .xcconfig
the debug icon uses the xcconfig conditional [config=Debug]


dimitre:
if approve I'll make the changes in iOS too

PR #6951, opened 2022-04-30
Remove redundant serialTest.ino and .cpp

dimitre:
I think this two files are safe to remove. the cpp one is similar to the one used in platformio, but still, it is in a folder called applet, so I think they are good to remove.
ping author @LeoColomb


LeoColomb (approved):
> it is in a folder called applet

Yeah, aren't all the file inside `applet` the one generated by the Arduino IDE/builder?
I honestly don't remember, what I don't see what can be wrong removing these files.

dimitre:
Another easy and approved by the author @ofTheo

PR #6973, opened 2022-05-12
Moving lines from Release and Debug to Shared and selector [config=]

dimitre:
One idea of having the configurations side by side, using the [config=*] selector to differentiate, 
If this alteration is welcome we can remove Release.xcconfig and Debug.xcconfig

```xcconfig
COPY_PHASE_STRIP[config=Debug] = NO
COPY_PHASE_STRIP[config=Release] = YES
```


ofTheo:
Not 100% sure its needed, but if it makes it easier to manage fine with it. 👍

danoli3:
Looks good !

Issue #6947, opened 2022-04-21
Proposal : ofGraphicsCairo.h - cairo functions outside ofGraphics.h


dimitre:
I'm proposing here a separation of all Cairo functionalities to a file outside of ofGraphics.h
Already tested here and works great.  I've called ofGraphicsCairo.h and ofGraphicsCairo.cpp

This paves the way of a modular OF where we can trim down unneeded functionalities per project.
I'm already testing some proof of concept with separate .xcconfig 
```xcode
//THIS HAS ALL THE HEADER AND LIBS FOR OF CORE
#include "../../../libs/openFrameworksCompiled/project/osx/CoreOF.xcconfig"
GCC_PREPROCESSOR_DEFINITIONS=OFLIB_USECAIRO
#include "../../../libs/openFrameworksCompiled/project/osx/Cairo.xcconfig"
```
related: https://github.com/openframeworks/openFrameworks/issues/6948

#### **`ofGraphicsCairo.h`**
```c++
#pragma once

#include "ofRectangle.h"

/// \}
/// \name Screen Saving
/// \{

/// \brief Begin rendering to a PDF file.
///
/// openFrameworks allows rendering of 2D graphics to pdf via the
/// ofCairoRenderer. ofBeginSaveScreenAsPDF() is called before drawing. When
/// done drawing call ofEndSaveScreenAsPDF() to output the file.
///
/// ~~~~{.cpp}
/// void ofApp::setup(){
///     ofBeginSaveScreenAsPDF("screenshot.pdf", false);
///     ofSetColor(54,54,54);
///     ofDrawEllipse(100,100,200,200);
///     ofEndSaveScreenAsPDF();
/// }
/// ~~~~
/// \sa End drawing with ofEndSaveScreenAsPDF()
///
void ofBeginSaveScreenAsPDF(std::string filename, bool bMultipage = false, bool b3D = false, ofRectangle outputsize = ofRectangle(0,0,0,0));

/// \brief Terminates draw to PDF through ofCairoRenderer and outputs the file.
/// \sa ofBeginSaveScreenAsPDF()
void ofEndSaveScreenAsPDF();

/// \brief Begin rendering to a SVG file.
/// \sa ofEndSaveScreenAsSVG(), ofBeginSaveScreenAsPDF()
void ofBeginSaveScreenAsSVG(std::string filename, bool bMultipage = false, bool b3D = false, ofRectangle outputsize = ofRectangle(0,0,0,0));

[...]

Issue #6948, opened 2022-04-22
Proposal: Different #define handling in ofMain.h


dimitre:
What I'm proposing here is having certain defines to see if some functionality is included or not, which can be based on platform as it is now, but can be used independent of platform too.
#undef is very useful here because we can make the exceptions independent of the previously set #defines.

I mean previously set because we can make this defines in the project itself using GCC preprocessor (tested here and working)
and it can be easily set per project. 

Ideas? suggestions? 

```c++
#define OF_LIB_USE_THREAD
#define OF_LIB_USE_CAIRO
#define OF_LIB_USE_SERIAL
#define OF_LIB_USE_SOUND

#ifdef TARGET_NO_SOUND
	#undef OF_LIB_USE_SOUND
#endif

#if defined(TARGET_EMSCRIPTEN)
	#undef OF_LIB_USE_THREAD
#endif 

#if defined( TARGET_OF_IOS ) | defined(TARGET_ANDROID) | defined(TARGET_EMSCRIPTEN)
	#undef OF_LIB_USE_SERIAL
	#undef OF_LIB_USE_CAIRO
#endif
```

Issue #6866, opened 2022-02-01
OpenFrameworks - room for improvement


dimitre:
First of all thank you for the incredible project and collective effort put in OF. 
I've been using for more than 10 years, and I've decided to read carefully the code to see if some things could be improved there.
And IMHO there are. so I'm opening this issue here and inviting you to discuss ideas or post new ones.

roymacdonald:
Hi @dimitre 
I agree with you. we discussed some of this things back on 2019 at the denver conference, but soon after covid came and, well, you know...

What I recall and think would be mostly useful was the idea of modularizing OF. Have a very basic core and all the rest is treated as addons, communications, networks, 3D, sound, video, graphics, etc. The way that OF is used right now will still be possible, but with this modularization one could easilly disable the modules that are not needed, which should compile faster and into smaller apps.
There were other things  we mentioned and were written in some documents, which I think @bakercp own. 
@ofTheo @ofZach @arturoc 

What ideas do you have in mind?

dimitre:
Sure @roymacdonald it would be great to have a way of sharing ideas and results from the conference, so others could pick up or add or discuss more.

dimitre:
### Folders naming and location
it would be great to change some parts of structure for a major version like 0.12.0
changing osx to macos in docs, scripts and templates

and moving templates to root, not inside scripts. it is just strange that this folder sits there, together with lots of .sh folders.

ofTheo:
Just fyi @dimitre there is a 0.12.0 task-list issue here: 
https://github.com/openframeworks/openFrameworks/issues/6692 

It outlines what we are aiming for, for 0.12.0. 
Any easy fixes for anything on that list would be helpful. And anything that should go on that list can be added as a comment on the issue. 

Thanks for all the help so far! 
Feel free to PR any things on the ofSite that seem out of date or inaccurate.

thomasgeissl:
hey hey, thanks for reaching out.

I am currently traveling and turns out train/bus rides work very well for refactoring.
I was discussing some package manager/project generator features with Theo a couple of months ago. Unfortunately, other gigs came in and I haven't found time to continue working on it. 
Basically, I am simplifying things and porting the UI stuff to c++, with the hope that everyone who knows oF could work on the package manager. 
[...]

PR #6985, opened 2022-05-18
Cairo functions from ofGraphics separated in different files.

dimitre:
So in near future we can build OF without Cairo.a, optionally, as discussed here
- https://github.com/openframeworks/openFrameworks/issues/6947


dimitre:
The idea itself was upvoted in the thread by @roymacdonald and @danoli3 
If anybody has the time to proofread this PR it would be great!

roymacdonald:
@dimitre cool. Is it just that you moved the Cairo stuff from ofGraphics.* Into ofGraphicsCairo.* ?
Why are you putting the include to ofGraphicsCairo.h in ofMain.h instead of ofGraphics.h?
I think that that include should go in the place that the code was extracted so not to beeak the dependecies in projects. Say, someone has a project and they use #include "ofGraphics.h" to use some of the cairo related functions, with what you are proposing it will not compile and that would suck. Figuring out what was making it not to compile might take some time yet it wont harm if you put #include "ofGraphicsCairo.h" in ofGraphics.h

dimitre:
Thanks @roymacdonald! it does make sense. 
For me it is the same, as it is or as you suggest. I'll make this change if others agree.

Personally I think it is a very remote possibility and I prefer this way because we can use preprocessor definitions to customize the OF to compile "Cairoless" soon.

I mean, most of the projects includes ofMain.h and lots of addons does the same.
and for 0.12 we can break theoretically compatibility a little :)

dimitre:
Any opinion on this one @openframeworks/core ?

dimitre:
Hey @ofTheo, finally it is passing all tests and ready to review.
Share your opinion on this

dimitre:
Reason behind this change: 
Sometimes we want to compile faster, or build a smaller binary file, have less time spent validating in app store, so it makes sense to aim for a modular OF where we can left out some unused functionality.
I love Cairo but I don't use it in 80% of my projects, so it is great to be able to left it out, as well as freeimage, fmod, etc.

Some related threads
https://forum.openframeworks.cc/t/of-lite/5170
https://forum.openframeworks.cc/t/openframeworks-lite/37256
https://forum.openframeworks.cc/t/bundling-osx-app-without-glut-and-without-fmodex-nasty/23503/4

I think there was some similar ideas shared in Denver conference in 2019 as posted here too:
https://github.com/openframeworks/openFrameworks/issues/6866#issuecomment-1053783484

danoli3:
Definitely let’s add some macros to define stuff out where not needed,
fmod, curl/ssl, Cairo. Will make future maintenance easier too

dimitre:
Finally, good to go @ofTheo

dimitre:
Hello @ofTheo I've updated this branch to work well with latest. 
I've tested with project generator and it seems good to go.
Any observations or objections on this one?

ofTheo:
Thanks @dimitre - is the idea to do another PR later which will selectively include cairo based on MACRO / #defines? 
If that's the case it looks good to me.

dimitre:
Yes, the idea is merge this one, and work on another which involves more decisions to uniformize defines.
here is the complete proposal for this one
https://github.com/openframeworks/openFrameworks/issues/6947

and here a proposition of changing the way ofConstants.h is organized which we can discuss soon
https://github.com/openframeworks/openFrameworks/issues/6948

Issue #6527, opened 2020-01-25
ofVideoPlayer isPlaying(), different behavior on ofVideoPlayer with or without Gstreamer


dimitre:
Using Gstreamer isPlaying() always returns true even when paused and using plain ofVideoPlayer it returns false when paused. It would be great to have a standard behavior.
see:
https://github.com/arturoc/ofxGStreamer/issues/27

roymacdonald:
Hi, I came across this issue as well when fixing [this](https://github.com/openframeworks/openFrameworks/pull/6517), although as I was digging under ofxGStreamer's hood, I figured out what @arturoc  [explained](https://github.com/arturoc/ofxGStreamer/issues/27), regardless of this it should be fixed so it makes sense.

ofTheo:
I think @arturoc is correct. 
would be good to standardize across platforms. 

I can take a look at macOS and Windows.  
Can we assume streamer is fine as is?

roymacdonald:
I am not sure what it correct or fine in this case. 
Having a function that is called `isPlaying()`which returns true even when it is paused feels a bit counter-intuitive. 
is there any other big difference between stopping and pausing a video player other than that the former moves the playhead back to the begining? 
calling `stop()`, as fas as I understand does not deallocate memory, thus, from the users perspective it is not really different  from calling `videoPlayer.setPaused(true);videoPlayer.setPosition(0);`. 
I think that `isPlaying()` should simply be the negation of `isPaused()` 
I did just a quick check and there is no consistent behavior in this regard. 
I agree upon standarizing across platforms but, which would be this standard?

dimitre:
yes, to me it is more intuitive that isPlaying returns false when paused.

ofTheo:
Hmm. 
I always understood isPlaying() to be related to play() stop() and isPaused() for toggling pause states. 

I think it makes sense that isPlaying is true unless stop has been called.

dimitre:
|  | isPlaying()  | isPaused()  |
|---|---|---|
| stopped  | false  | ?  |
| paused  | ?  | true  |
| playing  | true  | false  |

ofTheo:
okay this seems to be the last issue on #6483 

Here are the different implementations currently. 
Bolded for values that aren't the same. 

**GStreamer:**

|  | isPlaying()  | isPaused()  |
|---|---|---|
| stopped  | false  | **true**  |
| paused  | **true**  | true  |
| playing  | true  | false | 


[...]

PR #6986, opened 2022-05-20
ofVideoPlayer isPaused fix for macOS

dimitre:
now macOS follows windows model as discussed here
https://github.com/openframeworks/openFrameworks/issues/6527

basically the only change is:
after you call video.stop() 
video.isPaused() returns false instead of true

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationPlayer.mm
+++ b/libs/openFrameworks/video/ofAVFoundationPlayer.mm
@@ -285,10 +285,7 @@ void ofAVFoundationPlayer::stop() {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


ofTheo:
Thanks @dimitre ! 
Would you be comfortable taking a stab at the gstreamer implementation, so the platforms are all uniform?

dimitre:
Sure thing @ofTheo I'll take a look today. Do you think we group in the same PR?

ofTheo:
Thanks @dimitre! 
Maybe a new PR as its probably all platform dependent files? 

Going to do a quick test with the changes from this one and merge if all good.

Issue #6715, opened 2021-03-23
ofCairoRenderer stops working after setting ofScale(0, 0);


dimitre:
I've noticed ofCairoRenderer stops working after setting ofScale(0, 0)
Maybe there is some destructive scale happening there. easy to reproduce.
Thanks

arturoc:
I imagine what you mean is that it doesn't work afterwards even changing back the scale? It might be introducing some NaNs and never recover from it

dimitre:
@arturoc yes this is what I mean.

dimitre:
ok, almost there to close this one. I've discovered the ofScale uses the actual matrix as it was 100%.
so if you for example calls this every frame the circle shrinks
```c++
	ofScale(0.5, 0.5);
	ofDrawCircle(100,100,100);

```

dimitre:
one possible solution is adding some lines to setupGraphicsDefaults, so it resets the matrix in new frames.
But only if we start using cairo with the second optional parameter = true
```c++
ofSetCurrentRenderer(cairo, true);
```

```c++
//----------------------------------------------------------
void ofCairoRenderer::setupGraphicDefaults(){
	setStyle(ofStyle());
	path.setMode(ofPath::COMMANDS);
	path.setUseShapeColor(false);
	clear();
	
// new lines
	cairo_matrix_t matrix;
	cairo_matrix_init_scale(&matrix, 1.0, 1.0);
	cairo_matrix_init_translate(&matrix, 0.0, 0.0);
	cairo_set_matrix(cr,&matrix);
};
```

dimitre:
Closed by https://github.com/openframeworks/openFrameworks/pull/6991

PR #6991, opened 2022-05-23
ofCairoRenderer fix scale=0 / matrix transformation issues

dimitre:
This PR avoids scale=0 issues mentioned in this Issue https://github.com/openframeworks/openFrameworks/issues/6715
and fix some matrix transformation cumulative between frames.
like scale not resetting to 1.0 in a new frame, as others renderers do.


libs/openFrameworks/graphics/ofCairoRenderer.cpp
```diff
@@ -792,7 +792,9 @@ void ofCairoRenderer::translate(const glm::vec3 & p){
 
 //----------------------------------------------------------
 void ofCairoRenderer::scale(float xAmnt, float yAmnt, float zAmnt ){
+	if (xAmnt == 0 || yAmnt == 0) return;
```
ofTheo, line 795: I think we can skip this line and keep the rest of the PR. As then cairo behavior matches the other renderers. 
For example if you do: 

```
void ofApp::draw(){
    ofScale(0, 0, 1);
    ofDrawCircle(100, 100, 50); 

    ofScale(1, 1, 1);
    ofDrawCircle(200, 100, 50); 

}
```

With the regular OF renderer you won't see anything either.

dimitre: Great.
but without this line Cairo stops and never recovers if we set scale=0
a quick test to reproduce the issue
```c++
#include "ofApp.h"

shared_ptr<ofBaseGLRenderer> opengl;
shared_ptr<ofCairoRenderer> cairo;
ofTexture render;
bool useCairo = true;

void ofApp::setup(){
	ofRectangle rect = { 0, 0, (float)ofGetWidth(), (float)ofGetHeight() };
	opengl = ofGetGLRenderer();
	cairo = make_shared<ofCairoRenderer>();
	cairo->setupMemoryOnly(ofCairoRenderer::IMAGE, false, false, rect);
	render.allocate(rect.width, rect.height, GL_RGBA);
}

void ofApp::draw(){
	if (useCairo) {
		ofSetCurrentRenderer(cairo, true);
	}
	// draw here
	ofClear(0,255);
	ofSetColor(255,0,80);
	ofTranslate(100,100);
//	float s = cos(ofGetElapsedTimef() * .5)*.5 + .5;
	float s = 0.5;
	//	cout << s << endl;
	s = (mouseX-200) * .01;
	cout << s << endl;
	ofScale(s, s);
	ofDrawCircle(100,100,100);
	
	if (useCairo) {
		cairo->flush();
		render.loadData(cairo->getImageSurfacePixels());
		ofSetCurrentRenderer(opengl, true);
		ofSetColor(255);
		render.draw(0,0);
	}
}

void ofApp::keyPressed(int key) {
	if (key == 'a') {
		useCairo ^= 1;
	}
}

```

ofTheo: Hmm, I just tried this test with 0.11.2 and it works fine. 
The circle scales with the mouse even after `s` goes to 0. 

Does it not work with the current master?

dimitre: Yeah I should have split in a new issue for clarity. Thanks!

ofTheo: Been messing around with it and can't seem to get it working again once scale is set to 0.0. 
It must break something else in the matrix / rendering that isn't fixed with resetting the matrix with the `cairo_matrix_init_scale` 

Totally stumped though.

ofTheo: Been messing around with it and can't seem to get it working again once scale is set to 0.0. 
It must break something else in the matrix / rendering that isn't fixed with resetting the matrix with the `cairo_matrix_init_scale` 

Totally stumped though.

dimitre: Yeah I think we should have this guard in OF for stability until they solve in Cairo project. (if it happens)
I can insert a comment like:
```c++
// temporary band-aid for a issue where Cairo never recovers after setting scale = 0
// safe to remove once it is fixed 
if (xAmnt == 0 || yAmnt == 0) return;
```

dimitre: Let's merge this one and be one step closer to 0.12
Even if it is not perfect it is better than  _freezing Cairo_  haha
frozen_pyramids_by_hatemind-d6wu88x-3386732872

ofTheo: My only thought was to maybe try passing scale though at 0.000001 or something if 0 was requested, but didn't get a chance to test that yet. 

If that doesn't work ( ie: you can recover from that ), lets merge as is.

dimitre: Your idea worked fine
I'm using the smallest number a float can represent
```c++
	if (xAmnt == 0) xAmnt = std::numeric_limits<float>::epsilon();
	if (yAmnt == 0) yAmnt = std::numeric_limits<float>::epsilon();

```

PR #6992, opened 2022-05-23
xcodebuild - less verbose

dimitre:
using xcodebuild each of the three scripts used to output all env variables to the terminal
this PR disables this, and marks each build phase in terminal

Screen Shot 2022-05-23 at 12 02 52


ofTheo:
I wonder if instead of just **Bash script 01**  we could have something more descriptive: 
**Compiling openFrameworks** 
**Copying Resources** 
etc ?

ofTheo:
Perfect! 👍

PR #7005, opened 2022-05-27
using namespace std, removing

dimitre:
removing from core files


dimitre:
Hey @ofTheo I would like to hear your opinion on this one, thanks

PR #7009, opened 2022-05-31
removing unneeded .h

dimitre:
in this PR I've removed some more unneeded.h files and removed #include "ofConstants.h" from cpp files where it was already included in .h files.

the only exceptions left are ofFbo.cpp and ofLight.cpp (file isn't included in .h) 
it passes all tests.
what is your opinion @ofTheo ?


ofTheo (approved):
Looks great.  
Thanks @dimitre. Hope this reduces our core compiling time 🙂

PR #7010, opened 2022-06-01
fix(c++17) : deprecated std::iterator inheritance

### LINKED ISSUE #6977: ofPixels.h - std::iterator deprecated
This file uses std::iterator which is deprecated on c++17
This PR is great in replacing iterator in ofFileUtils.h
- https://github.com/openframeworks/openFrameworks/pull/6945

related issue:
- https://github.com/openframeworks/openFrameworks/issues/6926

oxillo:
fix #6977

as suggested in https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

It also compiles in c++11 (tested only with mingw64)

### DIFF
--- a/libs/openFrameworks/graphics/ofPixels.h
+++ b/libs/openFrameworks/graphics/ofPixels.h
@@ -458,7 +458,13 @@ public:
 
     /// \cond INTERNAL
 
-    struct ConstPixel: public std::iterator<std::forward_iterator_tag,ConstPixel>{
+    struct ConstPixel{
+		using iterator_category = std::forward_iterator_tag;
+		using value_type = ConstPixel;
+		using difference_type = std::ptrdiff_t;
+		using pointer = ConstPixel*;
+		using reference = ConstPixel&;
+
         ConstPixel(const PixelType * pixel, size_t bytesPerPixel, ofPixelFormat pixelFormat);
         const ConstPixel& operator*() const;
         const ConstPixel* operator->() const;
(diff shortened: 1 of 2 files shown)


2bbb:
maybe you (and the original author of this iterator) misunderstand `value_type` (and `reference`, `pointer`).
for example, `std::vector<int>::iterator` has `value_type` as `int`, `std::vector<int>::iterator` has `float`.
i.e., basically, `operator*()` returns `value_type &` and `operator->()` returns `value_type *`.

so, maybe `Line::value_type` is `std::string` and etc, etc...

ofTheo:
I'll admit that I don't really get this one! 
I think if we can get a clean solution that works for C++11 / C++17 onwards that would be ideal, but I'll leave it to @oxillo, @dimitre and @2bbb to decide :)

2bbb:
@oxillo 
In most cases, these codes will work fine. but if the user uses these with `iterator_traits`, some problems will come.
if this PR is first aid for remove deprecate warning, I agree to merge this.

I think, we need to fix some points about iterators.
(but, the biggest problem is what I have never used those features... 😉)

oxillo:
I've created #7011 to review the logic behind the ofBuffer iterators. 
This PR can be merged with the only goal to remove the deprecation warnings.

2bbb:
hi, @dimitre 

`difference_type` appear at `iterator::operator+(difference_type)` and non member function `operator+(iterator, difference_type)` and etc, etc...
so, basically `difference_type` is integer type.

if we use only simple increment operation, those type aliases will be not used. thus, it doesn't appear as bugs.
maybe we use with some std algorithms, then those will not work correctly. (but i can not think of concrete situation... 😖)

2bbb:
ref: https://www.internalpointers.com/post/writing-custom-iterators-modern-cpp

2bbb:
@dimitre 

great!

by the way, I found most simple example of std algorithm uses type aliases in iterator class.

so, `std::distance` returns distance between two iterator, and this type is `typename std::iterator_traits<some_iterator>::difference_type`. 

ref: https://en.cppreference.com/w/cpp/iterator/distance

dimitre:
I think this one needs a fix in ofFileUtils,
value_type for Line should be std::string, as we can see in the object operators

This is the main difference between the two PRs and I'll be happy to see any of them merged

ofTheo:
Thanks @dimitre - I'll merge yours then as it as the Line fix and we can always do a PR for anything we missed.

PR #7014, opened 2022-06-04
Msys2 : use libs from pkg-config

oxillo:
Prepare move to clang64 and ucrt64


ofTheo:
@oxillo do you think we should the same libs from our apothecary build forumulas?

oxillo:
@ofTheo  openframeworks/apothecary#226 remove msys2 from the build formulas - remaining ones are kiss, tess2 and videoInput.
MSYS2 being a rolling release as Arch Linux, I think it is better to follow system packages than rebuilding our own.
Now that MSYS2 has MINGW64, UCRT64 and CLANG64, I think it is even more justified.

PR #7018, opened 2022-06-08
Yet another std::Iterator replacement

dimitre:


dimitre:
@2bbb do you have any more suggestion on this one? Thank you!

2bbb:
@dimitre 

I think, basically, type aliases are ok.

about iterators like `ofBuffer::Line`, I can't judge what is correct behavior.
but It's true that it contains bugs anyway.

optionally, I think it makes sense to we think about really `iterator_cateogry` is `forward_iterator` or not.
if we can get simple and natural implementation about `operator--`, then it is `bidirectionaryl_iterator`. and if we can implement `operator[]` then it is `random_access_iterator`.

dimitre:
Thanks @2bbb in your avaliation this PR is worse as the actual code or it is equivalent? 
I'm all about getting rid of the warnings now

PR #7026, opened 2022-07-21
ofAVFoundationMediaPlayer - fix 'self' warning

dimitre:
In this PR I've addressed this warning
```
block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Wimplicit-retain-self]
```
by adding self-> to the variables. I've tested and it is working ok here, but it would be great to have some second opinion from @danoli3 or @2bbb to see if it is OK.


ofTheo:
Maybe @2bbb could confirm? 
( _I need to learn obj-c properly!_ )

danoli3:
Yeah basically without the self the completion handler boxes the booleans, which actually may lead to issues as it's just a copy of the bool bit
With self it forces the box of the ptr object, hopefully the original ptr and not a copy of the whole object haha

Yeah since this is recommended by Xcode I think all for it

danoli3 (approved):
Approved

PR #7030, opened 2022-07-24
Deleting empty addons.make from examples

dimitre:
Cleaning some empty addons.make files from examples folder


dimitre:
I think config.make and Makefile in examples are OK to remove too.

PR #7031, opened 2022-07-24
adding of.entitlements and vscode files to gitignore

dimitre:
in examples folder only so it doesn't exclude the template folder.


dimitre:
I've just added some more patterns to this unmerged PR, so it will ignore vscode files and new xcode project generator files in examples.

PR #7033, opened 2022-07-25
iOS updates 

dimitre:
Bundle identifier now present in Project.xcconfig
bin/data/ images removal (used on older launchscreen, now it uses a simple svg)


danoli3:
Looking good, 

What about this though:
```
** BUILD FAILED **
[6377](https://github.com/openframeworks/openFrameworks/runs/7502624861?check_suite_focus=true#step:5:6378)
warning: Capabilities for Signing & Capabilities may not function correctly because its entitlements use a placeholder team ID. To resolve this, select a development team in the emptyExample editor. (in target 'emptyExample' from project 'emptyExample')
```

ofTheo:
@dimitre I think the data folder got removed from the template which is breaking the rsync script 
```
building file list ... rsync: link_stat "scripts/templates/ios/bin/data/." failed: No such file or directory (2)
[7285](https://github.com/openframeworks/openFrameworks/actions/runs/3099301608/jobs/5018275347#step:5:7286)
```

dimitre:
Great. data folder is there but I removed its contents (old png files, not needed anymore) so maybe we have to update rsync script end from /. to / only (not tested)
do you know where the rsync scripts are? 
thanks

dimitre:
Yess, I've submitted this PR to address this https://github.com/openframeworks/openFrameworks/pull/7093

I'm now investigating another issues related to projectGenerator too

dimitre:
@ofTheo about rsync, I've disabled this and it is not the cause for failure.
I can compile using XCode with a real phone, there are no libs for simulator.
xcodebuild fails in the end. 
```
The following build commands failed:
	Ld /Users/z/Downloads/of_v0.11.2_ios_release/examples/ios/graphicsExample/build/Release-iphoneos/graphicsExample.app/graphicsExample normal (in target 'graphicsExample' from project 'graphicsExample')
```
I've downlaoded the latest release to see if it completes using xcodebuild and it fails too.

dimitre:
@ofTheo now I've noticed the Github Actions uses emptyExample to test with xcodebuild, so I've tried to repeat the same command in my setup, like
```
xcodebuild -configuration Release -project "/Volumes/tool/ofw/scripts/templates/ios/emptyExample.xcodeproj"  -target emptyExample -sdk 'iphonesimulator' ARCHS='x86_64'
```
and it completes OK locally, and fails remotely. 
do you have any idea why?

ofTheo:
@dimitre the error I am seeing in Github Actions is in this section:

https://github.com/openframeworks/openFrameworks/blob/master/scripts/templates/ios/emptyExample.xcodeproj/project.pbxproj#L310-L319

A quick fix to try would be appending the rsync line with

`2>/dev/null` 

So it reads: 

`"rsync -avz --exclude='.DS_Store' \"${SRCROOT}/bin/data/\" \"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}\" 2>/dev/null"`

I think that would fix the CI error.

dimitre:
Thank you @ofTheo you are right, it was data folder missing all the time :)

PR #7051, opened 2022-08-18
No ARC enforcing

dimitre:
this PR is meant to allow using non ARC projects so OF is compatible with addons like ofxSyphon and others.
Related: https://github.com/openframeworks/openFrameworks/pull/6889#issuecomment-1195667266

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -188,7 +188,7 @@ enum ofTargetPlatform{
 	#endif
 
 	#if defined(__OBJC__) && !__has_feature(objc_arc)
-// 		#error "Please enable ARC (Automatic Reference Counting) at the project level"
+		#warning "Please enable ARC (Automatic Reference Counting) at the project level"
 	#endif
 #endif
 
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -188,7 +188,7 @@ enum ofTargetPlatform{
 	#endif
 
 	#if defined(__OBJC__) && !__has_feature(objc_arc)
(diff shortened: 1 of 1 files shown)


ofTheo:
thanks @dimitre - do you think we should change it to a `#warning` ? 
I guess in the case of ofxSyphon there is nothing a user can do to fix except upstream in the addon?

cc @2bbb

dimitre:
Yeah a warning can work also, if you prefer.
I've tried to fix ofxSyphon myself but it is out of my reach, technically, now :)

2bbb:
I think writing `#error` macro  in every implementation files updated to ARC code (or creating new head for only wrote this macro and include from those) is better.
then, it allows add flag "-fno-objc-arc" to third party library.

but "#warning" is good too. then, it makes addon developers want to update to ARC.😉

dimitre:
What about just commenting out now and uncomment it before the release? or change to whatever is decided?

ofTheo:
@dimitre I think warning is fine as we'll have ARC on at the project level. 
So only people who disable ARC will see it.

libs/openFrameworks/utils/ofConstants.h
```diff
@@ -188,7 +188,7 @@ enum ofTargetPlatform{
 	#endif
 
 	#if defined(__OBJC__) && !__has_feature(objc_arc)
-		#error "Please enable ARC (Automatic Reference Counting) at the project level"
+// 		#error "Please enable ARC (Automatic Reference Counting) at the project level"
 	#endif
```
ofTheo, line 192: @dimitre want to change this to #warning and then we can merge?

dimitre: Yes, I don't now how to do it. any suggestion? 
#warning does nothing on xcode
so maybe it is #pragma warning?

PR #7052, opened 2022-08-18
XCode pbxproj - json ident

dimitre:
Finally ok. first converters I've tried double escaped some paths.


ofTheo:
awesome thanks!! :)

PR #7053, opened 2022-08-18
Github Actions updates - Caching libs and compiled code on macOs

dimitre:
opening this one just to share some tests and open discussion.

— I've noticed some jobs are uploading libs in the end. What about disabling it in everyday commits?
— ~~Bump macOS version to 12, ok for macOS and some strange issues on tvOS/iOS (framework CoreMotion not found) I can investigate further or try macOS 11 for tvOS/iOS~~ @ofTheo already fixed that
— change some scripts number of cores (some used to use one core, some four, github nowadays usually uses 2, so I've set to use all cores possible, unless the process was killed, so I've changed to two cores and worked ok)

Other things that could be improved in the future
there is a way of caching downloads from one session to another, so ofw libs could be cached saving download time and server transfer.
https://github.com/actions/cache
and there is a caching of apt packages too
https://github.com/marketplace/actions/cache-apt-packages


ofTheo:
@dimitre would be awesome to look into the caching stuff. 

When I  started  porting stuff from Travis over to Github I couldn't get the caching to work well and there was so much else to try and get working, I basically didn't implement it, but it could be added ( maybe selectively ) for things which are always the same and can take a long time. 

Same with cores. Happy to do anything to speed things up :) 
A smarter system will be faster and use less compute / server time ( environmental win ), so yeah and clean-up / speed-up is great. 

Bear in mind CI stuff is a beast and kind of a massive time suck, so some things are the way they are because they finally worked after weeks of trial and error 🙂. @danoli3 can attest to this 🙃

In terms of the libs we may not need them in the future if we get Github doing all the packaging. 
But for now I think it is best to leave.

dimitre:
Hello @ofTheo some improvements here in actions. I was able to cache 
— libs, so it is a faster process when it runs for the second time, 
— and compiled openframeworks too. I suppose if we cache compiled files we can save some precious minutes and it will compile the modified files anyway.

In fact osx action (without makefile) is already running in 2 min, makefiles are taking a lot more (17 min).
Questions:
Do you have any idea of some other folder we can save in cache to avoid recompiling (makefiles) entirely?
Tests only runs on makefiles on purpose? can we switch to only run on xcode build which is much faster now?
And lastly, why do we need to upload libs? is this for the nigtly builds? if yes I think we can do it selectively, like this
https://github.com/orgs/community/discussions/27128

ofTheo:
Thanks @dimitre! 

This is awesome. 

### Xcode vs Makefile 
It might  be good to try the tests on Xcode too and see if it is much faster?
If so then it  might be down to the makefile approach recompiling OF each time. 

### Uploading libs 
I think ultimately we won't need this as I plan to switch the nightly builds to Github Actions anyway, but it might be too much work to make just the uploading run nightly when switching it over completely wouldn't be that much more work. Maybe we can leave this for now?

dimitre:
Great! I'll do that, move the tests to XCode, uncomment the upload libs and bring back all the tests once it runs well.
It will be great to have special attention to macOs machine, once I've read they bill minutes x 10

ofTheo:
Awesome @dimitre 👍

> once I've read they bill minutes x 10

We don't get charged for CI runs, but even so making things faster and less wasteful is good for everyone ( environmental impact too ).

dimitre:
Yeah @ofTheo I'm all about saving watts and seconds of human energy. 
I'm pretty happy now both macOs actions finish under 8 minutes.

dimitre:
Not all builds upload libs to the server, just some of them. 
This is why I thought it was something unimportant
- [x] android
- [x] emscripten
- [ ] ios
- [x] linux  
- [ ] macos
- [ ] msys2
- [ ] vs

dimitre:
good to go

ofTheo:
Thanks @dimitre !!

PR #7062, opened 2022-08-28
update linearResampleTo() and hermiteResampleTo() in ofSoundBuffer class.

sadmb:
update resampling code in order to fix the following bugs.

1. They were crashed when end had negative number.
2. They only had 2 loops with high speed resampling

screenshot 2022-08-28 22 12 31
.

### DIFF
--- a/libs/openFrameworks/sound/ofSoundBuffer.cpp
+++ b/libs/openFrameworks/sound/ofSoundBuffer.cpp
@@ -341,7 +341,21 @@ void ofSoundBuffer::linearResampleTo(ofSoundBuffer &outBuffer, std::size_t fromF
 	}
 	
 	std::size_t start = fromFrame;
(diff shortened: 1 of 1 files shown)


dimitre:
Ping @roymacdonald

danoli3:
If negative values, suggests file format was signed and not unsigned, might
need to convert bit shift up to get valid representation of audio and not
clipped out the bottom spectrum

What input data type were you using in that example?

roymacdonald:
OF uses signed floats between -1 and 1. 
I would rather use an external library to deal with resampling as it is non trivial. I made an addon using libsamplerate which can be ported to all platforms and it gives me correct results all the time, while the of implementation fails often. I would love to see such implemented. If you give me the gonI would be happy to implement for all platforms

Issue #6592, opened 2020-06-07
bug with making contours from fonts 


ofZach:
one of my students was experimenting with this font and this code

https://www.dropbox.com/s/p02qyulbud26fqb/lato.ttf?dl=0
https://github.com/ofZach/RTP_SFPC_SUMMER20/tree/master/typeExample3

and found that the parsing looks a little weird, some screenshots when you zoom in

<img width="429" alt="Screen Shot 2020-06-07 at 7 04 09 AM" src="https://user-images.githubusercontent.com/142897/83967222-86952d80-a88d-11ea-9a4d-d21a698fa32c.png">
<img width="517" alt="Screen Shot 2020-06-07 at 7 04 01 AM" src="https://user-images.githubusercontent.com/142897/83967223-872dc400-a88d-11ea-8fce-d747441ca70b.png">

it seems that when you change the makeContoursForCharacter and make all the 64 to **64.0** instead it fixes things, so perhaps it's just a resolution issue?

<img width="398" alt="Screen Shot 2020-06-07 at 7 04 43 AM" src="https://user-images.githubusercontent.com/142897/83967232-9e6cb180-a88d-11ea-97dd-52f5575fd814.png">
<img width="455" alt="Screen Shot 2020-06-07 at 7 04 37 AM" src="https://user-images.githubusercontent.com/142897/83967233-9e6cb180-a88d-11ea-97d5-1b8375e2402a.png">

happy to submit a PR just wanted to mention it as an issue in case anyone has ideas.

dimitre:
I think this is related to the fact assumes as a default parameter for simplifyAmt the value of 0.3
so maybe every loaded font is simplified?

dimitre:
I've checked out the code and confirm simplifyAmt = 0.0 fixes this
This PR closes this issue
https://github.com/openframeworks/openFrameworks/pull/7065


![Screen Shot 2022-09-27 at 21 05 26](https://user-images.githubusercontent.com/58289/192658512-50c8fbdf-b998-42a0-92b9-69c4445f602c.png)

ofTheo:
closed by #7065

Issue #5138, opened 2016-06-23
ofTrueTypeFont snaps to pixels (no subpixel aliasing)


morphogencc:
It looks like `ofTrueTypeFont` doesn't render properly unless it's drawn at an integer pixel position due to subpixel aliasing.  This makes doing effects with moving text look clunky and low-framerate.

I posted a question about this on the forum, and @ofTheo mentioned rendering it to an `ofFbo` and then drawing it, which works great as a solution.  As this is something I do frequently with text, I was considering adding functionality to render to an FBO first as a method to `ofTrueTypeFont`, or perhaps even making it the default drawing method.

But first, I wanted to get a sense of whether there's a reason this hasn't been finished already.  I'm more than happy to write the code myself and send it a PR.

ofTheo:
Here is the thread in the forum which this came from. 
https://forum.openframeworks.cc/t/oftruetypefont-looks-clunky-like-slow-framerate-while-translating/23753/6

any reason we shouldn't use this type of approach to rendering the text out ( minus the translate ) vs the current one?

```
//do this once
    ofVboMesh mesh;
    mesh = font.getStringMesh("sentence",0,0);

// draw
    ofTranslate(x, y);
    font.getFontTexture().bind();
    mesh.draw();
    font.getFontTexture().unbind();
```

arturoc:
that's more or less what's happening behind the scenes in ofTruetypefont the problem is not how we do the translation but how the ttf are rendered. for this to work we would need subpixel rendering which works by loading the fonts as rgb instead of grayscale and then using a shader to do the antialiasing depending on the fraction of the position and the rgb value of the pixels.

it's like using the rgb components of each pixel as if they were small pixels themselves https://en.wikipedia.org/wiki/Subpixel_rendering

ofTheo:
ahh - yeah - I just made this following test: 

```
ofTrueTypeFont ttf;
ofVboMesh mesh;
ofFbo fboText;
ofRectangle textBounds;

//--------------------------------------------------------------
void ofApp::setup(){
    ttf.load("arial", 20);
    
    ofRectangle textBounds = ttf.getStringBoundingBox("hello slow moving text - as an fbo", 0, 0);
    fboText.allocate(textBounds.getWidth()+2, textBounds.getHeight()+2);
}

//--------------------------------------------------------------
void ofApp::update(){

}

//--------------------------------------------------------------
void ofApp::draw(){
[...]

PR #7065, opened 2022-09-03
fix ofTrueTypeFont::stringWidth

dimitre:
I'm submitting this PR to start a discussion.
stringWidth actually get width of getStringBoundingBox, which uses a mixture of glyphProperties advance and xmin+width, that are not the same thing.

this PR fixes getStringBoundingBox, makes stringWidth independent of the function, so it is less overhead.

address the issue discussed here
https://forum.openframeworks.cc/t/are-monospace-fonts-really-monospace/40358/3

### DIFF
--- a/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
+++ b/libs/openFrameworks/graphics/ofTrueTypeFont.cpp
@@ -231,7 +231,7 @@ static ofPath makeContoursForCharacter(FT_Face face){
 
 #ifdef TARGET_OSX
 //------------------------------------------------------------------
-static std::string osxFontPathByName(const std::string& fontname){
+static string osxFontPathByName(const string& fontname){
 	CFStringRef targetName = CFStringCreateWithCString(nullptr, fontname.c_str(), kCFStringEncodingUTF8);
 	CTFontDescriptorRef targetDescriptor = CTFontDescriptorCreateWithNameAndSize(targetName, 0.0);
 	CFURLRef targetURL = (CFURLRef) CTFontDescriptorCopyAttribute(targetDescriptor, kCTFontURLAttribute);
@@ -240,7 +240,7 @@ static std::string osxFontPathByName(const std::string& fontname){
 	if(targetURL) {
 		UInt8 buffer[PATH_MAX];
 		CFURLGetFileSystemRepresentation(targetURL, true, buffer, PATH_MAX);
-		fontPath = std::string((char *)buffer);
+		fontPath = string((char *)buffer);
 		CFRelease(targetURL);
 	}
 
@@ -254,7 +254,7 @@ static std::string osxFontPathByName(const std::string& fontname){
 #ifdef TARGET_WIN32
 #include <map>
 // font font face -> file name name mapping
-static std::map<std::string, std::string> fonts_table;
+static std::map<string, string> fonts_table;
 // read font linking information from registry, and store in std::map
 //------------------------------------------------------------------
 void initWindows(){
@@ -299,7 +299,7 @@ void initWindows(){
 	char fontsPath[2048];
 	SHGetKnownFolderIDList(FOLDERID_Fonts, 0, nullptr, &ppidl);
 	SHGetPathFromIDList(ppidl,&fontsPath);*/
-	std::string fontsDir = getenv ("windir");
+	string fontsDir = getenv ("windir");
 	fontsDir += "\\Fonts\\";
 	for (DWORD i = 0; i < value_count; ++i)
 	{
@@ -314,8 +314,8 @@ void initWindows(){
 
 			wcstombs(value_name_char,value_name,2048);
 			wcstombs(value_data_char,reinterpret_cast<wchar_t *>(value_data),2048);
-			std::string curr_face = value_name_char;
-			std::string font_file = value_data_char;
+			string curr_face = value_name_char;
+			string font_file = value_data_char;
 			curr_face = curr_face.substr(0, curr_face.find('(') - 1);
 			fonts_table[curr_face] = fontsDir + font_file;
 	}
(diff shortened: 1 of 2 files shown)


libs/openFrameworks/graphics/ofTrueTypeFont.cpp
```diff
@@ -1162,8 +1164,16 @@ void ofTrueTypeFont::drawCharAsShape(uint32_t c, float x, float y, bool vFlipped
...
+//	return getStringBoundingBox(c, 0,0).width;
+	int w = 0;
+	iterateString( c, 0, 0, false, [&]( uint32_t c, glm::vec2 pos ){
+		if ( c == '\t' ){
+			w += getGlyphProperties(' ').advance * spaceSize * TAB_WIDTH;
+		} else {
+			w += getGlyphProperties( c ).advance;
+		}
+	});
+	return w;
```
ofTheo, line 1176: Maybe this should be in `getStringBoundingBox`

libs/openFrameworks/graphics/ofTrueTypeFont.h
```diff
@@ -377,6 +377,13 @@ class ofTrueTypeFont{
     /// \returns current font direction
 	void setDirection(ofTrueTypeFontDirection direction);
 
+	float charWidth(uint32_t c) const{
+		return getGlyphProperties(c).width;
+	}
+	float charAdvance(uint32_t c) const{
+		return getGlyphProperties(c).advance;
+	}
+
```
ofTheo, line 386: We should have this in the cpp and maybe with better naming. 
`getCharWidth` `getCharAdvance` ?

dimitre: Sure! Other idea we can just make getGlyphProperties public, so this functions are not needed

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
dimitre: Yes I think the same. I'll take some time to test with right to left texts too, just to see if it works correctly

danoli3:
Amazing!!! I’ve discovered this bug only recently and it makes so much
sense seeing this fix!

I’ll test the changes as well

dimitre:
I've just tested and both this PR update and the original code doesn't return any width for right to left text.
so I think at least we don't have any regression
```c++
	cout << font.stringWidth("برامج") << endl;
```

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
dimitre: Done. I didn't use the function itself so we dont' call twice iterateString

libs/openFrameworks/graphics/ofTrueTypeFont.h
dimitre: I've just renamed for now, I can separate to cpp soon

dimitre:
Some new changes address this issue from 2016
https://github.com/openframeworks/openFrameworks/issues/5138

I think somebody used long variables (integer) to store 26.6 fixed-point data directly from Freetype, but instead stored bit shifting and converting but storing in long, so metrics are losing precision.

I've used functions to convert back and forth this kind of number to double or float, so things get much more precise.

danoli3:
Oh amazing! I’ll check this out see if solves all my glyph width issues

libs/openFrameworks/graphics/ofTrueTypeFont.cpp
ofTheo: Not sure I understand? `getStringBoundingBox` only has one iterateString. 

So couldn't the `stringWidth` function just do what it used to?

```
 	ofRectangle rect = getStringBoundingBox(c, 0,0);
 	return rect.width;
```

Sorry if I missed something?

ofTheo: Thanks @dimitre - type stuff can be really tricky. 🙂
My first thought is including the xMin is okay - as people are requesting the bounding box representing the full bounds of the represented string. 

I could see us adding more utilities in the future so placing two strings side by side ( with different typefaces ) is easier ( I have run into this ). 

But I think negative values in some cases with 0,0 origin can be okay.

dimitre: So if we keep using xMin in bounding box it makes sense to keep stringWidth as a separate independent function, so we can measure only the character advance in stringWidth

ofTheo: Ahh - okay, looking back at the original forum post I see the logic here. 👍
Let's do what you suggest and before the PR is merged we can do some tests, with [the case in the forum post](https://forum.openframeworks.cc/t/are-monospace-fonts-really-monospace/40358) and the existing examples.

dimitre:
@ofTheo here I used an example of a font with a negative xMin in relation to origin. 
It is Zapfino font present on macOs, but as we can see in this software the character width only consider the distance from origin to character advance.

We can see a bounding box there too disregarding font bleed to the left and right, 
I think we can use only advance to measure string and bounding box. It is simpler and 99% of the time we won't use this kind of font. 

If we consider xmin and xmax we have to consider if the character is the first or last in the string, as the other metrics we can know all from advance which considers kerning too.

Screen Shot 2022-09-20 at 21 30 25

ofTheo:
thanks for the example @dimitre 
I think your approach sounds good as long as you think it won't break people's current usage in a big way. 

we could also add additional functions for other metrics like bounding box with font bleed etc.

dimitre:
Yes I think it can be merged as it is since it fixes long time lurking integer issues, path simplification artifacts
And then when we have time we can improve functions, consider right to left texts, etc.

dimitre:
closes https://github.com/openframeworks/openFrameworks/issues/6592

PR #7074, opened 2022-09-14
ofOpenALSoundPlayer load check

NickHardeman:
Check if OpenAL successfully loaded the file before proceeding. I forgot to install codecs and without this check it caused an app crash.


ofTheo:
Thanks @NickHardeman!

PR #7083, opened 2022-09-16
unreachable code

dimitre:
XCode complains here because code is unreachable. of course, it begins with return 0;
should we comment out? remove?

https://github.com/openframeworks/openFrameworks/blob/4ccc0b20a6b138dc94bc719bc308e29d518a8e67/libs/openFrameworks/app/ofAppNoWindow.cpp#L44-L51

### DIFF
--- a/libs/openFrameworks/app/ofAppNoWindow.cpp
+++ b/libs/openFrameworks/app/ofAppNoWindow.cpp
@@ -44,10 +44,10 @@ void set_conio_terminal_mode()
 int kbhit()
 {
 	return 0;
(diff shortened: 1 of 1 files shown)


dimitre:
this seems to invalidate this piece of code here too
https://github.com/openframeworks/openFrameworks/blob/4ccc0b20a6b138dc94bc719bc308e29d518a8e67/libs/openFrameworks/app/ofAppNoWindow.cpp#L241-L257

ofTheo:
yeah - I noticed this one too. 
Good to have it commented out.

PR #7098, opened 2022-09-23
fix algeterror

azchatlanin:
Maybe it's better this way?


dimitre:
@ofTheo I think this is good to merge

ofTheo:
absolutely! - what a funny mistake.  🤦‍♂️
thanks @azchatlanin

PR #7099, opened 2022-09-26
Examples: OpenCV image classification example using Yolo5

ofTheo:
Screen Shot 2022-09-22 at 3 55 57 PM

This requires OpenCV 4.6. 
Will do a separate PR to apothecary to update OF to 4.6.


dimitre:
Woww that's great! 
With this apothecary PR I will be able to port some code made with ofxOpenCvDnnObjectDetection to ofxOpencv

PR #7110, opened 2022-09-28
of::filesystem PR

dimitre:
realtes to:
#6585 

specifically:
https://github.com/openframeworks/openFrameworks/issues/6585#issuecomment-1146430183


dimitre:
I finally finished the PR which uniformize of::filesystem for file paths, so I've marked as ready to review.
I know it is a big PR but once we get after this huge change we can rely a lot more on filesystem to simplify a lot other things too.
Today I've ported Assimp, textures, everything to use filesystem. now it is working OK here.
take a look when you have time @ofTheo @NickHardeman

dimitre:
I'm now separating things that are unrelated in small PRs, so this one will be easier to read

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
```diff
@@ -80,7 +80,7 @@ bool ofxAssimpModelLoader::load(string modelName, int assimpOptimizeFlags){
 	//	}
 
 	// loads scene from file
-	std::string path = file.getAbsolutePath();
+	std::string path = file.getAbsolutePath().string();
```
ofTheo, line 83: Curious - does `std::string path = file.getAbsolutePath();` not work with this change? 
Or is this just being more proper? 🙂

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
```diff
@@ -388,20 +389,35 @@ void ofxAssimpModelLoader::loadGLResources(){
...
+				auto modelFolder = ofFilePath::getEnclosingDirectory( file.path() );
+				auto relTexPath = ofFilePath::getEnclosingDirectory(texPathStr,false);
+				auto realPath = modelFolder / of::filesystem::path{ texPathStr };
+				
+				using std::cout;
+				using std::endl;
+				cout << "modelFolder : " << modelFolder << endl;
+				cout << "texPathStr : " << texPathStr << endl;
+				cout << "relTexPath : " << relTexPath.string() << endl;
+				cout << "realPath : " << realPath.string() << endl;
```
ofTheo, line 402: probably want to remove this before merging

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
dimitre: in this PR  getAbsolutePath now returns of::filesystem::path instead of string, so it is needed to convert to string in this case

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
dimitre: Sure thing :)

dimitre: done!

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
ofTheo: Hmm - I think that could break a ton of addons/projects/code etc if it doesn't auto convert. 

One thing that might work is to have `std::string` and `of::filesystem::path` versions of both with deprecation? 
Or maybe we could figure out a way to have path auto convert to string with = operator overload? 

I think without it we'll break a ton of stuff 🙂

dimitre: Yes you're right. I'm good about any of this. maybe keeping std::string for legacy compatibility, I can do that soon

dimitre: We cannot have both functions with same name and different return type. I'm unsure of how to proceed here. I know it is not your style but I would suggest we switch to of::filesystem and live with labor pains. 
What is your recommendation on this?

ofTheo: Hmm - I wonder if it just works? 
Could you maybe test in your branch what happens if you do: 

```
std::string path = file.getAbsolutePath();
std::cout << path << std::endl; 
```


For  me this works without error/warning:

```
    std::filesystem::path myPath;
    myPath = "test/something/yeah";

    std::string str = myPath;
    std::cout << " str is " << str << std::endl;
```

dimitre: Yeah it works. I've just pushed another commit.
Everything working here.

ofTheo: bummer - looks like VS is not so forgiving: 

```
ofxAssimpModelLoader.cpp(83,43): error C2440: 'initializing': cannot convert from 'std::filesystem::path' to 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' 
```

dimitre: Yes some platforms are complaining, like android. vs
why? maybe compiler version?

ofTheo: yeah - I am not sure why its okay on some and not others. 
probably a compiler or c++ implementation difference. 

I tried an = operator and conversion operator but both need to be declared in the class itself so that's no good ( as we don't have access to string or path. 

pinging @artificiel incase he has some insight here.

artificiel:
no windows here but interesting problem and seems like a case of "native" OS paths being std::string vs std::wstring:

https://stackoverflow.com/questions/57377349/implicit-conversion-between-stdfilesystempath-and-stdstring-should-it-hap
https://developercommunity.visualstudio.com/t/cannot-implicitly-cast-stdfilesystempath-to-stdstr/518455
https://stackoverflow.com/questions/58521857/cross-platform-way-to-handle-stdstring-stdwstring-with-stdfilesystempath

so...i guess a small namespace and using it to usurp std::filesystem::path is imaginable but hacky/brittle (and will take a bit of time to get right).

i agree things should break as little as possible; i also agree that the cost of moving forward is sometimes having to break a bit. 

within OF instead of using `.string()` the way out is to use `auto` and fix the break as it propagates. it will reveal places where utf-16 paths would have failed in one way or another. once OF is coherent, user code is only a matter of typing (again, to `auto` instead of `std::string`). (and that's presuming users already have code using std::filesystem::path).

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
dimitre: PR updated and conflicts solved. 
I think untangling the whole filesystem issue worths the occasional rewrite it can cause in some other places. for me personally this change is less dramatic than ARC for example, and it reverberated very well on core and addons.

any ideas on this one @artificiel ?

artificiel: just so i understand the core idea of the PR (ofConstants.h is an area i don't visit much):

1. there is some (pre-identified, perhaps historical or platform-specific) reason that makes std::filesystem not fit for all, so there is a strategy of aliasing of::filesystem to one of different ::filesystem based on compile-time (`#ifdef`) criteria;

2. the platform-specific std::filesystem::path is solved with `auto`;

3. then the PR also tackles the switch of std::filesystem to of::filesystem, as well as upgrade some common usage of "strings as paths" to the actual ::filesystem::path (but then question: for instance why in SVG it's still `void ofxSvg::load(std::string path)`? changing the arg to filesystem::path will not break code that pass a string. not a complaint! really a question to grab the scope -- if the goal is to get a functional strategy in place it's OK to do deep cleaning later).

dimitre: Motivation behind this PR: uniformize of::filesystem outside of std, instead of try to tuck everything inside std::filesystem (there are different versions of c++11, experimental c++17 and boost filesystem)
And another motivation is try to rely more on filesystem::path instead of converting from string to path and vice versa depending on the functions.

3 . If approved other leftovers can be done in separate PRs easier to read. I can see other places like 
```ofCairoRenderer::setup(const std::string & _filename``` and ```setupShaderFromSource```

Comment (artificiel) [reply addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp:83]: OK very clear! i'm not in a position to test things but as you propagated the changes in a good number of components with success it's clearly a positive move forward. my only comment (and maybe that falls in a "style" discussion) it so committing commented unused/updated code -- i understand the value of that while working, being able to quickly refer to the past and toggle "versions", but in committed core code it adds a bit of noise (even more so within large `#ifdef` blocks like in ofConstants.h which are hard to follow). it also provides traction for the of:: namespace.

Comment (dimitre) [reply addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp:83]: Totally agree @artificiel. I've just cleaned out the comments. Thanks for the reading.
@ofTheo I really think the potential cons are outweighted by pros of this change
And this can be useful for tackle some projectGenerator issues, which uses a mixture of filesystem and Poco::Path

**Comment by dimitre** [comment]
This PR is finally complete, 
it changes lots of files at once, so if we just left it open the tendency is conflicts arise from newer PRs like this : https://github.com/openframeworks/openFrameworks/pull/7361

I think having to append .string() to convert a path to a string is a good change, so people can consciously convert the path to string, or u8string when needed
```c++
    fs::path path{u8"愛.txt"};
    std::u8string path_string{path.u8string()};
```

But I would like to invite others to give opinion on this: @artificiel @ofTheo @2bbb @NickHardeman

2bbb:
hi! @dimitre 

I think what we have explicit way to convert is good.

addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
ofTheo: @artificiel some more context for the changes are in this issues: 
https://github.com/openframeworks/openFrameworks/issues/6585#issuecomment-1146430183 

Will add to the main PR description to link the issues. 

@dimitre - going to take a look at this today. 
Most nervous about changing the return types for the ofFile:: call from string to path.  I def agree path is better especially for international support, but thinking about all the things this could break ( similar of a break as going from ofPoint to glm::vec3 ). 

Going to see if there is a solution so we can have cake and eat it 🙂
Curious if anyone has creative ideas there.

artificiel: i must confess i'm not sure about what requires the of:filesystem (vs direct std::filesystem) but i'm presuming that has been sorted out and is the right thing to do — it does not change the fact that returning one or the other is different than std::string. 

1. direct implicit conversions such as `std::string p = my_dir.getAbsolutePath();` will fail on windows. but the current `getAbsolutePath()` method makes use of `.string()` to output strings:
https://github.com/openframeworks/openFrameworks/blob/edb72bd71f3a91cf8833ffc0a5e39df9e2cdca6f/libs/openFrameworks/utils/ofFileUtils.cpp#L1207-L1212
which means it forces the conversion of windows wstring to string, which is convenient but "incorrect"....

2. what is going to break globally is things like the usage of strings to construct ad-hoc paths:
```
std::string my_m = d.getAbsolutePath() + "/my_mov.mov"; 
```
but that usage should be definitely be shifted (not advocating this a good path strategy just pointing the "most-direct-adaptation" here):
```c++
auto my_m =  d.getAbsolutePath() / "my_mov.mov";
my_movie.load(my_m);
ofLogNotice("my_m") << my_m;
```
so these breaking changes are breaking on things that should be fixed.

the glm transition example is interesting, and i remember the moment of that change `OF_USE_LEGACY_VECTOR_MATH` seemed like a good thing, but in truth it's like making a deal with the credit card company: feels like a win today, but in the end you pay more anyhow. (and someone paid for all those #ifdefs's!)

(and the case of filesystem::path is even more necessary as it's a conceptual transition upgrading a bare string to a structured object, not a transition between somewhat "equivalent" systems).

one thing though: this means that all the interfaces and all the examples must be adapted — to a learning mind, a stale example is worse than no example... it should not be hard, but it has to be done coherently.

i will allow myself to re-phrase my take on the relationship between to OF source and one's projects: 
 - either the project is in development (in which case, either you are under deadline pressure and stick with a release in the "production mindset", or you are comfortable riding git and working on something that tolerates these moments to happen)
 - or the project is old, and either you stick to the old OF codebase (the copy you took with your project archive, or the freshly downloaded corresponding archive, hoping it still compiles), or try to work it up to today. considering there are 15-year old OF projects i find it "out of scope" to try to support legacy at all costs as the "responsibilities" of OF are small vs the evolution of OS API's, IDE's and C++ (especially since we are not talking about incremental upgrades -- if you pull out a 2010 project, you will not "upgrade" to each sequential OF, OS and compiler versions until you hit today, you will point your project to an OF tree and compile with your current setup. there are so many variables involved in that i don't think it's good to burden OF with additional work that makes older edge cases work transparently).

(i find the OF deprecation strategy very nice -- the old functions can hang in there for ever, and politely point out the better option to the user; unfortunately there is no way to do something like that with return types).

ofTheo:
Thanks for those thoughts! 

One thing we could do if we want to merge this sooner is keep the string return types, merge this PR and put the path return types in a seperate PR. 

This would give a bit more time to think about strategy. I agree though this is ultimately the way to go, just don’t want to break a ton of addons, examples and projects without a bit more planning.

dimitre:
Great @ofTheo I've just did that
converted all the return types to string.

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -23,46 +23,40 @@ using std::ostream;
...
-            return std::filesystem::canonical(ofFilePath::join(ofFilePath::getCurrentExeDir(),  "../../../data/")).string();
-        }catch(...){
-            return ofFilePath::join(ofFilePath::getCurrentExeDir(),  "../../../data/");
-        }
-    #elif defined TARGET_ANDROID
-        return string("sdcard/");
-    #else
-        try{
-            return std::filesystem::canonical(ofFilePath::join(ofFilePath::getCurrentExeDir(),  "data/")).make_preferred().string();
-        }catch(...){
```
ofTheo, line 41: Thanks @dimitre! 
The only thing I am wondering about is if we are losing anything with the removal of the try{ catch in the old defaultDataPath().  Do you know if the new behavior is the same or could there be some breakage here?

dimitre: @ofTheo, thanks for reading, I've reverted back to try and catch and pushed a new commit.
I'm not sure why I changed that before.

ofTheo: thanks @dimitre 
I think we need it for the #else section too ( last commit was just for `#if defined TARGET_OSX` 

Then this looks good to go! 👍
Thanks!!

dimitre: Great, done

ofTheo:
Thanks @dimitre ! 🎉🎉🎉👏

PR #7116, opened 2022-09-30
iOS Template Updates

dimitre:
eliminate some warnings by
ALWAYS_SEARCH_USER_PATHS = NO

Bundle identifier with iOS in reverse URL identifier
PRODUCT_BUNDLE_IDENTIFIER = cc.openFrameworks.iOS.$(TARGET_NAME)

"ENABLE_TESTABILITY": "YES", in Debug

Updating the template to latest version, so it will be easy to add new files (ofAVEngine, etc)


danoli3:
These changes stale now yeah @dimitre

PR #7118, opened 2022-09-30
Removing Quicktime code - Updates on make template

dimitre:
Removing unexistent file $(OF_LIBS_PATH)/videoInput/ from exclusions
Removing QTKit from frameworks in makefiles


ofTheo:
Awesome. Wondering if maybe these should just get deleted? 
We still have them in the tagged branches and I can't imagine anyone using them anymore.

dimitre:
yeah I'll delete them. I've moved Quicktime files to another folder in this PR too, or we can remove them totally if you prefer
they are now in $ofw/other/legacy

ofTheo (changes requested):
This looks good @dimitre 
If you can switch it to remove the files instead of moving them I think it is good to go. 👍

ofTheo (approved):
Looks great! 
Thanks @dimitre 👍

Once it passes the last check I'll merge it in.

PR #7121, opened 2022-10-01
WIP - Cleaning up .h files

dimitre:
I'm wondering about improving compiler times by changing some things in OF core:

minimizing including .h files in other .h files, like ofConstants.h where it is not needed.
forward declaration of objects where possible
moving all logic which still happens in .h files to cpp files

### DIFF
--- a/libs/openFrameworks/app/ofAppRunner.h
+++ b/libs/openFrameworks/app/ofAppRunner.h
@@ -13,40 +13,34 @@ class ofBaseApp;
 class ofBaseRenderer;
 class ofCoreEvents;
 
+using std::shared_ptr;
 
 void ofInit();
 void ofSetupOpenGL(int w, int h, ofWindowMode screenMode);	// sets up the opengl context!
-std::shared_ptr<ofAppBaseWindow> ofCreateWindow(const ofWindowSettings & settings);	// sets up the opengl context!
-std::shared_ptr<ofMainLoop> ofGetMainLoop();
-void ofSetMainLoop(const std::shared_ptr<ofMainLoop> & mainLoop);
+shared_ptr<ofAppBaseWindow> ofCreateWindow(const ofWindowSettings & settings);	// sets up the opengl context!
+shared_ptr<ofMainLoop> ofGetMainLoop();
+void ofSetMainLoop(const shared_ptr<ofMainLoop> & mainLoop);
 
 template<typename Window>
-void ofSetupOpenGL(const std::shared_ptr<Window> & windowPtr, int w, int h, ofWindowMode screenMode){
-	ofInit();
-	ofWindowSettings settings;
-	settings.setSize(w, h);
-	settings.windowMode = screenMode;
-	ofGetMainLoop()->addWindow(windowPtr);
-	windowPtr->setup(settings);
-}
+void ofSetupOpenGL(const shared_ptr<Window> & windowPtr, int w, int h, ofWindowMode screenMode);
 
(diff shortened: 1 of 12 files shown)


dimitre:
Hello @ofTheo @2bbb check this one when you have time.
My efforts here were:
- remove includes from .h files when possible
- change includes to forward declaration when possible.
- change some includes from .h to .cpp when possible (so there is less functionality to parse, things compile once and get stored in .o objects)
- Added include headers in cpp files where dependencies come from other .h files.
- made a change in ofConstants.h splitting to ofDeprecate.h so we can include just this bit of functionality if nothing else is needed from ofConstants.h

in my tests it shaved down about 3 seconds from the total compilation time (106s to 103s)

libs/openFrameworks/app/ofAppRunner.h
```diff
@@ -13,40 +13,34 @@ class ofBaseApp;
 class ofBaseRenderer;
 class ofCoreEvents;
 
+using std::shared_ptr;
```
2bbb, line 16: I don't recommend `using` declaration on header file.
It will affect to all codes include `ofAppRunner.h`.

dimitre: thank you @2bbb I'll be changing this

2bbb:
I suggest a point what is not critical about this topic but related including headers.

So, it is order of including.
I recommend to include order like:

* related detail headers (e.g. ofXXXVideoPlayer related to ofVideoBaseTypes.h)
* other openFrameworks core headers (like ofConstants, utils, etc...)
* third-party headers
* system headers (i.e. std libraries)

This order is helpful for find missing include files hide by implicit include.
For example:

some_features.h
```cpp
struct X {
    // implements
}

std::ostream &operator<<(std::ostream &os, const X &x);
```

main.cpp
```cpp
#include <iostream>
#include "some_features.h"

void main() {
    std::cout << "init" << std::endl;
    X x;
    std::cout << x << std::endl;
    return 0;
}
```

Now, some_features.h depends to iostream, but not include iostream.
Maybe those codes works for a while, but it will case compile error when using some_features.h alone.

If we write as:

```cpp
#include "some_features.h"
#include <iostream>

void main() {
    std::cout << "init" << std::endl;
    X x;
    std::cout << x << std::endl;
    return 0;
}
```

Then we can find missing include files in early stage.

dimitre:
@2bbb Nice ideas about includes organization, I agree with this ideas.
I'm now using a very nice tool called "include-what-you-use" which is basically a fork of clang who gives you suggestion of simplifying includes. I'll be using this on my next projects, as you can include only parts of openFrameworks and shave off 30% compile time.

if other people want to give a try, you can install via homebrew
```
brew install iwyu
```

and then use in config.make or Makefiles like this:
```
PROJECT_CXX = include-what-you-use
PROJECT_CC = include-what-you-use
CXX = include-what-you-use
CC = include-what-you-use
```

and it gives you a report of what can be changed / optimized

dimitre:
https://en.cppreference.com/w/cpp/language/pimpl

dimitre:
This one is being broken up in pieces, so PRs are easier to read 
https://github.com/openframeworks/openFrameworks/pull/7367
https://github.com/openframeworks/openFrameworks/pull/7366
https://github.com/openframeworks/openFrameworks/pull/7357

dimitre:
Most of the ideas here were broken in smaller PRs, so this is not relevant anymore

Issue #7076, opened 2022-09-16
Examples - Update


NickHardeman:
Some of the examples could be removed or refreshed / updated.
Below are some that I think could use some attention and I plan to update.
Will submit a PR per example. 

### 3d
- [ ] 3d / cameraRibbonExample 
- [ ] 3d / meshFromCameraExample ( spins too fast ).
- [ ] 3d / ofNodeExample
- [ ] 3d / quaternionArcballExample
- [ ] 3d / quaternionLatLongExample

### gl
- [ ] ~~gl / alphaMaskingShaderExample~~ ( remove, duplicate of shader / 07_fboAlphaMask)
- [ ] gl / areaLightExample
- [ ] gl / geometryShaderExample
- [ ] gl / multiTextureShaderExample
- [ ] gl / singleLightExample > change to gl/pointLightExample
- [ ] gl / transformFeedbackExample ( not included in OSX download? )

### shader
- [ ] shader / 07_fboAlphaMask

Thoughts on any other examples?

dimitre:
Great @NickHardeman!
One example I really wanted to change is serialExample. It can be more inviting for beginners to work with serial communication, withtout memset, memcpy etc.

dimitre:
I can work in an alternative approach, I use something like this
```c++
void ofApp::setup(){
	int baud = 9600;
	serial.listDevices();
	vector <ofSerialDeviceInfo> deviceList = serial.getDeviceList();
	// this should be set to whatever com port your serial device is connected to.
	// (ie, COM4 on a pc, /dev/tty.... on linux, /dev/tty... on a mac)
	// arduino users check in arduino app....
	serial.setup(0, baud); //open the first device
}

void ofApp::update(){
	if (serial.available()) {
		serial.readBytes(buffer, serial.available());
		serialData += buffer.getText();
		serial.flush();
	}
}
```

NickHardeman:
That definitely seems more approachable.

ofTheo:
Going to ping @ofZach @LingDong @golanlevin on this thread and see if there are any examples that students often ask for.

ofZach:
I'll think about this, but off the top of my head, some demos my students always really appreciate: 

- drawing into a high res fbo for high res image output
- examples that help explain how to do custom meshes / geometry and texture map them (ie, cut up a webcam into triangles, etc) 
- converting paths -> polylines, polylines to meshes,  etc 

[...]

PR #7124, opened 2022-10-04
Add transform feedback example to macos (osx) download

eduardfrigola:
As discussed in #7076. 
I think this is all that is needed in order to not be removed from macos download


ofTheo:
Thanks!!

PR #7126, opened 2022-10-04
ofMaterial set additional textures to support better model loading

ofTheo:
Been working on making the assimp model loader much more robust and useful. 
( will do a separate PR for that ) 

This PR adds texture support for loading: 
- emissive 
- occlusion 
- normal map / bump map 
- specular 
- ambient 
 
To the ofMaterial phong.frag shader. 
In the future we could add a pbr.frag shader for Physical Based Rendering but for now this adds quite a bit. 

You can see the emissive and normal map support this adds to this model: 

Screen Shot 2022-10-04 at 1 30 17 PM

And some of the details you get from the normal map in this model: 
Screen Shot 2022-10-04 at 2 02 15 PM


dimitre:
Looking gr8!

PR #7128, opened 2022-10-05
download_libs.sh from apothecary nightly

dimitre:
worked ok here locally let's see how it works on actions

### DIFF
--- a/scripts/dev/download_libs.sh
+++ b/scripts/dev/download_libs.sh
@@ -35,6 +35,7 @@ if [[ ! -d "$SCRIPT_DIR" ]]; then SCRIPT_DIR="$PWD"; fi
 download(){
     echo "Downloading $1"
     # downloader ci.openframeworks.cc/libs/$1 $SILENT_ARGS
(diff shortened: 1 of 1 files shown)


ofTheo:
thanks @dimitre !!

dimitre:
I still have one or two things to fix there
- I suppose this one
openFrameworksLibs_master_vs2017_64
was renamed to this
openFrameworksLibs_master_vs_64

- and this one is missing
openFrameworksLibs_master_msys2_mingw32
because here are our options on apothecary nightly:
openFrameworksLibs_master_msys2_ucrt64
openFrameworksLibs_master_msys2_clang64
openFrameworksLibs_master_msys2_mingw64

ofTheo:
Ahh yes.  

I think we can drop mingw32 ( meant to do that already ). 
For now we can have msys2 CI just do mingw64 and once that is all working we can do a PR for clang64 / crt64 ( maybe these aren't all needed? ). 

The vs name change is my mistake, but I think it works better? 
So happy to drop the 2017 from any scripts. 🙂

dimitre:
I've sent another PR, but as the actions arent' running on OF project we can see if it works here
https://github.com/dimitre/openFrameworks/actions

PR #7129, opened 2022-10-05
download_libs.sh vs2017 -> vs

dimitre:


ofTheo:
testing a slightly more involved branch here that also removes 32bit from the visual studio template. 
https://github.com/openframeworks/openFrameworks/compare/master...ofTheo:openFrameworks:remove-win32 

going to close this one if that is okay?

PR #7130, opened 2022-10-06
A dream about fbos summing using += operator

dimitre:
Last night I dreamt about OF, ofFbo had a += operator so you could sum layers like this
```c++
fbo += fbo2;
```

so I decided to make this quick PR 

example
```c++
#include "ofApp.h"

ofFbo fbo;
ofFbo fbo2;
ofFbo fbo3;

void ofApp::setup(){
	fbo.allocate(600, 600, GL_RGBA);
	fbo.begin();
	ofClear(0,0);
	ofSetColor(255, 0, 0);
	ofDrawCircle(300, 300, 200);
	fbo.end();
	
	fbo2.allocate(600, 600, GL_RGBA);
	fbo2.begin();
	ofClear(0,0);
	ofSetColor(0, 0, 255);
	ofDrawRectangle(100, 100, 300, 50);
	fbo2.end();

	fbo3.allocate(600, 600, GL_RGBA);
	fbo3.begin();
	ofClear(0,0);
	ofSetColor(0, 255, 0);
	ofDrawRectangle(200, 200, 300, 50);
	fbo3.end();

	fbo += fbo2;
	fbo += fbo3;
}

void ofApp::update(){

}

void ofApp::draw(){
	fbo.draw(0,0);
}

void ofApp::keyPressed(int key) {

}

```
Screen Shot 2022-10-06 at 16 23 03

### DIFF
--- a/libs/openFrameworks/gl/ofFbo.cpp
+++ b/libs/openFrameworks/gl/ofFbo.cpp
@@ -303,6 +303,14 @@ ofFbo::ofFbo(const ofFbo & mom){
 }
 
 //--------------------------------------------------------------
+void ofFbo::operator+=(const ofFbo & mom){
+	this->begin();
+	ofSetColor(255);
+	mom.draw(0,0);
+	this->end();
+}
+
+//--------------------------------------------------------------
 ofFbo & ofFbo::operator=(const ofFbo & mom){
 	if(&mom==this) return *this;
 	clear();
--- a/libs/openFrameworks/gl/ofFbo.h
+++ b/libs/openFrameworks/gl/ofFbo.h
@@ -48,6 +48,7 @@ public:
 	ofFbo();
 	ofFbo(const ofFbo & mom);
 	ofFbo & operator=(const ofFbo & fbo);
+	void operator+=(const ofFbo & fbo);
     ofFbo(ofFbo && mom);
     ofFbo & operator=(ofFbo && fbo);
 	virtual ~ofFbo();
--- a/libs/openFrameworksCompiled/project/Makefile
+++ b/libs/openFrameworksCompiled/project/Makefile
@@ -14,6 +14,12 @@
 # define the OF_SHARED_MAKEFILES location
 OF_SHARED_MAKEFILES_PATH=./makefileCommon
 
+# PLATFORM_CFLAGS 
+
+# PROJECT_CXX = include-what-you-use
+# PROJECT_CC = include-what-you-use
+# CXX = include-what-you-use
+# CC = include-what-you-use
 
 # include the global configuration file
 include $(OF_SHARED_MAKEFILES_PATH)/config.shared.mk


ofTheo:
love this idea. 
curious what others think? @ofZach @NickHardeman @2bbb 

I would definitely consider removing the ofSetColor in the += incase you would want to do that explicitly.

dimitre:
yeah I was in doubt about setcolor, I'll remove it now

NickHardeman:
I like the idea of operators. 
It makes sense using GL_RGBA, but a GL_RGB fbo would just draw right over the other. Not sure if this would be confusing or not.
I would also be interested in an equal operator to a color  like `fbo = ofColor::black` that simply begins, clears to color and then ends.

2bbb:
It seems like good!

Can we support more general rhs targets like ofImage, ofVideoPlayer, ofVideoGrabber?
(or more and more general, structs have draw method...?😊)

Yes, I know, making thing general makes complex.
However, I think it would be good if it could be generalized with a good balance.

ofZach:
one question -- what happens when FBOs are not the same size?

dimitre:
@ofZach it can be drawin in 0,0 coordinate. top left

dimitre:
I like the ofColor operator too!

dimitre:
@NickHardeman I think it is OK to just overwrite if there is no transparency or blend mode enabled for the two images.
Not an unexpected result.

2bbb:
This is just an idea.

`+=` is simple and understandable.
But `fbo += fbo1` and `fbo = fbo + fbo1` are not same. (I don't know there is good implementation for `fbo + fbo1`)
Maybe this is not so important. However, there is a point of concern.

And I think design of iostream is not so cool. (mainly about string formatting and iomanip...)
However I think those style is possibility those are better and extendable way to fbo processing.

like below:

```cpp
fbo << fbo1;
fbo << fbo2 << fbo3;
fbo << ofFboOpStyle::Fill << fbo4;
fbo << ofFboOpStyle::AspectFit << fbo5;
fbo << ofFboOpStyle::Offset(10, 10) << fbo6;
```

I know those may be over-featured or non-intuitive...
As a mere suggestion.

libs/openFrameworks/gl/ofFbo.cpp
```diff
@@ -302,6 +302,13 @@ ofFbo::ofFbo(const ofFbo & mom){
     }
 }
 
+//--------------------------------------------------------------
+void ofFbo::operator+=(const ofFbo & f){
+	this->begin();
+	f.draw(0,0);
+	this->end();
+}
+
```
ofTheo, line 311: might want to add a check that if the fbo isn't allocated you can an error/warning 
or you could have if it is not allocated it gets allocated to match first?

libs/openFrameworks/gl/ofFbo.cpp
```diff
@@ -302,6 +302,13 @@ ofFbo::ofFbo(const ofFbo & mom){
     }
 }
 
+//--------------------------------------------------------------
+void ofFbo::operator+=(const ofFbo & f){
```
ofTheo, line 306: arg should be fbo in header/cpp I think to be consistent.

libs/openFrameworks/gl/ofFbo.cpp
```diff
@@ -303,10 +303,10 @@ ofFbo::ofFbo(const ofFbo & mom){
...
 //--------------------------------------------------------------
-void ofFbo::operator+=(const ofFbo & f){
-	if (f.isAllocated()) {
+void ofFbo::operator+=(const ofFbo & fbo){
+	if (fbo.isAllocated()) {
 		this->begin();
-		f.draw(0,0);
+		fbo.draw(0,0);
 		this->end();
 	}
```
ofTheo, line 311: Feel like it should be something like: 

```
	//if we aren't allocated lets allocate to be the same as the input
	if (!this->isAllocated() && fbo.isAllocated()) {
 		this->allocate(fbo.settings); 
 	}
	if (fbo.isAllocated()) {
 		this->begin();
 		fbo.draw(0,0);
 		this->end();
 	}else{
 		ofLogWarning("ofFbo::operator+=") << " input fbo is not allocated, skipping += ";
 	}
```

ofTheo: @dimitre what do you think about the suggestion above?👆

This way it basically copies the fbo being += if the main fbo isn't allocated

artificiel:
what about  `fbo *= shader`, where` fbo` is passed through `shader`, transparently swapping a lazily-allocated buffer?

artificiel:
(posting like this as I'm not sure how to pull another user's branch then add to the PR?) 

ofFbo.h
```c++
	ofFbo & operator *=(const ofShader & shader) { return pass(shader); }
	ofFbo & pass(const ofShader & shader);
	std::unique_ptr<ofFbo> buffer; // sole overhead if method is unused is unallocated pointer
```
ofFbo.cpp
```c++
ofFbo & ofFbo::pass(const ofShader & shader) {
	
	// lazy  --  perhaps augment settings to make within allocate() if so desired
	if (!buffer) buffer = std::make_unique<ofFbo>();
	
	// conformant  --  if compare is too costly at every frame, hook into allocate to maintain sync
	if (buffer->settings != settings) buffer->allocate(settings);
	
	buffer->begin();
	ofClear(0,0,0,0);
	shader.begin();
	draw(0,0);
	shader.end();
	buffer->end();
	
	// perhaps this could be "texture swapped" with some clever GL code?
	begin();
	buffer->draw(0,0);
	end();
	
	return *this; // to enable things like : flatten_fbo += (layer_fbo *= layer_shader);
}
```
ofApp.cpp:
```c++
    fbo *= shader;
```

PR #7142, opened 2022-10-16
Example slitscan clock using ofVideoGrabber and ofTransform

danbz:
this additional example uses the oF video grabber to sample camera input. Then the code accesses the pixels in each frame from the camera to create a slitscan video. The image is created in two styles, progressively across the screen or as a ribbon, pushing pixels sampled from the center of the screen. The videopixels object is rendered to the screen as a flat, horizontal image or using ofTransform to display as a radial slitscan image . The timing of the scan is used in the style of a clock with each screen taking 1 minute to scan. the example also uses ofTrueTypeFont to overlay the system clock time.
 
Screenshot 2022-10-16 at 13 44 26


ofTheo:
Thanks @danbz 
These look great!

PR #7149, opened 2022-10-28
support for include and pragma include shader macros

patriciogonzalezvivo:
Adding support for both 

```glsl
#include "dependency.glsl"
```

```glsl
#pragma include "dependency.glsl"
```

to `ofShaders`

### DIFF
--- a/libs/openFrameworks/gl/ofShader.cpp
+++ b/libs/openFrameworks/gl/ofShader.cpp
@@ -414,6 +414,8 @@ bool ofShader::setupShaderFromSource(ofShader::Source && source){
  * https://www.opengl.org/discussion_boards/showthread.php/169209-include-in-glsl?p=1192415&viewfull=1#post1192415
  */
 
+typedef std::vector<std::string> StringList;
+
 //--------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/gl/ofShader.cpp
```diff
@@ -414,6 +414,8 @@ bool ofShader::setupShaderFromSource(ofShader::Source && source){
  * https://www.opengl.org/discussion_boards/showthread.php/169209-include-in-glsl?p=1192415&viewfull=1#post1192415
  */
 
+typedef std::vector<std::string> StringList;
+
 //--------------------------------------------------------------
```
ofTheo, line 419: Can we remove the typedef? 
Then I think it will be good to go! 🙂

patriciogonzalezvivo: Ups! Good catch!

ofTheo:
Thanks @patriciogonzalezvivo!  🙏

patriciogonzalezvivo:
Thank you guys!! This is going to be FUN!

PR #7150, opened 2022-11-01
swapping around closing the window before you delete it

danzeeeman:
fixes a crash on windows when you close the window


ofTheo:
Thanks @danzeeeman !!

PR #7159, opened 2022-11-11
Fix linux testallexamples

### LINKED ISSUE #6742: linux testAllExamples.sh fails

themancalledjakob:
testAllExamples.sh stumbled over README.md (and eventual qtCreator build-* directories), and then failed.

fix #6742

### DIFF
--- a/scripts/linux/testAllExamples.sh
+++ b/scripts/linux/testAllExamples.sh
@@ -11,7 +11,7 @@ do
     cd $category
     for example in $( ls . )
     do
-        if [[ "$example" == osx* ]]
+        if [[ "$example" == osx* ]] || [[ "$example" != *Example ]]
         then
             continue
         fi    
--- a/scripts/linux/testAllExamples.sh
+++ b/scripts/linux/testAllExamples.sh
@@ -5,7 +5,7 @@ cd ../../examples
 
 for category in $( ls . )
 do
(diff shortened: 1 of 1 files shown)


ofTheo:
thanks!! :)

PR #7160, opened 2022-11-11
Feature mt linux testallexamples

themancalledjakob:
add option to compile examples multithreaded


themancalledjakob:
whoops, some wrong commits sneaked in

PR #7164, opened 2022-11-15
ofMesh added more utility in removal functions (start/end index)

danoli3:
Changes to **ofMesh** 

Related to removal of vertices / normals / indices / textures (coords) 
Added new utility of new parameter remove functions to also have ability to remove between range of iterators from start to end.

Note: code is not index bound and uses iterators so > .size() is correct for end index 

Part of the evaluate and fix leaks PRS incoming

### DIFF
--- a/libs/openFrameworks/3d/ofMesh.h
+++ b/libs/openFrameworks/3d/ofMesh.h
@@ -196,6 +196,9 @@ public:
 	/// \brief Removes the vertex at the index in the vector.
 	void removeVertex(ofIndexType index);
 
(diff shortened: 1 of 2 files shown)


ofTheo:
Happy to merge this, I think maybe the naming should be tweaked though to removeNormals, removeIndices etc 

As we have: 
addVertex 
addVertices etc

danoli3:
Happy to make that change yep!

ofTheo:
Thanks @danoli3 !

PR #7166, opened 2022-11-16
ofxAssimpModelLoader - fix ambiguous load

### LINKED ISSUE #7165: [bug] ofxAssimpModelLoader has ambiguous load function
compile `3DModelLoaderExample` and get:
```bash
/home/jrkb/openFrameworks_master/examples/3d/3DModelLoaderExample/src/ofApp.cpp:96:28: error: call of overloaded ‘load(const char [12])’ is ambiguous
   96 |    model.load("penguin.3ds");
      |                            ^
In file included from /home/jrkb/openFrameworks_master/examples/3d/3DModelLoaderExample/src/ofApp.h:4,
                 from /home/jrkb/openFrameworks_master/examples/3d/3DModelLoaderExample/src/ofApp.cpp:1:
/home/jrkb/openFrameworks_master/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h:45:14: note: candidate: ‘bool ofxAssimpModelLoader::load(std::string, int)’
   45 |         bool load(std::string modelName, int assimpOptimizeFlags=OPTIMIZE_DEFAULT);
      |              ^~~~
In file included from /home/jrkb/openFrameworks_master/libs/openFrameworks/ofMain.h:6,
                 from /home/jrkb/openFrameworks_master/examples/3d/3DModelLoaderExample/src/ofApp.h:3,
                 from /home/jrkb/openFrameworks_master/examples/3d/3DModelLoaderExample/src/ofApp.cpp:1:
/home/jrkb/openFrameworks_master/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h:48:182: note: candidate: ‘bool ofxAssimpModelLoader::load(std::string, bool)’
   48 | DEPRECATED_MSG("ofxAssimpModelLoader::load(std::string modelName, bool bOptimize) is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(std::string modelName, bool optimize=false));
[...] following similar errors [...]

```
the issue is, that there are now multiple functions with name `load` accepting a single string as argument, since they have default values for both the `assimpOptimizeFlags int` and the `optimize bool`.
```cpp
// addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h

// the new function allowing to pass string and optional assimp flags
        bool load(std::string modelName, int assimpOptimizeFlags=OPTIMIZE_DEFAULT);

// the deprecated function allowing to pass string and optional boolean
        OF_DEPRECATED_MSG("...", bool load(std::string modelName, bool optimize=false));

```

Calling an assimpModelLoader.load("model.extension") will therefore fail, because it doesn't know which function to pick.

a possible solution would be to remove the default value from the deprecated function. Reasoning is, that the deprecated function should not be the default one, and at the same time old code using the deprecated function by specifically passing a boolean should still work... right?

here the possible fix:
```cpp
diff --git a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
index 00515df19..d24b264b1 100644
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
@@ -45,7 +45,7 @@ class ofxAssimpModelLoader{
         bool load(std::string modelName, int assimpOptimizeFlags=OPTIMIZE_DEFAULT);
         bool load(ofBuffer & buffer, int assimpOptimizeFlags=OPTIMIZE_DEFAULT, const char * extension="");
 
-        OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(std::string modelName, bool bOptimize) is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(std::string modelName, bool optimize=false));
+        OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(std::string modelName, bool bOptimize) is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(std::string modelName, bool optimize));
         OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(ofBuffer & buffer, bool optimize=false, const char * extension="") is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(ofBuffer & buffer, bool optimize=false, const char * extension=""));
 
         OF_DEPRECATED_MSG("ofxAssimpModelLoader::loadModel() is deprecated, use load() instead.", bool loadModel(std::string modelName, bool optimize=false));

```

Issue should be the same on all platforms/build targets.
Fix successfully tried on Linux Debian 11 with `libassimp-dev/stable,now 5.0.1~ds0-2 amd64` and should equally work on all other platforms/build targets as well.

PR with fix: #7166

themancalledjakob:
remove the default value from the deprecated function. Reasoning is, that the deprecated function should not be the default one, and at the same time old code using the deprecated function should still work.

Fix #7165

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
@@ -45,7 +45,7 @@ class ofxAssimpModelLoader{
         bool load(std::string modelName, int assimpOptimizeFlags=OPTIMIZE_DEFAULT);
         bool load(ofBuffer & buffer, int assimpOptimizeFlags=OPTIMIZE_DEFAULT, const char * extension="");
 
-        OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(std::string modelName, bool bOptimize) is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(std::string modelName, bool optimize=false));
+        OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(std::string modelName, bool bOptimize) is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(std::string modelName, bool optimize));
         OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(ofBuffer & buffer, bool optimize=false, const char * extension="") is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(ofBuffer & buffer, bool optimize=false, const char * extension=""));
 
         OF_DEPRECATED_MSG("ofxAssimpModelLoader::loadModel() is deprecated, use load() instead.", bool loadModel(std::string modelName, bool optimize=false));
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.h
@@ -46,7 +46,7 @@ class ofxAssimpModelLoader{
         bool load(ofBuffer & buffer, int assimpOptimizeFlags=OPTIMIZE_DEFAULT, const char * extension="");
 
         OF_DEPRECATED_MSG("ofxAssimpModelLoader::load(std::string modelName, bool bOptimize) is deprecated, use load(std::string modelName, int assimpOptimizeFlags) instead.", bool load(std::string modelName, bool optimize));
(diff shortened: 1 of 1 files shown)


themancalledjakob:
ha, of course!

ofTheo:
Thanks!!

Issue #7169, opened 2022-11-16
Header files missing in Windows for some first-party addons


c-mendoza:
It seems like the header files for ofxPoco and ofxOpenCv are not being downloaded when requested via `download_libs.ps1`. Not sure if this is a Windows-only issue.

The release version of OF does not have this problem, and is shipping with both the `include` and the `license` folders when appropriate.

ofTheo:
We switched the libs to being hosted on Github Releases - I am guessing those scripts would need to be updated with the correct url. See: https://github.com/openframeworks/openFrameworks/blob/master/scripts/dev/download_libs.sh#L38 

If you get it working could you submit a PR? 

Thanks so much!

ofTheo:
closes by #7173

PR #7173, opened 2022-11-17
Fix for Header files missing in Windows for some first-party addons (#7169)

c-mendoza:
Fix for #7169 

The download_libs.ps1 script was only copying the VS lib files of downloaded addon libs, which left out anything else, such as include directories. This fixes the script so that the complete library is copied.

It also updates the download URL to point to Github Releases.


ofTheo:
Thanks @c-mendoza !! 🙏

PR #7175, opened 2022-11-20
Emscripten video player fix

Jonathhhan:
This fixes ofVideoPlayer with Emscripten (and also the 3DPrimitivesExample patch, because it uses the videoGrabber).
Maybe allocate can be replaced with mallloc at some point?
https://github.com/emscripten-core/emscripten/issues/11547

The `assimpExample` patch works too, just not with the first model (which is the preset yet), because there are some shader issues which crashes the patch.
This `loadModel("Astroboy/astroBoy_walk.dae");` and the old man works without issues, the other two have issues, but still load. Did not change anything regarding the assimpExamples in this pull request, just wanted to mention it.

### DIFF
--- a/addons/ofxEmscripten/libs/html5video/lib/emscripten/library_html5video.js
+++ b/addons/ofxEmscripten/libs/html5video/lib/emscripten/library_html5video.js
@@ -120,7 +120,7 @@ var LibraryHTML5Video = {
     },
 
     html5video_player_pixel_format: function(id){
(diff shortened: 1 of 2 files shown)


ofTheo:
Thanks @Jonathhhan! 
Happy if you want to PR a malloc fix but I will merge this for now. 

Think maybe the assimp issue is the texture of the Fox is loaded from memory as opposed to the others which are from disk. 
If you see any relevant error messages for that def post an issue or to the Emscripten discussion: https://github.com/openframeworks/openFrameworks/discussions/7168

PR #6764, opened 2021-09-09
Update ofTrueTypeFont.cpp to make it work with Emscripten

Jonathhhan:
I made a pull request coming from this issue: https://github.com/openframeworks/openFrameworks/issues/6745
Not sure if my change has any drawbacks.

PR #7177, opened 2022-11-21
Emscripten true type font fix

Jonathhhan:
This pull request is related to this issue: https://github.com/openframeworks/openFrameworks/pull/6764
I seem to need it to make the fontExample work. Just wonder why it works for @themancalledjakob without this fix... 
https://dev.pointer.click/of/
So maybe I miss something, and this fix is not needed?


Jonathhhan:
Actually it would be nice not to merge this, and find out how it works without this pull request, because it does look much better without.

Jonathhhan:
I installed a clean Emscripten 3.1.26 (and not the audioWorklet branch) and now it works. So it seems the audioWorklet branch somehow breaks `FT_LOAD_FORCE_AUTOHINT`. Thats also the reason why it is working for @themancalledjakob. Close this for now. Should look into this again, if audioWorklets are a topic...

PR #7179, opened 2022-11-21
Emscripten_pointCloudExample fix

Jonathhhan:
Not ideal because the points are very small now, but it seems that `GL_POINT_SMOOTH` and `glPointSize()` do not exits in WebGL. https://stackoverflow.com/questions/66781512/why-doesnt-pointsize-default-to-a-standard-value-in-webgl
Same with the pointPickerExample.

### DIFF
--- a/examples/3d/pointCloudExample/src/ofApp.cpp
+++ b/examples/3d/pointCloudExample/src/ofApp.cpp
@@ -28,8 +28,8 @@ void ofApp::setup() {
 
 	ofEnableDepthTest();
 	#ifndef TARGET_EMSCRIPTEN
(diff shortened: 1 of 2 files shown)


ofTheo:
Ahh thanks @Jonathhhan - we could also make the example use rects instead of points. 
But will merge this for now.

Jonathhhan:
@ofTheo in theory it seems possible to use `glPointSize()` with GLES: https://registry.khronos.org/OpenGL-Refpages/es3.1/html/gl_PointSize.xhtml so it would be nice to make `glPointSize()` work with Emscripten and then to revert this commit.

PR #7181, opened 2022-11-21
Emscripten pointsAsTexturesExample fix

Jonathhhan:
Similar fix for the shaderExample... (While I wonder, if Emscripten not also adresses Open GL ES? But it works... - but maybe there is a better solution.)


ofTheo:
Thanks @Jonathhhan !!

PR #7182, opened 2022-11-22
Emscripten audioInputExample and opencvImageClassification fix

Jonathhhan:
If the videoGrabber is fixed the `opencvImageClassification` example changes are not needed anymore (the same for other examples that use ofVideoGrabber).


ofTheo:
I might wait on this one and see if we can fix the video grabber itself.

PR #7185, opened 2022-11-22
Emscripten meshFromCameraExample fix

Jonathhhan:

PR #7187, opened 2022-11-22
Revert "Emscripten meshFromCameraExample fix"

ofTheo:
Reverts openframeworks/openFrameworks#7185

@Jonathhhan 
Think actually we can't solve it this way as other platform grabbers will sometimes not return the requested size but the closest matching.  

I think we should probably figure out a way for emscripten to return the requested size immediately instead.


Jonathhhan:
@ofTheo alright, that would be much better.

Jonathhhan:
@ofTheo there is still the problem that the camera is not initialized on setup, but a little later. This affects some of the examples (like this one).
Maybe its normal for the web that it setups asynchronisly?

PR #7194, opened 2022-11-22
Emscripten billboard examples fix

Jonathhhan:
Just needed to change the header of the shader...


examples/gl/billboardExample/bin/data/shadersGLES3/Billboard.frag
```diff
@@ -0,0 +1,13 @@
+#version 300 es
+
```
ofTheo, line 2: Hey @Jonathhhan - Is this the only change needed in the shader?   
If so I might look at a core/code solution

Jonathhhan: @ofTheo yes, and `precision lowp float;` in the fragment shader.

ofTheo: Ahh okay - that does make it a bit trickier. 
Was thinking we could just do a header replace at the shader level for emscripten. 

the float thing is tricky as it prob isn't good to hide the precision from the user ( ie: replace `float` with `precision lowp  float'` for emscripten automatically )

Jonathhhan:
@ofTheo 
```
#ifdef GL_ES
// define default precision for float, vec, mat.
precision highp float;
#endif
```
This seems to work in the shader...

PR #7195, opened 2022-11-23
fix for emscripten not loading the first model

ofTheo:
the absolute path was breaking the file loading when the file didn't exist. 
this fixes it


Jonathhhan:
@ofTheo thanks. that works well now. only for model 2 and five i get this error from a shader:
```
[ error ] ofShader: GL_FRAGMENT_SHADER, offending line 1:
	    1	#extension GL_OES_standard_derivatives : enable
	    2	precision highp float;
	    3	#define IN varying
```
My problem is I can not find the shader, because it seems neither part of the example nor the addon.

PR #7196, opened 2022-11-23
fix models with normal maps not working in opengles

ofTheo:
models with normal maps weren't loading because `dFdx` and `dFdy` aren't defined by default in OpenGL ES. 
with this change you get much nicer rendering of the 2nd model. 

Screen Shot 2022-11-22 at 8 09 47 PM

### DIFF
--- a/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
@@ -2331,7 +2331,7 @@ static string defaultShaderHeader(string header, GLenum textureTarget, int major
 		ofStringReplace(header,"%extensions%","");
 	}
 #else 
-	ofStringReplace(header,"%extensions%","");
+	ofStringReplace(header,"%extensions%","#extension GL_OES_standard_derivatives : enable");
 #endif
 	if(textureTarget==GL_TEXTURE_2D){
 		header += "#define SAMPLER sampler2D\n";
--- a/libs/openFrameworks/gl/shaders/phong.frag
+++ b/libs/openFrameworks/gl/shaders/phong.frag
@@ -10,10 +10,6 @@ static const string fragmentShader = R"(
     IN vec4 v_color;
 #endif
 
-// needed for Perturb Normal on emscripten GLES
-#ifdef TARGET_OPENGLES
-    #extension GL_OES_standard_derivatives : enable
-#endif
 
     struct lightData
     {
--- a/libs/openFrameworks/gl/shaders/phong.frag
+++ b/libs/openFrameworks/gl/shaders/phong.frag
@@ -10,6 +10,10 @@ static const string fragmentShader = R"(
     IN vec4 v_color;
 #endif
 
+// needed for Perturb Normal on emscripten GLES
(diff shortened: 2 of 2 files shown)


Jonathhhan:
With this I get this error message:
```
[ error ] ofShader: GL_FRAGMENT_SHADER shader reports:
ERROR: 0:26: extension directive must occur before any non-preprocessor tokens in ESSL1

[ error ] ofShader: GL_FRAGMENT_SHADER, offending line 26:
	   24	// needed for Perturb Normal on emscripten GLES
	   25	#ifdef TARGET_OPENGLES
	   26	    #extension GL_OES_standard_derivatives : enable
	   27	#endif

```
The fox model works great now.

ofTheo:
Ahh thanks @Jonathhhan - yeah that makes sense where it is. 
I can fix it. 👍

ofTheo:
Repeating @Jonathhhan's report from #7195 

> @ofTheo thanks. that works well now. only for model 2 and 5 i get this error from a shader:
> 
> [ error ] ofShader: GL_FRAGMENT_SHADER, offending line 1:
> 	    1	#extension GL_OES_standard_derivatives : enable
> 	    2	precision highp float;
> 	    3	#define IN varying
> My problem is I can not find the shader, because it seems neither part of the example nor the addon.

Hmm, that is weird - I don't get the error on my system with 3.1.26. 
Maybe its more down to the browser / OS? 

@Jonathhhan - I'll see if there is a way to enable it if its available. Or better I'll remove it and find a non extension way to do `dFdx` `dFdx`

Jonathhhan:
This is what I get with Firefox (I got the above error with Chrome):
```
WebGL warning: linkProgram: Must have a compiled fragment shader attached:
SHADER_INFO_LOG:
WARNING: 0:1: 'GL_OES_standard_derivatives' : extension is not supported
ERROR: 0:82: 'dFdx' : no matching overloaded function found
ERROR: 0:82: '=' : dimension mismatch
ERROR: 0:82: '=' : cannot convert from 'const mediump float' to 'highp 3-component vector of float'
ERROR: 0:83: 'dFdy' : no matching overloaded function found
ERROR: 0:83: '=' : dimension mismatch
ERROR: 0:83: '=' : cannot convert from 'const mediump float' to 'highp 3-component vector of float'
ERROR: 0:84: 'dFdx' : no matching overloaded function found
ERROR: 0:84: '=' : dimension mismatch
ERROR: 0:84: '=' : cannot convert from 'const mediump float' to 'highp 2-component vector of float'
ERROR: 0:85: 'dFdy' : no matching overloaded function found
ERROR: 0:85: '=' : dimension mismatch
ERROR: 0:85: '=' : cannot convert from 'const mediump float' to 'highp 2-component vector of float'
```
Only tested with Linux (Ubuntu), but I already wonder why my browsers do not recognize the right GPU (testet with glInfoExample). With Firefox the renderer is GTX 980 and with Chrome the renderer is webkit WebGL. So maybe something with my settings or GPU drivers is wrong.

ofTheo:
Thanks!

I think I am going to just make a replacement function for those calls so we don't need to try and use the extension. 👍

PR #7198, opened 2022-11-23
Emscripten special keys fix

Jonathhhan:
This fixes special keys in Emscripten. Maybe I missed some special Keys, but most of them are working now. And maybe there is a better way to implement it (send the id directly to OF?).

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -30,7 +30,7 @@ void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
     EmscriptenWebGLContextAttributes attrs;
     emscripten_webgl_init_context_attributes(&attrs);
 
-/// when setting  explicitSwapControl to 0 it is emscripten that is in charge of swapping on each render call.
+/// when setting explicitSwapControl to 0 it is emscripten that is in charge of swapping on each render call.
     attrs.explicitSwapControl = 0;
     attrs.depth = 1;
     attrs.stencil = 1;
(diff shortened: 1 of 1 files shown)


Jonathhhan:
The check failed, after I added the comment. can remove it again (would be nice to set `emscripten_set_main_loop_timing()` from `ofApp.cpp`, but its also fine to set it in `ofxEmscriptenWindow.cpp`). Normally I leave it at 2, because 30 FPS seems most of the time enough and it makes audio stuff glitch less(compared to rendering every screen refresh(with request animation frame))...

ofTheo:
Awesome - thanks @Jonathhhan 
Is this good to go?

Jonathhhan:
@ofTheo thanks no drawbacks or issues so far, just not sure if the code can be simplified...

ofTheo:
Hey @Jonathhhan - it might be that you could make a local function for mapping from Emscripten keys to OF keys so that you don't have to do everything twice? 

Also good to check this for the full list of special keys we map:
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/events/ofEvents.h#L37-L91

addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
```diff
@@ -98,17 +99,107 @@ void ofxAppEmscriptenWindow::display_cb(){
...
-	if(key==0){
+	std::string id = keyEvent->key;
+	if(key == 0){
 		key = keyEvent->which + 32;
+	}else if(id == "ArrowLeft"){
+		key = 57356;
+	}else if(id == "ArrowUp"){
+		key = 57357;
+	}else if(id == "ArrowRight"){
+		key = 57358;
```
ofTheo, line 110: Might be good to use the ofKey defines for these? 
eg: `key = OF_KEY_RIGHT`

Jonathhhan: Definitely...

Jonathhhan: @ofTheo i think its better now and can be merged.

PR #7199, opened 2022-11-23
modelNoiseExample fix

Jonathhhan:
Fixes the desktop version for me, too.


ofTheo:
Thanks @Jonathhhan 
That is from fixing an old bug with the assimp model loader - so was broken on all platforms 👍

PR #7201, opened 2022-11-23
Emscripten slitscanRadialClockExample fix

Jonathhhan:


ofTheo:
I think a better fix for this one might be to wrap the whole slit scan section with a check or pixels.size() so that way the slit scan doesn't update until the first frame comes in:

```
    if( pixels.size() ){ 
        if (ofGetSeconds() > seconds){ // one second has elapsed
```

Including the section where the texture gets updated at the end of ofApp::update

Jonathhhan:
@ofTheo that makes sense...

ofTheo:
Thanks @Jonathhhan !!

PR #7203, opened 2022-11-24
Emscripten polylineBlobsExample fix

Jonathhhan:


Jonathhhan:
@ofTheo Thanks for merging. Actually not sure about the difference between initGrabber and setup. It seems that setup has an additional argument for using textures. So all examples I changed too initGrabber can also use setup, it should not make a difference and is up to you. I like the setup somehow better, and in a way it would be consistent with ofVideoPlayer if ther is a setUseTexture instead of the third argument from setup (but thats not rteally important, and maybe its good like it is).

Jonathhhan:
I can also make a PR, that reverts all ofVideoGrabber.setup()s/.initGrabber()s to the original methods(how it was before my changes), like it was intended.

Jonathhhan:
@ofTheo now I checked it. initGrabber() is deprecated:
```
		OF_DEPRECATED_MSG("Use setup instead",bool initGrabber(int w, int h){return setup(w,h);})
		OF_DEPRECATED_MSG("Use setup instead",bool initGrabber(int w, int h, bool bTexture));
```
And there is setUseTexture() method for the ofVideoGrabber.
So best would be, to change all `initGrabber()` to `setup()` ion the examples?

ofTheo:
hi @Jonathhhan - yes thanks! 
that would be helpful!! :)

PR #7204, opened 2022-11-24
Emscripten 07_fboAlphaMask fix

Jonathhhan:


ofTheo:
This might be a cleaner way to go and not need the #ifdef  

```
                auto& tex = maskFbo.getTexture();
		for( auto& tc : texCoords ){
                    tc = maskFbo.getTexture().getCoordFromPercent(tc.x, (1.0-tc.y));
		}
```

Jonathhhan:
Very good :)

PR #7206, opened 2022-11-25
Emscripten ofxVectorGraphics fix

Jonathhhan:
Makes the vectorGraphicsExample work with Emscripten...
And I actually wonder, why some of the GL methods are not recognized by Emscripten, while their OF counterpart is (like `glBegin(GL_LINE_STRIP)` and `ofMesh.SetMode(OF_PRIMITIVE_LINE_STRIP)`. It would be much better to make them work in Emscripten too, instead of changing examples and other addons because of that.


ofTheo:
👍

PR #7208, opened 2022-11-25
Emscripten multiTexture3dExample fix

Jonathhhan:
I added mainly the header of the shader (and some explicit float declarations).
Works not with all browsers / GPUs.
And I guess the rotation of the sphere is not totally correct, but better than before (the Y-axis was flipped).


ofTheo:
Thanks @Jonathhhan - going to test this first before merging!

ofTheo:
Thanks - this looks great! 👍

PR #7210, opened 2022-11-26
Fix style in example methods

TobiasGrothmann:
I just created a new project and got a little annoyed (not for the first time) with a single rogue space in the template:
```c++
mouseMoved(int x, int y )
```

I had a look at the [style guides](https://github.com/openframeworks/openFrameworks/wiki/oF-code-style) and removed *some* of the extra spaces in all templates and examples.
I say *some* because there still is a lot of inconsistency. I just wanted to get rid of this mouseMoved space that has haunted me so many times now ;)


TobiasGrothmann:
I rebased this to the newest master (3f457231d)

I know it's minor, but how is this still an issue after 3 years 🙈

PR #7211, opened 2022-11-26
Emscripten glInfoExampleFix

Jonathhhan:
The advanced version would be to use the Emscripten filesystem and download the glInfo as a .txt file (right now it just prints).


Jonathhhan:
I need to add this `#import <GL/glew.h>` to `ofConstants.h` to make it work.

PR #7214, opened 2022-11-28
Emscripten videoPlayerExample fix

Jonathhhan:


ofTheo:
@Jonathhhan - I think we shouldn't merge this one as it might break other platforms. 
using isFrameNew for drawing could mean flashing frames if the video framerate is lower than the app frame rate.  

We should probably just make things work if the camera or video hasn't uploaded the texture yet. Either with a dummy texture or track down the issue with returning an unallocated texture that breaks emscripten but not other platforms. 

( hope that makes sense 🙂 )

Jonathhhan:
@ofTheo yes, of course. That sounds good.

PR #7215, opened 2022-11-29
Emscripten isFrameNew()

Jonathhhan:
Since `isFrameNew()` does not work in Emscripten I decided to return true, so it does not break apps that were made with `isFrameNew()`. Instead `ofxEmscriptenVideoGrabber::isInitialized()` returns `html5video_grabber_ready_state(id)>=HAVE_ENOUGH_DATA` (like `ofxEmscriptenVideoPlayer::isLoaded()`).
I also changed all `initGrabber()` that I found to `setup()`.

### DIFF
--- a/addons/ofxEmscripten/src/ofxEmscriptenVideoGrabber.cpp
+++ b/addons/ofxEmscripten/src/ofxEmscriptenVideoGrabber.cpp
@@ -59,7 +59,7 @@ bool ofxEmscriptenVideoGrabber::setup(int w, int h){
 }
 
 bool ofxEmscriptenVideoGrabber::isInitialized() const{
-	return texture.isAllocated();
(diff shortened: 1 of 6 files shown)


ofTheo:
Thanks! I'm going to try and see if I can get isFrameNew working properly - if I get a fix I'll share it here so you can update and then I can just merge this all in.

Jonathhhan:
Here is some about new frame events, but it does not look too promising: 
https://stackoverflow.com/questions/68402476/event-for-every-frame-of-html-video
A solution would be, if that would be published: https://wicg.github.io/video-rvfc/

Jonathhhan:
Changed it, because I had some issues with examples that use `ofVideoPlayer()`.

PR #7216, opened 2022-11-30
One audioContext

Jonathhhan:
This pull request is more like a suggestion, because I am not sure if I removed too much.
There is now only one audioContext, one ffft, one stream and one mediaElement intead of arrays.
My guess is, that more is not needed. And it seems to work quite well (without issues) so far.


Jonathhhan:
Closed this in favor of: https://github.com/openframeworks/openFrameworks/pull/7217

PR #7217, opened 2022-11-30
Emscripten audio video revision

Jonathhhan:
This is basically what I changed end of last year.
I compared it with my recent changes and it seems more complete (but I changed a lot).
I can also split it up into smaller parts.

### DIFF
--- a/addons/ofxEmscripten/libs/html5audio/include/html5audio.h
+++ b/addons/ofxEmscripten/libs/html5audio/include/html5audio.h
@@ -2,29 +2,32 @@
 
 extern "C"{
 	typedef void (*html5audio_stream_callback)(int bufferSize, int inputChannels, int outputChannels, void * userData);
-	
+
 	extern void html5audio_list_devices();
 	extern int html5audio_context_create();
-	extern int html5audio_context_start(int context);
-	extern int html5audio_context_stop(int context);
-	extern int html5audio_context_samplerate(int context);
-	extern void html5audio_context_spectrum(int context, int bands, float * spectrum);
+	extern int html5audio_context_start();
+	extern int html5audio_context_stop();
+	extern int html5audio_context_samplerate();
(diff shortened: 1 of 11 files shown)


Jonathhhan:
Its not really a pull request, its more like a suggestion. It can be surely further optimized, but works quite well.

Jonathhhan:
Let my try to explain what the main changes are:

Audio:
There is only one audioContext().
There is only one fft.
There is only one stream.
There is only one media element.
Added soundPans: [].
There are no soundBuffers: [], no soundSources: [], no soundStartTimes: [], and no soundGains: [].

Video:
Added soundSources: [], added soundPans: []. This has the advantage that its possible to pass the audio from the video into OF for further processing, which is also possible with the stream or the soundplayer (at least with audioworklets ;) ).

Please tell me, if I reduced too much, but until now everything seems to work as before.

ofTheo:
@Jonathhhan thanks for this!
From a user perspective is everything the same? Are there things that used to work that won't work with this update? 
Or does this PR just add new features or provide more options?

Jonathhhan:
@ofTheo the idea is, to remove redundant stuff, fix some stuff and add some features.
From a user perspective everything is the same (+ some fixes and features).

ofTheo:
Okay great @Jonathhhan 👍 - if you want to just let me know when this is good to merge. 
I think currently there is a conflict in ofxEmscriptenVideoGrabber.cpp

Jonathhhan:
@ofTheo thank you for merging this. Actually I wanted to triple check it, to be sure there are no errors. But I checked it once again yesterday, and did not found anything so far. Maybe its better like that, instead of waiting for too long.

ofTheo:
Ahh yes! Was wondering if it was done. 
It could always be patched if anything comes up and probably good to get it in for testing too. 
🙂

PR #7229, opened 2022-12-03
List devices 2

Jonathhhan:
@ofTheo maybe I found with `EM_ASYNC_JS` a good solution for async loading with Emscripten. Its now in the .cpp files, because if its in the .js files the method needs to be added to  `-s ASYNCIFY_IMPORTS=html5video_list_devices`. If its in the cpp file called with `EM_ASYNC_JS` I only need to add the `-s ASYNCIFY=1` flag.
The problem now is, that I cannot change the return type of ofxEmscriptenVideoGrabber::listDevices() (same for the sound player). The type is `vector<ofVideoDevice> / vector<ofAudioDevice>` and I have a `vector<std::string>`. I also tried to make an additional method `ofxEmscriptenVideoGrabber::listDevices2() ` which works, but it needs ofxEmscriptenVideoGrabber to be defined and it breaks apps that are not made for Emscripten. Maybe the best solution would be to construct an `ofVideoDevice` instance from the string information? Then the return type can stay the same and it does break any apps?
And somehow I realized that I need `ALLOC_STACK -1` for returning a string correctly, maybe its for the other cases (get pixel format for example) also true.
`return allocate(intArrayFromString(string, ALLOC_STACK -1));`

(I guess, `EM_ASYNC_JS` makes also loading local files much easier...)

### DIFF
--- a/addons/ofxEmscripten/libs/html5audio/include/html5audio.h
+++ b/addons/ofxEmscripten/libs/html5audio/include/html5audio.h
@@ -2,8 +2,7 @@
 
 extern "C"{
 	typedef void (*html5audio_stream_callback)(int bufferSize, int inputChannels, int outputChannels, void * userData);
-	
-	extern void html5audio_list_devices();
+
 	extern int html5audio_context_create();
 	extern int html5audio_context_start(int context);
 	extern int html5audio_context_stop(int context);
--- a/addons/ofxEmscripten/libs/html5audio/lib/emscripten/library_html5audio.js
+++ b/addons/ofxEmscripten/libs/html5audio/lib/emscripten/library_html5audio.js
@@ -27,26 +27,6 @@ var LibraryHTML5Audio = {
     	}
     },
 
-    html5audio_list_devices: function(){
-        if (!navigator.mediaDevices.enumerateDevices) {
-            console.log("enumerateDevices() not supported.");
-        } else {
-            // List cameras and microphones.
-            navigator.mediaDevices.enumerateDevices()
-            .then((devices) => {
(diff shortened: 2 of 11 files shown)


Jonathhhan:
Ha. it actually just works for the video at the moment...

Jonathhhan:
Thats why I set the third argument of `emscripten_set_main_loop` to 0:
https://github.com/emscripten-core/emscripten/issues/16071

Jonathhhan:
> @Jonathhhan I think this looks good so far. Ideally you could break up the info from the string and then populate the ofSoundDevice fields ( like id, type, name etc ).
> 
> Sort of similar to:
> 
> https://github.com/openframeworks/openFrameworks/blob/e06be9a4c61a0bb7298e17bf36f9e4f4954e76c8/libs/openFrameworks/sound/ofRtAudioSoundStream.cpp#L95-L104

I tried that, but it did not work (probably because I was doing something wrong).
A thing is also, the devices have only 2 properties that are relevant: id and name.
And it is seperated between input and output devices, so the webcam is an input device too...
I think it is not that important to have it as an ofSoundDevice, but it would be nice to return the values name and id somehow.
That is how an input property looks like:

```
InputDeviceInfo
deviceId
: 
"2020e024037a4efe6e7f17fc5268e80b3f5d568e4b063a80b4c5f3affcd5a696"
groupId
: 
"bcd27dc100290d2b90c65ab8f62809583e80e4b06bbe4ceeceafb8e6ee486fd1"
kind
: 
"audioinput"
label
: 
"USB Audio Digital Stereo (IEC958)"
```

PR #7230, opened 2022-12-05
VS2019 Fixes : move some settings from emptyExample into props, add missing files

elliotwoods:


elliotwoods:
There are 4 commits mentioned but only the last one has any effect
Following on a bit from conversation in https://forum.openframeworks.cc/t/static-linking-not-building-needed-boost-version-in-vs-2017/32009/25

ofTheo:
Thanks @elliotwoods - looks great! 👍

PR #7232, opened 2022-12-06
ofMaterial PBR advanced example

NickHardeman:
Set the shader main functions for the materials without recompiling, PBR material with textures and clear coat. Cube map and a light with shadows.

materialPBRAdvancedExample


NickHardeman:
> Hey @NickHardeman thanks for all this great work! I'm having an issue on latest, on both examples :
> 
> ```
> Unknown type name 'ofCubeMap'
> ```

Hi @dimitre, what platform are you on? Did you use the project generator?

dimitre:
Yes macOS, command line project generator.

dimitre:
it runs if I add this to ofMain.h
#include "ofCubeMap.h"

NickHardeman:
Do you know what the command line pg compiles against? does it use the make files? 
I updated the openFrameworksCompiled/project/osx/openFrameworksLib.xcodeproj. Is there something else I need to update? I don't have much experience with the pg on the command line
Also, thanks for checking it out!

dimitre:
I think the electron app invokes the command line, so it should be the same output.

NickHardeman:
Thank you for testing and catching this. Submitted a pr that should fix it #7233

PR #7238, opened 2022-12-09
Emscripten mousedragged buttons

Jonathhhan:


ofTheo:
I think maybe simpler is to leave it how it was and then change the `0` ( last argument ) to `mouseEvent->button` 

eg:
```
if(ofGetMousePressed()){
 		instance->events().notifyMouseDragged(mouseEvent->targetX * (canvas_width / css_width), mouseEvent->targetY * (canvas_height / css_height),0);
```
to

```
if(ofGetMousePressed()){
 		instance->events().notifyMouseDragged(mouseEvent->targetX * (canvas_width / css_width), mouseEvent->targetY * (canvas_height / css_height), mouseEvent->button);
```

Let me know if that works or if you see possible issues with that. 

Thanks!! :)

Jonathhhan:
@ofTheo I tried that first. No idea why `mouseEvent->button` does not work there. But it would be much nicer.

PR #7241, opened 2022-12-09
Emscripten wheel event

Jonathhhan:
Not really sure, how to implement this right (but it outputs the y wheel value - not sure what the x wheel and z wheel value is). So this is more a suggestion than a pr...  y wheel value is -53/53 with chrome and -103/103 with firefox, so something with the scaling is not correct yet ( x wheel is -0/-0, z wheel 0/0).
Here is some information about the wheel event: https://emscripten.org/docs/api_reference/html5.h.html

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -310,7 +310,7 @@ int ofxAppEmscriptenWindow::mousemoved_cb(int eventType, const EmscriptenMouseEv
 }
 
 int ofxAppEmscriptenWindow::mousescrolled_cb(int eventType, const EmscriptenWheelEvent *wheelEvent, void *userData){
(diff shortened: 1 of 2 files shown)


ofTheo:
I think this could be good! 
I believe the scroll value is relative, so if going slowly you get smaller numbers and going quickly you get larger that is correct. The differences might be due to differences in polling rate ( so bigger changes are reported but less frequently ).

Jonathhhan:
Actually in this Emscripten pr the wheel value is divided throgh 100 too, which seem not to be unproblematic, but better than without...
https://github.com/emscripten-core/emscripten/pull/7968

ofTheo:
Awesome thanks @Jonathhhan !

Jonathhhan:
@ofTheo actually I am not sure if I did that right, because in the mentioned pr are two options:
`delta = event.wheelDelta / 120;` for mousewheel event.
And `delta /= 100;` for wheel event.
On the other hand it worked quite well, while testing.

PR #7244, opened 2022-12-11
Emscripten gamepad event class draft

Jonathhhan:
I tried to make a class for gamepad events, maybe its better than to include everything in `ofxEmscriptenWindow`.
Just add #include `ofxEmscriptenGamePad.h` to ofApp.h and call `gamePadData()` in `ofDraw()` for updating the gamepad events. Just need to find a way how to return the data, maybe someone has an idea?
This is why I closed (there is some information about gamepad events): https://github.com/openframeworks/openFrameworks/pull/7243
What do you think about that approach @ofTheo?


Jonathhhan:
Another idea is to make one class (`ofxEmscriptenEvents` for example) that holds all events, that are not connectet with `ofBaseEvents`. This way, I would know how to pass the values to `ofApp.cpp`...

danoli3 (approved):
Wicked ! Polling for Game Pads is per frame so should be Defined if needed

PR #7255, opened 2022-12-15
Emscripten replace allocate and some small fixes...

Jonathhhan:
Replaces allocate, which makes the flag `-s EXPORTED_RUNTIME_METHODS=allocate` redundant.
Fixes one bug that I introduced in the htmlaudio lib.
And one change in `ofSoundBaseTypes` for printing the devices only once.


ofTheo:
good to merge @Jonathhhan ?

Jonathhhan:
@ofTheo yes, tested it and did not experience any errors...

Jonathhhan:
(of course I made all the changes as good as I can, I just think sometimes that I am not experienced enough to find the best solution - still no errors found until now)

PR #7259, opened 2022-12-19
Emscripten WebMidi draft

Jonathhhan:
Its quite minimalistic, but seems to work well.
Here is an example patch:
https://github.com/Jonathhhan/ofEmscriptenExamples/blob/main/webmidi_draft/src/ofApp.cpp

### DIFF
--- a/addons/ofxEmscripten/libs/webMidi/library_webMidi.js
+++ b/addons/ofxEmscripten/libs/webMidi/library_webMidi.js
@@ -1,0 +1,70 @@
+var webMidi = null;
+var midiIn = null;
+var midiOut = null;
+
+function getMidiDevices() {
+	var midiInDevices = "";
+	var midiOutDevices = "";
+	function onMIDISuccess(midiAccess) {
+		webMidi = midiAccess;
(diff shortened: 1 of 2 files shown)


Jonathhhan:
@ofTheo you need MIDI devices to use it, similar to `ofxMidi`. Something like this does work , of course: http://coolsoft.altervista.org/en/virtualmidisynth
Buit since MIDI is not a core feature, this is not a core feature too (if I remember correctly, there was a discussion about making `ofxMidi` a core feature).

danoli3 (approved):
Amazing!

danomatika:
Why wouldn't you merge this into ofxMIDI? There is are base classes for input and output which can be subclasses for the various MIDI backends already, then the same project could work with MIDI hardware and virtual ports on desktop as well as WebMIDI in the browser.

Jonathhhan:
> Why wouldn't you merge this into ofxMIDI? There are base classes for input and output which can be subclasses for the various MIDI backends already, then the same project could work with MIDI hardware and virtual ports on desktop as well as WebMIDI in the browser.

Thats actually a very good idea.

Issue #7268, opened 2023-01-05
C++20 support (Linux)


Tomaqa:
I'm missing support for c++20 standard which I use in my project.
I verified that it requires only slight changes, and it can also resolve some existing issues such as #5641 or #7054.
Hopefully my pull request will follow.

Tomaqa:
Not sure I'm following.
The PR explicitly resolves this particular issue: https://eel.is/c++draft/diff.cpp17#class-2.
Moreover, the changed code is still completely fine for C++17 compilers, but now also for C++20 compilers. I do not see any violation of backward-compatibility.

I honestly do not see any downside of the proposed PR, except of that developers will no longer be allowed to use deprecated syntax (like templated destructors) ...

PR #7269, opened 2023-01-05
maintained C++20 compatibility for Linux (#7268)

Tomaqa:
I wonder why the fixed destructors relate specifically to the c++20 standard, they seem wrong in general to me


ofTheo:
Thanks @Tomaqa - do most of the examples build and run for you without issue with these changes?

Tomaqa:
Unfortunately I failed to compile even the `graphics/polygonExample` from the [install guide](https://openframeworks.cc/setup/linux-install/) (after using `projectGenerator/commandLine`), which unrelated to C++20.
It says `ofConstants.h: No such file or directory`.
I did not manage to find out how to resolve it...

Issue #7270, opened 2023-01-05
Do not require to run Linux `install_depencencies.sh` scripts as root


Tomaqa:
For security reasons, I think that running **whole** script as root is highly dangerous, or at least, may seem like that for the user.

Consider running only particular commands *inside* the scripts (e.g. `pacman`) with sudo explicitly.

PR #7271, opened 2023-01-05
used sudo inside install_dependencies scripts instead of requiring to…

Tomaqa:
… run the whole script as root; only for archlinux, el6 and fedora (#7270)


ofTheo:
Thanks! 
Adding to 0.12.1.  As there is some chance to break things with this change and also we'd probably need to do the same for Ubuntu too.

PR #7275, opened 2023-01-19
Emscripten float textures

Jonathhhan:
This makes float textures work with Emscripten.

### DIFF
--- a/libs/openFrameworks/gl/ofGLUtils.cpp
+++ b/libs/openFrameworks/gl/ofGLUtils.cpp
@@ -275,7 +275,7 @@ int ofGetGLTypeFromInternal(int glInternalFormat){
 		break;
 #endif
 
-#ifndef TARGET_OPENGLES
+#if not defined TARGET_OPENGLES || defined TARGET_EMSCRIPTEN
 		case GL_LUMINANCE32F_ARB:
 		case GL_LUMINANCE_ALPHA32F_ARB:
 		case GL_R32F:
--- a/libs/openFrameworks/gl/ofGLUtils.cpp
+++ b/libs/openFrameworks/gl/ofGLUtils.cpp
@@ -152,7 +152,7 @@ string ofGetGLInternalFormatName(int glInternalFormat) {
 int ofGetGLFormatFromInternal(int glInternalFormat){
 	switch(glInternalFormat) {
 			case GL_RGBA:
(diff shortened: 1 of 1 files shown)


ofTheo:
Thanks @Jonathhhan! 
( cc @NickHardeman - not sure if you needed float textures for some of the emscripten fixes you were doing )

Jonathhhan:
@ofTheo this enables also allocating fbos with GL_RGBA32F, I needed that for a case where I need pixel accuracy in a shader.
First I wanted to use GL_RGB32F which I am using for desktop, but somehow only GL_RGBA32F works for Emscripten (I guess, thats changeble with the right settings).

ofTheo:
Awesome!
Thanks @Jonathhhan 👍

Jonathhhan:
@ofTheo maybe they work for GLES in general (if so, my change needs to be adjusted)?

NickHardeman:
@Jonathhhan float fbo textures is exactly what I need right now. Any ideas about how to get GL_RGB32F?

Jonathhhan:
@NickHardeman sadly not, I tried it, but without luck (which is probably because my lack of knowledge). Actually the flag for RGB32F seems to be set, if I am not wrong, but without effect.

Jonathhhan:
@NickHardeman @ofTheo Maybe you just need to change line 173 to: `#if not defined TARGET_OPENGLES || defined TARGET_EMSCRIPTEN`?
I will try that this weekend.

NickHardeman:
Hi @Jonathhhan, I've tried a bunch of combinations and tried just using straight gl calls to create a frame buffer and attach a texture. That also only works with _GL_RGBA32F_. According to this document, seems like only _GL_GRBA32F_ is required for both render buffers and textures. And _GL_RGB32F_ is only required for textures.
https://www.khronos.org/opengl/wiki/Image_Format

Jonathhhan:
Hi @NickHardeman. Thanks, then it seems not to be available. But I think the only downside is that RGBA32F takes a quarter more space than RGB32F?

NickHardeman:
Yes agreed. What's kind of odd is that RGBA, RG, RED are supported. 🤷

Jonathhhan:
And if it works for GLES in general (did not test that), you can just leave the `#if not defined TARGET_OPENGLES || defined TARGET_EMSCRIPTEN` ifdefs that I made away...

Issue #7280, opened 2023-01-26
ios sound player and soundstream not working


fred-dev:
This is also a forum post (https://forum.openframeworks.cc/t/nightlies-and-ios-sound-input/41221) 

but maybe it is more relevant here. Sound input for iOS is not currently working.

When the `bool ofxiOSSoundStream::setup(const ofSoundStreamSettings & settings)`
function runs it has an error:

`Thread 1: EXC_BAD_ACCESS (code=1, address=0x464630b96110)`
On this line:

       ` [(__bridge SoundInputStream *)soundInputStream start];`

Trying to trace these changes back, it seems maybe they came from this commit:

https://github.com/openframeworks/openFrameworks/pull/6889

EDIT: After some more checking, the sound player also does not work, it exits on another bridged function in load:

`bool ofxiOSSoundPlayer::load(const std::filesystem::path& fileName, bool stream) {
    if(soundPlayer != NULL) {
        unload();
    }

    string filePath = ofToDataPath(fileName);
    soundPlayer = (__bridge void *)[[AVSoundPlayer alloc] init];
    BOOL bOk = [(__bridge AVSoundPlayer *)soundPlayer loadWithPath:[NSString stringWithUTF8String:filePath.c_str()]];
    
    return bOk;
}
`

Giving the EXC_BAD_ACCESS error on this line:
`BOOL bOk = [(__bridge AVSoundPlayer *)soundPlayer loadWithPath:[NSString stringWithUTF8String:filePath.c_str()]];
`

artificiel:
just fooling around with the 12-gauge in the dark as I don't have an iOS dev setup under hand: what happens if you try `__bridge_retained`:

`soundPlayer = (__bridge_retained void *)[[AVSoundPlayer alloc] init];
`

fred-dev:
Thanks for the suggestion, but that doesnt do it. 
`Cast of C pointer type 'void *' to Objective-C pointer type 'AVSoundPlayer *' cannot use __bridge_retained`
I also tried __bridge_transfer, "Moves a Core Foundation pointer to Objective-C with transfer of the ownership to ARC" but this gives the same error. 

I tried these with both the soundstream and sound player classes to no avail.

artificiel:
[...]

PR #7283, opened 2023-01-29
Fix ARC problem of ofSoundStream 

2bbb:
fix about memory management for `void *` with `__bridge_retained`, `__bridge_transfer`

related: #7280

### DIFF
--- a/addons/ofxiOS/src/sound/ofxiOSSoundPlayer.mm
+++ b/addons/ofxiOS/src/sound/ofxiOSSoundPlayer.mm
@@ -25,15 +25,19 @@ bool ofxiOSSoundPlayer::load(const std::filesystem::path& fileName, bool stream)
     }
 
     string filePath = ofToDataPath(fileName);
-    soundPlayer = (__bridge void *)[[AVSoundPlayer alloc] init];
+    __autoreleasing AVSoundPlayer *player = [[AVSoundPlayer alloc] init];;
     BOOL bOk = [(__bridge AVSoundPlayer *)soundPlayer loadWithPath:[NSString stringWithUTF8String:filePath.c_str()]];
-    
+    if(bOk) {
+        soundPlayer = (__bridge_retained void *)player;
+    }
+
     return bOk;
 }
 
 void ofxiOSSoundPlayer::unload() {
     if(soundPlayer != NULL) {
-        [(__bridge AVSoundPlayer *)soundPlayer unloadSound];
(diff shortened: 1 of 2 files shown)


artificiel:
hello! testing against nightly 20230129 with macOS13.1 / iOS 15.3.1:

- "soundPlayerExample" works for AVSoundPlayer but not ofSoundPlayer. no errors or warnings, just no sound on "synth" and "beats". (FYI ofSoundPlayer does work with the crude `_retained` attempt in #7280);

 - "audioOutputExample" works; 

 - "audioInputExample" works on the input side (but not the ofSoundPlayer part).

dimitre:
@2bbb I've added a commit to your PR, not sure if it is the right thing to do.
soundplayer object was loading audio file, and then being assigned to empty player object.
with this soundPlayerExample will be able to load and play all three audio files.

dimitre:
@artificiel can you please test it again pulling this PR? Now it is working for me. for microphone usage you can use this other PR too https://github.com/openframeworks/openFrameworks/pull/7287

artificiel:
I confirm the 3 audio files are playing in soundPlayerExample!

However there is still a strangeness with audioInputExample: the sound does not come in ("buffers received" stay at 0), unless the line `sound.load("sounds/beat.caf");` is commented out, at which point the sound comes in.

so something in loading the sound breaks the incoming audio stream? I tried moving the sound load from setup() in touchDown() like this:
```C++
void ofApp::touchDown(ofTouchEventArgs & touch){
    sound.setVolume(1.0);
    sound.load("sounds/beat.caf");
    sound.setLoop(true);
    sound.play();
}
```
and the microphone runs until the touch, at which point it stops streaming.

artificiel:
a bit more investigation: if the ofSoundStreamSetup() comes after the sound.load(), the microphone "wins" (audio input / no sound player). so it seems there is a conflict in the AVSession setups -- probably unrelated to the memory management, so perhaps this should be merged, and figure out the conflict in another PR?

dimitre:
I'm testing this example a little bit more and modernize some things there (ofSoundStream object, audioIn) and it makes difference depending of how object you initialize first. if I load the sound before the mic I can still hear the sound but with a very low loudness. 
Just wondering if iOS try to avoid feedback if recording and playing throught the speaker at the same time.
If I use headphones it seems the example works ok (loading sound first)

dimitre:
Hey @ofTheo I think this can be merged as it is now

PR #7293, opened 2023-02-02
Av engine sound player -redo

ofTheo:
Re-opening #6728 
Basically needs ARC fixes before we can merge this.


ofTheo:
Adding @artificiel and @2bbb as I think we are in need of some ARC fixes to get this merged / working. 

Incase its helpful when ofxSyphon moved to ARC they used this approach to avoid the void * definition in the CPP header file: https://github.com/astellato/ofxSyphon/pull/52/files#diff-44275e1a507b5ce7b49b5e24d5b86fee282f3787a0e9eb7777458792b36460aa

artificiel:
looking at the log, the ARC stuff is not worrying but this error grabs the eye:

```
error: 'AVAudioSession' is unavailable: not available on macOS
```
is it a case of missed conditional that should lead to use something else on macOS? or is there an expectation of a form of wrapper that maps the class onto something else?

just to make sure what is known/unknown/designed here before digging deeper.

NB apple mentions catalyst but not macOS
image

ofTheo:
Ahh yes - you're right @artificiel I think @danoli3 added that in the last commit. 
We'll probably need to #ifdef that out for macOS.  

I think at some point I got that working but then got stumped on the right way to convert the code to ARC friendly.

artificiel:
my approach to ARC is less subtle than @2bbb as they master __autoreleasing (I get away with plain __bridge and __bridge_retained but maybe that's not complete in terms of lifecycle vs the C++ allocations) so I'll defer on this!

ofTheo:
okay I think I got this all working with ARC and macOS. 

Here is the diff for the ARC changes - it seems to work well on macOS. 
https://github.com/openframeworks/openFrameworks/pull/7293/commits/4dc561a0f9844b4db2d995baa7010e33f2549286?w=1

@danoli3 @2bbb @artificiel - would be happy to have some eyes on this. 

It's not enabled by default - but you can make it default by uncommenting this line:
https://github.com/openframeworks/openFrameworks/blob/4dc561a0f9844b4db2d995baa7010e33f2549286/libs/openFrameworks/utils/ofConstants.h#L386-L387

artificiel:
also successfully tested: switching whilst playing, between internal speaker, 1/8 headphones, and USB2 interface (through clicking in the Sound Settings panel). the drum does not skip a beat.

ofTheo:
Thanks @artificiel! 

Yeah I got a lot of those warnings with other variables which were fixed by changing `_bIsPlaying` to `self.bIsPlaying` except that one errored when doing that and it wasn't clear what the right way to fix it was. 

@NickHardeman 

Thanks for catching that - should be fixed now!

ofTheo:
So I had to do it via:
`__typeof(self) __weak weak_self = self;` 

There is still the one instanced variable warning - but I can't make it a weak ref due to it not being accessed via self. 

I think this could be good to merge for now and we can test in the nightlies ( it is not enabled by default ).

artificiel:
+1 nightlies

dimitre:
Yess push the button

PR #7294, opened 2023-02-02
using const & on setSoundStream

dimitre:
so shared_ptr don't need to be incremented and decremented on the function


ofTheo:
Hey @dimitre, I think this change would  stop you being able to do: 

```
myStream.setSoundStream(make_shared<MyCustomSoundStream>()); 
```

I think the extra reference dereference wouldn't be as a much a hit as losing the functionality. 

( if my assumption is wrong - def re-open )

Issue #7291, opened 2023-01-31
iOS example audioInputExample: issue with audioIn / ofSoundPlayer order


dimitre:
a bit more investigation: if the ofSoundStreamSetup() comes after the sound.load(), the microphone "wins" (audio input / no sound player). so it seems there is a conflict in the AVSession setups -- probably unrelated to the memory management, so perhaps this should be merged, and figure out the conflict in another PR?

_Originally posted by @artificiel in https://github.com/openframeworks/openFrameworks/issues/7283#issuecomment-1409743155_

It seems something in soundPlayer changes the way ofSoundStream behaves.

artificiel:
indeed, the sound does work out of the headphones... probably related to the way each class interacts with `[AVAudioSession sharedInstance]`... 

https://github.com/openframeworks/openFrameworks/blob/625df269a462dcbc988ca0bcd90aafd0b66767bf/addons/ofxiOS/src/sound/SoundStream.m#L77

vs

https://github.com/openframeworks/openFrameworks/blob/625df269a462dcbc988ca0bcd90aafd0b66767bf/addons/ofxiOS/src/sound/AVSoundPlayer.m#L27

I tried adding `AVAudioSessionCategoryOptionDefaultToSpeaker` to both, with no difference. dealing with AVAudioSession and automatic audio routing seems tricky (lots of not-clearly-answered questions on stackoverflow). also no effect with:

```objc
[audioSession overrideOutputAudioPort:AVAudioSessionPortOverrideSpeaker
                                error:&audioSessionError];
```

the fact that the order of calls bet wound sound.load() and ofSoundStreamSetup() changes the behaviour feels like a bug, but is it really? or is it simply the result of the complicated (and potentially evolving) iOS behaviour for which it is difficult to provide an homogeneous OF-style interface? 

I guess there is a problem if OF wants to expose all the iOS routing possibilities (ex: airplay). as such, it would be good to be able to switch the sound routing from an OF interface. but maybe it's beyond the scope of OF, and it's a more of a matter of providing (and explaining) good default behaviours? and the tiny fraction of iOS OF projects that need more than simple playback or audio reactiveness will have to dig a bit into the sound classes to tweak them to taste?

danoli3:
Need to respond to the errors and reconfigure on changes

artificiel:
that's ideal, but as of right now ofSoundPlayer and ofSoundStream are mutually incompatible on iOS, due to their design/implementation.

[...]

PR #7301, opened 2023-02-06
make ofSoundStream systemSoundStream able to lazy initialize

2bbb:
related to #7298

### DIFF
--- a/libs/openFrameworks/sound/ofSoundStream.cpp
+++ b/libs/openFrameworks/sound/ofSoundStream.cpp
@@ -21,7 +21,10 @@
 #endif
 
 namespace{
-    ofSoundStream systemSoundStream;
+    ofSoundStream &getSystemSoundStream() {
+        static ofSoundStream _;
+        return _;
+    }
 }
 
 using std::shared_ptr;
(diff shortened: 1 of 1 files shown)


dimitre:
Thank you @2bbb 
I personally think it is great and we should merge it.

artificiel:
that's perfect for the hidden/builtin systemSoundStream.

but how to we "orient" the usage? is it logical to have a strategy using a builtin ofSoundStream singleton configured with global ofSoundStream* calls (which is probably good considering the hardware mapping), but at the same time a strategy using self-allocated instances? in other words, for what reason would someone declare an ofSoundStream in .h?

ofTheo:
>  in other words, for what reason would someone declare an ofSoundStream in .h?

@artificiel I think maybe if they want to use multiple streaming apis ( via setSoundStream ) or if they want to dynamically open and close input devices separately from output devices? 

I agree though the whole pattern is not very OF. 🙂

I am not opposed to this PR though as it gets rid of an extra unneeded initialization. 
Happy to merge unless there is a strong reason not to.

artificiel:
if the decision is to maintain the "global" systemSoundStream, yes merging this is definitely a gain! (although I still wonder if the design of this automatic instance is wanted, vs declaring an explicit ofSoundStream yourself in ofApp.h?)

I'm in support of removing lines of user code (which the global does) but then, for the global approach to make sense, the examples should :

 - focus the ofSoundStream usage on that "global" systemSoundStream with the ofSoundStream*() calls (could add an ofSoundStreamGetSystem() that returns a reference to the instance for eventual finer-grain tweaking);
 
- not provide precedent of declaring additional ofSoundStreams in ofApp.h -- such advanced use has a lot of moving parts and should be left to the discretion of the users that need to "go there";

 - not mix ofSoundPlayer and ofSoundStream (in order to circumvent #7291) -- and/or insert warnings about the ordering of calls, but it seems complicated to provide straightforward recommendations (or there are incompatibilities within the 2 classes, causing inconstant routing behaviour, at least on iOS) -- also, unless #6728 fixes those problems.

I can work on that (including mac, iOS and linux tests), but this effort should also take into account #6728 -- so everything gets tweaked coherently?

PR #7304, opened 2023-02-07
Github Actions : checkout v3, ubuntu 20.04

dimitre:


dimitre:
I've tried to update android to ubuntu 20.04, but it seems it is failing to install gcc-6
Is there some alternative to this @danoli3 ?

dimitre:
Actions are not working in specific days, due to the fact Ubuntu 18.04 is being phased out
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

This PR updates everything possible except from Android which needs gcc-6 and script fails on 20.04.

Issue #7306, opened 2023-02-08
transform feedback example emscripten


Jonathhhan:
I tried to adapt the shaders. It compiles fine, but then I get this error: `WebGL: INVALID_OPERATION: bindBufferBase: transform feedback is active`
Here is a Github issue about that: https://stackoverflow.com/questions/12017175/what-can-cause-gldrawarrays-to-generate-a-gl-invalid-operation-error

Issue #7242, opened 2022-12-09
ofShader header replacement for glsl version


ofTheo:
There are quite a few shader examples which would work on GLSL / Emscripten with just a different glsl versions header. 

For example the multiTexture3dExample just needs these two lines for OpenGL ES, but currently it means we have to have two different shader folders for the different platforms: 

https://github.com/openframeworks/openFrameworks/blob/master/examples/gl/multiTexture3dExample/bin/data/shadersGLES/shader.frag#L1-L3

```
#version 300 es
precision lowp float;
```

A possible better solution would be to have a shader header macro which we parse out ( like the includes ) and auto generate the header based on the Major and Minor versions. 

eg:

```
OF_GLSL_SHADER_HEADER

// we are going to use 2 lights
// we also have 2 ofLights in ofApp
#define MAX_LIGHTS 2

uniform sampler2D mapDiffuse;
uniform sampler2D mapNormal;
..........
```

With the changes below to ofShader 

```
	// look for OF_GLSL_SHADER_HEADER header placeholder 
	// this will swap the glsl version based on the OpenGL version set in main.cpp 
	// note this won't always work, but is handy for switching between compatible versions of GLSL based on the system 
	if( ofIsStringInString(source.source, "OF_GLSL_SHADER_HEADER") ){
		auto glRenderer = dynamic_pointer_cast<ofBaseGLRenderer>(ofGetCurrentRenderer());
		if( glRenderer ){
			string header = "#version " + ofGLSLVersionFromGL(glRenderer->getGLVersionMajor(), glRenderer->getGLVersionMinor());
			#ifdef TARGET_OPENGLES 
				//TODO: we should have a getGLSLExtensions for OpenGLES 
				header += "\n#extension GL_OES_standard_derivatives : enable";
				header += "\nprecision mediump float;";
			#endif 
			ofStringReplace(source.source, "OF_GLSL_SHADER_HEADER", header);
		}
	}
```

and an updated 

```

string ofGLSLVersionFromGL(int major, int minor){
#ifdef TARGET_OPENGLES
	if(major>=2){
		return "300 es";
	}else{
		return "ES1";
	}
#else
	switch(major){
	case 3:
		if(minor==0){
			return "130";
		}else if(minor==1){
			return "140";
		}else if(minor==2){
			return "150";
		}else{
			return ofToString(major*100+minor*10);
		}
	case 4:
		return ofToString(major*100+minor*10);
	default:
		return "120";
	}
#endif
}
[...]

PR #7309, opened 2023-02-08
Transform feedback animated example emscripten (gles)

Jonathhhan:
Not the most elegant ifdefs (maybe someone can make them prettier...), but it works well...
Basically I just changed the shader headers and the .glsl to .vert and added an empty .frag, because it seems to be needed.

And I wonder why the much simpler tranformFeedbackExample returns an error when running (and shows a black screen): https://github.com/openframeworks/openFrameworks/issues/7306

### DIFF
--- a/examples/gl/transformFeedbackAnimatedExample/bin/data/GLES/pointSprite.frag
+++ b/examples/gl/transformFeedbackAnimatedExample/bin/data/GLES/pointSprite.frag
@@ -1,0 +1,18 @@
+#version 300 es
+
+precision lowp float;
+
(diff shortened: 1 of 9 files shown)


ofTheo:
Looking at the file difference they are almost the same and make me think we should just get this implemented so we don't need two versions: https://github.com/openframeworks/openFrameworks/issues/7242

Jonathhhan:
@ofTheo the only other thing that seems to be important for GLES is, that there needs to be a .frag file, even if it is empty except the headers...

PR #7313, opened 2023-02-08
ofxAssimpModelLoader - animation speed now working and OF_LOOP_PALINDROME

dimitre:


ofTheo:
Awesome! 
Only thing I think might be needed is to change speedFactor back to 1.0 on reload/reset?

dimitre:
Sure! I'll update soon

PR #7314, opened 2023-02-08
pixelBufferExample Emscripten

Jonathhhan:
This also demonstrates how to download files with Emscripten...


ofTheo:
@Jonathhhan - thanks for this! 

I think it would actually maybe be a better approach to work the file saving code into ofSaveImage etc
Is this tricky to do?

Jonathhhan:
@ofTheo that would be much better :) Not sure how tricky it is, but I guess it should be doable...

PR #7323, opened 2023-02-10
Fixes for programmable renderer and open gl es

NickHardeman:
Updates multiTextureShaderExample.


ofTheo:
Wondering if we can just make both targets use non ARB textures?  
Then the difference will be just the header right?

NickHardeman:
Yeah would make things more straight forward, but thought it would be helpful to show how to deal with ARB as well.

NickHardeman:
Converted to non ARB textures and added the OF glsl header token.

PR #7326, opened 2023-02-12
Cairo functionality to addon -> ofxCairo

dimitre:
This PR moves code and libraries to an addon. 
Basically moved everything cairo to ofxCairo

This can serve as a discussion right now and it needs another modification on installing libraries, changing path from libs to addons/ofxCairo/libs


dimitre:
from the code it seems Cairo was never used with emscripten, here in the ofMain.h
Yes it would be great if you can make it work.
```c++
#if !defined( TARGET_OF_IOS ) & !defined(TARGET_ANDROID) & !defined(TARGET_EMSCRIPTEN)
	#include "ofCairoRenderer.h"
	#include "ofGraphicsCairo.h"
#endif
```

Jonathhhan:
@dimitre the difficult part is (for me) to write the shell script for compiling (I never did that before), but I will give it another try.

dimitre:
@ofTheo please share your thoughts on this one. 
If it is welcome I can work in two more details (template and script) to complete this.

dimitre:
@ofTheo I understand your latest comment was meant to this PR. so cc @ofZach to know his opinion on this one.

ofTheo:
Thanks @dimitre - moved it here 🙂:

@dimitre I will defer to others who use Cairo more ( eg @ofZach  ), but I have a feeling this would be quite a breaking change for a lot of projects/examples and maybe even lesson plans. 

I do like that we can now compile OF without cairo but moving core functionality to an addon might break too many things.

ofZach:
maybe can you describe the motivation for moving Cairo to an addon?  I think it's generally not something I use too often but I do find pdf / svg export from OF helpful and dip into Cairo for 2d line rendering from time to time, especially with students who are used to how nice thick lines can look in canvas.

dimitre:
I just love Cairo and use from time to time. For me it is like ofxAssimpModelLoader. It is great functionality but doesn't need to be compiled in every OF program, being a Core Addon instead. 
related: https://github.com/openframeworks/openFrameworks/discussions/7034

Maybe it is more the idea of moving some parts from the core to addons to make a modular OF 
as discussed previously in forums and OF conference.

ofZach:
ah that makes sense!   I think in that context it seems directionally right in intent and part of a larger discussion of making OF more modular and lighter in the core (but also worth discussing as part of a larger plan).    I think even seeing that alternative renderers are possible (by making it more of an add-on) may be helpful.

ofTheo:
Yeah, I think if we starting fresh this would absolutely be the way to go. 

I think more of the issue is how to remove features like this in a way that is not painful to users. 

It might be that we could try and put out a 0.12.0 soon or LTS/Stable ( version 🙂 ) and start working on a more minimal and modular OF in a separate branch.  But would be great to have a discussion / plan in place first.

dimitre:
Yes we definetly need a release. 0.11.3 or 0.12. 
I'm afraid 0.12 todo list still needs some months of work

Jonathhhan:
> Yes we definetly need a release. 0.11.3 or 0.12. I'm afraid 0.12 todo list still needs some months of work

I am also for a minor release if it decreases the release pressure. Or 0.12 and leave the things that are still on the list for 0.12.1...

dimitre:
PR got messed up over time ~ conflicts.
I'll be closing this one as it will be easier to start again if this functionality is desired.

PR #7328, opened 2023-02-13
ofThreads Emscripten

Jonathhhan:
This PR enables `ofThreads` for Emscripten. The three examples are also fixed.
Only thing is, at least the `tess2` and `FreeImage` libs need to be compiled with pthreads / atomics support too (I guess all libs that are used with `ofThread` need `pthread` support).
Which is already working, if I set the flag in `apothecary`. But I do not know how to automate that...

The `threadedImageLoaderExample` does not load the image from the web, also not with the desktop version (but I guess that it is a CORS-policy issue with Emscripten).

Here a PR to make the libs multithreaded (except FreeImage):
https://github.com/Jonathhhan/apothecary/tree/MultiThread2

I also had to make some small changes in opencv, I hope it does not break anything: https://github.com/Jonathhhan/opencv/tree/Emscripten


Jonathhhan:
Tested it a little more and everything `ofThread` related seems to work quite well.

Jonathhhan:
@ofTheo I solved the apothecary issues https://github.com/Jonathhhan/apothecary/tree/MultiThread2, so both PRs together should make multithreading with Emscripten work.

And I rrealized one interestiing thing: Building openCV only works as it is up to 3.1. 21 and it does not work with 3.1.24 anymore (did not test the versions between). I guess this is because of a llvm update, but not sure: https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md
They are also working on updating openCV.
But the libs compiled with 3.1.21 can still be used with the current Emscripten.

ofTheo:
Cool thanks @Jonathhhan I'll merge the apothecary PR first then. 
In terms of OpenCV as long as we stay with the current setup in apothecary it is all good for now?

Jonathhhan:
> In terms of OpenCV as long as we stay with the current setup in apothecary it is all good for now?

yes.

ofTheo:
🎉

PR #7330, opened 2023-02-14
Add a default header token for glsl shaders closes #7242

### LINKED ISSUE #7242: ofShader header replacement for glsl version

ofTheo:
Useful for having examples on Desktop and Emscripten / Web use the same shader files with minimal changes. 
Uses medium precision for emscripten and defines TARGET_OPENGLES. 

Once this is merged we could update: 

tagging @Jonathhhan @NickHardeman 
#7309 and #7323 should be cleaner with this approach 

closes #7242


Jonathhhan:
@ofTheo thanks. That works great.
I only have to enable the extension before specifying float / int:
```
            if( versionStr != "ES1" ){
                header += "#extension GL_OES_standard_derivatives : enable\n";
            }
            header += "precision mediump float;\n";
            header += "precision mediump int;\n";
```
Updated the animated feedback example and had to change a few things, but much better now.

ofTheo:
Thanks @Jonathhhan 

I switched the header order around. 

I left the extension in as I have seen other docs saying ES 2 is needed for it: 
https://registry.khronos.org/OpenGL/extensions/OES/OES_standard_derivatives.txt 

And I remember our CI had issues before with the standard derivates not being defined without it.

PR #7331, opened 2023-02-15
TransformFeedbackAnimatedExample

Jonathhhan:
Updated with the OF_GLSL_SHADER_HEADER flag. Should work for a few other examples, too.


ofTheo:
Awesome - thanks!!

PR #7341, opened 2023-02-17
replaces all GLES shaders from shader examples with the OF_GLSL_SHADER_HEADER flag

Jonathhhan:
I needed to do some additional changes like `ofDisableArbTex();` for desktop apps that use textures, so it is more a proposal. But tested it twice with Emscripten and Ubuntu.


ofTheo:
Awesome! 

I was thinking about doing something like the core shader replacement where we define `SAMPLER` as `texture` or `texture2DRect` but I think doing that risks making our shaders less and less glsl. 

I think this change looks good. 

@dimitre @NickHardeman ?

dimitre:
I think this change looks good too!

ofTheo:
@Jonathhhan would love to merge this. 
Do we still need the ofDisableArb changes for the relevant examples?

NickHardeman:
The OF_GLSL_SHADER_HEADER makes the shader code a lot cleaner. @Jonathhhan  Thank you for updating the examples.
Just tested on OSX found the following:

- 04_simpleTexturing: openGL2 ->shaders convert sampler2DRect > sampler2D, needs uniform vec2 resolution

- 05_alphaMasking: openGL2 -> shaders convert sampler2DRect > sampler2D

- 06_multiTexture: openGL2 -> shaders convert sampler2DRect > sampler2D

- 07_fboAlphaMask: openGL2 -> shaders need OF_GLSL_SHADER_HEADER, convert sampler2DRect > sampler2D

- 08_displacementMap: openGL2 -> shaders need OF_GLSL_SHADER_HEADER, convert sampler2DRect > sampler2D

- 09_gaussianBlurFilter: openGL2 -> shaders need OF_GLSL_SHADER_HEADER, convert sampler2DRect > sampler2D, need uniform texwidth and texheight and divide by those uniforms

Jonathhhan:
@NickHardeman good point. I only updated the GLES / openGL3 shader, did not think about openGL2...

Jonathhhan:
> @Jonathhhan Do we still need the ofDisableArb changes for the relevant examples?

@ofTheo I guess, it is needed without further changes and I am not sure how to avoid it. Maybe someone else?

Jonathhhan:
@NickHardeman I had another idea, that is only to use `ofDisableArbTex` with the programmable renderer:
```
	if(ofIsGLProgrammableRenderer()){
		ofDisableArbTex();
		shader.load("shadersGL3/shader");
	}else{
		shader.load("shadersGL2/shader");
	}
```
Not sure if it makes sense and if it works for you. I am also open for changing the GL2 shaders, but it did not really work for me.

NickHardeman:
Hi @Jonathhhan, will `settings.glesVersion=2;` enable the programmable renderer on iOS and Android as well as Emscripten?

Jonathhhan:
> will `settings.glesVersion=2;` enable the programmable renderer on iOS and Android as well as Emscripten?

@NickHardeman I would guess so, but I do not have the setup to test it at the moment.

NickHardeman:
@Jonathhhan ah me neither. Unsure if iOS supports arb textures. I can help with the updates of the GL2 shaders.  
Not sure about the best approach for helping... if all the GL3 shaders are ready to go. Maybe it can be merged and then I can send a separate PR for the GL2 shaders?

Jonathhhan:
@NickHardeman the GL3 shaders are ready to go, GL2 should work as before...

PR #7342, opened 2023-02-17
Emscripten mouse / touch event improvement

Jonathhhan:
This PR enables mouse / touch events outside of the canvas. Maybe needs to be tested a little more, because it does not behave exactly like the desktop app. But it solves a problem that was bugging me for some time.
Maybe it would be the best solution to limit the mousePressed and mouseMoved events to the canvas?
I already tried it with an if condition:
```
int ofxAppEmscriptenWindow::mousedown_cb(int eventType, const EmscriptenMouseEvent *mouseEvent, void *userData){
float x = EM_ASM_INT(return canvas.getBoundingClientRect().left);
float y = EM_ASM_INT(return canvas.getBoundingClientRect().top);
if(mouseEvent->targetX - x >= 0 && mouseEvent->targetX - x <= EM_ASM_INT(return canvas.width) && mouseEvent->targetY - y >= 0 && mouseEvent->targetY - y <= EM_ASM_INT(return canvas.height)){
	instance->events().notifyMousePressed(mouseEvent->targetX - EM_ASM_INT(return canvas.getBoundingClientRect().left), mouseEvent->targetY - EM_ASM_INT(return canvas.getBoundingClientRect().top), mouseEvent->button);
	return 0;
	}
}
```
But somehow that does not work...

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -345,6 +345,7 @@ int ofxAppEmscriptenWindow::mouseleave_cb(int eventType, const EmscriptenMouseEv
 	int canvasWidth, canvasHeight;
 	emscripten_get_canvas_element_size("#canvas", &canvasWidth, &canvasHeight);
         double cssWidth, cssHeight;
+	emscripten_get_element_css_size("#canvas", &cssWidth, &cssHeight);
 	instance->events().notifyMouseExited(mouseX * canvasWidth / cssWidth, mouseY * canvasHeight / cssHeight);
 	return 0;
 }
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -47,8 +47,8 @@ void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
     _renderer = make_shared<ofGLProgrammableRenderer>(this);
     ((ofGLProgrammableRenderer*)_renderer.get())->setup(2,0);
 
-    emscripten_set_keydown_callback("#canvas",this,1,&keydown_cb);
-    emscripten_set_keyup_callback("#canvas",this,1,&keyup_cb);
+    emscripten_set_keydown_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW,this,1,&keydown_cb);
+    emscripten_set_keyup_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW,this,1,&keyup_cb);
     
     emscripten_set_mousedown_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW,this,1,&mousedown_cb);
     emscripten_set_mouseup_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW,this,1,&mouseup_cb);
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -12,68 +12,68 @@
 
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


ofTheo:
Amazing! 

I noticed a lot of the camera drag to pan around stuff doesn't work currently on mobile. Does this fix it? 

I think maybe mouse pressed / released should be limited to if it is in the viewport?
This would be similar to how desktop apps work ( mouse move is regardless of if in app window. press/drag/release is only if in the window/viewport ).

Jonathhhan:
I updated the PR. There is only one bug with mouse button 2. It does not capture outside of the window. But its already stuck in the Emscripten lib `library_html5.js`:
```
  // Copies mouse event data from the given JS mouse event 'e' to the specified Emscripten mouse event structure in the HEAP.
  // eventStruct: the structure to populate.
  // e: The JS mouse event to read data from.
  // target: Specifies a target DOM element that will be used as the reference to populate targetX and targetY parameters.
  $fillMouseEventData__deps: ['$JSEvents', '$getBoundingClientRect', '$specialHTMLTargets'],
  $fillMouseEventData: function(eventStruct, e, target) {
#if ASSERTIONS
    assert(eventStruct % 4 == 0);
#endif
    {{{ makeSetValue('eventStruct', C_STRUCTS.EmscriptenMouseEvent.timestamp, 'e.timeStamp', 'double') }}};
    var idx = eventStruct >> 2;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.screenX / 4 }}}] = e.screenX;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.screenY / 4 }}}] = e.screenY;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.clientX / 4 }}}] = e.clientX;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.clientY / 4 }}}] = e.clientY;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.ctrlKey / 4 }}}] = e.ctrlKey;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.shiftKey / 4 }}}] = e.shiftKey;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.altKey / 4 }}}] = e.altKey;
    HEAP32[idx + {{{ C_STRUCTS.EmscriptenMouseEvent.metaKey / 4 }}}] = e.metaKey;
    HEAP16[idx*2 + {{{ C_STRUCTS.EmscriptenMouseEvent.button / 2 }}}] = e.button;
    HEAP16[idx*2 + {{{ C_STRUCTS.EmscriptenMouseEvent.buttons / 2 }}}] = e.buttons;
```

Jonathhhan:
So it seems normal JS behaviour, that the right mouse button behaves differently: https://stackoverflow.com/questions/42336777/right-click-drag
Maybe there is a workaround, but this is as good as I can get it.
Right now mouse pressed and moved is limited to the viewport. I like if mouse released works outside of the viewport, it is for example helpful if I drag a slider and release it outside of the window (otherwise the slider is still dragged, if the mouse is released outside of the window).

ofTheo:
Cool!
I'm happy if you want to make the mouseReleased outside viewport work. 
Otherwise will merge this as is. 🙂

Jonathhhan:
> I'm happy if you want to make the mouseReleased outside viewport work.

Thanks. Its already like that, so I would say that its fine to merge.

Jonathhhan:
@ofTheo maybe it would make sense not to limit mouseMoved and mousePressed by default, but let the user choose and ifdef if necessary (and wanted)? Really not sure about the best behaviour...

ofTheo:
@Jonathhhan - I think if in doubt we should emulate the desktop behavior. 

eg: 
- mouseMove not being constrained and goes negative left of the view / above the view 
- mousePressed only inside view 
- mouseReleased inside and outside view ( checked and desktop does this ) 
- mouseDragged constrained to view

Jonathhhan:
@ofTheo alright. I guess, now it is like the desktop behaviour...

PR #7344, opened 2023-02-18
General changes (namespace, deprecated loadFile function)

dimitre:
Changes in some more files 
This is meant to remove unrelated changes in this big PR - https://github.com/openframeworks/openFrameworks/pull/7110

### DIFF
--- a/addons/ofxGui/src/ofxSliderGroup.cpp
+++ b/addons/ofxGui/src/ofxSliderGroup.cpp
@@ -1,5 +1,4 @@
 #include "ofxSliderGroup.h"
-using namespace std;
 
 template<class VecType>
 ofxVecSlider_<VecType>::ofxVecSlider_(ofParameter<VecType> value, float width, float height){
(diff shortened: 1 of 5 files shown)


addons/ofxGui/src/ofxSliderGroup.cpp
```diff
@@ -130,7 +129,7 @@ ofxColorSlider_<ColorType> * ofxColorSlider_<ColorType>::setup(ofParameter<ofCol
     ofxGuiGroup::setup(value.getName(), "", 0, 0);
     parameters.clear();
 
-	const string names[4] = {"r", "g", "b", "a"};
+	const std::tring names[4] = {"r", "g", "b", "a"};
```
ofTheo, line 132: think this might be a typo? 
tring -> string

ofTheo: good to merge once this is fixed :)

PR #7357, opened 2023-02-27
ofMainloop cleanup

dimitre:
removing ofConstants.h, moving to cpp file
moving logic to cpp file
removing ofConstants.h from ofWindowSettings.h, not needed there but needed here.
Untangling includes
some minor format fixes.

### DIFF
--- a/libs/openFrameworks/app/ofMainLoop.cpp
+++ b/libs/openFrameworks/app/ofMainLoop.cpp
@@ -5,10 +5,12 @@
  *      Author: arturo
  */
 
-#include <ofMainLoop.h>
+#include "ofMainLoop.h"
 #include "ofWindowSettings.h"
 #include "ofAppBaseWindow.h"
 #include "ofBaseApp.h"
+// #include "ofConstants.h"
+
 
 //========================================================================
 // default windowing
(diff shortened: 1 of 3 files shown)


dimitre:
@artificiel please share your thoughts on this one.
the plan is minimize including ofConstants.h in other .h files in small PRs so they are readable.

artificiel (commented):
good for the cleanup;

in terms of style if you want to modernize all-at-once:

```c++
	for(auto i: windowsApps){
 		std::shared_ptr<ofAppBaseWindow> window = i.first;		 		std::shared_ptr<ofAppBaseWindow> window = i.first;
 		std::shared_ptr<ofBaseApp> app = i.second;		 		std::shared_ptr<ofBaseApp> app = i.second;
```
can become:
```c++
	for (const auto &[window, app] : windowsApps){
```
(same with the other loop a bit below)

also instead of the laborious sequential double-negative :
```c++
		if(window == nullptr){
			continue;
		}
		if(app == nullptr){
			continue;
		}
```
can be more expressive as:
```c++
if (window && app) { ... }
```

dimitre:
Yes great love all the suggestions. 
the first one has the benefit of not increasing / decreasing shared_ptr count. Is this compatible with c++11 ?
Maybe this suggestions fits well in a separate PR, so it is less things to read @ofTheo?

artificiel:
[] structured bindings are c++17 (i realize i work in '17 but: is OF tied to '11? (with filesystem etc it seems 17 is a baseline?))

artificiel:
also since a while `of.cxxFlags: ['-std=c++17' ]` seems required by QtCreator  : https://forum.openframeworks.cc/t/undefined-reference-to-oftodatapath/41467/2

dimitre:
I think it is compatible with c++11 so it runs on OS older than macos 10.15.
for filesystem it uses c++11 experimental and boost filesystem too.
Maybe something to drop in the future?

artificiel:
ah yes i remember at least with Xcode C++11 mentions a warning that bindings are C++17 but honors them.
that being said, it works as well with .first and .second — my comment is more about simplifying with auto whenever it's possible.

PR #7363, opened 2023-03-03
audioWorklets proposal 2

Jonathhhan:
This time only one Emscripten file is changed: https://github.com/emscripten-core/emscripten/compare/main...Jonathhhan:emscripten:audioWorkletOF_3.1.32-2
Maybe it is possible to replace that file automatically from the OF side?
It still can be optimized, but I think it is on the right track (and it works quite well)...
Here an example: https://audioplayer.handmadeproductions.de/


ofTheo:
Could you use the last argument UserData for the inputs outputs etc? 

ie:

```
struct AudioStruct{
  int numInputs = 0; 
  int numOutputs = 0; 
  ofBuffer inBuffer; 
  ofBuffer outBuffer; 
}

```
// This function will be called for every fixed 128 samples of audio to be processed.
EM_BOOL ProcessAudio(int numInputs, const AudioSampleFrame *inputs, int numOutputs, AudioSampleFrame *outputs, int numParams, const AudioParamFrame *params, void *userData)
{
#ifdef REPORT_RESULT
	++audioProcessedCount;
#endif

AudioStruct * audioData = (AudioStruct *) userData; //everything you need in the struct 
audioData->outBuffer = ..... 
```

Then when you start it you pass in the data as the last argument like: 

```
        audioData.inputs = 2; 
        audioData.outputs = 0; 
	emscripten_start_wasm_audio_worklet_thread_async(context, wasmAudioWorkletStack, sizeof(wasmAudioWorkletStack), WebAudioWorkletThreadInitialized, &audioData);
```

For example?

Jonathhhan:
@ofTheo thanks a lot, that looks promising and I will try that.

ofTheo:
@Jonathhhan awesome! Glad to hear! 👍

Jonathhhan:
There is a recent Emscripten changes, which makes it more clear which userData flows where: 
https://github.com/emscripten-core/emscripten/issues/18989

But I still did not found a way to pass the data where I want.
> However the userData pointer that is passed to a call to emscripten_start_wasm_audio_worklet_thread_async() is not the userData pointer that will flow in to ProcessAudio().

Jonathhhan:
Closed because of a new draft: https://github.com/openframeworks/openFrameworks/pull/7727

PR #7383, opened 2023-03-15
More of::filesystem work

dimitre:


dimitre:
@NickHardeman this PR changes some things in ofCubeMap
Basically I've ported path operations from string to filesystem::path 
Take a look when you have the time
Thanks

dimitre:
@ofTheo can you double check this one when you have time? Thanks

dimitre:
Oww nice catch. I've made a mistake with filenames there.
Already fixed in new commit. I've added the missing parameter too.

NickHardeman:
The caching path for the irradiance map and the prefiltered map seems like it's getting merged; outputting 
_**modern_buildings_2_1k_irr_32.exr_pre_128.exr**_

In ofCubeMap.cpp, line 403;
`cacheIrrName += ("_pre_"+ofToString(data->settings.preFilterRes,0)+".exr");`
should be 
`cachePrefilterName += ("_pre_"+ofToString(data->settings.preFilterRes,0)+".exr");`

dimitre:
oh thanks @NickHardeman 😓 ctrl-c ctrl-v issues
it should be fixed now

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -1854,10 +1849,9 @@ string ofFilePath::getCurrentExePath(){
 }
 
 //------------------------------------------------------------------------------------------------------------
-// MARK: - near future
-//of::filesystem::path ofFilePath::getCurrentExeDir(){
-std::string ofFilePath::getCurrentExeDir(){
-	return ofFilePath::getEnclosingDirectory(ofFilePath::getCurrentExePath(), false);
+of::filesystem::path ofFilePath::getCurrentExeDir(){
```
NickHardeman, line 1852: Not sure if this will break some previous projects because it no longer outputs a string with a trailing slash, ie:
( just showing the end part of the path, though it they both use absolute paths).
`materialPBR/bin/materialPBRDebug.app/Contents/MacOS/`
vs
`materialPBR/bin/materialPBRDebug.app/Contents/MacOS"`

NickHardeman (commented):
I can't really speak to the PG changes but left some comments in a few other places. It's a big change going from string file system functions to path. Just want to be thorough for legacy support.

libs/openFrameworks/sound/ofOpenALSoundPlayer.cpp
```diff
@@ -457,7 +457,7 @@ bool ofOpenALSoundPlayer::stream(const of::filesystem::path& fileName, std::vect
 
 bool ofOpenALSoundPlayer::readFile(const of::filesystem::path& fileName, std::vector<short> & buffer){
 #ifdef OF_USING_MPG123
-	if(ofFilePath::getFileExt(fileName)!="mp3" && ofFilePath::getFileExt(fileName)!="MP3"){
+	if(ofToLower(fileName.extension().string())==".mp3"){
```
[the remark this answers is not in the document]
dimitre: Thanks. PG changes aren't really part of this PR, it is just the fact I've checked out a more recent version of PG and it updates submodules on this commit.
Well thanks for the reading, another lack of attention on this detail. already fixed.

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -1695,24 +1691,23 @@ std::string ofFilePath::addTrailingSlash(const of::filesystem::path & _path){
...
 	}
-//	return path;
 	return path.string();
 #endif
 }
 
 
 //------------------------------------------------------------------------------------------------------------
+// FIXME: - deprecate and suggest using of::filesystem::path.extension() instead
 string ofFilePath::getFileExt(const of::filesystem::path& filename){
```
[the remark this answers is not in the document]
dimitre: Yes agreed, removed the comment. 
I've noticed api names are different ofFilePath::getFileExt and ofFile::getExtension
it would be great to deprecate over time things that are covered by std::filesystem, so there are less things to mantain and to document too :)

libs/openFrameworks/utils/ofFileUtils.cpp
dimitre: Good point. we can append preferred_separator at the end. 
and revert back to string. I thought it was more used as an internal core function.

NickHardeman: @dimitre Not sure how frequently it is used by externally. I don't think I have ever used it, but I just noticed that it was different and didn't want it break any previous projects :)

dimitre: I was curious myself too and I've seen it is used in different projects
https://github.com/search?q=ofFilePath%3A%3AgetCurrentExeDir%28%29&type=code

NickHardeman (approved):
I ran the following functions using the previous file system approach and using this PR and it all output the same values. Not sure if there are any other edge cases, maybe with ofSystem. But as far as I can tell it renders the same functionality.
```
cout << "ofFilePath:: ext: " << ofFilePath::getFileExt("modern_buildings_2_1k.exr") << endl;
cout << "ofFilePath:: ext: " << ofFilePath::getBaseName("data/modern_buildings_2_1k.exr") << endl;
	
string path = "cache/sticker.SVG";
cout << "filename: " << ofToLower(ofFilePath::getFileExt(path)) << endl;
	
cout << "current exe: " << ofFilePath::getCurrentExeDir() << endl;
	
cout << "trailing slash: " << ofFilePath::addTrailingSlash(path) << endl;
cout << "leading slash: " << ofFilePath::addLeadingSlash(path) << endl;
	
cout << "trailing slash: " << ofFilePath::addTrailingSlash(path+"/") << endl;
cout << "leading slash: " << ofFilePath::addLeadingSlash("/"+path) << endl;
	
cout << "extension: " << ofFilePath::getFileExt(path) << endl;
cout << "data path: " << ofToDataPath(path, false) << endl;
cout << "data path abs: " << ofToDataPath(path, true) << endl;
	
cout << "dir exists: " << ofDirectory::doesDirectoryExist("empty") << endl;
cout << "dir exists slash: " << ofDirectory::doesDirectoryExist("empty/") << endl;
	
string fullPath = ofToDataPath("empty", true);
cout << "dir exists: " << ofDirectory::doesDirectoryExist(fullPath) << " : " << fullPath << endl;
fullPath = ofToDataPath("empty/", true);
cout << "dir exists slash: " << ofDirectory::doesDirectoryExist(fullPath) << " : " << fullPath << endl;
	
cout << "remove extension: " << ofFilePath::removeExt(path) << " : " << path << endl;
	
cout << "current dir: " << ofFilePath::getCurrentWorkingDirectory() << endl;
```

ofTheo:
Thanks all!! 👏
Going to do a bunch of testing over the next few days to make sure it can handle everything well.

dimitre:
@ofTheo this PR just make some further ports to of::filesystem and doesn't change anything dramatic internally related to paths

dimitre:
@ofTheo @NickHardeman in this PR I've reverted back fileUtils test now

dimitre:
Hello @ofTheo can you please review this one when you have the time?
I've simplified and separated everything that was not ofFileUtils in other PRs, and reverted back tests fileUtils as it was before my changes.

in this one I'm proposing naming functions that returns of::filesystem::path with FS at the end, so we can keep the classic ones outputting strings, but internally we are one step closer to having fs::path from end to end.

dimitre:
@NickHardeman @ofTheo I've updated this PR and I think it is now good for merge or to be discussed more.
I would like to know @artificiel thoughts on this. 
The idea of this PR is to minimize internal conversion from fs::path to string back and forth

artificiel:
great to move further along with paths!

one thing I'm wondering: the return type of `defaultDataPath` changes from string to path. I think that is the main public API change? (which is fine IMO).

I wonder if there a specific reason not do the same with `getCurrentExeDir()` instead of creating a "parallel" `getCurrentExeDirFS()`, as paths convert gracefully into strings? 

in a similar vein I'm also wondering about the methods that take a path and return a string such as:

```c++
string ofFilePath::getBaseName(const of::filesystem::path & filePath)
```

(maybe not the same PR but quickly looking at it seems a lot of static ofFilePath could be deprecated in favour of operating directly on of::filesystem::path instances (which would have the additional effect of reducing the need for some functions such as trailing slashes which are remnants of using strings to represent paths)).

(and probably not the same PR haha but it seems ofFileUtils() could be broken up in ofBuffer, ofFile, ofDirectory...)

dimitre:
Yes, I think defaultDataPath is just part of the core internals, not public API, is this right?
And yes my favorite option was just convert everything to return of::filesystem::path, 
I don't remember exactly why we were keeping return objects the same. 
maybe it was not for automatic conversion but for comparison.

I'm good about using like this or converting return types to path. What do you think @ofTheo ?

dimitre:
Hey @ofTheo I would love to know your opinion, as I'm planning to make other changes after this one.

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -1819,9 +1811,9 @@ string ofFilePath::getFileName(const of::filesystem::path& _filePath, bool bRela
 }
 
 //------------------------------------------------------------------------------------------------------------
-string ofFilePath::getBaseName(const of::filesystem::path& filePath){
-	// FIXME: is this still useful?
-	return ofFile(filePath,ofFile::Reference).getBaseName();
+// FIXME: - suggest using stem() instead
+string ofFilePath::getBaseName(const of::filesystem::path & filePath){
+	return filePath.stem().string();
```
ofTheo, line 1816: can we get a test that shows what happens with something like: 
```
myDir/myOtherDir
myDir/myOtherDir/
myDir/myOtherDir/myFile 
```

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -25,29 +25,27 @@ namespace{
...
 	#if defined TARGET_OSX
 		try{
-			return of::filesystem::canonical(ofFilePath::getCurrentExeDir() / of::filesystem::path("../../../data/")).string();
+			return of::filesystem::canonical(ofFilePath::getCurrentExeDirFS() / of::filesystem::path("../../../data/"));
 		}catch(...){
-			return (ofFilePath::getCurrentExeDir() / of::filesystem::path("../../../data/")).string();
+			return (ofFilePath::getCurrentExeDirFS() / of::filesystem::path("../../../data/"));
 		}
 	#elif defined TARGET_ANDROID
 		return string("sdcard/");
```
ofTheo, line 36: should this be a fs return not a string?

ofTheo:
@dimitre - it might be good to update the fileUtils tests to include extra tests that reflect the changes ( exe directory etc ). 
And maybe other tests for future PRs that can account for a lot of edge cases ( joining paths, missing trailing slashes, making relative etc ) 

Anything with filesystem we should be extra careful with as bugs in FS can cause all sorts of problems.  

I added a couple of notes above. 

Should be easy to merge soon. Thanks!!

libs/openFrameworks/utils/ofFileUtils.cpp
dimitre: Yes it can be a string. 
my reasoning about using fs: I think this is not part of the public API, 
so the core can use fs::path instead of string, this would avoid windows wstring issues.

libs/openFrameworks/utils/ofFileUtils.cpp
dimitre: I've just tested this code on master and this PR
```c++
	for (auto & s : {
		string{ "myDir/myOtherDir" },
		string{ "myDir/myOtherDir/" },
		string{ "myDir/myOtherDir/file" },
	}) {
		cout << ofFilePath::getBaseName(s) << endl;
	}
```

and the results are the same. I'm wondering if the second test is correct.
```
myOtherDir

file
```

ofTheo: thanks for testing - at least they are the same!! 

not saying we should base our decisions on php 😁
but basename with a folder that has a / returns the folder name: 
https://www.php.net/manual/en/function.basename.php 

be curious if others had thoughts on this, but maybe that is a new issue/bugfix separate from this PR ?

dimitre: I've tested a variant with 		
```cout << of::filesystem::path(s).filename() << endl;```
and surprisingly the results are the same.

Comment (artificiel) [reply libs/openFrameworks/utils/ofFileUtils.cpp:1816]: to be clear `std::filesystem::path::stem` returns an empty string on a trailing /:
```
path: "/foo/bar.txt", stem: "bar"
path: "/foo/.bar", stem: ".bar"
path: "foo.bar.baz.tar", stem: "foo.bar.baz"
path: "foo/bar/", stem: ""
```
the OF `ofFilePath::getBaseName()` docs are ambiguous and does not mention the case of the trailing slash, so we can consider relying on it to represent a "directory" a form of undefined behaviour...

but as mentioned above most if not all usage of the `ofFilePath` stuff should be removed in favour of operating on std::filesystem::filepath instances (which probably covers all the functionality provided by static ofFilePath functions), and making sure the `filepath` nature of paths is preserved integrally up to the last minute (the function calls involved with the FS access itself). as mentioned elsewhere, both to reduce the problems linked to wstring, and also to stop reasoning about paths as strings.

so I guess this means for historical reasons we want `ofFilePath::getBaseName()` to behave like before (PHP-style or not) and we want to deprecate it and remove it from core.

PR #7388, opened 2023-03-15
ofTexture map to unordered_map

dimitre:
unordered_map can have an improvement performane over map, with the same funcionality less the automatic reordering.

### DIFF
--- a/libs/openFrameworks/gl/ofTexture.cpp
+++ b/libs/openFrameworks/gl/ofTexture.cpp
@@ -5,7 +5,7 @@
 #include "ofGLBaseTypes.h"
 #include "ofBufferObject.h"
 #include "ofMesh.h"
-#include <map>
+#include <unordered_map>
 
 #ifdef TARGET_ANDROID
 #include "ofAppAndroidWindow.h"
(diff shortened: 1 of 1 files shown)


libs/openFrameworks/gl/ofTexture.cpp
```diff
@@ -121,8 +121,8 @@ void ofDisableArbTex(){
 }
 
 
-static std::map<GLuint,int> & getTexturesIndex(){
-	static std::map<GLuint,int> * textureReferences = new std::map<GLuint,int>;
+static std::unordered_map<GLuint,int> & getTexturesIndex(){
+	static std::unordered_map<GLuint,int> * textureReferences = new std::unordered_map<GLuint,int>;
```
artificiel, line 125:
```c++
static auto * textureReferences = new std::unordered_map<GLuint,int>;
```

and the function could even be:
```c++
static auto & getTexturesIndex(){
``` 

as it is an exact example of `auto` as future-proof type polymorphism -- it turns out here the code using `getTexturesIndex()` is quite localized but if some user or addon code was written against an API explicitly expecting `std::map`, it would have to be revised for `std::unordered_map`. however these 2 classes are interchangeable, hence specifying `auto` as a return type would make a transition to any compatible class transparent. the type becomes an implementation detail instead of a burden on the contract.

`auto` is not only about typing/maintaining less repetitions of the same type, but more importantly letting the system adapt itself at compile-time.

artificiel (commented):
`static auto * textureReferences = new std::unordered_map<GLuint,int>;`

Issue #6129, opened 2018-09-18
ofDirectory::sort() is slow! 


ofTheo:
I noticed that with old projects which have a lot of png sequences that a project that used to take 20-30 seconds to load are now taking 15+ minutes. 

All I did was add sort() after calls to listDir() ( due to the non-sorted directories Apple lovingly introduced #5852 ) 

Taking the current dirListExample, calling sort() with listDir() makes a directory listing takes more than 10x longer than not calling sort(), but then getting randomly ordered results. 

I noticed that if I sort the files first as strings instead of as a vector ofFile::References it is almost as fast as not sorting at all. 

I think this might be due to the extra overhead of having files loaded into memory ( even as references ). 

I need to put together a PR for this, the main issue I see is that I am sorting inside of listDir before even creating the files vector. The plus side is you get sorted results by default without having to call sort(); But calling sort() will still be slow.  I could prob get the fix to work in sort() too without losing the performance improvement. It might just need to rebuild the file list which is doable. 

This is a comparison for sorting 110 files. 
```
Debug
 time diff for file list with current sort: 0.349263 seconds 
 time diff for file list with string based sort: 0.0325761 seconds 

= 10.7x faster with string based sort 

Release
 time diff for file list with current sort: 0.339156 seconds 
 time diff for file list with string based sort: 0.0316965 seconds 

= 10.7x faster with string based sort 
 ```

For 792 files:

```
Release:
 time diff for file list with current sort: 4.0761 seconds 
 time diff for file list with string based sort: 0.180382 seconds 

= 21x faster with string based sort
```

Going to post some more thoughts to this thread.

**Comment by arturoc** [comment]
i think when we open a file we open it by default for reading it so we are opening a handle to the file which can be quite expensive. when listing the dir i think we open it only in "reference" mode which doesn't open the file but there might be a bug somewhere where we are opening the file for read instead?

otherwise perhaps sorting over std::fs::path is probably more expensive than using strings but the fact that is so much is surprising

**Comment by ofTheo** [comment]
Here is some more timing. 
I think also part of the issue is the function calls inside of the natural compare function ( ofToString, ofToInt, getBaseName() ). 

If this is getting called for each compare it could be a source of some of the overhead.  

```
//------------------------------------------------------------------------------------------------------------
static bool natural(const ofFile& a, const ofFile& b) {
	string aname = a.getBaseName(), bname = b.getBaseName();
	int aint = ofToInt(aname), bint = ofToInt(bname);
[...]

PR #7399, opened 2023-03-17
ofDirectory::listDir performance improvement

dimitre:
still a draft. other things to do
- [ ] test devices
- [ ] implement hidden files

as it breaks api, better to implement in a major version change. 0.12.0 ?

### DIFF
--- a/examples/input_output/dirListExample/src/ofApp.cpp
+++ b/examples/input_output/dirListExample/src/ofApp.cpp
@@ -3,22 +3,26 @@
 //--------------------------------------------------------------
 void ofApp::setup(){
 
-	dir.listDir("images/of_logos/");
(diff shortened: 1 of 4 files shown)


ofTheo:
@dimitre going to do a PR just for the sorting speed for #6129 

ListDir is still quite slow and I think a PR which doesn't break the api would be good. 
I'll add some thoughts to this PR in comments on how we could do that.

dimitre:
Nice. I think it is possible to update ofDirectory not to open file immediatly. this would speed up everything.

ofTheo:
sounds good @dimitre! - PR #7509 speeds up sort by quite a bit ( even more for larger directories ) and doesn't change current behavior.

PR #7402, opened 2023-03-17
Updates to example/shader so that shaderGL2 will work with non arb tex

NickHardeman:


NickHardeman:
Added GL2 compatibility to #7341

Jonathhhan:
@NickHardeman thanks for merging and sorry that I created some extra work with my last change (changing the `ofDisableArbTex` position)...

NickHardeman:
@Jonathhhan , thank you for the updates! A quick fix :)

PR #7403, opened 2023-03-18
More .h work

dimitre:
I'll be building on this PR for untangling .h files. the first thing I've noticed is ofBaseMaterial is was declared in ofMaterial.h, so I'm proposing of separating in another .h file to follow the logic of all other base types.

— creating libs/openFrameworks/gl/ofMaterialBaseTypes.h file so it follows the same logic of other basetypes, 
nowadays in master we have the header declared in ofMaterial.h and cpp logic is inside ofBaseTypes.cpp so thigs are crossed. with this new arrangement we can decrease the inclusion of the entire ofMaterial.h file.

— Uniformization of includes order and spaces. first OF ones, then library ones like GLM, then c++ ones like <map> and finally forward declarators after a line break. 
—moved ofConstants.h to the last position in OF includes, just to be sure one include file is not including it throught another.
example
```c++
#pragma once

#include "ofMesh.h"
#include "ofNode.h"
#include <map>

class ofTexture;
class ofVboMesh;
class ofRectangle;
```


libs/openFrameworks/3d/ofNode.h
```diff
@@ -1,10 +1,10 @@
 
 #pragma once
 
-#include "ofConstants.h"
 #include "ofParameter.h"
 #include <array>
 #include "glm/mat4x4.hpp"
+#include "ofConstants.h"
```
ofTheo, line 7: might be good to have this above the glm line just incase it catches any non initialized errors

dimitre:
great @ofTheo I suppose there is a tiny speed change if any.
The idea is to keep working on .h changes after this PR is merged, so components start to get more independent from one another, maybe split ofConstants.h logic in more than one file

ofTheo:
good to go when your ready @dimitre

PR #7405, opened 2023-03-19
Old Linux - gcc5 fix

dimitre:
I've noticed nightly builds runs great in an old linux distro here (GCC5) but installing from git doesn't work because of scripts. as we don't have gcc5 in the name of the libraries I'm proposing an update in scripts, so they all download gcc6 ones and it seems to work great here.


dimitre:
closed by https://github.com/openframeworks/openFrameworks/pull/7406

PR #7406, opened 2023-03-19
Linux - enable gcc 5.4 to use c++17

dimitre:
it is not a perfect implementation but at least it allows us to use c++17 with gcc5.4
there is some missing logic there if the compiler major version is higher than 4 but not yet 7.


ofTheo:
Thanks @dimitre ! 
fyi -I don't think filesystem support is there until gcc6 and then its still part of the experimental implementation - just want to make sure that is understood and not relating to this issue?

dimitre:
I've double checked here and it is running 100% here. I'm not sure which filesystem implementation is being picked but everything good with c++17

PR #7411, opened 2023-03-21
More .h work - splitting ofConstants in more files

dimitre:
here the idea is splitting ofConstants.h in different smaller .h files

ofConstants.h - calls internally every other smaller .h file and include the specific platform includes.
ofFS.h - all about of::filesystem
ofDeprecated.h - only the definitions to make functions deprecated
ofPlatforms.h - all about platforms definitions only. 

This is a first idea, so feel free to give suggestions, file naming and paths.

motivation: with this we can add to .h files only the needed functionality without having to include the entire ofConstants.h
This one can potentially improve compile time of OF project. 

Only some ofConstants.h were replaced in this PR. if approved further work will be done


ofTheo:
If we go this route I wonder if we should include glm via ofMathConstants.h or another similar way.  So we don't need a ton of `#define GLM_FORCE_CTOR_INIT` all over the place?

dimitre:
Yeah I agree. For me it is the same. I don't mind having GLM_FORCE_CTOR_INIT all over too.
If someday we decide to zero initialize everything we can incrementally cleaning GLM_FORCE_CTOR_INIT for each file too

danoli3:
Would GLM_FORCE_CTOR_INIT be better suited in a PCH if it's everywhere?

dimitre:
@danoli3 exactly. somebody suggested using project generator to add this to all projects. I can help with that once we know what to add and where (I suppose it is addDefine function on PG but not sure)

PR #7413, opened 2023-03-22
Substitute defines DEG_TO_RAD / RAD_TO_DEG with ofDegToRad / ofRadToDeg

dimitre:
More
Substituting MIN / MAX  to std::min std::max
FLT_EPSILON to std equivalent
pi, half_pi, two_pi by glm ones

now we can remove the unused constants later on a breaking release 0.12.0

### DIFF
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -60,7 +60,7 @@ void ofCamera::setupPerspective(bool _vFlip, float fov, float nearDist, float fa
 	ofRectangle orientedViewport = getRenderer()->getNativeViewport();
 	float eyeX = orientedViewport.width / 2;
 	float eyeY = orientedViewport.height / 2;
-	float halfFov = PI * fov / 360.0f;
+	float halfFov = glm::pi<float>() * fov / 360.0f;
 	float theTan = tanf(halfFov);
 	float dist = eyeY / theTan;
 
@@ -129,7 +129,7 @@ bool ofCamera::getOrtho() const {
 
 //----------------------------------------
 float ofCamera::getImagePlaneDistance(const ofRectangle & viewport) const {
-	return viewport.height / (2.0f * tanf(PI * fov / 360.0f));
+	return viewport.height / (2.0f * tanf(glm::pi<float>() * fov / 360.0f));
 }
 
 //----------------------------------------
--- a/libs/openFrameworks/3d/ofCamera.cpp
+++ b/libs/openFrameworks/3d/ofCamera.cpp
@@ -98,7 +98,7 @@ void ofCamera::setupOffAxisViewPortal(const glm::vec3 & topLeft, const glm::vec3
 	setLensOffset(lensOffset);
 	setAspectRatio( glm::length(bottomEdge) / glm::length(leftEdge) );
 	auto distanceAlongOpticalAxis = fabs(glm::dot(bottomLeftToCam, cameraLookVector));
-	setFov(2.0f * RAD_TO_DEG * atan( (glm::length(leftEdge) / 2.0f) / distanceAlongOpticalAxis));
+	setFov(2.0f * glm::degrees( atan( (glm::length(leftEdge) / 2.0f) / distanceAlongOpticalAxis) ) );
 }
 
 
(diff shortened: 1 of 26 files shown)


libs/openFrameworks/gl/ofLight.cpp
```diff
@@ -224,7 +224,7 @@ float ofLight::getSpotlightCutOff() const{
 
 //----------------------------------------
 void ofLight::setSpotConcentration( float exponent ) {
-    data->exponent = CLAMP(exponent, 0, 128);
+    data->exponent = ofClamp(exponent, 0, 128);
```
artificiel, line 227: how about std::clamp? it is smarter with types. actually perhaps ofClamp should be deprecated?

libs/openFrameworks/math/ofMath.h
```diff
@@ -284,7 +284,7 @@ float ofRadToDeg(float radians);
 ///
 /// \param degrees An angle in degrees.
 /// \returns the angle in radians.
-float ofDegToRad(float degrees);
+float ofDegToRad(const float & degrees);
```
artificiel, line 287: will this &ref not remove the possibility of writing directly 

```c++
auto a = ofDegToRad(90);
```

dimitre: totally agreed. reverted and pushed a commit

libs/openFrameworks/gl/ofLight.cpp
artificiel: ah! yes it's c++17, but I thought it was allowed for 0.12?

ofTheo:
looks great @dimitre!

libs/openFrameworks/3d/ofCamera.cpp
```diff
@@ -60,7 +60,7 @@ void ofCamera::setupPerspective(bool _vFlip, float fov, float nearDist, float fa
 	ofRectangle orientedViewport = getRenderer()->getNativeViewport();
 	float eyeX = orientedViewport.width / 2;
 	float eyeY = orientedViewport.height / 2;
-	float halfFov = PI * fov / 360;
+	float halfFov = PI * fov / 360.0f;
```
ofTheo, line 63: any reason to not replace PI internally with `glm::pi<float>()` ? 
( but still keep the define for legacy use )

dimitre: Yeah we can do it. I'll substitute PI and TWO_PI / M_TWO_PI by the glm constants.
maybe we have to have some more glm includes

dimitre: done in latest commit, passing tests. should we merge it?

ofTheo:
Hey!  I think there are two fixes need at lines 511 and 512 of of3DGraphics. Pi is used in instead of two_pi 

https://github.com/openframeworks/openFrameworks/pull/7413/commits/6783807e264ca01e2f8daab4de040256507b1eeb#diff-d65e3c3d754334a7189b5c27b80d2f6c4760d9357b1c5b8639c99082ab1acfeeR511

The rest looks great!

libs/openFrameworks/graphics/ofGraphics.cpp
```diff
@@ -398,11 +398,11 @@ void ofBackgroundGradient(const ofColor& start, const ofColor& end, ofGradientMo
 		gradientMesh.addVertex(glm::vec3(center, 0.f));
 		gradientMesh.addColor(start);
 		int n = 32; // circular gradient resolution
-		float angleBisector = TWO_PI / (n * 2);
+		float angleBisector = glm::two_pi<float>() / (n * 2);
```
NickHardeman, line 401: Do all of the numbers need to be floats? Same for line 405 in ofGraphics.cpp
float angleBisector = glm::two_pi<float>() / ((float)n * 2.0f);

libs/openFrameworksCompiled/project/osx/config.osx.default.mk
```diff
@@ -78,7 +78,7 @@ ifndef MAC_OS_STD_LIB
 endif
 
 ifndef MAC_OS_CPP_VER
-    MAC_OS_CPP_VER = -std=c++17
+    MAC_OS_CPP_VER = -std=c++20
```
ofTheo, line 81: @dimitre was this meant to slip in? 🙂

dimitre: oh no! 😓I'll be reverting now in a new PR

dimitre: done and pushed to master. good catch @ofTheo thanks

Issue #7350, opened 2023-02-22
Update the emscripten template to allow for starting app without any html decoration


ofTheo:
Currently all Emscripten builds produce the same sort of html output. 

We've been discussing having Debug show the current view, but Release builds being full window by default. 

Something like the following:

Debug:

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/144000/220701284-6a217440-b409-4965-8ce8-7c3a9e3a5c04.png">

Release:

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/144000/220702004-d98bceb5-ce2a-407b-835f-f91025786cca.png">

or

Release:

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/144000/220702242-cf184357-8667-4612-8404-f0f6ba52c44c.png">

Some other options would be to have it set via code. It could be done via the window settings in main.cpp, or a static call the later would keep the main.cpp non emscripten specific.  

ie: 

`ofxEmscripten::setLayout(ofxEmscripten::Debug);`
`ofxEmscripten::setLayout(ofxEmscripten::Minimal);`
`ofxEmscripten::setLayout(ofxEmscripten::FillWindow);`

( not sure if that is the correct api approach, but just to give an idea )

cc @dimitre @Jonathhhan @NickHardeman 

Note: I don't think it is possible to have an emscripten app start in Full Screen, but that might be worth looking into at the same time.

NickHardeman:
+1 for Fill Window as default for Release build.

Jonathhhan:
There are two basic fullscreen options: "real" fullscreen and soft fullscreen. Soft fullscreen seems to be possible as a default.
https://glow3d.com/blog/2019/10/fullscreen-in-emscripten
Both have several options. Use fullscreen with resize canvas option (you can toggle that in the browser). It works very well for most examples, but some behave kind of strange (billboardRotationExample for example).
Another option is to scale everything, but that can cause artifacts from stretching / resizing.
A third option would be to keep the original size in the center of the screen and have a black border.

Jonathhhan:
Here are some example that do not look as "expected" with resize canvas (but most do):
```
advanced3dexample
modelnoiseexample
all cv examples
simpleeventsexample
billboardrotationexample
fbotrailsexample
mtextureshaderexample
pixelbufferexample
shaderexample
textureexample
screengrabexample
fontshapesexample
fontsexample
graphisexample
rectalignmentandscaling
dirlistexample
[...]

PR #7418, opened 2023-03-23
Emscripten Soft Fullscreen

NickHardeman:
Related to https://github.com/openframeworks/openFrameworks/issues/7350

- Added window resize events 
- Added soft fullscreen is set to OF_FULLSCREEN in main.cpp
- Touch can emulate mouse
- Set fullscreen scale strategy
- Disable default browser touch move

```
#if defined(TARGET_EMSCRIPTEN)
	auto ewin =  dynamic_pointer_cast<ofxAppEmscriptenWindow>(ofGetCurrentWindow());
	ewin->setTouchSimulatesMouse(true);
	ewin->preventDefaultBrowserTouchMoveBehavior();
	ewin->setFullscreenScaleStrategy(EMSCRIPTEN_FULLSCREEN_SCALE_ASPECT);
#endif
```

See windowed example:
https://nickhardeman.com/temp/OFExamples/AssimpExampleWindowed/

Soft fullscreen:
https://nickhardeman.com/temp/OFExamples/AssimpExample/

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -14,19 +14,22 @@ using namespace std;
 
 ofxAppEmscriptenWindow * ofxAppEmscriptenWindow::instance = NULL;
 
-
+//------------------------------------------------------------
 ofxAppEmscriptenWindow::ofxAppEmscriptenWindow(){
 	instance = this;
 }
 
+//------------------------------------------------------------
 ofxAppEmscriptenWindow::~ofxAppEmscriptenWindow() {
 	if(context != 0){
 		emscripten_webgl_destroy_context(context);	
 	}
 }
 
+//------------------------------------------------------------
 void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
     setWindowShape(settings.getWidth(),settings.getHeight());
+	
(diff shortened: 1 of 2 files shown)


ofTheo:
either way I think we should merge this and as there is a lot of useful functionality and we can always add/change stuff in another PR.

NickHardeman:
I would prefer the full screen to change resolution and fill the screen like you mentioned.

Not sure where this javascript function is defined, is it part of the emscripten library?
`onclick='Module.requestFullscreen(document.getElementById("pointerLock").checked,document.getElementById("resize").checked)'`

The soft fullscreen in OF is using this scale strategy.  
https://github.com/openframeworks/openFrameworks/pull/7418/files#diff-f3ce187f3e25e12b72b32629cd308f41075d821206fa7fedd69401ae03f91944R112-R114

NickHardeman:
I think it may be this from the generated assimExample.js
image

ofTheo:
ahh - yes, that looks like it. 👍

with my changes I basically don't need it, except to restore the canvas size afterwards - so the aspect checkbox uses that function instead of the others because they don't restore the aspect ratio afterwards. 
https://github.com/openframeworks/openFrameworks/pull/7424/files#diff-90b4970da5d99976d71e783c808f0f98ba27a1eb13dd31850a6f3ba3e0323745R101

Maybe I can use the other functions then and just add some code to save/restore the canvas size?

PR #7424, opened 2023-03-24
soft fullscreen for iOS and option to resize via checkbox in html template

### LINKED ISSUE #7423: Fix fullscreen button for iOS in emscripten
connected to #7350 

iOS doesn't support canvas / div fullscreen on iPhone according to this chart:
https://caniuse.com/?search=fullscreen

proposal:

keep the current functionality which works on desktop and add code to make the canvas fill the window for browsers that don't support the fullscreen functionality. 

Still need to add the resize fullscreen option in from the original.

eg:

```
        <script>
            function goFullScreen(){
                    var canvas = document.getElementById("canvas");
                    if(canvas.requestFullScreen)
                        canvas.requestFullScreen();
                    else if(canvas.webkitRequestFullScreen)
                        canvas.webkitRequestFullScreen();
                    else if(canvas.mozRequestFullScreen)
                        canvas.mozRequestFullScreen();
                    else{
                        //fake it
                        document.getElementById('header').style.display = 'none'; 
                        document.getElementById('output').style.display = 'none';  
                        var canvas = document.getElementById('canvas'); 
                        canvas.width=window.innerWidth;
                        canvas.height=window.innerHeight;
                    }
            }
        </script>
```

ofTheo:
closes #7423

It's not a perfect solution but it seems much more functional. 

We can improve the OF fullscreen as discussed here:
https://github.com/openframeworks/openFrameworks/pull/7418#issuecomment-1482087752 

but this makes the html button and checkboxes a little more useful / functional

### DIFF
--- a/libs/openFrameworksCompiled/project/emscripten/template.html
+++ b/libs/openFrameworksCompiled/project/emscripten/template.html
@@ -96,38 +96,66 @@
         outline: none;
       }
     </style>
+    <script>
+      function goEmscriptenFullscreen(resize){
+          Module.requestFullscreen(0, resize);
+      }
+      function tryFullScreen(aspect, resize){
+          var canvas = document.getElementById("canvas");
+          if( resize ){
+              canvas.width=screen.width; 
+              canvas.height=screen.height;
+          }
+          if(canvas.requestFullScreen){
+              if( aspect )goEmscriptenFullscreen(resize);
+              else canvas.requestFullScreen();
+          }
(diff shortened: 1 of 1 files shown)


NickHardeman:
I wonder if these full screen enter and exit events would fire in OF? That would be helpful for resize events.
https://emscripten.org/docs/api_reference/html5.h.html?highlight=emscriptenfullscreenstrategy#c.EmscriptenFullscreenChangeEvent

PR #7425, opened 2023-03-24
make pthreads not on by default for emscripten. add notes in config.make

### LINKED ISSUE #7412: Emscripten Wasm.memory Exception
I am using the latest nightly OF build with Emscripten. I can compile everything and run the project locally, however when I upload to my server, it throws the following error and does not run. I have tested on OSX, both Safari and Chrome.  
I noticed that an additional _3DPrimitivesExample.worker.js_ file has been created which seems different than previous versions of OF. 
Here is an example:
https://nickhardeman.com/temp/OFExamples/3DPrimitivesExample/3DPrimitivesExample.html

```
var INITIAL_MEMORY = Module["INITIAL_MEMORY"] || 134217728;
assert(INITIAL_MEMORY >= 65536, "INITIAL_MEMORY should be larger than STACK_SIZE, was " + INITIAL_MEMORY + "! (STACK_SIZE=" + 65536 + ")");
if (ENVIRONMENT_IS_PTHREAD) {
    wasmMemory = Module["wasmMemory"]
} else {
    if (Module["wasmMemory"]) {
        wasmMemory = Module["wasmMemory"]
    } else {
        wasmMemory = new WebAssembly.Memory({
            "initial": INITIAL_MEMORY / 65536,
            "maximum": 2147483648 / 65536,
            "shared": true
        });
        if (!(wasmMemory.buffer instanceof SharedArrayBuffer)) {
            err("requested a shared WebAssembly.Memory but the returned buffer is not a SharedArrayBuffer, indicating that while the browser has SharedArrayBuffer it does not have WebAssembly threads support - you may need to set a flag");
            if (ENVIRONMENT_IS_NODE) {
                err("(on node you may need: --experimental-wasm-threads --experimental-wasm-bulk-memory and/or recent version)")
            }
            throw Error("bad memory")
        }
    }
}
```
image

### LINKED ISSUE #7421: Disable threading for emscripten by default - add options and documentaiton to make file
Just creating an issue for this @Jonathhhan @NickHardeman 

Discovered in #7412 

We should just comment out the flag in the makefile and add notes on the .htaccess requirements. 
Might be good to add a section to the Setup Guide too on how to enable it. 

https://openframeworks.cc/setup/emscripten/

ofTheo:
closes #7421
closes #7412  

@Jonathhhan this code below works for me.  
Basically you just need to uncomment two lines in the project config.make - so that way you can build some projects with pthreads if needed and others without. 

Happy if there is another / better way to enable threading, but given that it errors unless you set the server up correctly feels like we should default to off for now.

### DIFF
--- a/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk
+++ b/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk
@@ -64,8 +64,8 @@ PLATFORM_REQUIRED_ADDONS = ofxEmscripten
 ################################################################################
 
 # Code Generation Option Flags (http://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html)
-PLATFORM_CFLAGS = -pthread
-PLATFORM_CXXFLAGS = -Wall -std=c++17 -Wno-warn-absolute-paths -pthread
+PLATFORM_CFLAGS =
+PLATFORM_CXXFLAGS = -Wall -std=c++17 -Wno-warn-absolute-paths
 
 ################################################################################
 # PLATFORM LDFLAGS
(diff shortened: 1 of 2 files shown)


Jonathhhan:
@ofTheo thanks. I also think this is the best possible solution.

PR #7426, opened 2023-03-24
Make platform exclusions

dimitre:
Testing a way of setting platform exclusions per project
so a specific project can be compiled without many parts of OF, example: 3d, video, graphics
or libraries, like boost, poco, etc.

I'm testing some configurations in projectGenerator config.make to see how much of OF it can exclude
it is a program who doesn't need most of the things from core


libs/openFrameworks/app/ofAppRunner.cpp
dimitre, line 1: I've removed ofImage and ofTruetype here, they were only used to shutdown truetype and freeimage, and they can be declared as forward declarations as ofURLFileLoaderShutdown already is.
I see no disadvantage of doing this and the advantage is we can build projects without Freeimage and Freetype when needed.

libs/openFrameworksCompiled/project/makefileCommon/config.shared.mk
dimitre, line 1: now ALL_OF_CORE_SOURCE_PATHS can be filtered by CORE_EXCLUSIONS

.github/workflows/build-macos.yml
dimitre, line 1: not meant to be in this PR but it would be great to cache this path too.

dimitre:
I've now extended the possibility to other platforms also, 
if you have the time please take a look, some proofreading @ofTheo or @NickHardeman

Jonathhhan:
> I'm testing some configurations in projectGenerator config.make to see how much of OF it can exclude
> it is a program who doesn't need most of the things from core

would it work the same way with `make` and linux and without the project generator (I have no idea)?

dimitre:
@Jonathhhan sorry I dont' fully understand your question.

libs/openFrameworks/app/ofAppRunner.cpp
ofTheo: @dimitre - not sure I totally understand.  
How is:  `ofTrueTypeShutdown()` getting called now if this PR is merged?

dimitre: yes this is invoked from ofTrueTypeFont.cpp, it is overridden. anyway there is no code inside this function anymore

dimitre: You can check the way ofURLFileLoaderShutdown is being utilized, that has a forward declarator there, and then it was overridden with the proper function on ofURLFileLoader.cpp

ofTheo: Ah I see - yeah the function does nothing. 🙂
And `ofCloseFreeImage` is still getting called in the exit callback anyway. Cool!

dimitre: Yes and the nice thing it allows us to build OF without truetypefont when needed too.

PR #7428, opened 2023-03-24
Assimp example update for Emscripten

NickHardeman:


Jonathhhan (commented):
Hey @NickHardeman,
is this:
```
#if defined(TARGET_EMSCRIPTEN)
#include "ofxAppEmscriptenWindow.h"
#endif
```
Now needed for every example or just the assimp one?

NickHardeman:
Just for the assimp one because it calls functions in the window class to enable touch to mouse events to allow controls for easy cam

NickHardeman:
The prevent default touch behavior and pass mouse events for first touch have been made default via #7479

PR #7430, opened 2023-03-24
Actions - test update msys2

dimitre:
I'm now testing installing some needed packages by the action itself on github, instead of installing by scripts.
scripts will run anyway so hopefully everything is already installed

the idea behind it is maybe github caches internally or have a mirror for faster installation, or even to save bandwidth. And if the action start caching packages in the future it will already benefit from that.


oxillo (changes requested):
as msystem is set in setup-msys2, it is not needed to have :

```yaml
env:
  MSYSTEM: ${{matrix.sys}}
```

in the next steps

dimitre:
thanks @oxillo I've pushed a commit. please check if it is ok

dimitre:
Yeah. github today changed certificates so they cannot download things from github itself.
https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/

dimitre:
Well now github has a certificate problem. even our avatars are broken :)

oxillo (changes requested):
for pacboy packages, you need to suffix the package name with :p. That way it will install the package needed for the selected msystem

for example 
`assimp` should become `assimp:p`

Your old version of matrix (and matrix.sys) was fine.

dimitre:
ok that's not very helpful unless there was some real caching involved.

PR #7431, opened 2023-03-25
FUNDING.yml

dimitre:
add funding to link with open_collective initiative


ofTheo:
Awesome!! Great idea.

dimitre:
now there is a heart button :)
Screen Shot 2023-03-25 at 14 37 16

dimitre:
You can setup the same for the organization also. not sure if it is only for github funding or other platforms too

PR #7435, opened 2023-03-26
Windows - load image with filenames other than english

dimitre:


dimitre:
Hey @oxillo please check this PR when you have the time. I don't know much about windows but I've tried to fix path encodings there, it seems to be working.

dimitre:
PS: there is a test there meant to be removed before merge

oxillo:
@dimitre the UTF16 encoding of filenames in Windows is generally hard to handle. I'm not an expert on that. 
I've tested on MINGW64 and UCRT64. The system dialogs works fine in both environments. So I think the filepaths are correctly handled. 
The only problem is on console output that doesn't work in MINGW64 but it was expected.

dimitre:
Nice. it worked here in a file called consolação.jpg on which didnt load before. I've put this file temporarly on this pr on examples/dimitre/Test folder.
It is latin so I'm not sure about other languages
@artificiel any opinion on this one?

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -182,16 +182,21 @@ static int getJpegOptionFromImageLoadSetting(const ofImageLoadSettings &settings
 template<typename PixelType>
 static bool loadImage(ofPixels_<PixelType> & pix, const of::filesystem::path& _fileName, const ofImageLoadSettings& settings){
 	ofInitFreeImage();
-
+	
+#if defined(TARGET_WIN32) || defined(__MINGW64__) || defined(__CYGWIN__) || defined(TARGET_WINVS)
+	#define OF_TARGET_WINDOWS
```
artificiel, line 187: is this the place to `#define` the platform? it seems what happens in `ofConstants.h` should be used and not duplicated here?

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -182,16 +182,21 @@ static int getJpegOptionFromImageLoadSetting(const ofImageLoadSettings &settings
...
 	auto uriStr = _fileName.string();
+#else
+	auto uriStr = _fileName;
+#endif
```
artificiel, line 191: this creates a situation where on windows `uriStr` is a `std::string`, whereas elsewhere it is an `of::filesystem::path` — which happens to be OK because both class support the later `.c_str()` class but is confusing with the `Str` suffix. moreover, the destination type for URIetc functions being `const char *` (all future uses of `uriStr` are with a `.c_str()` call) it would be simpler to call 
```
uriStr = _fileName.string().c_str(); // char * everywhere
```
however: when the `_fileName` is a windows path containing inconvertible unicode characters, the `.string()` method will throw an exception. all `.string()` calls should be try/catched.

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -209,23 +214,35 @@ static bool loadImage(ofPixels_<PixelType> & pix, const of::filesystem::path& _f
 		return ofLoadImage(pix, ofLoadURL(_fileName.string()).data);
 	}
 
-	auto fileName = ofToDataPath(_fileName, true);
+	auto fileName = ofToDataPathFS(_fileName, true);
```
artificiel, line 217: variables names like "filename" should better represent the underlying object, in this case "filepath"

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -209,23 +214,35 @@ static bool loadImage(ofPixels_<PixelType> & pix, const of::filesystem::path& _f
...
 	bool bLoaded = false;
 	FIBITMAP * bmp = nullptr;
 
 	FREE_IMAGE_FORMAT fif = FIF_UNKNOWN;
+#ifdef OF_TARGET_WINDOWS
+	fif = FreeImage_GetFileTypeU(fileName.wstring().c_str(), 0);
```
artificiel, line 223: `std::filesystem::path::c_str()` will produce a native-appropriate `char *` so it's not necessary to pre-transform with wstring().

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -358,12 +375,12 @@ static bool saveImage(const ofPixels_<PixelType> & _pix, const of::filesystem::p
...
 	ofFilePath::createEnclosingDirectory(_fileName);
-	auto fileName = ofToDataPath(_fileName);
+	auto fileName = ofToDataPathFS(_fileName).string().c_str();
 	FREE_IMAGE_FORMAT fif = FIF_UNKNOWN;
-	fif = FreeImage_GetFileType(fileName.c_str(), 0);
+	fif = FreeImage_GetFileType(fileName, 0);
 	if(fif == FIF_UNKNOWN) {
 		// or guess via filename
-		fif = FreeImage_GetFIFFromFilename(fileName.c_str());
+		fif = FreeImage_GetFIFFromFilename(fileName);
```
artificiel, line 383: there is a `FreeImage_GetFIFFromFilenameU()`, why not maintain the `#ifdef windows` pattern? instead of forcing the `fileName` into a `.string()` (again risking data loss/exception)
```
fif = FreeImage_GetFIFFromFilenameU(_fileName().c_str()); // works on windows
```
(same applies below to other "U" suffixed-functions)

libs/openFrameworks/graphics/ofImage.cpp
```diff
@@ -209,23 +214,35 @@ static bool loadImage(ofPixels_<PixelType> & pix, const of::filesystem::path& _f
 		return ofLoadImage(pix, ofLoadURL(_fileName.string()).data);
```
artificiel, line 214: .string() should be here try/catched (and it is correct tp down convert, as URI/URLs are not wstring-compatible)

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -1925,17 +1925,18 @@ void ofSetDataPathRoot(const of::filesystem::path& newRoot){
 }
 
 //--------------------------------------------------
-// MARK: - near future
-//of::filesystem::path ofToDataPath(const of::filesystem::path & path, bool makeAbsolute){
 std::string ofToDataPath(const of::filesystem::path & path, bool makeAbsolute){
+	return ofToDataPathFS(path, makeAbsolute).string();
```
artificiel, line 1929: .string() should be try-catched

libs/openFrameworks/utils/ofSystemUtils.cpp
```diff
@@ -328,7 +300,10 @@ static int CALLBACK loadDialogBrowseCallback(
     std::string defaultPath = *(std::string*)lpData;
     if(defaultPath!="" && uMsg==BFFM_INITIALIZED){
 		wchar_t         wideCharacterBuffer[MAX_PATH];
-		wcscpy(wideCharacterBuffer, convertNarrowToWide(ofToDataPath(defaultPath)).c_str());
+		
+		
+		std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+		wcscpy(wideCharacterBuffer, converter.from_bytes(ofToDataPath(defaultPath)).c_str());
```
artificiel, line 306: out of my comfort zone here but I wonder if the native handling does not simplify things as `defaultPath` is now an `filesystem::path`? right now it looks as it's converted to a std::string then converted back to wide buffer.

libs/openFrameworks/graphics/ofImage.cpp
dimitre: Yes sure. this was the motivation of opening that discussion about WIN32 and a more organized define. 
something that can represent when OS Windows are being used

artificiel: ah! well yes then I like your `OF_${CATEGORY}_${TAG}` pattern at the end of #7450.

maybe for the sake of the FS/wstring work, simply add a `#define OF_OS_WINDOWS` at line 89 of `ofConstants.h` and work with that for now? (or use `TARGET_WIN32` knowing a refactor is coming)

libs/openFrameworks/graphics/ofImage.cpp
dimitre: OK I think this one is resolved now

libs/openFrameworks/graphics/ofImage.cpp
dimitre: I've changed now to c_str(), so one more tiny step of fs::path compatible.
ofLoadURL receives a string parameter anyway, and I wonder if fs::path is a good choice for dealing with URLs too.
Anyway it is much more difficult to find problems with foreign characters in URLs than in files.

libs/openFrameworks/graphics/ofImage.cpp
dimitre: Yeah I just left as it was in OF. later on there is a function which calls filename, so it is the same for me
FreeImage_GetFIFFromFilename(fileName)

libs/openFrameworks/graphics/ofImage.cpp
dimitre: https://github.com/openframeworks/openFrameworks/pull/7495

libs/openFrameworks/utils/ofFileUtils.cpp
```diff
@@ -1775,7 +1775,8 @@ std::string ofFilePath::getEnclosingDirectory(const of::filesystem::path & _file
 	if(bRelativeToData){
 		fp = ofToDataPath(fp);
 	}
-	return addTrailingSlash(fp.parent_path());
+//	return addTrailingSlash(fp.parent_path());
+	return fp.parent_path().string();
```
ofTheo, line 1779: This change seems problematic - could break a lot of stuff

ofTheo:
think this one should wait for after 0.12.0 as I see some possible issues here with the changes required.

libs/openFrameworks/graphics/ofImage.cpp
artificiel: well, URL should indeed be strings and not paths (I may be confused with the PR's) but as I see it currently `ofLoadUrL()` takes a string which is correct:

https://github.com/openframeworks/openFrameworks/blob/6ad07c350b7af1a7974966af7368e5f7027bdb74/libs/openFrameworks/utils/ofURLFileLoader.cpp#L383

the design problem is that `loadImage` takes "something that could be a path or an URL", and tries to extract an URL from the argument, and if it's not an URL (no http scheme) defaults to a file. that logic is flawed: if `loadImage()` is passed a `filesystem::path`, it's undoubtedly a path (no messing with URI); if it is passed a `string` (and no need to support wstring, as URL are not legal wstring), then the URL dance can take effect, and if it's not an URL, upgrade it to a native path.

ofTheo: can we have:
` ofImage::loadImage( filesystem::path filePath  ); ` and ` ofImage::loadImage( string urlArg ); ` or would that be ambiguous ? ( for backwards compatibility reasons )

libs/openFrameworks/utils/ofFileUtils.cpp
dimitre: Yes, for sure. I've reverted back

libs/openFrameworks/utils/ofFileUtils.cpp
dimitre: I've reverted this file as it is in master now. 
This change is still needed, I'll do in a separate PR finding all fuctions ending with FS in the name (which have a parallel string equivalent)

dimitre:
Hello @ofTheo I think this one is good to go.
Resolves a dread windows problem where you cannot load files in ofSystemLoadDialog with any accentuation and pass correctly to load an ofImage.

There is a test here:
https://github.com/dimitre/ofTests/tree/main/WinEncoding

https://github.com/openframeworks/openFrameworks/pull/7435#issuecomment-1515304092

dimitre:
@danoli3 can you double check this one if you have the time?
btw, thanks for the other windows PR merging, OF is almost ready to work with multiple languages

dimitre:
Hello @roymacdonald can you please test this PR here in your windows setup?
it can be tested with this code here
https://github.com/dimitre/ofTests/blob/main/WinEncoding/src/ofApp.cpp

roymacdonald:
@dimitre I used some other image that I renamed with portuguese characters and it works.
The title in the file dialog seems wrong. But not sure if that is because of my system not setup in portuguese or what.
I have a HiDPI screen, the whole OF window is displayed very small, and the dialog's size is hard coded, so it is quite unusable


of0
of1
of2

dimitre:
Nice, thanks for that. It will probably work if made differently as the other dialog worked.
maybe using ```converter.from_bytes(windowTitle).c_str()``` directly instead of using std::wstring

dimitre:
I've tried a test in a windows machine now, and using latest I have some errors on ofRandomDistributions.h, line 146, 790, 830, 1000. 
Is the latest running ok for you @roymacdonald ? 
cc: @artificiel

artificiel:
@dimitre can you post the code and the errors?

dimitre:
@artificiel for me it shows errors in this lines compiling the core, always on words "or" and "and" in templates like this one
```std::enable_if_t<!is_random_engine_v<A> and !is_random_engine_v<B>, T>```
I've noticed ofParameter.h recent changes are also failing in Windows for me
Strangely it seems to be working ok on tests. so maybe it is somehting related to project settings configuration

**Comment by artificiel** [comment]
ah well a bit of googling led me to learn that:

> While 'and', 'or' and 'not' are part of the C++ standard, Microsoft Visual C++ (MSVC) disables them by default, unlike GCC or Clang.

there are 2 solutions:

1. the `/permissive-` compiler flag enforces closer conformance to the C++ standard. 

2. `#include <iso646.h>`

if you can test both solutions and confirm they are both viable, I'll let those more knowledgeable with MSVC decide the course of action, but on a theory level, considering the cross-platform nature of OF, maybe `/permissive-` is not a bad idea?

also note:

> Starting with Visual Studio 2019, enabling certain standard versions (like /std:c++20) automatically enables /permissive-, affecting projects that mix legacy and modern C++.

I think the new OF default is C++2b?, so `/permissive-` would already be activated, also in the GitHub tests?

and please post also the ofParameter.h errors, in case it's not the same thing.

**Comment by artificiel** [comment]
thinking a bit more, for backwards-compatibility with existing project, perhaps `#include <iso646.h>` is a good idea as it does not require a project setting and does not hurt anything if `/permissive-` is also enabled, so both solutions can be applied in parallel.

**Comment by dimitre** [comment]
@roymacdonald thanks I've fixed the dialog title issues in latest and updated the repo with more tests and more filenames.
But there is another strange issue:
When opening filenames containing cyrillic in wchar_t / wstring it quits with a kernel panic. 
it would be great to have more opinions from windows experienced ones. 
ping : @openframeworks/windows @artificiel @danoli3

**Comment by roymacdonald** [comment]
Interesting. What ever it is that causes a kernel panic it must not be nice. No idea from my side. I mostly work on macos

**Comment by artificiel** [comment]
@dimitre do you have a trace on that crash? maybe the cyrilic are actually truly non-convertible wide (a lot of unicode is convertible wide->narrow, everything on the [basic plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) will not cause an exception with `.string()`). also what makes things complicated is the encoding of the source files — if you are getting an error from data received by the OS/FS, but not if you type the same strings in a .h, it may indicate that the source files are themselves being converted, so typed-in unicode chars are actually "preconverted" by the text editor. i don't have an MSVC env to try things...

one way to isolate the source encoding is by passing actual bytes like

```c++
std::wstring s = L"Hello, \u4E16\u754C!"; // "Hello, 世界!"
```

so whatever happens to the intepretation of the glyphs in the file, the string will be built with the actual codepoints (and as far as i can tell, those are true wides, but i did not find a super clear reference about that). if you can capture the wchar that trigger your crash they would be helpful for future testing.

(it would be very great if the MSVC tests in OF github CI could be augmented and also fail. right now it seems MSVC is passing everything although it should not... maybe a "hosted"/VM thing?)

dimitre:
@artificiel I don't have the trace. I only have occasional access to a slow windows "box"
the crash is related to cyrillic in some file name (jpg) I'm trying to load. 
the tests are here if somebody else wants to check, including some filenames https://github.com/dimitre/ofTests/tree/main/WinEncoding
I'll be adding chinese / japanese or korean filenames to see if it runs ok when I test on windows again.

NickHardeman:
Hi @dimitre, testing on Windows 10 with your branch, 
I get the following error in the debug console
[ error ] ofImage: loadImage(): couldn't load image from "consolaÃ§Ã£o.jpg"

image

danoli3:
Maybe a FreeImage problem

NickHardeman:
@artificiel This is the configuration property I found for unicode and it was already set for the openframeworksLib and WinEcoding project.
image
Not sure how check the source encoding for the file... I tried opening the image with VS, but there wasn't an option for Advanced Save Options

artificiel:
@NickHardeman the file encoding i meant was not for the image itself but the .cpp that contains the quoted string! 

https://learn.microsoft.com/en-us/visualstudio/ide/how-to-save-and-open-files-with-encoding?view=vs-2022

NickHardeman:
image
image
I tried the following, but the errors persist.

NickHardeman:
Passing in a wstring to the load function works.
image

NickHardeman:
Not sure how, but passing in a narrow string to a path constructor and then passing that to image.load() also works.
```
std::string str = "consolação.jpg";
std::filesystem::path p{ str };
image.load(p);
```

NickHardeman:
If I remove the const and reference & from the image load function, then it works with a narrow string.
`bool ofImage_<PixelType>::load( of::filesystem::path fileName, const ofImageLoadSettings & settings){`
https://github.com/openframeworks/openFrameworks/blob/fcc4fba835c7eebabcc182b2ef5b5eb0fe747122/libs/openFrameworks/graphics/ofImage.cpp#L785
std::string str = "consolação.jpg";
image.load(str);

dimitre:
The issue is important (OF accepting multiple languages) so it can be used for teaching

But FreeImage must go, so this PR does not makes sense anymore.
it is abandonware now. Related:
https://bbs.archlinux.org/viewtopic.php?id=307499

I'm using on ofWorks fork the mango library: very fast, maintained and with a nice c++ api. I was considering 3 ones (Freeimage / FreeimageRe, OpenImageIO and Mango) but mango was the first I could build so it is the winner.

PR #7455, opened 2023-03-31
Fix for the default osx makefile not honoring the user's plist values and not setting the application icon

### LINKED ISSUE #7437: macOS Window blurry when building with make instead of xcode
This is the window after I built using `make Release`: (fyi I'm using VSCode)
image

This is the window after I built in xcode:
image

The xcode build made the window twice as small and also sharper
But when I build it using `make` command in vscode terminal, I get a blurry window that is twice as big

I tried enabling High Resolution in the plist
image

Still had the issue.
(I made the project with the of Project Generator, template VSCode)
(Project wasn't touched, its just the default window you get after creating the project)

System information:
OS: macOS Ventura 13.2.1 (22D68)
CPU: Apple M1
Display: Built-in Retina Display 13.3-inch (2560 × 1600)
OF Version: of_v20230326_osx_nightly

dannylol19:
Added features and fixes to the default osx makefile
Fixed NSHighResolutionCapable in openFrameworks-Info.plist not being honored by make
Fixed make not loading the default of.icns for the app
Added a feature that lets user's use their own icons for the app

This commit made it so if NSHighResolutionCapable in the user's
openFrameworks-Info.plist was true, the value would carry on into the
real Info.plist in the .app

This commit also fixed the issue that make never loads the
of.icns and of_debug.icns into .app/Resources, and the Info.plist
CFBundleIconFile key never being set

And lastly, if the user creates an icons directory in their project folder,
and puts a release.icns and/or debug.icns in it, then the makefile will
use those icons for the release or debug build it creates.

Closes #7437

### DIFF
--- a/libs/openFrameworksCompiled/project/osx/config.osx.default.mk
+++ b/libs/openFrameworksCompiled/project/osx/config.osx.default.mk
@@ -372,18 +372,17 @@ afterplatform: $(TARGET_NAME)
 	@echo '  <string>$(BIN_NAME).app</string>' >> bin/$(BIN_NAME).app/Contents/Info.plist
 	@echo '  <key>CFBundleExecutable</key>' >> bin/$(BIN_NAME).app/Contents/Info.plist
 	@echo '  <string>$(BIN_NAME)</string>' >> bin/$(BIN_NAME).app/Contents/Info.plist
-
 # App icons
 	@echo '  <key>CFBundleIconFile</key>' >> bin/$(BIN_NAME).app/Contents/Info.plist
 ifeq ($(RUN_TARGET), RunRelease)
 # Use the user's custom icon in their project's icons folder
 # If you want your own custom icon just make an icons/debug.icns file and/or icons/release.icns
 	@cp $(OF_LIBS_PATH)/openFrameworksCompiled/project/osx/of.icns bin/$(BIN_NAME).app/Contents/Resources/of.icns
-	if [ -a icons/release.icns ]; then cp icons/release.icns bin/$(BIN_NAME).app/Contents/Resources/of.icns; fi;
+	@if [ -a icons/release.icns ]; then cp icons/release.icns bin/$(BIN_NAME).app/Contents/Resources/of.icns; fi;
 	@echo '  <string>of.icns</string>' >> bin/$(BIN_NAME).app/Contents/Info.plist
 else
 	@cp $(OF_LIBS_PATH)/openFrameworksCompiled/project/osx/of_debug.icns bin/$(BIN_NAME).app/Contents/Resources/of_debug.icns
-	if [ -a icons/release.icns ]; then cp icons/debug.icns bin/$(BIN_NAME).app/Contents/Resources/of_debug.icns; fi;
+	@if [ -a icons/release.icns ]; then cp icons/debug.icns bin/$(BIN_NAME).app/Contents/Resources/of_debug.icns; fi;
 	@echo '  <string>of_debug.icns</string>' >> bin/$(BIN_NAME).app/Contents/Info.plist
 endif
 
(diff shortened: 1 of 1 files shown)


ofTheo:
Thanks @dannylol19 

I wonder if it might be better to simply use openFrameworks-Info.plist if it exists?  ( ie if it exists copy it to Contents/Info.plist, instead of writing out a new one? )

That way any values set there can be changed by the user and copied info the App bundle.

dannylol19:
> Thanks @dannylol19
> 
> I wonder if it might be better to simply use openFrameworks-Info.plist if it exists? ( ie if it exists copy it to Contents/Info.plist, instead of writing out a new one? )
> 
> That way any values set there can be changed by the user and copied info the App bundle.

Good idea, and initially I was planning on doing something like that
But I realized some things:
From my understanding, the openFrameworks-Info.plist is more of an xcode template if anything because it contains variables like `$(TARGET_NAME)` or `$(ICON_NAME)` which the xcode build just ends up replacing for the value of those variables.

The value of those variables are set in the xcode build settings menu.
They can be retrieved in the terminal by running `xcrun xcodebuild -showBuildSettings`

What we could do is modify the projectGenerator template to make the openFrameworks-Info.plist not have the $() variables but rather put the real values into it, that way using the actual openFrameworks-Info.plist will work
But we should only do that if we know the user is using `make` since with xcode would be better with the $() variables.

Do you have any ideas on how we can get around this?

Btw: I just saw the build-osx (osx, makefiles) failed because it couldn't find the NSHighResolutionCapable value in the openFrameworks-Info.plist so to fix that I just have to make a check to see if it exists first before trying to use the value

ofTheo:
@dannylol19 

Ahh yeah - maybe the easiest solution for now is to just use the OF plist and then overwrite the values that need overwriting with plistbuddy. 

plist buddy has a - [set command](https://blog.bravewaffles.io/Plistbuddy-Guide/)
So we could use it to set `<key>CFBundleName</key>` from `$(TARGET_NAME)`  to `$(BIN_NAME)` 

?

dimitre:
@ofTheo I agree with maybe centralizing in one plist, 
but in this case maybe this can be merged as it is, and then upgraded in another PR if possible?

ofTheo:
Here's a simplified example based on @dannylol19 PR which copies the OF plist first then does find and replace on the variables. 

The nice thing with this is you can override anything in the plist. It only sets the values if they haven't been set by the user. 

It would still need the local icon support but just wanted to make a simple example.

```
	@cp openFrameworks-Info.plist bin/$(BIN_NAME).app/Contents/Info.plist
	
# App icons
 ifeq ($(RUN_TARGET), RunRelease)
	@cp $(OF_LIBS_PATH)/openFrameworksCompiled/project/osx/of.icns bin/$(BIN_NAME).app/Contents/Resources/of.icns
 else
	@cp $(OF_LIBS_PATH)/openFrameworksCompiled/project/osx/of_debug.icns bin/$(BIN_NAME).app/Contents/Resources/of.icns
 endif
 
	@sed -i '' 's/\$$(DEVELOPMENT_LANGUAGE)/English/g' bin/$(BIN_NAME).app/Contents/Info.plist
	@sed -i '' 's/\$$(ICON_NAME)/of.icns/g' bin/$(BIN_NAME).app/Contents/Info.plist
	@sed -i '' 's/\$$(EXECUTABLE_NAME)/$(BIN_NAME)/g' bin/$(BIN_NAME).app/Contents/Info.plist
	@sed -i '' 's/\$$(TARGET_NAME)/$(BIN_NAME)/g' bin/$(BIN_NAME).app/Contents/Info.plist
	@sed -i '' 's/\$$(PRODUCT_BUNDLE_IDENTIFIER)/cc.openFrameworks.$(BIN_NAME)/g' bin/$(BIN_NAME).app/Contents/Info.plist
	@sed -i '' 's/\$$(VERSION)/1.0/g' bin/$(BIN_NAME).app/Contents/Info.plist
	
#	@echo '<?xml version="1.0" encoding="UTF-8"?>' > bin/$(BIN_NAME).app/Contents/Info.plist
#	@echo '<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">' >> bin/$(BIN_NAME).app/Contents/Info.plist
#	@echo '<plist version="1.0">' >> bin/$(BIN_NAME).app/Contents/Info.plist
#	@echo '<dict>' >> bin/$(BIN_NAME).app/Contents/Info.plist
#	@echo '  <key>CFBundleGetInfoString</key>' >> bin/$(BIN_NAME).app/Contents/Info.plist
#	@echo '  <string>$(BIN_NAME).app</string>' >> bin/$(BIN_NAME).app/Contents/Info.plist
```

ofTheo:
closing in favor of 7460
thanks @dannylol19 for finding this and getting the fix started!

PR #7456, opened 2023-03-31
XCode - openFrameworks files by reference (blue folder)

dimitre:
so we don't need to edit the template if we add / remove files from the core.
other advantage is the template folder gets a little bit lighter if we have to edit it

Screen Shot 2023-03-31 at 14 13 55

### DIFF
--- a/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj
+++ b/scripts/templates/osx/emptyExample.xcodeproj/project.pbxproj
@@ -3,770 +3,6 @@
 	"objectVersion": "54",
 	"archiveVersion": "1",
 	"objects": {
-		"E42963292163EDD300A6A9E2": {
-			"path": "../../../libs/openFrameworks/gl/ofMaterial.cpp",
-			"isa": "PBXFileReference",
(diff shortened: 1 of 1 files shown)


dimitre:
Any opinion on this one @ofTheo @artificiel ?

artificiel:
don't see a problem with that.

thinking about it: is it even required since the CoreOF.xcconfig is included? it's nice to be able to fish in the directory for reference, but maybe not so for beginners? (and "jump to definition" will still work, which is the better way to fish). if it can be omitted it should.

dimitre:
I think it is nice to browse .h files in the project when working with OF projects.

ofTheo:
looks great! thanks @dimitre

dimitre:
I'm wondering about making the "src" folder a reference too, and alter PG so it doesn't handle the code adding anymore.
PG is great to add another sources in  "src" folder, but having ofApp.h and ofApp.cpp hardcoded is not a good idea since some tests only use main.cpp. It also simplify the way PG handles ios / macOS differences.

ofTheo:
yeah - I think that makes sense too @dimitre !

Issue #7470, opened 2023-04-05
Link errors in Windows with latest nightly


c-mendoza:
I'm getting link errors with the latest nightly in Windows 10:

```
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IBaseFilter
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IMediaSeeking
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IGraphBuilder
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IMediaControl
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IMediaEvent
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IMediaPosition
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_IBasicAudio
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol MEDIATYPE_Video
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol MEDIASUBTYPE_RGB24
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol MEDIASUBTYPE_RGB32
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol CLSID_FilterGraph
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol FORMAT_VideoInfo
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol CLSID_SampleGrabber
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol IID_ISampleGrabber
2>openframeworksLib.lib(ofDirectShowPlayer.obj) : error LNK2001: unresolved external symbol CLSID_NullRenderer
2>bin\videoPlayerExample.exe : fatal error LNK1120: 15 unresolved externals
[...]

PR #7472, opened 2023-04-05
Fix for include lib in direct show player

NickHardeman:
Because the includes were moved from ofVideoGrabber.h to ofVideoGrabber.cpp, the ofDirectShowGrabber.h file was not included. This file included "videoInput.h" from libs/videoInput / include which had the following at the top. 
```
#ifdef _MSC_VER
#pragma comment(lib,"Strmiids.lib")
#endif
```
Added the above to ofDirectShowPlayer.cpp to fix compilation error.


ofTheo:
Cool yeah - I think this is the correct fix. 
I think the changes I suggested in #7470 just revealed that we were relying on ofDirectShowGrabber to provide the needed #pragma via videoInput - but making both files work without the other as this does looks like the right way to go. 

then all your changes @dimitre can stay as is. 🙂

Issue #7473, opened 2023-04-06
Linux: ofxOSC SEGFAULT, when receiving many messages


revilo196:
I'm getting a segfault of the ofxOSC plugin when receiving a large amount of messages at once (~250). This happens if for e.g if I reload all parameters of my project.

I traced the segfault to [ofxOscReceiver.cpp](https://github.com/openframeworks/openFrameworks/blob/da5f12874dd0ddbb85f10b08abda4c89c4f29fb9/addons/ofxOsc/src/ofxOscReceiver.cpp#L146) line 146.  Here the `address` is read out of bounds.

this issue could be fixed by correcting the bound check 2 lines earlier.

But it is a simple fix:
`if(address.size() >= i+1)`
has to be changed to
`if(address.size() > i+1)`

revilo196:
closed with #7474

PR #7474, opened 2023-04-06
bugfix ofOSC receiver segfault index out of bounds

revilo196:
only a very small change to fix an issue that occurs for me when receiving a large amount of messages. 

this fixes issue #7473 by correcting the bounds check.

Tested on Linux and Window

this is my first pull request, I'm not sure if this belongs into the master or the patch-release branch.


ofTheo:
Thanks @revilo196! 
Looking at the code I don't really understand the difference in behavior - but if you are still getting data and no segfaults. Then 👍

artificiel:
@revilo196 i have a question: are you sending empty ofParameterGroups?

and are you confirming you are receiving all the data?

no doubt you what you have fixed is/was a bug, but it seems to be an edge case that "should not happen" (hence why it got to lie dormant for so long), which maybe hides another problem that your code might be good at activating.

PR #7479, opened 2023-04-10
Emscripten Window defaults to passing mouse events for first touch 

NickHardeman:
and preventing default touch event listener in document.

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -96,6 +96,9 @@ void ofxAppEmscriptenWindow::update(){
 	if (bSetMainLoopTiming) {
 		// choose the render speed with the second argument. 1 is for every screen refresh, 2 for every second, 3 for every third. and so on...
 		emscripten_set_main_loop_timing(1, 2);
+		// disable default touch behaviour 
+		preventDefaultBrowserTouchMoveBehavior();
+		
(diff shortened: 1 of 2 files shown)


addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
```diff
@@ -96,6 +96,9 @@ void ofxAppEmscriptenWindow::update(){
 	if (bSetMainLoopTiming) {
 		// choose the render speed with the second argument. 1 is for every screen refresh, 2 for every second, 3 for every third. and so on...
 		emscripten_set_main_loop_timing(1, 2);
+		// disable default touch behaviour 
+		preventDefaultBrowserTouchMoveBehavior();
+
```
ofTheo, line 101: going to merge this in - but I wonder if we would want the option to toggle this on and off - maybe make it a window setting? or make this function pass a bool like  `bDisableTouchMove`

NickHardeman: Yeah definitely something to consider. Might be best to set the boolean before the first update() call since the `preventDefaultBrowserTouchMoveBehavior()`  adds a listener on the html document.

Jonathhhan:
@ofTheo @NickHardeman 
> toggle this on and off
I would second that. An disadvantage (that I just dicovered) is, that it is not possible to zoom into the browser with touch controls.

PR #7482, opened 2023-04-11
XCode project template - run script phases outside xcode project

dimitre:
Here I'm implemented the idea discussed with @artificiel of having XCode run script phases joined together in an outside .sh file.
This have some advantages:
- if the scripts are updated, modernized you dont' need to regenerate all projects using projectGenerator, they will always pick the latest from the OF.
- a little bit faster build times. XCode always write each .sh to a temporary folder anyway.
- Better editing, one can use the code editor of choice directly instead of copying pasting from xcode tiny textfields.
- not having to reconvert xcode template to json in the case the scripts need editing

### DIFF
--- a/scripts/osx/xcode_project.sh
+++ b/scripts/osx/xcode_project.sh
@@ -1,0 +1,171 @@
+#!/bin/zsh
+
+#MARK: COPY RESOURCES
+msg() {
+    # result=$(($1 + $2))
+    # echo "Result is: $result"
+	# echo "\033[32;1;4m2$1\033[0m"
+	echo "💾 $1"
+}
+
+
+msg "Copying Resources - Icon and libfmod"
+
+mkdir -p "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Resources/"
+# Copy default icon file into App/Resources
+rsync -aved --delete --exclude='.DS_Store' "$ICON_FILE" "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/Resources/"
+# Copy libfmod and change install directory for fmod to run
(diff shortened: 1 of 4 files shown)


dimitre:
Il'l copy @danoli3 here to see if there is any downside to it. I've changed everything to run on zsh instead of sh.

dimitre:
@artificiel I've made into functions but the if has to be inside because we always have else.
I think it is good to go

dimitre:
now it is showing the path for the script running on console, and how to change variables if needed
What do you think of the path @ofTheo, @artificiel ? 
scripts/osx/xcode_project.sh

```
---------------------------------------------------------------
Running ../../../scripts/osx/xcode_project.sh
---------------------------------------------------------------
💾 Copying Resources - Icon and libfmod
sending incremental file list

sent 60 bytes  received 12 bytes  144.00 bytes/sec
total size is 57,107  speedup is 793.15
sending incremental file list

sent 57 bytes  received 12 bytes  138.00 bytes/sec
total size is 2,693,408  speedup is 39,034.90
---------------------------------------------------------------
💾 Bundle data folder disabled 
can be enabled with OF_BUNDLE_DATA_FOLDER=1 in Project.xcconfig 
---------------------------------------------------------------
💾 Codesign Disabled
---------------------------------------------------------------
💾 Bundling dylibs disabled 
can be enabled with OF_BUNDLE_DYLIBS=1 in Project.xcconfig 
---------------------------------------------------------------

```

artificiel:
looks good!

since the script becomes a "property" of the OF tree, can the script path be based on OF_PATH instead of relative to the project? `${OF_PATH}/scripts/osx/xcode_project.sh`

dimitre:
@artificiel yeah in fact this is the variable being used there. in some cases it is relative.  
Maybe there is a way of showing the bash script path, it can be done in a subsequent PR.

I'm wondering about making libfmod optional later, maybe icon too.

PR #7487, opened 2023-04-13
fix warning with replacing instance variable to @property

2bbb:
related to #7293


ofTheo:
Amazing! 

Does it matter that the Engine is a singleton ( effectively ) - ie do we potentially have an issue with the bool becoming a property? 

Note: I barely understand obj-c 😁

ofTheo:
Great! Yeah just wanted to check as I think the AVAudioEngine is shared across all players. 👍

Issue #7492, opened 2023-04-15
push_back to emplace_back


dimitre:
substitute when possible, so we can avoid creation of a temporary object

PR #7493, opened 2023-04-16
vector push_back to emplace_back

dimitre:
closes https://github.com/openframeworks/openFrameworks/issues/7492


artificiel:
hello! i am not certain there is a specific gain to change all the `push_back` to `emplace_back` because there are narrow conditions where `emplace_back` performs better from the (modern) optimized `push_back`, but at the cost of a template (`emplace_back` is templated) which might affect compile time and binary size, and in some case allow problems that `push_back` would choke at compile-time while `emplace_back` will reveal only at run-time.

https://quuxplusone.github.io/blog/2021/03/03/push-back-emplace-back/
https://gumeo.github.io/post/emplace-back/

dimitre:
Thanks about your perspective @artificiel. 
I've noted emplace_back is more strict to types conversion too which can be good to identify some conversions happening later and repeatedly on code. for example ofMesh mixes unsigned int, unsigned long and ofIndexType in a vector. 
So emplace_back brought to attention they can be cast to proper types before insertion, and later I've changed to an insertion list, so the vector doesn't get resized many times.

ofMesh is one object that optimizations can benefit all OF because it is used everytime, everyframe.
I've noticed ofTexture creates an ofMesh object everytime it is called, and it is used all around.

I'll be happy to revert back any of this alterations if needed too.
Would you recommend keeping any specific file as it is today?

```diff
-			newFaces.push_back(i1);
-			newFaces.push_back(i12);
-			newFaces.push_back(i13);
-			newFaces.push_back(i2);
-			newFaces.push_back(i23);
-			newFaces.push_back(i12);
-			newFaces.push_back(i3);
-			newFaces.push_back(i13);
-			newFaces.push_back(i23);
-			newFaces.push_back(i12);
-			newFaces.push_back(i23);
-			newFaces.push_back(i13);

+			newFaces.insert(newFaces.end(), {
+				i1,
+				i12,
+				i13,
+				i2,
+				i23,
+				i12,
+				i3,
+				i13,
+				i23,
+				i12,
+				i23,
+				i13
+			});
```

artificiel:
it is true that stuff that happens at every frame should be looked at. if every `ofTexture` creates an `ofMesh` every frame it is definitely good to study the path, and if there is a way to circumvent memory allocation (or whatever non-deterministic calls) it is ideal (perhaps it is the design that should be adjusted more than implementation details). but always under the measurements of a real-world benchmark. some attention should also be given to compile time (perhaps also binary size but personally that's not a concern).

you also raise a valid point about types and I am surprised to see all the warnings are disabled (at least in Xcode). I would expect the explicit conversions to be reported so it would catch things like an `uint32_t` iterating a `uint64_t` (aka `size_t`). activating them generates warnings at many places including in libraries: ex: there is a warning in utf8 that has been fixed since upstream (perhaps we should update that lib?):

https://github.com/nemtrif/utfcpp/blob/2ad995746bf1731d5e21cde47c9c3deff56bdbc2/source/utf8/cpp17.h#L73

`Implicit signedness conversion`, `Implicit conversion to 32 bits`, `Implicit float conversion`, `Sign comparison` reveal  a bunch of implicit and and comparisons that are mismatched. not much to do with performance and while 32 bits may seem enough for a mesh, `size_t` is the type returned by `getNumVertices()`.
 
otherwise, for the `emplace_back()` changes in this PR I don't have specific recommendations just wanted to raise the template point.

artificiel:
ha! interesting an unsigned 64 bits int which is not an unsigned 64 bits int! but yes, `size_t` => `unsigned long long` => will eventually be 128bits, it makes sense that `is_same` reports differently (and it makes sense to adopt `size_t` for counters as it will adapt).

my central point was to match what the function gives, and that case use `size_t` (and not `uint32_t`). 

but should we be doing?
```c++
for (decltype(mesh.getNumVertices()) j = 0; j < mesh.getNumVertices(); j ++) {...}
```

2bbb:
I think better way is using simply `for(size_t j = 0; j < mesh.getNumVertices(); j++) { ... }`.
or, `for(auto j = 0ul; j < mesh.getNumVertices(); j++) { ... }`. then type of `j` is decided to `unsigned long` (but not `size_t`).
from C++23, we can use `0uz` and `0z`.

Issue #7378, opened 2023-03-12
download_libs.sh deletes libs for alternative targets


Jonathhhan:
At least with Linux and Emscripten and svgtiny (i am almost sure, that it was not the case, short ago).

dimitre:
this is the only recent change 
https://github.com/openframeworks/openFrameworks/commit/f71884f12fde7dc6f9a3ea0fff2ebb33e3e60c4d
can you please check if it works as it was before?

Jonathhhan:
@dimitre I tested different versions (0.11.2, master and nightly). And found out that it seems to be the normal behaviour. The shell scripts (and the dev folder) are actually only included in master. NIghtly and release do not include the scripts but have both Linux and Emscripten libs included.

dimitre:
ok. so maybe download_libs.sh doesn't delete anything, it is just the way OF releases are packed right?

Jonathhhan:
`download_libs.sh` does delete alternative libs (at least Linux / Emscripten). But maybe those shell files are only meant for development or there is another reason for deleting them?

dimitre:
Yes now I understand the issue. I think the scripts are so old that were considering just different platforms, but emscripten you can install alongside macos or linux, so maybe it is time to change this behavior.

danoli3:
This is fixed

PR #7499, opened 2023-04-18
Emscripten for msys2 and vs 

ofTheo:


dimitre:
Great Theo! I'll link another emscripten issue maybe easy to resolve here.
Installing libs for emscripten in macos and potentially other platforms remove the native libs
https://github.com/openframeworks/openFrameworks/issues/7378

PR #7507, opened 2023-04-19
Proposal: a way of not bundling FMOD in XCode

dimitre:
I've added an optional variable that can be passed so the XCode doensn't bundle FMOD
if you add in Project.xcconfig or App.xcconfig
OF_IGNORE_FMOD = 1


dimitre:
what about the feature and the naming @ofTheo @artificiel?
it is something I would use now in projectGenerator, so less SSD wear

ofTheo:
Yes! Great. 
Can we change it to OF_NO_FMOD ? 

Also would be great to enable the ofAVEngineSoundPlayer on macOS / iOS / tvOS if this is defined 
On Windows @NickHardeman is working on a MediaEngine Sound Player which could also be enabled if this is defined.

dimitre:
Yes, much better OF_NO_FMOD

dimitre:
PR updated.

ofTheo:
Going to do a separate PR for setting it to switch between fmod and  avengine something alongs the lines of:

```
//FOR AV ENGINE SOUND PLAYER UNCOMMENT TWO LINES BELOW
//OF_NO_FMOD=1
//USER_PREPROCESSOR_DEFINITIONS="OF_NO_FMOD=1"
GCC_PREPROCESSOR_DEFINITIONS=$(inherited)$(USER_PREPROCESSOR_DEFINITIONS)
```
( shame it needs two lines but the second allows us to do `#ifdef OF_NO_FMOD` in the cpp ) 

Then in the build OF script we can pass it through to OF core

```
xcodebuild -project "$OF_PATH/libs/openFrameworksCompiled/project/osx/openFrameworksLib.xcodeproj" -target openFrameworks -configuration "${CONFIGURATION}"  CLANG_CXX_LANGUAGE_STANDARD=$CLANG_CXX_LANGUAGE_STANDARD MACOSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET GCC_PREPROCESSOR_DEFINITIONS=$USER_PREPROCESSOR_DEFINITIONS
```

And finally in ofConstants.h 

```
//------------------------------------------------ soundplayer
//MAC_OS and IOS uncomment to enable AVEnginePlayer
#ifdef OF_NO_FMOD
    #undef USE_FMOD
    #if defined(TARGET_OF_IOS) || defined(TARGET_OSX)
        #define OF_SOUND_PLAYER_AV_ENGINE
    #endif
    //add other platforms here 
#endif
```

artificiel:
a bit late to the party but considering https://github.com/openframeworks/openFrameworks/discussions/7450 it might make sense to insert a "subsystem" layer to give more structure to the defines and have `OF_SOUNDPLAYER_NO_FMOD`

PR #7514, opened 2023-04-20
Script to copy recursively {config.make,Makefile} to each example folder

dimitre:
I'm using this to help testing RPI4 
maybe it can be interesting to have it on repository
not sure about the file name


artificiel:
just to understand well: this is a more "adaptive/generic" version of the "update multiple" of PG (in the sense that it does not take a specific template?)

dimitre:
This is a script I'm using to quickly compile using make in some freshly installed OF without having to compile PG. make build is really similar in all linux flavors, so I got all examples compilable in no time

PR #7523, opened 2023-04-24
Windows Media Foundation and XAudio Sound Player

NickHardeman:
Needs more thorough testing. Tested on Windows 10 and VS 2019.


ofTheo:
cool - looks like it might just need the same MF  / XAudio2 libs added for msys2 
https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/msys2/config.msys2.default.mk#L236

libs/openFrameworksCompiled/project/msys2/config.msys2.default.mk
```diff
@@ -236,6 +236,7 @@ PLATFORM_CORE_EXCLUSIONS += $(OF_LIBS_PATH)/utf8/%
 PLATFORM_LIBRARIES += ksuser opengl32 gdi32 msimg32 glu32 dsound winmm strmiids #dxguid
 PLATFORM_LIBRARIES += uuid ole32 oleaut32 setupapi wsock32 ws2_32 Iphlpapi Comdlg32
 PLATFORM_LIBRARIES += freeimage boost_filesystem-mt boost_system-mt
+PLATFORM_LIBRARIES += mfreadwrite xaudio2
```
NickHardeman, line 239: Yeah I added some here, but maybe there needs to be more

NickHardeman: Going to add the remaining mf libs for msys2, since the video player is not included in msys2 at the moment

NickHardeman:
The class also inits and shuts down MF Media Foundation and calls CoInitializeEx. The same as the MediaFoundation Video Player class. This code should be consolidated into a single place. If the mf sound player works with msys2, maybe it can be there? Or in ofConstants.h? The media foundation video player is not included in msys2 so it shouldn't just be in there.

ofTheo:
I think if you pull master branch in now - the VS test will pass

libs/openFrameworks/sound/ofMediaFoundationSoundPlayer.h
```diff
@@ -0,0 +1,276 @@
...
+
+namespace of {
+	struct MFSourceReaderNotifyCallback {
+	public:
+		virtual void OnSourceReaderEvent(HRESULT hrStatus, DWORD dwStreamIndex,
+			DWORD dwStreamFlags, LONGLONG llTimestamp, IMFSample* pSample) = 0;
+	};
+}
+
+class ofMediaFoundationUtils {
```
NickHardeman, line 24: Moved some media foundation utils into the MF sound player class so that it could be shared with the Media Foundation video player.

PR #7531, opened 2023-04-26
Update linux64 - filesystem to run ok with Ubuntu 16.04

dimitre:
so it runs ok with c++17 and filesystem in Ubuntu 16.04


ofTheo:
Cool. Let's see if this passes the CI. 
I might remember some newer gcc versions erroring if that line is included.  But I might be misremembering

dimitre:
all tests ok here, let's merge it?

PR #7532, opened 2023-04-26
Fixing GLFW issue in Ubuntu 16.04 - GLFW_TRANSPARENT_FRAMEBUFFER

dimitre:
Ubuntu 16.04 parked GLFW in version 3.1.2
so this PR checks if the version is exactly 3.1.2 and ignores the line causing error.


ofTheo:
I think the function was added in 3.3 so maybe more compatible ( for other linux distros ) would be something like?  

#if (GLFW_VERSION_MAJOR >= 3 && GLFW_VERSION_MINOR > 2) || (GLFW_VERSION_MAJOR > 3 )

dimitre:
OK but this logic will grab version 4.1.x

dimitre:
Yes, updated! now it works too with 18.04 (GLFW 3.2.1-1)

Issue #6895, opened 2022-03-15
ofToggleFullscreen() crash when set fullscreen via window header's fullscreen button


thomasgeissl:
On osx, oF 0.11.2 my app crashes when i
first set the window fullscreen via the window bar
then call ofToggleFullscreen

```
2022-03-15 20:21:45.354 ofPackageManager[96562:4193583] NSWindowStyleMaskFullScreen cleared on a window outside of a full screen transition. Called from (
        0   AppKit                              0x00007fff22b92dca __25-[NSWindow setStyleMask:]_block_invoke + 125
        1   AppKit                              0x00007fff22b92cf3 NSPerformVisuallyAtomicChange + 132
        2   AppKit                              0x00007fff22b92c00 -[NSWindow setStyleMask:] + 170
        3   ofPackageManager                    0x0000000104e4971c _ZN15ofAppGLFWWindow13setFullscreenEb + 108
        4   ofPackageManager                    0x0000000104e4c474 _Z18ofToggleFullscreenv + 41
        5   ofPackageManager                    0x0000000104d778b9 _ZNSt3__110__function6__funcIZN7ofEventI11ofEventArgsNS_15recursive_mutexEE13make_functionIN8ofxImGui3GuiEEENS_10shared_ptrIN2of4priv8FunctionIS3_S4_EEEEPT_MSF_FvRS3_EiEUlPKvSH_E_NS_9allocatorISM_EEFbSL_SH_EEclEOSL_SH_ + 37
        6   ofPackageManager                    0x0000000104ea805f _ZN7ofEventI14ofKeyEventArgsNSt3__115recursive_mutexEE6notifyERS0_ + 149
        7   ofPackageManager                    0x0000000104ea65ea _ZN12ofCoreEvents14notifyKeyEventER14ofKeyEventArgs + 862
        8   ofPackageManager                    0x0000000104e48885 _ZN15ofAppGLFWWindow11keyboard_cbEP10GLFWwindowiiii + 1057
        9   ofPackageManager                    0x0000000104d8a7e4 _Z26ImGui_ImplGlfw_KeyCallbackP10GLFWwindowiiii + 40
        10  AppKit                              0x00007fff22ccb9c8 -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 6482
        11  AppKit                              0x00007fff22cc9e56 -[NSWindow(NSEventRouting) sendEvent:] + 347
[...]

PR #7538, opened 2023-04-27
fix macOS issue where fullscreen activation is broken after maximizing with green button

dimitre:
more description on this issue
closes - https://github.com/openframeworks/openFrameworks/issues/6895

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -843,12 +843,20 @@ void ofAppGLFWWindow::setFullscreen(bool fullscreen){
 //	setWindowShape(windowW, windowH);
 
 #elif defined(TARGET_OSX)
+	
+	NSWindow * cocoaWindow = glfwGetCocoaWindow(windowP);
+	if (([cocoaWindow styleMask] & NSWindowStyleMaskFullScreen) == NSWindowStyleMaskFullScreen) {
+		if (targetWindowMode == OF_FULLSCREEN) {
+			targetWindowMode = OF_WINDOW;
+			[cocoaWindow toggleFullScreen:nil];
+		}
+	}
+
+	
 	if( targetWindowMode == OF_FULLSCREEN){
 		//----------------------------------------------------
 		[NSApp setPresentationOptions:NSApplicationPresentationHideMenuBar | NSApplicationPresentationHideDock];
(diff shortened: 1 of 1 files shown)


ofTheo:
cool! 
don't merge just yet - found some edge cases where this fails.

ofTheo:
So for me this doesn't work going from green + fullscreen to window via setFullscreen(false)

(  using this code in ofApp ) 

```

    if(key == 'f'){
        ofSetFullscreen(true);
    }else if( key == 'w'){
        ofSetFullscreen(false);
    }else if( key =='t' ){
        ofToggleFullscreen();
    }
}
``` 


An approach which accounts for all situations that seems to work well is: 

make the top of setFullscreen this instead ( removing it from lower down ) 

```
//------------------------------------------------------------
void ofAppGLFWWindow::setFullscreen(bool fullscreen){
	if (fullscreen){
		targetWindowMode = OF_FULLSCREEN;
	}else{
		targetWindowMode = OF_WINDOW;
	}
 
    #if defined(TARGET_OSX)
	NSWindow * cocoaWindow = glfwGetCocoaWindow(windowP);
 	if (([cocoaWindow styleMask] & NSWindowStyleMaskFullScreen) == NSWindowStyleMaskFullScreen) {
 		if (targetWindowMode == OF_WINDOW) {
                    targetWindowMode = OF_WINDOW;
                    [cocoaWindow toggleFullScreen:nil];
 		}
 	}
    #endif
```

and 

Add this to the end of: `void ofAppGLFWWindow::resize_cb(GLFWwindow* windowP_, int w, int h) {` 
```
     #if defined(TARGET_OSX)
	NSWindow * cocoaWindow = glfwGetCocoaWindow(windowP_);
 	if (([cocoaWindow styleMask] & NSWindowStyleMaskFullScreen) == NSWindowStyleMaskFullScreen) {
            instance->settings.windowMode = OF_FULLSCREEN;
 	}else{
            instance->settings.windowMode = OF_WINDOW;
        }
    #endif
```

This accounts for even getting out of fullscreen using the green + - by using the resize callback to figure out if we are still in Apple fullscreen mode or not. 

It also stays in Apple fullscreen if you call setFullscreen(true) which I think is the desired behavior

dimitre:
Thanks @ofTheo! I'll re-run one test that fails often. tested here working great

ofTheo:
Thanks @dimitre for finding this issue and getting it resolved!

PR #7546, opened 2023-05-04
replace implementation of ofFilePath::makeRelative from custom to of::filepath::relative

2bbb:
@dimitre 

I think this is simple way.
but, please check feature i think is correct or not...

### DIFF
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -1938,7 +1938,7 @@ std::string ofFilePath::makeRelative(const of::filesystem::path & from, const of
     of::filesystem::path ret;
     try {
         ret = of::filesystem::relative(pathTo, pathFrom);
-    } catch(std::filesystem::filesystem_error &err) {
+    } catch(of::filesystem::filesystem_error &err) {
         ofLogError("ofFilePath::makeRelative") << err.what();
     }
 //	return ret;
--- a/libs/openFrameworks/utils/ofFileUtils.cpp
+++ b/libs/openFrameworks/utils/ofFileUtils.cpp
@@ -1935,25 +1935,14 @@ string ofFilePath::getUserHomeDir(){
 std::string ofFilePath::makeRelative(const of::filesystem::path & from, const of::filesystem::path & to){
 	auto pathFrom = of::filesystem::absolute( from );
 	auto pathTo = of::filesystem::absolute( to );
-	of::filesystem::path ret;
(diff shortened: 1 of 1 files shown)


dimitre:
Yes this is much needed. and I think we should remove ofFilePath::makeRelative from the core, if there is any.
question: is it needed to make them absolute before comparison? or can we just use the paths provided by the function?

ofTheo:
curious if this is closing an issue? 
If this PR does fix broken code - then great to merge, but would be good to check that it doesn't change paths in anyway. 

fyi - its currently failing on rpi / arm

```
/home/runner/work/openFrameworks/openFrameworks/libs/openFrameworks/utils/ofFileUtils.cpp: In static member function ‘static std::__cxx11::string ofFilePath::makeRelative(const path&, const path&)’:
/home/runner/work/openFrameworks/openFrameworks/libs/openFrameworks/utils/ofFileUtils.cpp:1940:15: error: ‘relative’ is not a member of ‘of::filesystem’
         ret = of::filesystem::relative(pathTo, pathFrom);
               ^~
```


> remove ofFilePath::makeRelative from the core 

I'd like to keep all the existing functions as is for a bit please! 🙂
Spent so many hours fixing bugs related to some of the filesystem changes - would be great to get things stable for a while before 0.12.0 release.

2bbb:
@dimitre 

maybe making absolute is not needed.

@ofTheo 

ok. I agree what you say. I close this, and try to update after releasing 0.12.

ofTheo:
Thanks @2bbb - wasn't sure if this was a PR to fix: 
https://github.com/openframeworks/projectGenerator/issues/368 

Or just a general fix / cleanup. 
If it is more general then I think def good to wait until after 0.12.0 

Thanks!!

2bbb:
@ofTheo 
I tried to test PG's problem with this replacing, but this wasn't solve it.
PG's process around relative / absolute path is so messy... i don't understand what causes....
but, maybe problem hides on PG's code (not oF core), I think, maybe... maybe...

At first, we need to clean up codes on PG and create unit tests on PG what we need, and well document platforms, templates, etc, etc...

ofTheo:
Ah okay!! 
Thanks for the update. 

Totally agree about better tests. 
Some unit tests in the CI checking abs and relative paths / external source etc would help a lot and then we could also check a PR isn't adding a regression. 

Thanks for all your help on this!!

dimitre:
Great. I can investigate further today what is happening on PG. 
I'll take a look on @2bbb messages I think he have a reproduceable case to test what is happening on XCode. with this I can try to solve the issue.

PR #7561, opened 2023-06-21
Emscripten player update

Jonathhhan:
The main change is, that the audioPlayer now only creates one audio element for each player and not for each audio file that is loaded. Some other fixes and some cleanup too...


Jonathhhan:
Now also multiplay is enabled for the soundPlayyer...

Issue #7570, opened 2023-07-12
Fresh install in Ubuntu, install make too


dimitre:
I've just installed Ubuntu 23.04 and noticed "make" isn't built in or in the install scripts. 
it would be great if we add to the install scripts

ofTheo:
oh crikey - yeah def something to fix for 0.12.0

PR #7571, opened 2023-07-13
Ubuntu : add make to packages to install

dimitre:
I suppose this is the change that should be made.
Can you double check it @ofTheo ?
Closes : https://github.com/openframeworks/openFrameworks/issues/7570


ofTheo:
looks good to me @dimitre 👍

PR #7574, opened 2023-07-21
add setup argument to listen on specific ip

2bbb:
for example:

let's setup 2 receiver with

```cpp
        receiver1.setup(26666, "localhost"); // same to receiver1.setup(26666, "127.0.0.1");
        receiver2.setup(26666, "192.168.2.105");
```

then we can receive message on same port with different receivers like:

```cpp
        while(receiver1.hasWaitingMessages()) {
            ofxOscMessage m;
            receiver1.getNextMessage(m);
            ofLogNotice("receiver1") << m;
        }
        
        while(receiver2.hasWaitingMessages()) {
            ofxOscMessage m;
            receiver2.getNextMessage(m);
            ofLogNotice("receiver2") << m;
        }
```

I think this feature is helpful to make secure on large systems and complex network structures.
and this update is compatible simple `setup` with only given port.

### DIFF
--- a/addons/ofxOsc/src/ofxOscReceiver.cpp
+++ b/addons/ofxOsc/src/ofxOscReceiver.cpp
@@ -28,11 +28,12 @@ ofxOscReceiver& ofxOscReceiver::copy(const ofxOscReceiver &other){
 }
 
 //--------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ofTheo:
looks good to me! :)

PR #7594, opened 2023-08-10
Updated arm runner for Bullseye closes #7573

### LINKED ISSUE #7573: linux-armv6: Github action build fails on <optional>
I'm using `<optional>` and the armv6 build chokes on the header. I can see `-std=c++17` in the args above so that seems correct in terms of language version, but somehow I get:

```
/home/runner/work/openFrameworks/openFrameworks/addons/ofxOsc/src/ofxOscReceiver.h:5:20: fatal error: optional: No such file or directory
 #include <optional>
                    ^
```
other platforms are fine. ideas? in addition to `c++17` GCC 7.1 is the minimal version for `<optional>` (but I don't know how to know which gcc is being used by GitHub)

https://github.com/artificiel/openFrameworks/actions/runs/5560129779/jobs/10156882889#step:6:451

ofTheo:
closes #7573 

going to wait on merging for caching improvements as mentioned here:
https://github.com/pguyot/arm-runner-action/issues/79


ofTheo:
Hmm turns out it's not the caching that takes forever but compile ( 50 minutes to compile OF empty example and all addons example ). This is because QEMU is emulating the processor as opposed to cross compiling ( which we currently use and takes 3 minutes ). 

50mins is def not practical - so going to look at other options.

dimitre:
Hey @ofTheo just as a curiosity: why linuxarmv6l instead of linuxaarch64?

PR #7597, opened 2023-08-15
bundle identifier fix (rfc1034 to allow spaces in names)

dimitre:
more info here:
closes https://github.com/openframeworks/projectGenerator/issues/403


ofTheo:
@dimitre does `$TARGET_NAME:rfc1034identifier` also work?

As the other will will have different values for Debug and Release.

dimitre:
My bad, that was the intention. I'll update this PR now

dimitre:
Yes it works well, I've pushed the update. Generated a project called "space sketch" and compiled. inspecting inside .app I've noticed the Info.plist have well formed data now: 	
```xml
<string>cc.openFrameworks.space-sketch</string>
```

ofTheo:
looks good to me ( as long as it passes macOS checks )

PR #7602, opened 2023-08-19
[WIP] Set application icon in glfw

### LINKED ISSUE #7590: MSYS2 : default application icon displayed at runtime
GLFW applications only display default application icon whatever the icon used (OF, OF debug or user specified).
The icon is correctly displayed in the Windows explorer.

Looking at the code, icon is only set on Linux and using native calls.
GLFW now offers glfwSetWindowIcon() that could be used cross-platform.

oxillo:
Fix #7590


oxillo:
@ofTheo I didn't manage to have Xcode to add it ! As GLFW do not support window icon for MacOs (and Wayland), I've thought it won't be a problem not to include it (use on these platform is guarded by #define)

ofTheo:
Added a comment above but happy to merge once conflicts are resolved.

Issue #7555, opened 2023-06-04
Very teeny tiny issue in materialPBRAdvanced


neilmendoza:
Just came across all the new rendering features and they look amazing! Thanks @NickHardeman! Was playing around with **materialPBRAdvanced** and it has a commented out line in **main.vert** that let's you modify the geometry. The shadows don't match up if you uncomment that line as though as these shaders are omitted on the shadow pass. Maybe worth adding an extra comment about it to make it clear why that's the shadows aren't 100% correct there. The example works as is though so feel free to close this issue and ignore...

NickHardeman:
Hi @neilmendoza, Thank you. I can add a comment about the geometry so it's more clear. :)

neilmendoza:
Awesome, thanks!

Wonder if it might make sense to have the possibility to pass in a custom depth shader for situations where you do want to change the geometry on the GPU, e.g. this example or mesh instancing.

NickHardeman:
Ah yes, that would definitely be beneficial. Were you thinking passing the custom depth shader to the light or using a material? 
Or maybe both 😉 
Something like 
`material.setShadowDepthPass(depthShaderBuffer.getText(), GL_VERTEX_SHADER, "shadow.vert");`
and the shadow class
`light.getShadow().setDepthPass( &myCustomDepthShader );`
after you're done, `light.getShadow().setDefaultDepthPass();`

So a custom depth pass shader could be used via material (internally it would automatically set the depth shader on the shadow) or via the shadow class directly.

neilmendoza:
The material would probably make the most sense for the situations I'm thinking about, where lighting stays the same but some objects in the scene have their geometry modified on the GPU. Either would work though and who knows what crazy use cases might pop up.

neilmendoza:
So I've been playing around with hacking this functionality into a project by doing...

```
light.beginShadowDepthPass(j);
dynamic_pointer_cast<ofGLProgrammableRenderer>(ofGetGLRenderer())->getCurrentShader().end(); customDepthShader.begin();
...
customDepthShader.end();
dynamic_pointer_cast<ofGLProgrammableRenderer>(ofGetGLRenderer())->getCurrentShader().begin(); 
light.endShadowDepthPass(j);
```

[...]

PR #7613, opened 2023-08-25
Custom shaders during a shadow pass.

NickHardeman:
This allows using custom shaders during a shadow pass. The materials do not support custom shadow passes yet. Maybe that can be addressed in the next minor release. I tested using the shadow and material examples but could probably use a bit more.

Shadow shader functions have been added, so you can setup a shader using a function within the shadow class
`shadow.setupShadowDepthShader(ofShader& ashader, const std::string aShaderMain);`

The shader function for sending to the frag shader accepts a world position
```
vec3 worldPosition = (modelMatrix * vec4(position.xyz, 1.0)).xyz;
sendShadowDepthWorldPosition(worldPosition);
```

An example vert shader based on the MaterialPBRAdvanced Example
```
#if defined(SHADOW_DEPTH_PASS)
uniform mat4 modelMatrix;
in vec4 position;
uniform float iElapsedTime;
#else
OUT vec3 v_localPos;
#endif

void main (void){
	vec4 npos = position;
	npos.xyz += 3.0 * vec3( cos(iElapsedTime+position.y*1.25), 0.0, 0.0);
#if defined(SHADOW_DEPTH_PASS)
	vec3 worldPosition = (modelMatrix * vec4(npos.xyz, 1.0)).xyz;
	sendShadowDepthWorldPosition(worldPosition);
#else
	v_localPos = position.xyz;
	sendVaryings(npos);
	gl_Position = modelViewProjectionMatrix * npos;
#endif
	
}
```

Example usage:
```
if( light.shouldRenderShadowDepthPass() ) {
	int numShadowPasses = light.getNumShadowDepthPasses();
	light.beginShadowDepthPass();
	mDepthShader.begin();
	mDepthShader.setUniform1f("iElapsedTime", ofGetElapsedTimef());
	renderScene(true);
	mDepthShader.end();
	light.endShadowDepthPass();
}
```

### DIFF
--- a/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
@@ -1290,6 +1290,19 @@ void ofGLProgrammableRenderer::bind(const ofShader & shader){
     if(currentShader && *currentShader==shader){
 		return;
     }
+	
+	if(bIsShadowDepthPass) {
+		// if we are not the shadow shader, lets unbind it
+		if(!settingDefaultShader && currentShadow && currentShader ) {
+			// lets assume it's bound?
+			if(!bCustomShadowShader) {
+				glUseProgram(0);
+			}
+			// we are assuming that since it's a custom depth shader, it all will be taken care of ...
+			bCustomShadowShader=true;
+		}
+	}
+	
 	glUseProgram(shader.getProgram());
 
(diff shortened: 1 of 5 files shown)


NickHardeman:
Addresses issue #7555

ofTheo:
the broken arm check is fixed now in apothecary will re-run in a sec to confirm

PR #7632, opened 2023-09-03
Bugfix emscripten audio loading

ofTheo:
fix audio loading/playback from index.data 
we might need to add in some other audio mime-types but this seems to account for most of them. cc @Jonathhhan


Jonathhhan:
@ofTheo not sure if those are mime types, but these additional containers seem to be accepted:
.adts, .caf, .flac, .mp3, .mp4, .webm...

ofTheo:
Thanks @Jonathhhan - I think mp4 and mp3 work with the mpeg tag. 
I can look up and add the others.

ofTheo:
okay I think this should be good to go once it passes the tests.

Issue #7626, opened 2023-09-01
emscripten - video doesn't play in safari if autoplay is enabled


ofTheo:
Was debugging video playback not working in emscripten and tracked it down to Safari preferences don't autoplay video. 
This gets added automatically for any site you visit if you have it set globally. 

<img width="771" alt="image" src="https://github.com/openframeworks/openFrameworks/assets/144000/bded6d64-76fa-4e17-b227-84a492830b22">

Looking at other suggestions I tried muting the video, but that didn't help. 
Wondering if we could at least detect it and add notice to the webpage to unblock video for that site? 

This is what it looks like on safari: 

![image](https://github.com/openframeworks/openFrameworks/assets/144000/e20d96b1-63cb-4005-8709-152d6c7f9fe9)

Changing the website preference makes it work fine. 

tagging @Jonathhhan @artificiel

Jonathhhan:
@ofTheo since there is no sound in the video, it could work, if volume is set to 0 (not tested yet):
https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide


> As a general rule, you can assume that media will be allowed to autoplay only if at least one of the following is true:
> 
> The audio is muted or its volume is set to 0
> The user has interacted with the site (by clicking, tapping, pressing keys, etc.)
> If the site has been allowlisted; this may happen either automatically if the browser determines that the user engages with media frequently, or manually through preferences or other user interface features
> If the autoplay [Permissions Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Permissions_Policy) is used to grant autoplay support to an [<iframe>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) and its document.

Jonathhhan:
For audio with autoplay there is a warning in the js console: `The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page.`

Jonathhhan:
@ofTheo 
setting
`fingerMovie.setVolume(0);`
before
`fingerMovie.play();`
works for me.

artificiel:
I'm really a tourist in JS but this seems promising? support is not complete on all platforms but upcoming — i presume an empty list would be returned for an audio-less video:

  https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/audioTracks 
  https://caniuse.com/?search=audiotracks

[...]

PR #7637, opened 2023-09-05
Fix video emscripten

### LINKED ISSUE #7627: emscripten - regression from #7617 - video files are not found
I think we have a regression from this #7617 
Not sure why but since #7617 ( cc @artificiel  ) moving the html to a directory and not copying the data/ folder videos now don't load. 

```
emrun bin/opencvExample 

Now listening at http://0.0.0.0:6931/
 ofxEmscriptenVideoPlayer::load 
127.0.0.1 - - [01/Sep/2023 10:04:41] code 404, message File not found: /Users/theo/Downloads/of_v0.12.0_osx_release/examples/computer_vision/opencvExample/bin/opencvExample/data/fingers.mp4
127.0.0.1 - - [01/Sep/2023 10:04:41] "GET /data/fingers.mp4 HTTP/1.1" 404 -

```

copying `data/` into `opencvExample/` fixes the issue. 

@Jonathhhan as he has some insight into how file loading is working from index.data 

but this might warrant a patch-release.

ofTheo:
add in auto muting for emscripten 
switch in mp4 for opencvPeopleDetection example to make compatible with emscripten 

with these changes we should be good to 
close #7627 #7626


Jonathhhan:
@ofTheo first I thought exactly that is the solution, but then I saw it works for Safari only (deleted my comment because of that):
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/audioTracks

ofTheo:
ahh thanks @Jonathhhan - I can make it so it detects just for safari.

Jonathhhan:
I have another Idea, to check it with the audio context. I will have a try...

ofTheo:
awesome - thanks @Jonathhhan. feel free to do a PR for it if it works :)

PR #7640, opened 2023-09-06
Emscripten mute video

Jonathhhan:
@ofTheo this is working for me.

```
            if (!(VIDEO.player[player_id].mozHasAudio || Boolean(VIDEO.player[player_id].webkitAudioDecodedByteCount) || Boolean(VIDEO.player[player_id].audioTracks && VIDEO.player[player_id].audioTracks.length))) {
                VIDEO.player[player_id].muted = true;
            }
```
I think it is not needed for ` html5video_player_load_url`, because there needs to be some user action before, which should enable autoplay too. 
And I removed muting if volume <= 0 because its silent anyway (not sure about that).

### DIFF
--- a/addons/ofxEmscripten/libs/html5video/lib/emscripten/library_html5video.js
+++ b/addons/ofxEmscripten/libs/html5video/lib/emscripten/library_html5video.js
@@ -150,8 +150,8 @@ var LibraryHTML5Video = {
             GLctx.texParameteri(GLctx.TEXTURE_2D, GLctx.TEXTURE_WRAP_T, GLctx.CLAMP_TO_EDGE);
             VIDEO.player[player_id].textureId = texId;
             
-            if (typeof VIDEO.player[player_id].audioTracks == "undefined") {
-                VIDEO.player[player_id].muted = true;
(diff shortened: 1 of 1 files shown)


Jonathhhan:
Still some issues. I try to solve them.

ofTheo:
> And I removed muting if volume <= 0 because its silent anyway (not sure about that).

Ah I was thinking that the muting might need to be explicitly set to true for the browser to allow it?

Jonathhhan:
My suggestion seems to work well for Safari and Firefox, but not for Chrome and Edge (the video needs to play for a short time, before it is possible to detect audio).
So I would suggest now, just to set the volume in ofApp.cpp to 0 with `.setVolume(0)`, if it is known that the video is without sound.
Autoplay in the web seems to be a controversial topic: 
http://www.punkchip.com/autoplay-is-bad-for-all-users/
http://www.punkchip.com/autoplay-is-still-bad-for-all-users/

Jonathhhan:
> I would suggest for now, just to set the volume in ofApp.cpp to 0 with .setVolume(0), if it is known that the video is without sound.

ofTheo:
cool I can revert my changes if that seems to make sense?

Jonathhhan:
@ofTheo yes, I think reverting makes sense (because in Firefox and Chrome a video with sound would be muted with your change).

PR #7643, opened 2023-09-07
Custom depth shader in materialPBRAdvanced Example

NickHardeman:
Screen Recording 2023-09-06 at 8 11 59 PM mov


ofTheo:
Ha! 

I wonder could the depth noise be on a key press / toggle?

NickHardeman:
I added it on a toggle (w) and is disabled by default.

PR #7646, opened 2023-09-07
Emscripten revert video volume

Jonathhhan:
Not sure,  but I think it is not needed...


ofTheo:
okay to leave as is? 
was thinking the muting might be good incase it helped with the autoplay detection ( ie if you set the volume to 0 )

Jonathhhan:
@ofTheo if volume is set to 0, it autoplays without being muted:
https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide#autoplay_availability

ofTheo:
yeah but in safari it only mentions muting 
so I figured why not error on the safe side 😁 

image

https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari

If you think it 100% isn't needed happy to merge.

Jonathhhan:
@ofTheo good to know. Looks like it is needed for Safari (can not test it at the moment).

Issue #264, opened 2010-11-18
ofClear uses inconsistent arguments


kylemcdonald:
ofClear(r,g,b) is inconsistent with ofBackground(r,g,b) because the alpha is 0 by default -- meaning you always have to explicitly declare the alpha to clear the background.

furthermore, because all the arguments have default values there is not/cannot exist an ofClear(gray, alpha) to match ofSetColor style

julapy:
was about to log an issue for this very same thing when i remembered you guys chatting about this.

i found it very confusing that `ofClear(255)` clears the fbo and makes it transparent.
when i think ofClear, it is like filling the fbo with a color value.
so writing `ofClear(255)` should make the fbo white.
on the other hand, `ofClear(255, 0)` should make the fbo transparent.
this is in line with the way ofColor class works.

+1 for making ofClear always default to opaque.

required changes below.

`void ofClear(float r, float g, float b, float a=255);`
in https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofGraphics.h#L243

`void ofClear(float brightness, float a=0);`
in https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/graphics/ofGraphics.h#L254

tgfrerer:
Renates to #3566

oxillo:
Hi,

We are trying to clean out the many issues that we have.
Many versions of OF have been released since this was issued.

Can you help us by checking if this issue is still present in the latest version ?

In case of inactivity, we'll close this issue by end of january, 2020.

Thanks

Issue #5421, opened 2017-01-19
ofSetColor(ofFloatColor) down-samples back to 8 bit


armadillu:
I'm making very slow trail effects, and i'm noticing a hard limit at 0,0039 alpha values when playing with float fbo's. This also happens to be 1 / 256, so looking into ofSetColor() it converts back to ofColor (char) no matter what you feed it (ofColor or ofFloatColor).

Also found an old forum post about this;

https://forum.openframeworks.cc/t/incorrect-float-colors-gl-rgba32float-fbos-banding/21200/2

danoli3:
Oh shiiit yeah that's a pretty big bug!

arturoc:
probably ofSetColor should only accept ofFloatColor everything else will convert to that but ofFloatColor will stay as it is, then the renderers should set it as floats

dimitre:
I suppose this one wasn't touched yet. for me it is the most serious issue in OF.

dimitre:
cc: @ofTheo @NickHardeman @artificiel

artificiel:
@arturoc's suggestion is of course the right direction but there might be a slight can of worms as `ofStyle` defines `.color` as `ofColor,` and is written as such (char) beside the calls to `glColor4f()`.

I think the proper solution would also need upgrading `ofStyle.color` (and `.bgColor`) to `ofFloatColor`? which will probably reveal more stuff... It might be a far-ranging project. 

but I agree probably necessary. with 10-bit color becoming more and more accessible, perhaps the float usage should become more encouraged (I find myself often working in float to keep the same brain state with shaders; the only advantage to 8-bit is the familiarity (and proliferation of the pattern due to web colours)).

https://github.com/openframeworks/openFrameworks/blob/b369a1861ee0bd82973fe85adafc93054fbd21dd/libs/openFrameworks/gl/ofGLRenderer.cpp#L1058-L1080

artificiel:
just took another look and the implications are far-reaching, e.g:
https://github.com/openframeworks/openFrameworks/blob/b818ea0304a0c442239fa3bf0a25e85f608fcc39/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp#L330-L334
so for things to be homogenous ofPath needs to be retouched... it basically means reworking all internal use of `ofStyle` to expect/respect `ofFloatColor`. (otherwise it would be a half-done upgrade, if something like ofPath color stays 8 bits)

[...]

Issue #3566, opened 2015-01-17
deprecate ofBackground or refactor to mean ofClear


kylemcdonald:
this suggestion comes from a discussion with @ofzach last week.

right now we have a few functions for working with the background in OF:

0. `ofClear()`: clears the background color right away
0. `ofBackground()`: clears the background color right away, and saves the background color
0. `ofSetBackgroundColorAuto()`: enables or disables whether the saved background color is used to clear the screen at the beginning of each frame
0. `ofSetBackgroundColor()`: changes the saved background color without clearing the screen

having so many functions creates confusion. the only reason that anything besides `ofClear()` exists is because processing sets an example of putting the `background()` call in `setup()` as if it is a property of the document rather than an action.

i suggest we only use `ofClear()`. alternatively, we can change the name of `ofClear()` to `ofBackground()` and note in the changelog that it doesn't work in `setup()` anymore.

tgfrerer:
+0.75 for keeping `ofClear()` and getting rid of all `ofBackground*` related items. 

* Getting rid of all the ofBackground logic would make it easier to maintain the renderers in the long run, and less error prone. 
* Also, `ofClear` does what it does, is an action verb and would feels more natural in terms of semantics. 
* Having only one point of entry will also make it easier to keep track on the GL / renderers side of things.

If i remember correctly, the screen buffer is cleared automatically to the window default color, and opaque alpha, while a frame buffer (FBO) is accumulating (not cleared) by default. 

So, if we had an example that goes like this:

`
void ofApp::draw(){
ofColor backgroundColor = ofColor(ofColor::darkgray);
ofClear(backgroundColor);
// ... things are drawn
}
`
The use of `ofClear` and how to set the background colour would be demonstrated in two lines. 

----

-0.25 for getting rid of `ofBackgound*`

* An argument for keeping an automated background color value is that if we want to maintain the current auto-clear for the screen buffer, setting a background color for the screen buffer by clearing manually would mean the screen buffer gets cleared twice (once to of-default, then to user-color) before anything gets drawn. In OpenGL glClear is one of the most expensive functions, and it feels like a waste.

---- 

I tend to think that people who mostly use the screen buffer tend to be new / or not too interested in OpenGL and performance and will appreciate anything that makes getting something drawn easy. And that therefore, anything screen buffer related should be written so that beginners feel welcome and can keep focus on their task.

[...]

PR #7650, opened 2023-09-09
Use floats for colors internally. 

NickHardeman:
All of the renderers accept floats for color arguments in the range 0-1 and use ofFloatColor instead of ofColor. 

ofSetColor functions that receive an ofFloatColor argument:
```
void ofSetColor(const ofFloatColor & color);
void ofSetColor(const ofFloatColor & color, float _a);
```

ofSetFloatColor functions have been added to accept floats between 0-1 or ofFloatColor
```
void ofSetFloatColor(float r, float g, float b);
void ofSetFloatColor(float r, float g, float b, float a);
void ofSetFloatColor(float gray);
void ofSetFloatColor(const ofFloatColor & color);
void ofSetFloatColor(const ofFloatColor & color, float _a);
```

Similar for ofClear
`void ofClear(const ofFloatColor & c);`

```
void ofClearFloat(float r, float g, float b);
void ofClearFloat(float r, float g, float b, float a);
void ofClearFloat(float brightness, float a);
void ofClearFloat(float brightness);
```

### DIFF
--- a/libs/openFrameworks/app/ofAppNoWindow.cpp
+++ b/libs/openFrameworks/app/ofAppNoWindow.cpp
@@ -151,23 +151,23 @@ private:
 	void setDepthTest(bool){};
 
 	// color options
-	void setColor(int r, int g, int b){}; // 0-255
-	void setColor(int r, int g, int b, int a){}; // 0-255
-	void setColor(const ofColor & color){};
-	void setColor(const ofColor & color, int _a){};
-	void setColor(int gray){}; // new set a color as grayscale with one argument
+	void setColor(float r, float g, float b){}; // 0-1
+	void setColor(float r, float g, float b, float a){}; // 0-1
+	void setColor(const ofFloatColor & color){};
+	void setColor(const ofFloatColor & color, float _a){};
+	void setColor(float gray){}; // new set a color as grayscale with one argument
 	void setHexColor( int hexColor ){}; // hex, like web 0xFF0033;
 
 	// bg color
-	ofColor getBackgroundColor(){return ofColor(200);}
-	void setBackgroundColor(const ofColor & color){}
+	ofFloatColor getBackgroundColor(){return ofFloatColor(200.f/255.f);}
+	void setBackgroundColor(const ofFloatColor & color){}
 	bool getBackgroundAuto(){
 		return true;
 	}
-	void background(const ofColor & c){};
+	void background(const ofFloatColor & c){};
 	void background(float brightness){};
-	void background(int hexColor, float _a=255.0f){};
-	void background(int r, int g, int b, int a=255){};
+	void background(int hexColor, int _a=255){};
+	void background(float r, float g, float b, float a=1.f){};
 
 	void setBackgroundAuto(bool bManual){};		// default is true
 
--- a/libs/openFrameworks/gl/ofGLBaseTypes.h
+++ b/libs/openFrameworks/gl/ofGLBaseTypes.h
@@ -319,7 +319,7 @@ public:
 	/// \brief Set the global ambient light color.
 	///
 	/// \param c The color to set this renderer to use as ambient lighting.
-	virtual void setGlobalAmbientColor(const ofColor& c)=0;
+	virtual void setGlobalAmbientColor(const ofFloatColor& c)=0;
 
 	/// \brief Enable a light at a specific index.
 	///
--- a/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
+++ b/libs/openFrameworks/gl/ofGLProgrammableRenderer.cpp
@@ -328,7 +328,7 @@ void ofGLProgrammableRenderer::draw(const ofPolyline & poly) const{
 
 //----------------------------------------------------------
 void ofGLProgrammableRenderer::draw(const ofPath & shape) const{
-	ofColor prevColor;
+	ofFloatColor prevColor;
 	if(shape.getUseShapeColor()){
 		prevColor = currentStyle.color;
 	}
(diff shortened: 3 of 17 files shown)


artificiel:
super back-end work!

for the user-facing interface, would this make things too flexible?
when you are working in a 0-255 range for some reason, but want to take advantage of more than 8 bits?

```c++
ofSetColor(float r, float g, float b, float a); // range 0-255
```

in essence, ofSet(char)Color => 0-255; ofSetFloatColor => 0-1

could also provide flavours ofSetShortColor() = 0-65535

and so since we're processing based on type, maaaaybe instead of introducing additional functions, whatever gets thrown in is explicitly treated as T.
```c++
ofSetColor<char>() // default, equivalent to ofSetColor()
ofSetColor<float>()
ofSetColor<short>()
```

NickHardeman:
Currently the ofSetColor( int ) functions remain the same and there are some specific ofSetFloatColor functions
```
void ofSetFloatColor(float r, float g, float b);
void ofSetFloatColor(float r, float g, float b, float a);
void ofSetFloatColor(float gray);
void ofSetFloatColor(float gray, float _a);
```

I added an `ofSetColor( ofColor_<T> )` function that accepts any ofColor type and normalizes it.  
When I tried the following:
```
template<typename T>
void ofSetColor( const ofColor_<T>& acolor, T alpha )
```
There was a lot of issues with having to cast the second argument when calling the method, ie.
`ofSetColor( ofColor(200), (unsigned char)255);`
So I removed that approach as it seemed quite verbose and would break existing code using `ofSetColor(200, 200)` because the first argument gets converted to an ofColor via ofColor's equals operator and thus require the cast for the second argument.

artificiel:
ah yes, otherwise requires a maze of `std::enable_if_t` -- sometimes justified but here it's pretty OK to have ofSetFloatColor(), especially since passing an ofFloatColor gets properly honoured.

since it's a far-reaching change, it would be great to have @armadillu (as the originator of #5421) and @dimitre (who wanted to see this happen) take a look at the code and get concrete tests results?

NickHardeman:
added the following because it seemed strange to have ofSetFloatColor not accept ofFloatColor as an argument
```
void ofSetFloatColor(const ofFloatColor& acolor, float a);
void ofSetFloatColor(const ofFloatColor& acolor);
```

armadillu:
This seems like the right approach to me. The only thing I would add is maybe define a macro somewhere in ofConstants indicating that floatColor is now the default, in case someone is targeting a moving version of OF and can ```#ifdef``` their source in addons/projects accordingly.

Something along the lines of ```#OF_USE_LEGACY_VECTOR_MATH```, but without the option to disable.

I am in the middle of a house move with renovations, but I'll try test this when I have a minute.

NickHardeman:
In this pull request, `ofClear` has also changed to have more consistent arguments with ofSetColor ( as discussed in #264 ). The default alpha arguments of 0 were removed and the default is now opaque (1 or 255 respectively).
 Previously calling `ofClear(255)` would have 0 applied to the clear call since `ofClear( int brightness, int alpha=0)` was the default argument.
https://github.com/openframeworks/openFrameworks/blob/7cf3cfeb99b3e304dfbfe01399b37737c4a8ecd8/libs/openFrameworks/graphics/ofGraphics.h#L254-L278

```
void ofClear(float r, float g, float b, float a);
void ofClear(float r, float g, float b);
void ofClear(float brightness, float a);
void ofClear(float brightness);
void ofClear(const ofColor & c);
void ofClear(const ofFloatColor & c);
void ofClearAlpha();

void ofClearFloat(float r, float g, float b);
void ofClearFloat(float r, float g, float b, float a);
void ofClearFloat(float brightness, float a);
void ofClearFloat(float brightness);
void ofClearFloat(const ofFloatColor & c);
```

NickHardeman:
Screenshot 2023-09-12 at 9 57 29 AM

ofTheo:
I guess with the ofClear thing it might be a breaking change if people were using `ofClear(255)` to clear a transparent fbo right?

NickHardeman:
> I guess with the ofClear thing it might be a breaking change if people were using `ofClear(255)` to clear a transparent fbo right?
> 
> Everything else looks good though as far as I can tell :D

Yes, now the default is opaque. I can revert to 0 alpha as the default argument to avoid the change in functionality.

artificiel:
this should be the occasion to close #264 and #3566 so a forward-looking decision should be made.

the only consensus is that ofClear(int) (single arg) is confusing. that is mostly because the expectations differ widely when it's used as ofBackground, and when it's used in the FBO draw context, and all camps are right.

what about making ofClear(int) (single argument) make alpha 255 if > 0? so people writing ofClear(0) get ofClear(0,0) = transparent black (like @memo wants) and people writing ofClear(255) get ofClear(255,255) = opaque white like @julapy expects. and writing ofClear(128) gets you 128, 255 which seems reasonable. 

in any context, who would expect ofClear(255) to be transparent white? and in a RGBA FBO context, who would expect ofClear(0) to be opaque black?

and make ofClear(r,g,b) opaque — if you're not sending alpha you're not concerned about it, and in an RGBA FBO context you will have the rgba mindset anyway so you'll put the 4 args.

(and thinking about it, deprecating ofClear(int) (single argument) would perhaps be envisioned as it is ambiguous and not such a great timesaver?)

ofTheo:
Thanks @NickHardeman @artificiel 
Looking at the [search results](https://github.com/search?q=ofClear%28255%29&type=code) of `ofClear(255);` 

I think we shouldn't change the current behavior but maybe deprecate in favor of a more clear approach.

That way it's an easy ( non breaking ) merge.

NickHardeman:
ok, the current method has a default argument which makes it strange to deprecate in favor of using a 2 argument function
Current method
`void ofClear(float brightness, float a=0);`
It can be broken out into two methods
```
void ofClear(float brightness, float a);
OF_DEPRECATED_MSG("Use ofClear(brightness, alpha) instead",void ofClear(float brightness));
``` 
And the ofClear(float brightness) still defaults to 0.0f
```
void ofClear(float brightness){
	ofClear(brightness, brightness, brightness, 0.f);
}
```
So using previous calls to `ofClear(255);` should call the `ofClear(float brightness)` which calls `ofClear( brightness, brightness, brightness, 0.0f )` with 0.0f for the default argument.

`ofClear( r, g, b);` now defaults to 255.f instead of the previous method `void ofClear(float r, float g, float b, float a=0); ` which falls in line with `ofSetColor`, where `ofSetColor(gray)` and `ofSetColor(r,g,b);` both default to 255 for alpha.

artificiel:
I'd be curious to know how many of those ofClear(255) are within an FBO, and which are also actually in an alpha blending context, in which they expect transparent white...!

removing the default alpha a above plus deprecating ofClear(int) "single arg" in favour ofClearOpaque(int) and ofClearTransparent(int) that sets a=limit() or 0 respectively (obviously the problem is not that people don't understand the nuance, but to unambiguously match the implicit expectations).

the warning could be something like "use ofBackground to set the background color of the window, or use ofClearOpaque or ofClearTransparent to manage transparency in a blending context".

artificiel:
just re-read the above and actually probably that the ofClear ambiguity stems from glClear, which is not "clearing" in the colloquial sense. the expecting result of the user-action is more akin to "filling".

with the above opaque/transparent naming proposal, it becomes easy to follow, but maybe the should be ofFillOpaque() and ofFillTransparent()? or maybe that's too close to the very different ofFill (which should probably be ofSetFill() BTW).

assuming it stays ofClear*,  ofClearTransparent() and ofClearOpaque() (no argument) could be introduced and map to clear and opaque black. so uses of ofClear(0) would become ofClearTransparent() in a blending context, and ofBackground(0) in a plain drawing context.

the people currently using ofClear(255) in a blending context that cared about the actual color will discover that their white expectations were probably not being met.

NOW, what is strange is ofClearAlpha() sets the alpha to 1... I guess the expectation is "too clear of transparency" implying that "being alpha" is "being transparent", but it's a strange exploitation of the terminology, perhaps coming from someone too close to the glClear usage.

ofTheo:
This all looks good to me! 
Going to merge this in unless there are objections.

dimitre:
Excellent! Happy to see it merged! 
There are some small warnings on the core now, about ofClear with just brightness.

smeyfroi:
@NickHardeman just had to drop a note that this PR is awesome: able to make subtle transparency effects without resorting to glsl for a current project. 👍

dimitre:
Just one minor observation: the warning about ofClear being deprecated can be misleading.
Is there some way of specifying that what is deprecated is ofClear with just one parameter?
Screenshot 2023-11-22 at 16 52 09

artificiel:
there is already a better message:
```c++
OF_DEPRECATED_MSG("Use ofClear(brightness, alpha) instead", void ofClear(float brightness));
```
but as noted in https://github.com/openframeworks/openFrameworks/issues/7673 Xcode seems to mishandle that macro since a while. (I confirm same behaviour in 14.2)

[[deprecated]] works (it's what's used in the new ofRandom stuff) so perhaps a cleanup task of converting bespoke OF_DEPRECATED_MSG to proper [[deprecated]]

PR #7674, opened 2023-09-22
Bugfix Cubemap PBR for mobile devices

NickHardeman:
Some shader optimizations, naming conventions and support for float16 textures to work properly on open gl es mobile devices. 

I have checked on iOS via Safari and DuckDuckGo.
https://nickhardeman.com/temp/OFExamples/MaterialPBR/

This is what should be output
image

### DIFF
--- a/examples/gl/materialPBR/src/main.cpp
+++ b/examples/gl/materialPBR/src/main.cpp
@@ -4,7 +4,7 @@
 //========================================================================
 int main( ){
 // PBR only works with the programmable renderer
-#ifdef TARGET_OPENGLES
+#ifdef TARGET_EMSCRIPTEN
 	ofGLESWindowSettings settings;
 	settings.glesVersion = 3;
 #else
--- a/examples/gl/materialPBR/src/main.cpp
+++ b/examples/gl/materialPBR/src/main.cpp
@@ -3,17 +3,21 @@
 
 //========================================================================
 int main( ){
-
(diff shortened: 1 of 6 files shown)


examples/gl/materialPBR/src/main.cpp
```diff
@@ -3,17 +3,21 @@
 
 //========================================================================
 int main( ){
-
+// PBR only works with the programmable renderer
+#ifdef TARGET_OPENGLES
```
ofTheo, line 7: This might need to be emscripten as PI doesn't like gles 3 
Makes me think we should have some good system for fallbacks, like if you ask for 3 but it can't be provided fallback to 2 with a warning.

ofTheo:
Wowza - this is awesome. 
Made a small note, but everything looks good. Thanks @NickHardeman

examples/gl/materialPBR/src/main.cpp
NickHardeman: Yeah agreed. Would be great if the default was the programmable renderer with the most recent GL version available for your targeted platform. I have yet to test, but iOS is OpenGL ES right? so it should/might work.

ofTheo: Yeah definitely.  A  platform agnostic `windowSettings.glLatest = true;` would be super helpful. 
Probably also good to then talk about defaulting to programmable and having to request legacy GL if needed.

Issue #7665, opened 2023-09-19
Windows Video Player doesn’t report back texture sizes in setup.


danzeeeman:

danzeeeman:
I have a simple app that takes a video texture and maps in onto a sphere and it fails during setup but works in update.

ofTheo:
Ahh yeah this might be with the new Media Foundation player. 
cc @NickHardeman 

going to add this to 0.12.1 

You can always `#define OF_VIDEO_PLAYER_DIRECTSHOW` at the top of ofConstants if you want to use the old player.

NickHardeman:
Ah yes, looks like the events get processed via the update function even during a sync load. So the width and height only get processed after the first update call. I can look into updating it for sync load.

NickHardeman:
Closed by #7675. Width and height are reported back on sync loading.

PR #7675, opened 2023-09-23
Windows Media Foundation Video Player Width and Height

NickHardeman:
Set width and height immediately if loading sync


NickHardeman:
Should fix #7665

Issue #7677, opened 2023-09-25
VS Code as default Linux IDE / better general IDE


ofTheo:
At this point it is a much easier IDE to setup and use, but the documentation for getting it working and the way it's a hidden option in PG adds some pain points. 

We've already added VS Code templates for all the examples by default but making a new project involves a bunch of extra steps on Linux. 

Here are a few of the different steps needed to make it a more usable IDE on Linux. 

TODO: 
- [x] PG: Enable it as a template / project by default in PG. 
- [x] PG: Can add paths to the properties.json to better show errors in code etc. (done, needs testing)
- [x] Packaging: PG frontend is not executable on linux - needs +x 
- [ ] Scripts: running compilePG.sh on Linux should let you know that you can run the frontend GUI without installing it. 
- [x] Site: Guide on using VS Code as an IDE. I know @oxillo has started a [VS Code guide for msys2](https://github.com/openframeworks/ofSite/blob/master/content/setup/vscode.md) which we could add to.

dimitre:
VS Template will be much more useful if we can list the selected addons in the sidebar instead of all addons folder

ofTheo:
Yes @dimitre - agree! 
I think the main thing will be to get it to be a project and not a template - templates basically just get copied over vs being generated.

dimitre:
I'm starting to work on this one @ofTheo 
thinking about adding ofTargetPlatform ```OF_TARGET_VSCODE``` in OF

ofTheo:
Awesome!! Tagging @NickHardeman too as we've been working with VS Code a lot. 
I think most of the work will be inside the `.vscode/c_cpp_properties.json`. 

I think basically doing something similar to the custom `AddonIncludes` array and then including them in the different platform sections. 

<img width="1062" alt="image" src="https://github.com/openframeworks/openFrameworks/assets/144000/cf39c738-8c45-45d3-98c5-529bf3079452">

So you could have: 
```
                "${AddonIncludes}",
                "${ExtraSrcIncludes}",
                "${ProjectIncludes}"
```
etc

NickHardeman:
We have been using the C/C++ Extension and there is a disconnect between the make files and IntelliSense. Not sure of the best way of integrating the two, but a project can compile via make but throw a ton of errors for Intellisense.
[...]

PR #7692, opened 2023-09-29
adding OF_TARGET_VSCODE

dimitre:
to prepare updates in projectGenerator
related:
https://github.com/openframeworks/openFrameworks/issues/7677

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -50,7 +50,8 @@ enum ofTargetPlatform{
 	/// \brief Compiled to javascript using Emscripten.
 	/// \sa https://github.com/kripken/emscripten
 	OF_TARGET_EMSCRIPTEN,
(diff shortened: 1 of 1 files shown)


oxillo:
VSCode is just an IDE. I do not understand why we should target an IDE in OF Core.
I would just keep this local to the projectGenerator.

dimitre:
I agree it is weird mixing ide and platforms.
we can modify projectGenerator so it stops using this ofTargetPlatform entirely, this is one alternative.

projectGenerator is already mixing platform and build systems, like:
```
    OF_TARGET_MINGW,
	/// \brief 32- and 64-bit x86 architecture using Visual Studio on Windows OS.
	OF_TARGET_WINVS,
```

what do you think @ofTheo ?

ofTheo:
Totally agree it should be a PG only define. 

It only really makes sense there as in OF/OF there is no real way to know if a project is being compiled with VS Code vs Make.

PR #7694, opened 2023-10-03
Better vscode json

ofTheo:
As discussed in #7677 this updates the vscode json file to do better code completion and have the VS Code syntax highlighting match make.


scripts/templates/vscode/.vscode/c_cpp_properties.json
oxillo, line 1: This configuration assumes that the project is given location compared to OF_ROOT.
I would rather have an OF_ROOT in the env part that indicates the location of OF relative to the project. That would also make the generation of the template easier.
I would also change the order of OF_LIBS_ROOT and OF_INCLUDE so OF_INCLUDE can depend of OF_LIBS_ROOT

```json
{
  " env" : {
       "OF_ROOT" : "${workspaceFolder}/../../..",
       "OF_LIBS_ROOT":"${OF_ROOT}/libs",
       "OF_INCLUDE":["${OF_LIBS_ROOT}/openFrameworks/**"],

```

For MSYS2 configuration, it is still refering MINGW32...
Let's call it `MINGW64` instead of `Win32` so people can duplicate it for UCRT64 or else
Having "c:/msys64/mingw64/include/**" in includePath is enough and will include all libs such as cairo, glew,...
Compilerpath shall also be "c:/msys64/ming64/bin/g++.exe"

ofTheo: Thanks @oxillo that all sounds like great suggestions and actually will make it easier for @dimitre as he's working on getting proper PG support. So this way he would just need to update OF_ROOT if its not in the default location and the rest will work. 

Will make these changes. 
Thanks!

dimitre: good to merge?

ofTheo:
👍
Feel free to. 🙂

PR #7701, opened 2023-10-08
ofxSvg initialize with filename

dimitre:
This allows for direct initialization and file loading in the likes of ofImage.
Yes I didn't know about this ofImage until recently
```c++
ofxSvg logoSvg { "_material/logo.svg" };
ofImage logo { "_material/logo2.png" };

```

### DIFF
--- a/addons/ofxSvg/src/ofxSvg.h
+++ b/addons/ofxSvg/src/ofxSvg.h
@@ -17,11 +17,10 @@
 
 class ofxSvg {
 public:
-	~ofxSvg();
-	
 	ofxSvg();
 	ofxSvg(const of::filesystem::path & fileName);
-
+	~ofxSvg();
+	
 	float getWidth() const;
 	float getHeight() const;
 
--- a/addons/ofxSvg/src/ofxSvg.h
+++ b/addons/ofxSvg/src/ofxSvg.h
@@ -17,7 +17,11 @@
 
 class ofxSvg {
 public:
-	ofxSvg();
+
+	ofxSvg() = default;
+	~ofxSvg() = default;
+	ofxSvg(const ofxSvg & a) = default;
+
 	ofxSvg(const of::filesystem::path & fileName);
 	
 	float getWidth() const;
--- a/addons/ofxSvg/src/ofxSvg.h
+++ b/addons/ofxSvg/src/ofxSvg.h
@@ -18,6 +18,9 @@
 class ofxSvg {
 public:
 	~ofxSvg();
+	
+	ofxSvg();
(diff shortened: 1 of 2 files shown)


addons/ofxSvg/src/ofxSvg.h
```diff
@@ -17,15 +17,17 @@
 
 class ofxSvg {
 public:
+	ofxSvg();
+	ofxSvg(const of::filesystem::path & fileName);
 	~ofxSvg();
-
+
```
ofTheo, line 23: I would actually get rid of the ofxSvg() and ~ofxSvg() and let them just be default implementations. 

I was going to suggest using the default keyword but I think you don't need any of the other destructors / constructors if you are just defining:
`	ofxSvg(const of::filesystem::path & fileName);`

dimitre: I think ~ofxSvg() is being used, there is a paths.clear() there

dimitre: and I'm wondering if we need ofxSvg() to use it in a traditional way lets say
```
ofxSvg svg;
```

ofTheo: > and I'm wondering if we need ofxSvg() to use it in a traditional way lets say
> ofxSvg svg;

ahh yeah, in that case default constructor, default destructor and default copy constructor might be good too. 
https://en.cppreference.com/w/cpp/language/default_constructor 

> I think ~ofxSvg() is being used, there is a paths.clear() there

that should just happen automatically when the class is destroyed - so I think paths.clear and ~ofxSvg()  implementation is not needed.

dimitre: good! I'll test here

dimitre: Nice! 
destruct the destructor ok
and tidy up file loading. now it correctly checks if file is not found.

ofTheo:
@dimitre - sorry I wasn't able to easily paste what I meant. 
so for the .h you would just do this and for everything with `= default` you don't do anything in the .cpp:

```
class ofxSvg {
	public:
		
		ofxSvg() = default;
		~ofxSvg() = default;
		ofxSvg(const ofxSvg & a) = default;

		//keep as is in your PR: 
		ofxSvg(const of::filesystem::path & fileName);
```

dimitre:
Thank you @ofTheo, check it again please

ofTheo:
Looks great @dimitre - I do see an error with the VS CI checks.
Once that's fixed I think its good to merge :)

Issue #7552, opened 2023-05-30
ofImage loads sRGB 16bpp PNGs, but fails on greyscale 16bpp PNGs using ofx 11.2


bbogart:
I'm working with various deep colour and float images, mostly in opencv, but I sometimes use ofx to render them in various ways. I just realized things were working fine with my RGB 16bpp PNGs (using version 11.2) but my 16bpp greyscale PNGs don't load using ofImage.

Note, I expect the 16bit versions are converted / scaled down to 8bit, which is OK, I'd just like to be able to load and render them without having to manually convert to 8bit.

See this simple test: [16bitTest.tar.gz](https://github.com/openframeworks/openFrameworks/files/11606097/16bitTest.tar.gz)

Since RGB 16bit are working, I suspect this would be an easy fix.

Ben

dimitre:
I've tested here with your PNG in macOS and I can reproduce the issue. the grayscale doesn't draw.
it would be great to be able to work with 16bits greayscale image. 
Did you try using tiff also?
one good place to start is to find if freeimage supports this format, or if it is supported in more recent versions

danoli3:
Updated FreeImage in new libs should have it

dimitre:
closed by https://github.com/openframeworks/openFrameworks/pull/7708

PR #7708, opened 2023-10-13
Load 16 bit grayscale images downsampled to 8bit

NickHardeman:
Related to #7552


dimitre:
Great @NickHardeman ! 
idea: how about notifying with ofLogNotice or ofLogWarning when it is downsampled?

NickHardeman:
Maybe ofLogVerbose?

ofTheo:
Ah but wouldn’t this PR break loading 16 bit grayscale images into ofShortPixels ?

NickHardeman:
The added line only executes if it's an ofImage / unsigned char due to this pixel type check in the if statement. So the functionality for images other than ofImage remains the same.
`if(sizeof(PixelType)==1`
Attached is an example for testing with ofImage, ofFloatImage and ofShortImage for the color and gray images.
[sRGB16bit.zip](https://github.com/openframeworks/openFrameworks/files/12894203/sRGB16bit.zip)

PR #7726, opened 2023-10-19
ofNode with function to set position, rotation and scale

NickHardeman:
The new functions will reduce the number of matrix multiplications when needing to set the position, rotation and scale. 
Previously each call to position, rotation or scale called createMatrix() internally, which resulted in 3 mat4 multiplications.

```
node.setPosition(pos);
node.setRotation(quat);
node.setScale(scale);
```
Would be 9 mat4 matrix multiplications.

With the proposed methods, 
`node.setPositionRotationScale(pos,quat,scale);`
Only performs 3 mat4 matrix multiplications.

### DIFF
--- a/libs/openFrameworks/3d/ofNode.cpp
+++ b/libs/openFrameworks/3d/ofNode.cpp
@@ -283,6 +283,29 @@ glm::vec3 ofNode::getScale() const {
 }
 
 //----------------------------------------
+void ofNode::setPositionRotationScale( const glm::vec3& p, const glm::quat& q, const glm::vec3& s ) {
+	this->position = p;
+	this->orientation = q;
+	this->scale = s;
+	createMatrix();
+	onPositionChanged();
+	onOrientationChanged();
+	onScaleChanged();
+}
+
+//----------------------------------------
+void ofNode::setPositionRotationScale( const glm::vec3& p, const glm::vec3& eulerAngles, const glm::vec3& s ) {
+	this->position = p;
+	glm::quat q(glm::radians(eulerAngles));
+	this->orientation = q;
+	this->scale = s;
+	createMatrix();
+	onPositionChanged();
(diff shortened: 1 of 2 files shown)


artificiel:
there is no `node.setRotation()` method — it's `setOrientation()`... so perhaps go with the flow and `setPositionOrientationScale()`?

also it is useful to have complete argument names like `position`, `orientation`/`orientationEulerAngles` and `scale` as it pops better information in IDEs. (I see the current ofNode impl is using single-letter variables names, which is not ideal — they should also be expanded to complete names)

and I would suggest these overloads as it is frequent to scale linearly (and also frequent not to scale):
```
setPositionOrientationScale(const glm::vec3& position, const glm::quat& orientation, float scale = 1.0f );
setPositionOrientationScale(const glm::vec3& position, const glm::vec3& orientationEulerAngles, float scale = 1.0f );
```

ofTheo:
Could we use this approach and bundle that js file in the folder package?
https://stackoverflow.com/a/72628095

artificiel:
ah apologies I did not notice the member variables were the same (I was concerned only with the interface). as you suggest although `this->` will technically work to disambiguate, shadowing in general is not a good pattern.

i am team-trailing-underscore-for-member-variables (as https://google.github.io/styleguide/cppguide.html#Variable_Names) but this is a more involved topic warranting it's own focus, so forget my parameter rename suggestion and use your original parameters in the style of the current code.

artificiel:
@ofTheo I think your JS comment/link fell into the wrong issue

ofTheo:
Oops!! Thanks 😁

ofTheo:
oops - sorry I thought this was good to merge. 
maybe another PR to close #7726

NickHardeman:
I can send another PR with the updated function names

PR #7742, opened 2023-11-01
macOS: updates to version and custom app icon handling

danomatika:
This PR implements changes to the projectGenerator templates proposed in https://github.com/openframeworks/projectGenerator/issues/429 and https://github.com/openframeworks/projectGenerator/issues/430 but ported here:

* separate Project.xcconfig VERSION into the default MARKETING_VERSION and CURRENT_PROJECT_VERSION variables to match Xcode UI
* add examples to new method for setting custom app icons as well as info in transitioning from previous ICON_FILE_PATH variable

Ping @dimitre @ofTheo


danomatika:
Hmm not sure why it's pulling in that old commit which I believe was already merged last year.

dimitre:
My suggestion is we keep VERSION instead of MARKETING_VERSION, so it won't break lots of projects already using this.
The word marketing itself is kind of strange to me, because I work in a lot of different projects but none of them is a product.
I've been using version a lot to check if changes are in sync in other machines, using syncthing

danomatika:
MARKETING_VERSION is the built-in Xcode variable which is what the Version box in the settings UI uses, regardless of issues with its naming. I found it confusing/annoying that I could no longer set the version and build in Xcode as with any other ios/mac app development. By using VERSION, this common pattern is broken in favor of editing a settings file and, coincidentally broke this in existing projects before 0.12.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Nov 2, 2023, at 2:28 PM, Dimitre ***@***.***> wrote:﻿
My suggestion is we keep VERSION instead of MARKETING_VERSION, so it won't break lots of projects already using this.
The word marketing itself is kind of strange to me, because I work in a lot of different projects but none of them is a product.
I've been using version a lot to check if changes are in sync in other machines, using syncthing

dimitre:
ok, great now I understand. I thought it was just custom variables, but it seems MARKETING_VERSION is actually internally used in XCode.

My only suggestion is keep it uncommented, this way the value set in Project.xcconfig is being reflected on XCode UI automatically

danomatika:
> My only suggestion is keep it uncommented, this way the value set in Project.xcconfig is being reflected on XCode UI automatically

I disagree as IMO you would only set it in the Project.xcconfig if you *do not* set it in Xcode, since I believe it would override the UI. My feeling is that the majority of people would which is what the comments are meant to indicate.

danomatika:
Looking in the [xcconfig reference](https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project/), there doesn't appear to be a way to conditionally set a variable if it's not empty. I was hoping to do something like the following like with `#include? "App.xcconfig"`:
~~~
MARKETING_VERSION = VERSION?
~~~
Unfortunately this doesn't work and the file format is not as advanced as a shell script etc.

ofTheo:
@danomatika I think its okay to leave as is, but just make it clear in comments that it is recommended to set these values in the Xcode UI.  The main use case I can see for needing to set in Project.xcconfig would be if people were using the command line `xcodebuild` command then overriding in `Project.xcconfig` would make sense

dimitre:
Yes but I'm talking about a blank state (recently generated by projectGenerator) and the version is blank.
Everything will be overwritten only App.xcconfig survives projectGenerator updates
so I think it makes sense to use a default uncommented app version to start with for most projects or examples where version doesn't matter

dimitre:
@danomatika sorry if I didnt express correctly, my point was left this line uncommented:
```
MARKETING_VERSION = 0.1.0
```

ofTheo:
So I tried this today. 

Setting the value in the Xcode GUI does work as expected:

image

But if you set it in Project.xcconfig it doesn't override it ( the Gui value overrides the Project.xcconfig value ). 

So I think it's safe to do @dimitre suggesting of having `MARKETING_VERSION` uncounted - otherwise no version is set.

danomatika:
@ofTheo Thanks for checking. So setting the variable in the xcconfig is not a big deal as it acts as a default and is overridden  by the Xcode UI? Good.

Also, I noticed that I used the wrong Build variable name, is should be `CURRENT_PROJECT_VERSION`.

Latest commit corrects these issues.

ofTheo:
Awesome - thanks @danomatika

PR #7746, opened 2023-11-03
fixed regeneration of plist and icns at compile MacOS

Nejrup:
This also fixes the need to 'allow your application to access microphone' at every compile.


ofTheo:
Amazing thanks!!

PR #7756, opened 2023-11-16
ofLoadImage for ofTexture will detect HDR and EXR 

NickHardeman:
Detect and load HDR and EXR to ofTexture from file paths using either ofPixels or ofFloatPixels. Will also set the pixels to 4 channels for broader OpenGL ES support. There is also an added function for adding a flip in y argument.

`bool bLoadOk = ofLoadImage(*tex, "excellentHdrImage.hdr", mirrorY );`

Previously, something like below would be necessary.
```
std::string ext = ofFilePath::getFileExt( "excellentHdrImage.hdr" );
ext = ofToLower(ext);
bool hdr = (ext == "hdr" || ext == "exr");
if( hdr ) {
	ofFloatPixels fpix;
	bLoadOk = ofLoadImage(fpix, apath);
	if( bLoadOk) {
		#if defined(TARGET_OPENGLES)
		// GL_RGB32F, GL_RGBA32F and GL_RGB16F is not supported in Emscripten opengl es, so we need to set to GL_RGBA16F or GL_RGBA32F. But GL_RGBA32F is not supported via opengl es on most mobile devices as of right now.
			if(fpix.getNumChannels() != 4 ) {
				// set alpha to 1.
				fpix.setImageType( OF_IMAGE_COLOR_ALPHA );
			}
		#endif
		if( mirrorY ) {
			fpix.mirror(mirrorY, false);
		}
		tex->loadData(fpix);
	}
} else {
	ofPixels pixels;
	bLoadOk = ofLoadImage( pixels, apath );
	if( bLoadOk ) {
		if( mirrorY ) {
			pixels.mirror(mirrorY, false);
		}
		tex->loadData(pixels);
	}
}
```

### DIFF
--- a/libs/openFrameworks/graphics/ofImage.cpp
+++ b/libs/openFrameworks/graphics/ofImage.cpp
@@ -343,12 +343,42 @@ bool ofLoadImage(ofFloatPixels & pix, const ofBuffer & buffer, const ofImageLoad
 }
 
 //----------------------------------------------------------------
-bool ofLoadImage(ofTexture & tex, const of::filesystem::path& path, const ofImageLoadSettings &settings){
-	ofPixels pixels;
-	bool loaded = ofLoadImage(pixels, path, settings);
-	if(loaded){
-		tex.allocate(pixels.getWidth(), pixels.getHeight(), ofGetGLInternalFormat(pixels));
-		tex.loadData(pixels);
+bool ofLoadImage(ofTexture & tex, const of::filesystem::path& path, const ofImageLoadSettings &settings ) {
+	return ofLoadImage( tex, path, false, settings );
+}
+
+//----------------------------------------------------------------
+bool ofLoadImage(ofTexture & tex, const of::filesystem::path& path, bool bFlipInY, const ofImageLoadSettings &settings){
+	bool loaded = false;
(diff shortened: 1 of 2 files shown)


NickHardeman:
The openFrameworks/tests/io/loadImage/src/main.cpp test seems to be failing because it's loading from HTTPS. I was only able to test with make files on OSX. Failed with HTTPS and success via HTTP image.

NickHardeman:
When I add 
`PLATFORM_LDFLAGS += -lcurl`
to `config.osx.default.mk` then it works correctly via OSX make files.

ofTheo:
oops - yes! that's a much needed fix. feel free to include :)

dimitre:
Nice! I suppose after this PR merged the function ```ofCubeMap::load``` can be simplified

NickHardeman:
@ofTheo ok, is config.osx.default.mk the correct spot to add it? 
@dimitre definitely! Also allows for ofMaterial and ofxAssimp to load exr textures easily, ie normal maps that are exr.

ofTheo:
@NickHardeman yes - definitely the right place. 
probably here: https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/osx/config.osx.default.mk#L171

PR #7767, opened 2023-11-21
App Cateogory : Games

dimitre:
removed the app category entry in .plist
and added in xcconfig


artificiel:
i confirm this works in enabling game mode on an OS-fullscreened window in macOS14

PR #7777, opened 2023-11-30
fix macos runner test - segmentation fault with ofNodeRegression test

ofTheo:
I think this is a caching issue. 
I ran this action in my repo with the cache commented out and it all works fine. 

https://github.com/ofTheo/openFrameworks/actions/runs/7050263655/job/19190560975

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -80,8 +80,8 @@ jobs:
     - name: Download libs
       # continue-on-error: true
       run: ./scripts/$TARGET/download_libs.sh
-    - name: rm-dev
-      run: sudo rm -rf /Library/Developer
+#    - name: rm-dev
+#      run: sudo rm -rf /Library/Developer
 
     # this step is not needed here because the script is empty
     # - name: install


ofTheo:
So confused - in my repo the action is running without error ( left side ) but in this repo its erroring ( right side ) 

image

Might try killing all caching

.github/workflows/build-macos.yml
```diff
@@ -30,59 +30,58 @@ jobs:
...
-          libs/tess2/**
-          libs/uriparser/**
-          libs/utf8/**
-          addons/ofxAssimpModelLoader/libs/**/
-          addons/ofxKinect/libs/**/
-          addons/ofxOpenCv/libs/**/
-          addons/ofxSvg/libs/**/
-          addons/ofxPoco/libs/**/
-
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('scripts/apothecary/apothecary/formulas/**/*.sh') }}
```
ofTheo, line 77: @dimitre I think for this to work correctly the apothecary submodule would need to be updated earlier in this script

dimitre: Glad you found the issue. maybe it was caching a previous version of apothecary libs.

PR #7779, opened 2023-12-03
More .h cleanup

dimitre:
This PR optimizes further the .h includes
And it adds some more ```GLM_FORCE_CTOR_INIT``` where needed

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
@@ -5,6 +5,7 @@
 #include "ofPixels.h"
 #include "ofGraphics.h"
 #include "ofConstants.h"
+#include "ofMatrix4x4.h"
 
 #include <assimp/cimport.h>
 #include <assimp/scene.h>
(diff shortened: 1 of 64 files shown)


dimitre:
Tests are good here, I think it is good to merge. Any second opinion is welcome
@ofTheo @artificiel

artificiel:
looks good -- one minor thing: clang-format sorts the #includes in alphabetical order, as include order should not be critical. if desired groups of includes can be separated by a newline, and they will not be mixed, so it's still possible to separate eg: system includes, OF includes, "my" includes or other conceptual grouping.

(NB: I'm using the automator approach to bind ctrl-opt-cmd-i to reformat the selection as described here: https://www.apulsoft.ch/blog/clang-format-automator-quick-action/)

also i was under the impression that the `GLM_FORCE_CTOR_INIT` issue was to be "correctly" solved with a project flag? because otherwise even if the core is totally consistent, it does not guarantee addons or library will get the #define prior to their use?

dimitre:
I agree with GLM_FORCE_CTOR_INIT and I can do that separately on projectGenerator. The complicated side of this operation is we have to test manually all platforms because it will compile OK and potentially even pass tests when using non proper initialized GLM values, so I'll need help on this one.
As we don't have this functionality now I'll be adding everywhere glm is being included and there is not ofConstants.h included

I agree with organizing includes order, separating with line breaks. I'm temporarly using ofConstants.h in the end (of OF includes)  to assure less and less nested dependencies among files, even if it is more declarative. 

I can make a subsequent PR with clang format and includes order to tidy up and be easier to review

dimitre:
Nice one @ofTheo we can use something like pragma warning

PR #7784, opened 2023-12-04
More .h cleanup #6

dimitre:


dimitre:
PR is now passing tests. I think it is good to merge,
opinions are welcome @artificiel @ofTheo

artificiel:
ah! it looks good on 2 lines, keeping the function at it's correct indentation. I vote for that.

there are some comments that perhaps predate the first deprecation such as:
```c++
    // getInterpolated
    [[deprecated("Use member method getInterpolated() instead.")]]
    ofVec2f interpolated( const ofVec2f& pnt, float p ) const;
```
they can be also removed -- and actually perhaps also even remove the doxygen for the deprecated stuff?

and also maybe move the deprecated functions together at the bottom of the .h so they don't clutter the IDE helper menus and split them with a `// MARK: DEPRECATED CEMETERY` so it gives a hint in the IDE. I don't see a problem with keeping deprecated methods around if they don't hinder code reading or productivity (of course this is assuming maintaining the deprecated is zero overhead).

if we want to "time out' the deprecations it would make sense to add a deprecation date, and determine that every over N years stuff can get deleted:

```c++
    [[deprecated("Use member method getInterpolated() instead.")]] /* 2023-12-04 */
```
there is also the possibility of instead of being friendly and calling the new functions within the deprecated, have the deprecated `static_assert` false. this is a bit more gentle than ripping the function out, and it forces the adoption of the new method, else you don't compile.

so friendly-deprecate for N years, then `static_assert(false,"message")` after N years, then rip-out after N*2 or something like that?

thinking about it... adding the date (even maybe just the year to keep it simple) is probably not a bad idea whatever the policy ends up to be as it's easier than digging git-blame to figure it out later.

artificiel:
ah sorry: `static_assert` is applied even not-called code; it will not work for this kind of non-template usage, but it's the idea of throwing an error if the function is used (maybe it's only runtime).

dimitre:
Nice. I've removed the // getInterpolated, maybe it was just a helper leftover from who was writing OF_DEPRECATED_MSG
I agree with  moving down every deprecated comment and marking , it can be done in a separate PR in all files at once, after we finish converting to [[deprecated]]

My personal opinion is all about removing deprecated code after a while (like the idea of expiration date) 
but I know OF books were written with previous API and it is something to consider

artificiel:
it's true that it's better if books still compile. as I said I see no problem in carrying long-deprecated stuff, as long as it's not in the way (when 25% of a header is deprecation it makes it hard to read through if you are relying on the header as "documentation"; if they are lumped at the end it's more clear that you can stop worrying when your reach them).

Issue #7780, opened 2023-12-03
guiExample / ofParameter ofColor error :  no matching function for call to 'ofSetColor'


dimitre:
```
examples/gui/guiExample/src/ofApp.cpp:56:2: error: no matching function for call to 'ofSetColor'
        ofSetColor(color);
        ^~~~~~~~~~
../../../libs/openFrameworks/graphics/ofGraphics.h:56:6: note: candidate function not viable: no known conversion from 'ofxColorSlider' (aka 'ofxColorSlider_<unsigned char>') to 'int' for 1st argument
void ofSetColor(int gray);
     ^
../../../libs/openFrameworks/graphics/ofGraphics.h:71:6: note: candidate template ignored: could not match 'ofColor_' against 'ofxColorSlider_'
void ofSetColor(const ofColor_<T> & acolor) {
     ^
../../../libs/openFrameworks/graphics/ofGraphics.h:57:6: note: candidate function not viable: requires 2 arguments, but 1 was provided
void ofSetColor(const ofColor & acolor, int a);
     ^
../../../libs/openFrameworks/graphics/ofGraphics.h:39:6: note: candidate function not viable: requires 3 arguments, but 1 was provided
void ofSetColor(int r, int g, int b);
     ^
../../../libs/openFrameworks/graphics/ofGraphics.h:55:6: note: candidate function not viable: requires 4 arguments, but 1 was provided
void ofSetColor(int r, int g, int b, int a);
     ^

```

artificiel:
it seems this template (seems recent, with the #7650) fails to be "deduced" as a candidate for the ofColor operator of ofxColorSlider because it tries to process the ofxColorSlider itself:

https://github.com/openframeworks/openFrameworks/blob/dd09e975ad1166ea0cb48edeae35c1f75dc9ed4b/libs/openFrameworks/graphics/ofGraphics.h#L70-L74

https://github.com/openframeworks/openFrameworks/blob/dd09e975ad1166ea0cb48edeae35c1f75dc9ed4b/addons/ofxGui/src/ofxSliderGroup.h#L52

adding an explicit `ofSetColor(ofColor aColor);` method (one without the alpha param) fixes the specific case of the reported bug, but not sure how to properly write things for the templated ofxColorSlider to follow the templated ofColor as it is presumed it should?

NickHardeman:
[...]

PR #7788, opened 2023-12-05
ofSetColor functions without Template

NickHardeman:
Explicit functions for different ofColor types.
Related to #7780


dimitre:
+1 for this one. seems good (I didnt' test yet)

PR #7803, opened 2023-12-08
Emscripten cors

Jonathhhan:
This PR makes it possible to run multi thread Emscripten without configuring the server with:
```
Header set Cross-Origin-Embedder-Policy "require-corp"
Header set Cross-Origin-Opener-Policy "same-origin"
```
Now I add only:
```
RemoveType .gz
AddEncoding x-gzip .gz
AddType application/wasm .wasm
```
To my `.htaccess` file.
See the bottom of the page: https://emscripten.org/docs/compiling/WebAssembly.html
It is not necessary, but `AddType application/wasm .wasm` avoids a warning and seems to be faster.
The suggested `Options Multiviews` does not work for me, so I leave it away.
Would be nice, to set that with the script too (but my attempt did not work).

Multithreading is now set as default (maybe change the default to `USE_PTHREADS=0` like it was before?).

### DIFF
--- a/addons/ofxEmscripten/libs/cors/coi-serviceworker.js
+++ b/addons/ofxEmscripten/libs/cors/coi-serviceworker.js
@@ -1,0 +1,143 @@
+/*! coi-serviceworker v0.1.7 - Guido Zuidhof and contributors, licensed under MIT */
+let coepCredentialless = false;
+if (typeof window === 'undefined') {
+    self.addEventListener("install", () => self.skipWaiting());
+    self.addEventListener("activate", (event) => event.waitUntil(self.clients.claim()));
+
+    self.addEventListener("message", (ev) => {
+        if (!ev.data) {
+            return;
+        } else if (ev.data.type === "deregister") {
+            self.registration
+                .unregister()
+                .then(() => {
+                    return self.clients.matchAll();
+                })
+                .then(clients => {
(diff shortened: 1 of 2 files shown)


ofTheo:
Awesome - thanks @Jonathhhan !!
Should we leave pthreads on and see if it affects things?  

so in theory no .htaccess file is needed ( apart for the warnings / loading speed ) ?

Jonathhhan:
@ofTheo 

> Should we leave pthreads on and see if it affects things?
> 

I would say yes.

> so in theory no .htaccess file is needed ( apart for the warnings / loading speed ) ?

Exactly (tested it, but only with one Computer).

This is the warning without `AddType application/wasm .wasm` (which is also there without multi threading): 
```
(Index):1 wasm streaming compile failed: TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
printErr @ (Index):1
(anonym) @ index.js:1
Promise.then (async)
(anonym) @ index.js:1
Promise.then (async)
instantiateAsync @ index.js:1
createWasm @ index.js:1
(anonym) @ index.js:1
(Index):1 falling back to ArrayBuffer instantiation
```

ofTheo:
Great!! thanks @Jonathhhan! 
Yes, got those same warnings too - good to know the way to deal with them.

PR #7805, opened 2023-12-09
Bugfix vbo mesh instanced example emscripten

NickHardeman:
Relies on #7804


NickHardeman:
Not sure why this is showing super old commits.

PR #7807, opened 2023-12-09
Latest Libraries Download / Install -b (Automation Test) 

danoli3:
This test will install libraries automatically using the latest library builds with latest fixes, security, latest compile flags, c++17

Changes to download_libs to use -b 

Initial Test:
- Visual Studio 2022 (x64 / ARM64)
- Linux (64)


danoli3:
Tests successful

PR #7809, opened 2023-12-09
Fix vbo instance example to run with Emscripten

NickHardeman:
Relies on #7808 

https://nickhardeman.com/temp/OFExamples/vboMeshDrawInstancedExample/


Jonathhhan:
Nice.

PR #7815, opened 2023-12-11
More .h work #7

dimitre:
Some more .h cleanup
Removing previously commented out includes
Due to recent merged PR [[deprecated]] some inclusions of ofConstants.h are not needed anymore.
There is opportunity of removing more (some functions header defined for specific targets)

both submodules updated to latest

### DIFF
--- a/libs/openFrameworks/app/ofAppBaseWindow.h
+++ b/libs/openFrameworks/app/ofAppBaseWindow.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "ofWindowSettings.h"
+// MARK: Target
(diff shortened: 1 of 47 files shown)


ofTheo:
@dimitre - just wanted to say huge thanks for all this. 
Noticeably faster compile times on all platforms I've tested with so far. 🙏

dimitre:
Yes I'm pretty happy about that too. 
Obviously there are lots of other improvements counting and maybe github actions are faster now, 
but it is great to see msys2 time now running in 15 mins instead of 28 like last year

PR #7819, opened 2023-12-12
VS download addons fix paths

danoli3:
Fixes addon paths when using download_libs.sh 

https://github.com/openframeworks/apothecary/issues/315


dimitre:
@danoli3 I'm having other issue here on latest master and download_libs.sh
some __imp_ errors like:
unresolved external symbol __imp_uriParseUriA

PR #7821, opened 2023-12-13
Dev Scripts - Download Project Generator Latest / Download Latest Libraries macOS/vs

danoli3:
Download Project Generator nightly binary and installs it to oF Root 
- scripts/osx/download_projectGenerator.sh 
- scripts/vs/download_projectGenerator.sh 

Download Latest Libraries macOS - Bleeding
- scripts/osx/download_latest_libs.sh 

Download Latest Libraries vs - Bleeding
- scripts/vs/download_latest_libs.sh 

Existing download_libs.sh no longer used for targetting -b bleeding

### DIFF
--- a/scripts/dev/download_libs.sh
+++ b/scripts/dev/download_libs.sh
@@ -154,17 +154,27 @@ EOF
             ARCH=clang64
         fi
     fi
+
+    if [ "$PLATFORM" == "osx" ]; then
+        ARCH=x86_64
+    fi
 fi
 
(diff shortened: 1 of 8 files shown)


danoli3:
@dimitre this work for you ?

dimitre:
GREAT! That is a very useful addition. 
I think we should test and maybe change the guide install from git
I'll do some testing soon
we can think of removing projectGenerator and apothecary submodules after this one. What do you think @danoli3 ?

PR #7826, opened 2023-12-14
ofxAssimp  ADDON_ADDITIONAL_LIBS from ofxAssimpModelLoader

dimitre:


dimitre:
Even not passing the tests (vs test is failing for other resons) I'll be merging this so it is bundled in next nightly

Issue #7834, opened 2023-12-22
[actions] build-linux64 tries to upload to ci.openframeworks.cc


dimitre:
Maybe it is unnecessary now:
```yml
    - name: Upload Libs
      run: scripts/ci/upload_of_lib.sh;
      env:
        GA_CI_SECRET: ${{ secrets.CI_SECRET }}
```

ofTheo:
oh yeah! definitely.

dimitre:
closed by https://github.com/openframeworks/openFrameworks/pull/7835

PR #7835, opened 2023-12-23
[actions] remove updating ci.openframeworks.cc in linux

dimitre:
closes 
- https://github.com/openframeworks/openFrameworks/issues/7834

projectGenerator submodule update to latest


ofTheo:
@dimitre - I think this is good to merge. 
the msys2 check failing has been like that for a few days

PR #7837, opened 2023-12-23
Removing Boost paths from OF XCode project

dimitre:
As minimum requirement is now c++14 instead of c++11 we can remove boost from core


dimitre:
In this PR I've removed the boost links, so it will compile faster
and I've organized better the xcconfig files, so they are similar in tvos, macos and ios. 
removed duplicate entries so there will be no shadowing on ```GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR``` for example.
Ready to merge

danoli3:
Yeah sweet lets get this is in, i'm about to rework a lot of the configs as well with all the new xcframeworks and libs

PR #7838, opened 2023-12-23
comment out boost from install scripts, install_dependencies.sh using one line apt get

dimitre:
This makes linux actions runs in less time.


dimitre:
@danoli3 I want your opinion on this one as it changes install scripts
I've noticed if we install one package at a time it spends some time on "reading database" so if we convert this to oneliner it can shave off about 2 minutes for installing dependencies.
Do you see any downside?
this is the specific change:
```diff
- installPackages ${PACKAGES}
+ apt-get -y -qq install ${PACKAGES}
```

dimitre:
This is finally ready to go @ofTheo. 
Don't mind the number of commits I've splitted this PR in other ones.

PR #7842, opened 2023-12-26
[actions] msys2 fix cannot find -lglfw3dll

dimitre:
This one fix the issue with msys2 cannot find -lglfw3dll issue
And preinstall dependencies, caching using pacboy packages speeding up tests.
it also stop installing automatically unused items like poco, boost and tools


dimitre:
@ofTheo @danoli3 I'll be merging this one so we have msys2 tests back. 
Please review the changes when you have time to see if it is ok.

PR #7849, opened 2024-01-03
Make - using automatic number of cores

dimitre:
two changes:

internally build core using multiple cores on make (in compile.project.mk)
and change the -j2 to -j to detect automatic number of available cores

this can improve github actions, and overall performance


dimitre:
@ofTheo I'm merging this one as it have great effect on compile times. 
Core is now being compiled with multiple threads using make

ofTheo:
awesome!! thanks @dimitre 🤘

PR #7860, opened 2024-01-10
[actions] - testing one action with multiple jobs for tests

dimitre:
All jobs in one action.
advantages: 
all jobs will be cancelled if any one errors
less verbosity to read on actions tab


dimitre:
another good thing is all actions can be cancelled clicking one cancel only

Issue #7760, opened 2023-11-19
[macOS][bug] verticalSync problem since macOS13 (GLWF-inherited issue)


artificiel:
macOS13 introduces jitter in the timing of update/draw cycle of GLFW apps https://github.com/glfw/glfw/issues/2249. it has (at least) 2 concrete negative effects in OF as demontrated in https://github.com/openframeworks/openFrameworks/issues/7752#issuecomment-1817035425

- sometimes the jitter pushes the `draw()` later than the actual frame cycle, stuttering/dropping a frame

 - `ofGetElapsedTimef()` is not reliable to control phase-based animations requiring isochronous steps

until it is fixed within GLFW ([seems hard](https://github.com/RustyMoyher/glfw_MacVsyncFix)) — or at OF-level by working around the Timer ([as inspiration see the google deepmind workaround](https://github.com/google-deepmind/mujoco/commit/2f0fb1e4ef81ae4eb281e09fb67c6b36f27dc427)) , or by implementing a metal backend, or other — `ofSetVerticalSync(false)` and using `ofSetFrameRate(nnn)` is recommended. using [macOS14 game mode in fullscreen](https://eclecticlight.co/2023/10/18/how-game-mode-manages-cpu-and-gpu/)  also helps as seen in https://github.com/openframeworks/openFrameworks/issues/7752#issuecomment-1817035425.

the uneven timing issue can be worked around by using something like [`ofGetExpectedPaintTimef()`](https://github.com/openframeworks/openFrameworks/issues/7758), but that will not prevent dropped frames.

this issue will be updated as the GLFW and OF evolution affects the behaviour.

danomatika:
Does this issue happen with both `ofSetVerticalSync(true)` and `ofSetFrameRate(60)`? I don't *quite* remember why, but I feel like there was an issue with sync only and I added setting the frame rate as well.

artificiel:
yes, adding the framerate is better top the point of things being mostly unnoticeable, if the CPU load is low and if you don't relay of ofGetElapsedTimef()

timing with vsync, frameRate(60), OF fullscreen:
<img width="1374" alt="image" src="https://github.com/openframeworks/openFrameworks/assets/2172370/d10d82de-eba6-4a33-987e-74ab2a67ff35">

timing without vsync,  frameRate(60), OF fullscreen:
[...]

PR #7867, opened 2024-01-18
FPS timing with chrono

dimitre:
This PR has the intent of modernize OF framerate managing. 
The idea is using only std::chrono for two reasons : let std::chrono manage all time divisions in the best way possible, and unifying code across platforms.
it was tested in macOS and reduced FPS drift compared to ofTimer implementation


dimitre:
cc @artificiel

ofTheo:
@dimitre - Awesome! 
Might be good to test on Windows and Linux before merging.

dimitre:
Great! I'll be testing in the next days and report back here.

dimitre:
@artificiel can you please test this PR with your tests mentioned here?
- https://github.com/openframeworks/openFrameworks/issues/7760#issuecomment-1820046935

I think this PR can be a slight improvement to actual FPS. 
and here is also an alternative way of counting FPS

```c++
#include <chrono>
using namespace std::chrono;
using namespace std::chrono_literals;

struct fpsCounter {
public:
	int nAverages = 20;
	using space = std::chrono::duration<long double, std::nano>;
	time_point<steady_clock> lastTick;
	steady_clock::duration onesec = 1s;
	std::vector <space> intervals;
	space interval;
	space average;
	bool firstTick = true;
	int cursor = 0;

	void tick() {
		if (firstTick) {
			firstTick = false;
			lastTick = steady_clock::now();
			return;
		}

		interval = steady_clock::now() - lastTick;
		lastTick = steady_clock::now();
		if (intervals.size() < nAverages) {
			intervals.emplace_back(interval);
		} else {
			intervals[cursor] = interval;
			cursor = (cursor+1)%nAverages;
		}
	}
	
	double getFps()  {
		average = std::reduce(intervals.begin(), intervals.end());
		return (double)intervals.size() * onesec / average;
	}
	
	float get() {
		average = std::reduce(intervals.begin(), intervals.end());
		return (float)intervals.size() * onesec / average;
	}
};
```

dimitre:
finally a decent way of testing fps drift, measuring with ofGetFrameRate, and with chrono.
you can toggle (pressing key "t") between OF classic FPS control and new one to see differences in drift.
```ofw/apps/devApps/fps```
Screenshot 2024-05-08 at 17 02 29

dimitre:
now alternative fps counter has a subtle low pass filter (average of 4 last results) so both modes can be compared better. 
it is a great improvement in macOS. I'll be testing on Ubuntu soon 
Screenshot 2024-05-09 at 00 36 51

dimitre:
Now finally tested in Windows, Linux and macOS, 
Tests show up to 25x more precision, from 10 milliframes drift to 0.3

The idea is using sleep_until and wake processor early (about 35ms early), 
so processor has time to go from low energy (sleep) to high energy and calculate the remaining time in a while loop.

PR #7872, opened 2024-01-25
Remove -mwindows linker flag for msys2 build and adjust OPT flags

Segfault1602:
On MSYS2 build, the -mwindows flag was preventing the logs from appearing in the terminal.

Also noticed that some release build were building optimizing for size instead of speed on MSYS2 and OSX.

### DIFF
--- a/libs/openFrameworksCompiled/project/msys2/config.msys2.default.mk
+++ b/libs/openFrameworksCompiled/project/msys2/config.msys2.default.mk
@@ -129,9 +129,6 @@ ifeq ($(findstring MINGW64,$(MSYSTEM)),MINGW64)
 	PLATFORM_LDFLAGS += -Wl,--disable-dynamicbase,--disable-high-entropy-va,--default-image-base-low
 endif
 
-ifndef DEBUG
-	PLATFORM_LDFLAGS += -mwindows
-endif
 ifeq ($(findstring OF_USING_STD_FS, $(PLATFORM_DEFINES)),OF_USING_STD_FS)
 	PLATFORM_LDFLAGS += -lstdc++fs
 endif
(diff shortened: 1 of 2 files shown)


dimitre:
I'll be merging this one, as I've seen it was already discussed on the forums

danoli3:
Cool lets see how this goes. 
```
Code Generation Options

-O0, -O1, -O2, -O3, -Ofast, -Os, -Oz, -Og, -O, -O4
Specify which optimization level to use:

-O0 Means “no optimization”: this level compiles the fastest and generates the most debuggable code.

-O1 Somewhere between -O0 and -O2.

-O2 Moderate level of optimization which enables most optimizations.

-O3 Like -O2, except that it enables optimizations that take longer to perform or that may generate larger code (in an attempt to make the program run faster).

-Ofast Enables all the optimizations from -O3 along with other aggressive optimizations that may violate strict compliance with language standards.

-Os Like -O2 with extra optimizations to reduce code size.

-Oz Like -Os (and thus -O2), but reduces code size further.

-Og Like -O1. In future versions, this option might disable different optimizations in order to improve debuggability.

-O Equivalent to -O1.

-O4 and higher - Currently equivalent to -O3
```

We initially choose -Os after I had discussions with some Apple engineers on what they use for their libraries
-Os is also O2 opitimization. 

-O3 Like -O2, except that it **enables optimizations that take longer to perform or that may generate larger code (in an attempt to make the program run faster).** The may larger code thing here was the main factor.

Since Payload size is more a problem for iOS apps (<150/200mb was 50mb before ) for non wifi downloads / appstore size. 
I think -Os still probably the most ideal. 

Happy to leave this in for now and I'll get app size comparisons over the next weeks! For the Libraries we are using -OS. 
However, actually I don't think we did science and checked the sizes. So I will do that

Good contribution @Segfault1602 !!!

Segfault1602:
Unless I mess that up, iOS and Android app should still be set at -Os.

PR #7876, opened 2024-01-29
Update iPhoneGuiExample by using the storyboard

azuremous:
Update by using the storyboard, limit the size of the GUI view to make touch events possible in OF.


dimitre:
thank you!

PR #7880, opened 2024-01-31
Update iosOrientationExample with UIAlertController

azuremous:
Replace UIActionSheet with UIAlertController and UIAlertAction


dimitre:
Thanks!

PR #7885, opened 2024-02-03
Fix #7874 : [MSYS2] remove GLM from PKGCONFIG libs

### LINKED ISSUE #7874: [msys2] Recent GLM update doesnt carry its pkg-config metadatas
The install script will pick latest packages, it seems glm package doesn't embed the relevant pc file, thus breaking the further make commands.

It builds fine rolling back to glm-0.9.9.x, like so:

```
wget https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-glm-0.9.9.8-2-any.pkg.tar.zst
pacman -U mingw-w64-x86_64-glm-0.9.9.8-2-any.pkg.tar.zst --noconfirm
```

oxillo:
Fix #7874 : [MSYS2] remove GLM from PKGCONFIG libs

As of GLM 1.0, .pc file not provided anymore. However, it was referencing standard include path. Should work also with previous version of GLM.


danoli3:
@oxillo can you confirm working now on master

Issue #7891, opened 2024-02-09
[actions] - linux arm failing : GPG Error NO_PUBKEY


dimitre:
```
W: GPG error: http://deb.debian.org/debian bullseye InRelease: 
The following signatures couldn't be verified because the public key 
is not available: NO_PUBKEY 0E98404D386FA1D9 NO_PUBKEY 
6ED0E7B82643E131 NO_PUBKEY 605C66F00D6C9793
```

dimitre:
@danoli3 I'm not sure if it is related but the latest raspberry pi os now is bookworm

No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm

danoli3:
Yeah massive Raspberry Pi changes for the RiPi5 post launch.

They went nuclear on the APIs and other repos so will need to update

dimitre:
we can try to add this step in GH action as a measure to have them running
```bash
sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 0E98404D386FA1D9
sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6ED0E7B82643E131
sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 605C66F00D6C9793
```

PR #7897, opened 2024-02-12
[actions] fix RPI

dimitre:
new pubkeys - close https://github.com/openframeworks/openFrameworks/issues/7891
comment out linuxarmv6l - it should be updated in code to use glfw instead

### DIFF
--- a/.github/workflows/build-linux-arm.yml
+++ b/.github/workflows/build-linux-arm.yml
@@ -44,7 +44,7 @@ jobs:
         # key: ${{ matrix.os }}-${{ matrix.type }}
         key: ${{ matrix.cfg.target }}
     - name: New Keys
-      run: 
+      run:
         - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 0E98404D386FA1D9
         - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6ED0E7B82643E131
         - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 605C66F00D6C9793
--- a/.github/workflows/build-linux-arm.yml
+++ b/.github/workflows/build-linux-arm.yml
@@ -44,11 +44,11 @@ jobs:
         # key: ${{ matrix.os }}-${{ matrix.type }}
         key: ${{ matrix.cfg.target }}
     - name: New Keys
-      run:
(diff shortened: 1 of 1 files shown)


ofTheo:
so this means 64bit only for arm - I guess that's okay :)
I think we had discussed this before but just want to make sure we're okay dropping non 64bit for arm.

danoli3:
From the Raspberry Pi 2 Model B v1.2 onwards, including the Raspberry Pi 3 and newer models, the devices come with ARMv8 processors capable of running in both 32-bit (ARMv7) and **64-bit (ARM64)** modes. However, the default operating system for these devices, Raspberry Pi OS, was primarily offered in a 32-bit version until recent updates began to support 64-bit operation as well.**

So we only devices not capable of running ARM64 software:

ARMv6 architecture (specifically, the ARM1176JZF-S processor)
- Raspberry Pi 1 models and the 
- Raspberry Pi Zero series are based on the .

- Raspberry Pi 2 Model B v1.1 utilizes an ARMv7 architecture (with the Cortex-A7 processor), which also does not support ARM64.

ARM64 supporting Pi Models:
Raspberry Pi 3 Model B - The first in the lineage to support ARM64, though its default OS has long been 32-bit.
Raspberry Pi 3 Model B+ - Enhanced connectivity and performance over its predecessor, with ARM64 support.
Raspberry Pi 3 Model A+ - A more compact and cost-effective variant, still wielding ARM64 capabilities.
Raspberry Pi 4 Model B - A significant leap in performance, with up to 8GB of RAM, supporting ARM64 natively.
Raspberry Pi 400 - A unique keyboard-integrated computer, fully supporting ARM64.
Raspberry Pi Zero 2 W - A compact, yet powerful device for its size, bringing ARM64 to the Zero series.
Raspberry Pi 5 - Just released BCM2712 quad-core Arm Cortex A76 processor @ 2.4GHz (3x faster than Pi4 btw)

Yeah I think we good to drop armv6 armv7

danoli3:
Make sure scripts / install and deploy scripts have execute privileges

dimitre:
@ofTheo I think if we update armv6l and armv7l to use GLFW directly instead of EGL we can re-add the runners

ofTheo:
okay - that would mean we'll have to disable armv6 from nightly builds too I think. 
maybe I'll take a quick stab at getting armv6 runners working again and if not will merge this PR?

ofTheo:
fixed by #7901

Issue #7865, opened 2024-01-18
audioOutputExample crashes in iOS


lmtanco:
Hi there! 

I am using nightly version of_v20240109_ios_release. When running audioOutputExample it crashes, tested on iOS 16.7.2, iPad Pro, built with Xcode 14.3.1. 

I put breakpoints in `setup`, `update` and `draw`. The breakpoint in `draw` is sometimes reached, most of the time it's not. It crashes in a Thread for which I have no source code, see screenshot. 

![image](https://github.com/openframeworks/openFrameworks/assets/1093084/4f4549c0-a49e-4780-8047-8413c53d8e0a)

Many many thanks, 

Luis.

lmtanco:
Testing on iPhone 8, iOS 16.3, the crash seems to consistently be in ofGLRenderer::drawRectangle: 

![image](https://github.com/openframeworks/openFrameworks/assets/1093084/4e4ee422-8fa9-4ac2-a079-0e391296d9a0)

danoli3:
Try Programmable Renderer 
main.mm:
settings.glesVersion = OFXIOS_RENDERER_ES2; // type of renderer to use, ES1, ES2, etc.

Try and offset the Draw for 3 frames (int counter) - debug if its not initialised yet in 
Make sure OpenGL ES is added in project in frameworks


Alternative try and see if ios/GLKit example works for your setup

lmtanco:
Hi there! 
@danoli3 sorry I totally missed your message of Jan 22! 🙏 
This no longer crashes in my iPhone 8 iOS 16.3, and is fixed in the current nightly thanks to #7902
![IMG_881442ED6AE9-1](https://github.com/openframeworks/openFrameworks/assets/1093084/c8fe4f40-33c1-42af-8d4f-02876e3ffabd)

Many many thanks! ❤️ 

P.S. It doesn't actually sound!  I want to test it on the iPad which I don't currently have with me, then if I'm sure it's not my fault I would open a different issue?

PR #7902, opened 2024-02-19
audioOutputExample: Initialize sampleRate and refactor lAudio, rAudio to std::vector

azuremous:


azuremous:
I think this will solve this problem as well.
[#7865](https://github.com/openframeworks/openFrameworks/issues/7865)

PR #7907, opened 2024-03-10
download_libs downloader using wget2

dimitre:
So all files can be downloaded faster, in parallel
if wget2 is installed locally
if not it will fallback to curl, then wget
cc @danoli3


danoli3:
Nice1!

dimitre:
I think it is finally passing tests. I've simplified the functions to invoke the downloader, so multiple files can be passed easily. wget2 and curl download faster in parallel, wget download in series.

ofTheo:
@dimitre if this is good to merge - looks good to me :)

scripts/dev/download_libs.sh
```diff
@@ -9,278 +12,284 @@ BLEEDING_EDGE=0
...
 
+PARAMS=""
 for PKG in $PKGS; do
-    download $PKG
+	PARAMS+="https://github.com/openframeworks/apothecary/releases/download/$REPO/$PKG "
 done
+# echo " PARAMS $PARAMS"
+
+
+if command -v wget2 2>/dev/null; then
```
danoli3, line 230: Should just do this in downloader 
https://github.com/openframeworks/openFrameworks/blob/master/scripts/dev/downloader.sh#L25

scripts/msys2/buildAllExamples.sh
danoli3, line 1: do we need 755 
"755 means you can do anything with the file or directory, and other users can read and execute it but not alter it. Suitable for programs and directories you want to make publicly available. 644 means you can read and write and execute the file or directory and other users can only read it"

scripts/dev/download_libs.sh
danoli3: if wget2, else curl, else wget 
```
if command -v wget2 2>/dev/null; then 
        wget2 $@ 2> /dev/null;
elif command -v curl 2>/dev/null; then 
        curl -LO --retry 20 -O -s $@; 
else 
        wget -q $@ 2> /dev/null; 
fi
```

danoli3 (changes requested):

scripts/msys2/buildAllExamples.sh
dimitre: reverting this one. I've used chmod +x to execute on my system, to test some stuff on macOS from msys2 scripts

dimitre:
Closing this in favor of https://github.com/openframeworks/openFrameworks/pull/8002

PR #7908, opened 2024-03-10
moving ofVAArgsToString from ofUtils to ofLog

dimitre:
it is the only place that uses in OF Core, so it is ok to move
so it is less warnings of snprintf and less inclusion of templates in other files.


dimitre:
@ofTheo opinion on this one?

ofTheo:
Thanks @dimitre - looks good to me! :)

Issue #7791, opened 2023-12-05
RPi5 Bookworm arch64 / incompatibility list


siteswapjuggler:
So far so good but here come the first errors : 

- svgExample doesnt compile, see [svgExample_logs.txt](https://github.com/openframeworks/openFrameworks/files/13567090/svgExample_logs.txt)
- shader examples crash when the program run either from CLI with the "sudo startx [prog] -- -s off" command line or desktop environment

fieldOfView:
It seems that the svgExample can be fixed by editing `addons/ofxSvg/addon.config.mk`:

Add 
```
linuxaarch64:
        ADDON_LIBS = libs/svgtiny/lib/linuxaarch64/libsvgtiny.a
        ADDON_LIBS += libs/libxml2/lib/linuxaarch64/libxml2.a
```
after the similar chunk for `linuxarmv7l`

siteswapjuggler:
I'll give it a try thanks a lot for your answer :)

siteswapjuggler:
Yep it works perfectly with this editing, we should probably search for other missing references to xml and svgtiny for linuxaarch64 in the other addons too.

Thanks for resolving that :)

fieldOfView:
> we should probably search for other missing references to xml and svgtiny for linuxaarch64 in the other addons too

I searched for similar missing linuxaarch64 lemmas, but could not find any. So I created a PR: #7917

fieldOfView:
Fix for the SVG addon has been merged.

siteswapjuggler:
Hi I did test the fix for the SVG with a fresh install of the latest commit, it's ok to me. 

By th way the original issue was listing two different problems so there is still one to go. But I imagine we can close this one and start a new one with this remaining error : 

* shader examples crash when the program run either from CLI with the "sudo startx [prog] -- -s off" command line or desktop environment

PR #7917, opened 2024-04-21
Fix ofxSvg on 64 bit ARM linux architecture

fieldOfView:
This PR fixes the bundled ofxSvg running on 64 bit ARM linux, eg Raspberry Pi running a 64 bit OS.

I have checked other missing linuxaarch64 lemmas for other included addons, but could not find any sofar.

Contributes to https://github.com/openframeworks/openFrameworks/issues/7791


fieldOfView:
I don't see how my change could make the linuxaarch64 build fail like it does. Is that build otherwise healthy in the main branch?

dimitre:
Thank you @fieldOfView

PR #7930, opened 2024-05-06
ofAVEngineSoundPlayer fix of::fs

dimitre:
tests won't catch anything here. I'll test locally and merge


dimitre:
cc @ofTheo @danoli3 I've made an ugly ifdef just to comment out everything that is not macOS 
It is now running and can be tested more.

Issue #6226, opened 2019-02-02
ofPixels - artifacts / crashes with .resize()


johanjohan:
Since ofPixels.cpp has not been updated for over a year in the master branch, i therefore assume the following issue has not been fixed. i am using of 0.10.1 with win10 vs2017.

I am resizing images/pixels with the following function.
That leads to issues:

- OF_INTERPOLATE_NEAREST_NEIGHBOR crashes on large sizes
- OF_INTERPOLATE_BILINEAR: produces grey only pixels. well yes, "not implemented yet" in ofPixels
- OF_INTERPOLATE_BICUBIC produces some  singular green pixel/blue artifacts 

calling the same function using image.resize() produces best results

question: i assume that i may use pixelsRes = _pixels; to create a full twin copy, right?
if yes, then there is the described issue in the pixel resizeTo() function.

CODE: 
```c++
#include "ofApp.h"

/*
	this function aims to either
		- crop the image to the largest possible dimensions respecting _aspect 
			then scales the longest side to _longestSideLengthPx
		- stretch it respecting _aspect
			then scales the longest side to _longestSideLengthPx
*/

static void resize(
	ofPixels	 &_pixels,
	const float	 &_longestSideLengthPx,
	const ofVec2f	 &_aspect,
	const bool	 &_bStretch, // otherwise crop
	const ofInterpolationMethod	&_interp	// OF_INTERPOLATE_NEAREST_NEIGHBOR OF_INTERPOLATE_BILINEAR
) {
	ofScaleMode _scaleMode = _bStretch ? OF_SCALEMODE_FILL : OF_SCALEMODE_FIT;
	ofRectangle rectPixels(0, 0, _pixels.getWidth(), _pixels.getHeight());
	ofRectangle rectAspect(0, 0, _aspect.x, _aspect.y);
	rectAspect.scaleTo(rectPixels, _scaleMode);

	ofPixels pixelsRes;
	if (_scaleMode == OF_SCALEMODE_FIT) { // crop
		_pixels.cropTo(pixelsRes, rectAspect.x, rectAspect.y, rectAspect.width, rectAspect.height);
	}
	else { // stretch
		pixelsRes = _pixels;	
		
		// Q: is this a complete copy with new identical pixels?

		// looks like it does copy all params plus the pixels:
		//template<typename PixelType>
		//void ofPixels_<PixelType>::copyFrom(const ofPixels_<PixelType> & mom) {
		//	if (mom.isAllocated()) {
[...]

PR #7936, opened 2024-05-07
fix ofPixels <float> allocation issue

dimitre:
minimal changes to make it work
so now the pixels buffer is allocated with the exact number of bytes used.
it used to work ok with char because sizeof(char) is 1

related to some part of
 https://github.com/openframeworks/openFrameworks/issues/6226
addresses crash / artifacts


dimitre:
copying @ofTheo @NickHardeman

dimitre:
Merging this one, I consider a critical flaw

PR #7952, opened 2024-05-14
Core small changes

dimitre:


libs/openFrameworks/gl/ofTexture.cpp
```diff
@@ -1128,35 +1141,50 @@ ofMesh ofTexture::getMeshForSubsection(float x, float y, float z, float w, float
 	// to constantly add a 2 pixel border on all uploaded images
 	// is insane..
 
-	GLfloat offsetw = 0.0f;
-	GLfloat offseth = 0.0f;
```
ofTheo, line 1132: hmm - I am really not super in favor of introducing this pattern across the whole core. 

If it is for optimization I wouldn't bother as the compiler will do the optimization for you.

libs/openFrameworks/gl/ofTexture.cpp
```diff
@@ -1128,35 +1141,50 @@ ofMesh ofTexture::getMeshForSubsection(float x, float y, float z, float w, float
...
+	GLfloat offsetw { 0.0f };
+	GLfloat offseth { 0.0f };
```
ofTheo, line 1145: hmm - I am really not super in favor of introducing this pattern across the whole core.

If it is for optimization I wouldn't bother as the compiler will do the optimization for you.

libs/openFrameworks/gl/ofTexture.cpp
```diff
@@ -1128,35 +1141,50 @@ ofMesh ofTexture::getMeshForSubsection(float x, float y, float z, float w, float
...
-	quad.getTexCoords()[0] = {tx0,ty0};
-	quad.getTexCoords()[1] = {tx1,ty0};
-	quad.getTexCoords()[2] = {tx1,ty1};
-	quad.getTexCoords()[3] = {tx0,ty1};
+	quad.getTexCoords() = {
+		{tx0,ty0},
+		{tx1,ty0},
+		{tx1,ty1},
+		{tx0,ty1},
+	};
```
ofTheo, line 1177: to me, this feels like a good use case as it does in one line what is being done in 4 previously.

libs/openFrameworks/gl/ofTexture.cpp
```diff
@@ -1128,35 +1141,50 @@ ofMesh ofTexture::getMeshForSubsection(float x, float y, float z, float w, float
...
 	if (!ofGLSupportsNPOTTextures() && bTexHackEnabled) {
 		offsetw = 1.0f / (texData.tex_w);
 		offseth = 1.0f / (texData.tex_h);
 	}
 	// -------------------------------------------------
 
-	auto topLeft = getCoordFromPoint(sx, sy);
-	auto bottomRight = getCoordFromPoint(sx + sw, sy + sh);
+	auto topLeft { getCoordFromPoint(sx, sy) };
+	auto bottomRight { getCoordFromPoint(sx + sw, sy + sh) };
```
ofTheo, line 1154: again, just feels unnecessary vs: `auto topLeft = getCoordFromPoint(sx, sy);`

libs/openFrameworks/gl/ofTexture.cpp
dimitre: Agreed! I'll be reverting back this simple ones.
It makes sense only sometimes, like creating a shared_ptr, or a more complex object at once, to avoid an extra copy

libs/openFrameworks/gl/ofTexture.cpp
dimitre: Thanks. I've reverted back others. 
I've changed this one line, I think we get one less increment on shared_ptr this way
```
std::shared_ptr<ofBaseGLRenderer> renderer { ofGetGLRenderer() };
```

PR #7960, opened 2024-05-15
Vector graphics fix

Jonathhhan:
changes string to std::string.


dimitre:
I think this one could be replaced by of::filesystem::path

dimitre:
Hello @Jonathhhan do you want to update it to filesystem? 
if not we can merge as it is and then rework later

Jonathhhan:
@dimitre I tested it with of::filesystem, but maybe I do something wrong (at least the SWIG bindings do not work with that). So I would say, leave it as is for now.

PR #7966, opened 2024-05-16
ofFpsCounter with more precision using straight std::chrono

dimitre:
ofFpsCounter with more precision using straight std::chrono when possible.
filterAlpha is being calculated with double.
the removal of some includes shows other files were depending of the includes included here, so now they are more explicit.

this PR:
- return better precision in ofFpsCounter
- fixes one error where getLastFrameFilteredNanos is equal to getLastFrameNanos
- Simplify headers inclusion in other files.


dimitre:
Finally! ready to go @ofTheo

ofTheo:
oh cool @dimitre - going to really hammer it this week! 
be great to get some extra eyes on this too as its pretty core to a lot of things :) 

@artificiel @2bbb @NickHardeman

dimitre:
Nice. It is a direct replacement line by line, and keeping the same logic from original fps counter.
Using std::chrono allows the same code for all platforms, with the maximum precision possible allowed by the time resolution in a specific platform.
As it was possible to remove ofUtils.h include I had to add back includes in lots of other OF parts, which is good. 
The most direct the dependencies are, less things for preprocessing and compiling.

NickHardeman:
@dimitre, I just did a simple test with setting the fps of the app and recalculating via ofFpsCounter and it looked good to me. 

While looking at the PR, I noticed these namespace lines at the top of `ofTimeFps.h`
https://github.com/openframeworks/openFrameworks/blob/12f4a84f2e1d8838b6e472aba85e374ebfaabf56/libs/openFrameworks/utils/ofTimerFps.h#L10-L11

I think the appropriate `std::chrono`'s should be prepended to the vars in the .h file and the above lines moved to `ofTimeFps.cpp`. What do you think? Otherwise the `using namespace std::chrono; and using namespace std::chrono_literals;` is global.
```
std::chrono::time_point<std::chrono::steady_clock> wakeTime;
std::chrono::time_point<std::chrono::steady_clock> lastWakeTime;
```

dimitre:
Thank you Nick, good catch. I'll remove this from .h files

artificiel:
this PR circumvents `ofTime`, which has platform `#ifdefs` etc (I think that pre-dates std::chrono?). 

could the work here expand into simplifying the std::chrono in `ofTime` itself? so the timing functions internals are all based on the same core, in the sense that `ofGetElapsedTimef()` itself gets "upgraded"?

dimitre:
@artificiel good idea, ofTime can be updated so all internal operations are done using std::chrono.
I still think this one can be done with straight chrono as it will reduce .h includes

danoli3 (approved):
Awesome keen to try this out

danoli3:
Yeah if you want to wrap it with the ofTime stuff now or later up to you

danoli3:
okay if we passing on all targets lets get it in

danoli3:
could you rebase it

dimitre:
Rebased. No conflicts here

PR #7976, opened 2024-05-20
[test] - msys2 improvements

dimitre:
tried to replace LD with LLD, but I found it is linking with g++
anyway, changing some configurations on msys2 already have improvements on compiling and linking times.


dimitre:
@danoli3 I have a doubt on MSYS2
I've seen the make tries to copy dll files to the project folder at the end of compilation
but looking at openFrameworksLibs_master_msys2_mingw64 we have .a file extensions
should we update the makefiles, or apothecary files?

PR #7989, opened 2024-05-22
Fix ofPixels_<float> resize

janimatic:
Fixes ofPixels_<PixelType>::resizeTo for float pixels (and hopefully every type of pixels) simply replacing bytesPerPixel by channelsFromPixelFormat(pixelFormat) to find indices.
as described here https://github.com/openframeworks/openFrameworks/issues/6226#issuecomment-2099906869


dimitre:
Thank you @janimatic

PR #7991, opened 2024-05-22
[WIP] Android template

dimitre:
This is an alternative Android template to go together with projectGenerator project
and test generating projects for @danoli3 and @tctr fork
related:
https://github.com/openframeworks/projectGenerator/issues/500

The idea here is work in an alternative template / projectGenerator project to get things right without overwriting classic android templates and projects. if things work ok we can rename and use as the main one


danoli3:
Hooley dooley, nice work! Yeah the final frontier of the Android upgrades, the Project Generator updates and templates yay!

I had a quick glance looking great, incorporates all the things. 
Almost fixed that ci for vs, one moment

tctr:
That looks ok for me !

PR #7889, opened 2024-02-07
macOS Xcode Project Updates to XCFrameworks C++17 - Bleeding Edge

danoli3:
macOS / OSX Bleeding Edge Update

Built with latest Binaries 
Screenshot 2024-02-08 at 12 22 14 am

Presenting Apple XCFrameworks 
- macOS initial project changes for latest bleeding edge libraries and formatted in xcframeworks format.

XCFrameworks provide libraries in structured formats that allows for multiple architectures, targets all within the same project without causing linking errors and does provide target specific headers

Extra information to be packaged with each binary is a unique hash, framework version number, build date, checksums of binaries. With the latest changes comes sweeping security and vulnerability patches beyond count. 

iOS / tvOS / MacCatalyst to follow soon

PR #8017, opened 2024-06-12
Use getDeviceIds function to list audio devices instead of relying on contiguous IDs

### LINKED ISSUE #8013: ofSoundStream broken with RtAudio 6.0.0
I recently tried compiling a project that uses openFrameworks (and also ofSoundStream) on Arch Linux and was running into crashes involving audio device selection. After a lot of time spent debugging I realized that it's because the RtAudio API around device listing and selection has changed in RtAudio 6.0.0 (released around a year ago), which is the version of the library in Arch Linux's repository (well, it's actually 6.0.1 but the result is the same).

The main difference is that you cannot rely on devices being between `0` and `RtAudio::getDeviceCount() - 1` and instead must get the list of IDs from `RtAudio::getDeviceIds`. The only functional change needed to support this is changing the for loop in `ofRtAudioSoundStream::getDeviceList` to `for (unsigned int i: audioTemp.getDeviceIds()) { ...`. The application I was trying to compile also made the assumption about device IDs being contiguous from `0` to `getDeviceCount() - 1`, so presumably this would be a breaking change without wrapper functions to hide this new behavior. Maybe the solution is just to stick to a version of RtAudio < 6.0.0 but I at least wanted to document this issue here in case someone else came across it.

EvergreenTheTree:
Fixes #8013 

[As of RtAudio 6.0.0][1] you cannot assume that devices will fall in the range [0, getDeviceCount() - 1]. Doing so will potentially cause errors or out of bounds array accesses depending on the platform.

I am not super familiar with this project, but as far as I can tell I couldn't find any documentation or example projects that would need to be updated as a result of this change. However, I do know of at least two projects ([audiostellar][2] and [ofxPDSP][3]) that make this assumption, so it is definitely worth noting in the change log.

[1]: http://www.music.mcgill.ca/~gary/rtaudio/classRtAudio.html#ab1b5af34ddfe84ae2043d1f09c8638d5
[2]: https://gitlab.com/ayrsd/audiostellar
[3]: https://github.com/macramole/ofxPDSP

### DIFF
--- a/libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
+++ b/libs/openFrameworks/sound/ofRtAudioSoundStream.cpp
@@ -81,9 +81,8 @@ std::vector<ofSoundDevice> ofRtAudioSoundStream::getDeviceList(ofSoundDevice::Ap
 		if(audioTemp.getCurrentApi()!=rtAudioApi && rtAudioApi!=RtAudio::Api::UNSPECIFIED){
 			return deviceList;
 		}
-		auto deviceCount = audioTemp.getDeviceCount();
 		RtAudio::DeviceInfo info;
-		for (unsigned int i = 0; i < deviceCount; i++) {
+		for (unsigned int i: audioTemp.getDeviceIds()) {
 			try {
 				info = audioTemp.getDeviceInfo(i);
 			}


danoli3:
msys2: 
[mingw-w64-ucrt-x86_64-rtaudio](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-rtaudio?repo=ucrt64)
using 5.2.0-1 for some reason... 

macOS will be fixed for 6.0.1 after merge of https://github.com/openframeworks/openFrameworks/pull/7889

EvergreenTheTree:
I'll see about contributing an update to RtAudio 6.0.1 on msys2's end; based on the PKGBUILD it seems like it should be relatively straightforward. Also, the Linux build is failing due to what appears to be a missing dependency (curl specifically). Is that something that needs to be fixed too?

danoli3:
Please tag your head change, then rebase it ontop of current upstream
master.

What we can do is use RT audio defines for the meantime set in code while
we await the fix for msys2.

EvergreenTheTree:
The rtaudio 6.0.1 package update has now been merged into msys2 and is in the msy2 repositories. This branch should be okay to merge now.

danoli3:
Figured out a dynamic way using the version Define
```
#if RTAUDIO_VERSION_MAJOR >= 6
		for (unsigned int i: audioTemp.getDeviceIds()) {
#else
		auto deviceCount = audioTemp.getDeviceCount();
		for (unsigned int i = 0; i < deviceCount; i++) {
#endif

```

Issue #7951, opened 2024-05-13
ofAVFoundationVideoPlayer thread priority


dimitre:
Thread Performance Checker: Thread running at QOS_CLASS_USER_INTERACTIVE waiting on a lower QoS thread running at QOS_CLASS_DEFAULT. Investigate ways to avoid priority inversions
PID: 27389, TID: 7269077
```
Backtrace
=================================================================
3   XP                                  0x00000001009976a8 -[ofAVFoundationVideoPlayer loadWithURL:async:] + 432
4   XP                                  0x0000000100a5a1d4 _ZN20ofAVFoundationPlayer10loadPlayerENSt3__112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEEb + 612
5   XP                                  0x0000000100a5a58c _ZN20ofAVFoundationPlayer4loadENSt3__112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEE + 144
6   XP                                  0x000000010097ff84 _ZN13ofVideoPlayer4loadENSt3__112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEE + 420
7   XP                                  0x00000001006ee8dc _ZN10ofxMicroUI9videoList9updateValEv + 328
8   XP                                  0x00000001006f12cc _ZN10ofxMicroUI5radio3setERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 616
9   XP                                  0x00000001006e44a0 _ZN10ofxMicroUI4loadERKNSt3__112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEE + 1712
10  XP                                  0x00000001006e7334 _ZN10ofxMicroUI10loadPresetERKNSt3__112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEE + 732
11  XP                                  0x000000010070266c _ZNSt3__110__function6__funcINS_6__bindIM10ofxMicroUIFvNS_12basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEEEJPS3_RKNS_12placeholders4__phILi1EEEEEENS7_ISI_EEFvS9_EEclEOS9_ + 72
12  XP                                  0x00000001006f12ac _ZN10ofxMicroUI5radio3setERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 584
[...]

PR #8018, opened 2024-06-13
macOS video player thread priority fix

dimitre:
Closes 
https://github.com/openframeworks/openFrameworks/issues/7951


dimitre:
@ofTheo @NickHardeman seems ok to merge? thanks

NickHardeman:
Does this line in unloadVideoAsync also need to change? 
https://github.com/openframeworks/openFrameworks/blob/dee2eb7044aa37053d0b473c195bd6abecd786bc/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m#L442
I just tested with the change in the PR and the current unloadVideoAsync quickly as is and it seemed to unload ok, so not sure if it needs to change or not.

PR #8023, opened 2024-06-22
Fix for Emscripten not being able to render wireframe. 

### LINKED ISSUE #7833: 3DPrimitivesExample lighting broken in emscripten
Not sure what changed in the last few days, but the 3DPrimitivesExample is rendering black in emscripten. Works fine on desktop builds though. 

You can see it in the gallery here: https://openframeworks.cc/examples/3d/3DPrimitivesExample/ 

image

NickHardeman:
It renders the same as openGL ES using GL_LINES and is not visually the same as polygonMode.
Fixes #7833


danoli3:
Very cool!

PR #8027, opened 2024-06-25
linux fix missing libcurl

danoli3:


danoli3:
package managers ...

danoli3:
@dimitre I disabled the optimisation no-update stuff that was the cause

dimitre:
great! I think they fixed ubuntu 24.04 runners.
it would be good to re-enable cached apt packages if possible, it shaves some time from actions

PR #8028, opened 2024-06-26
pg use nightly vs

danoli3:
Fixes download ci script to link to correct projectGenerator download (nightly)
Fixes Issue was traced to addon_config.mk in ofxOSC common excluding win32 folder... causing recent problems in builds and debugging in projectGenerator... who would have thought


danoli3:
found all the PG issues for VS in addon_config.mk in ofxOSC common excluding win32 folder... D: lol

PR #8045, opened 2024-07-12
iOS Latest Libraries Update 

danoli3:
Updated.

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -262,15 +262,16 @@ enum ofTargetPlatform{
 
 
 #ifdef TARGET_OF_IOS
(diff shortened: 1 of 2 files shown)


danoli3:
iOS Core Project: 
+ Core Library Added 
- Brotli [rfc7932](https://datatracker.ietf.org/doc/html/rfc7932) - a generic-purpose lossless compression algorithm (prior embedded as sub-dependancy) - now usable most platforms 
All DJs use AIFF though right ... right 
- openSSL [3.0.14](https://www.openssl.org) 
Latest Security Fixes

danoli3:
ofConstants.h 
- Xcode Obj-C ARC error. Converted to warning with solutions to fix obj-c++ even with arc on 
```add '-fno-objc-arc' in Build Phases -> Compile Sources -> Compiler Flags.```

PR #8050, opened 2024-07-15
nightly packages - msys2 add clang64, ucrt64, add linux arm64 linuxaarch64 

danoli3:
Also fix:
```
#define OF_NO_FMOD
	#ifndef OF_NO_FMOD
		#define OF_NO_FMOD
	#endif
	```


danoli3:
Okay merging with these critical linux fixes

PR #8056, opened 2024-07-23
Emscripten 3.1.74 - Latest Libraries 

danoli3:
## Latest Potions | Libraries: 
EMSDK - 3.1.64 https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md
- ofxEmscripten updated with stricter return types on wrapper callback functions to [EM_BOOL](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3163---071224) 
- make file changes for new SDK 
- WASM Assembly
- c++17 / c17
```scripts/emscripten/download_latest_libs.sh```

Alternative Apothecary Potions available 
- pThread on or off 
- memory64 enabled or standard 
https://github.com/openframeworks/apothecary/releases/tag/bleeding

- 
| Name       | Version | Build Time              | Type       | Git URL                                         | Binary       | Binary SHA                                                   |
|------------|---------|-------------------------|------------|------------------------------------------------|--------------|--------------------------------------------------------------|
| libpng16   | 1.6.43  | 2024-07-23 T13:22:41Z   | emscripten | http://git.code.sf.net/p/libpng/code           | libpng16.a   | 4675a8d8164bdce13839fec93ebfaa45b1a0dbf3ee8c693411ba40a366ebbd37 |
| fmt        | 8.1.1   | 2024-07-23 T12:00:00Z   | emscripten | https://github.com/fmtlib/fmt                  | fmt.a        | 12ab8d9f4b4f8df97c4f2a1a9d9f8c4c12ab8d9f4b4f8df97c4f2a1a9d9f8c4c |
| svgtiny    | 0.1.7   | 2024-07-23 T14:00:00Z   | emscripten | https://github.com/toolchain/svg-tiny-lcd      | svgtiny.a    | d2cb8a7b6c1b6d8f8b8a7b6c1b6d8f8b8a7b6c1b6d8f8b8a7b6c1b6d8f8b8a7b |
| libxml2    | 2.9.12  | 2024-07-23 T15:00:00Z   | emscripten | https://gitlab.gnome.org/GNOME/libxml2.git     | libxml2.a    | e3c3a7b6d8f8b8a7b6c1b6d8f8b8a7b6c1b6d8f8b8a7b6c1b6d8f8b8a7b6c1b6 |
[...]

PR #8067, opened 2024-08-01
Emscripten 3.1.65 fix

Jonathhhan:
Not sure, if this is backward compatible, but it works wit the current Emscripten version.
Also, the apothecary libs need to be rebuild with the current Emscripten version.

### DIFF
--- a/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
+++ b/addons/ofxEmscripten/src/ofxAppEmscriptenWindow.cpp
@@ -50,21 +50,21 @@ void ofxAppEmscriptenWindow::setup(const ofGLESWindowSettings & settings){
     _renderer = std::make_shared<ofGLProgrammableRenderer>(this);
     ((ofGLProgrammableRenderer*)_renderer.get())->setup(2,0);
 
-    emscripten_set_keydown_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW,this,1,&keydown_cb);
-    emscripten_set_keyup_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW,this,1,&keyup_cb);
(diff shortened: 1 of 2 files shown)


danoli3:
You were able to build and test in browser this working ?

Jonathhhan:
@danoli3 actually I need to test it a little more. A lot of errors are resolved, but there is an issue with emrun. Similar to this one: https://github.com/emscripten-core/emscripten/issues/14128 which was not there before. But it should work, if uploaded to a webserver (did not test that, yet).

danoli3:
Please test, I might also, as I've been chasing my tail to find the right settings to work in Site. 

I was crashing on getGLError before that using just updated EM_BOOL stuff as above, but ended up finding a MAIN function in the old freeimage library on the old version so thats now fixed 
https://github.com/openframeworks/openFrameworks/pull/8056

Anyways I believe the old binding ways maybe causing something to break bad. I opted to just port us to GLFW and have working code to fix up but multi-window stuff was really, exciting 

I'm happy to merge this first on old libs if it's working as is with, as EM_SDK it kinda forces latest in most example scripts to download it or run it. 

I also was looking at your other PR's with the audio stuff, lets get those in once isolate issues with workers, might be a good idea to just create template project makes for each unique feature like that, as it does become a core change requirement... however we could define somewhere, not exactly sure how the ```config.emscripten.default.mk```

danoli3:
@Jonathhhan how did you go any working test? I might checkout your PR and see myself

Jonathhhan:
I am sorry. I am away from my computer for a good week, and did not manage to test it before. Will definitely come back to this. Until then it is still possible to use older Emscripten versions...

Jonathhhan:
I am sorry. I am away from my computer for a good week now, and did not manage to test it before. Will definitely come back to this. Until then it is still possible to use older Emscripten versions...

PR #8073, opened 2024-08-07
OF default plane with less resolution (rows, cols)

dimitre:
I've noticed the default OF plane comes with 6 cols, 4 rows
this plane is used for ofLight::draw (areaLight) for example
This is a suggestion of using a 1x1 plane as a default, and increase resolution if / when needed.


danoli3:
should be square regardless of resolution of vertices

PR #8079, opened 2024-08-10
XCode CoreOF - fix osx template pointing to macos core

dimitre:


danoli3:
Yeah probably. 

Error is fmod I think should be deployed to macOS folder when downloading OSX.
The issue is the makefile is deploying to the macos folder in the final output hence the change in Xcode but guess not needed for the osx project

Issue #8082, opened 2024-08-18
move libs/json/include/json.hpp to libs/json/include/nlohmann/json.hpp


2bbb:
when other libraries has `json.hpp`, `#include "json.hpp"` can't load `nlohmann::json`.

i think moving `include/json.hpp` to `include/nlohmann/json.hpp` and replace
`#include "json.hpp"`
to
`#include "nlohmann/json.hpp"`
or
`#include <nlohmann/json.hpp>`

for reference:
`#include <json.hpp>` works sometime.
but i think this is not complete solution, maybe...

oxillo:
👍 

MSYS2 is already using #include <nlohmann/json.hpp> in ofJson.h

```cpp
#if !defined(TARGET_MINGW)
	#include <json.hpp>
#else
	#include <nlohmann/json.hpp> // MSYS2 : use of system-installed include
#endif
```

Also, examples from nlomann site are using #include <nlohmann/json.hpp>. So, I think this is the right move.

2bbb:
this fix need to use apothecary...?
if so, i don't know apothecary enough, please fix anyone! ☺️

danoli3:
Can do

dimitre:
Nice! should this should be changed also? 
in ofJson.h
```c++
#if !defined(TARGET_MINGW)
	#include <json.hpp>
#else
	#include <nlohmann/json.hpp> // MSYS2 : use of system-installed include
#endif
```

danoli3:
Yes if someone could make the PR for that once the releases deployed to
merge

danoli3:
Nice I think this changed exposed a few major automation test problems! XD

PR #8083, opened 2024-08-18
fix nlohmann json include path

2bbb:
related: #8082


danoli3:
This is good minus the platforms still pointing to nightly apothecary libs. Will fix that in another PR

PR #8089, opened 2024-08-21
downloader [4.2.4 ~ 4.2.5] - windows - wget2 disabled, remove old download files if local outdated 

danoli3:
Noticing some issues with wget2 still on windows causing json headers in a few cases to not be extracted due to the zip file not completing, inconsistent 
- Fixed and forced remove of local download tar/zip if found (should be done automatically with clobber but to be safe

### DIFF
--- a/scripts/dev/downloader.sh
+++ b/scripts/dev/downloader.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-VERSION=4.2.4
+VERSION=4.2.5
 printDownloaderHelp(){
 cat << EOF
     
@@ -93,6 +93,7 @@ check_remote_vs_local() {
     if [ -z "$modified" ]; then
             echo "  [downloader] failed to retrieve last-modified header from remote ["$REMOTE_URL"] ... Proceeding with download"
             CHECK_RESULT=0
+            rm -f $LOCAL_FILE
             return
     fi
   if [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


dimitre:
Usually I have to delete downloads folder so libs install correctly. 
Yesterday I couldn't install the ones in ./scripts/macos folder.
should we name it _downloads?

danoli3:
That’s a problem with wget2.
I’m yet to write upstream report
wget2 is now buffered with support of cURL should fix the issues. I would
just use cURL though it’s in parallel now

dimitre:
I think it is the caching issue. If I run the same script again it returns like this and stops
```
 openFrameworks download_libs.sh v2.5.0
 -----
 [openFrameworks downloader v4.2.7] ... 
curl: (2) no URL specified
curl: try 'curl --help' or 'curl --manual' for more information
  [downloader] enabled brotli/zlib losslesss compression response
  [downloader] check if local == remote: [https://github.com/openframeworks/apothecary/releases/download/bleeding/openFrameworksLibs_bleeding_macos_1.tar.bz2]
  [downloader] Found download cache.
  [cache] [openFrameworksLibs_bleeding_macos_1.tar.bz2]
  [downloader] Remote size:[126MB] | Local size:[126MB]
```

danoli3:
whats your system config: 
-wget2 installed
-curl installed 

can you open dev/downloader.sh and set it to VERBOSE=1 

that cURL check if you have wget2 that the url is failing for should resolve the cache issue but that looks like it thinks the files are the same and will not download them... since its to the bit, its solid compare for new release. 

was this issue happening previously, it might be the connection close for wget2. I'll disable that and set timer for wget2. it will still close the port if curl header checks done however

PR #8092, opened 2024-08-23
GLFW transparent window background touch/click pass through

danoli3:
As mentioned on the forums here:
https://forum.openframeworks.cc/t/transparent-window-app-with-mouse-click-pass-through/43816

# GLFW Window clear transparency - mouse passThrough 
 - GLFW 3.4 + supports click pass through with these commands

Changes additional bools for GLFW settings.
- bool mousePassThrough; // input pass through for transparency window
- bool floating;  // type of window 

Thanks to @artificiel for documenting floating / passThrough info 

usage would be something like in main.cpp:
```
settings.transparent = true;
settings.mousePassThrough = true;
settings.floating = true;
settings.decorated = false;
```

Example:

transparentWindow


danoli3:
I made an example as this requires a bit of main.cpp setup and understanding of the limitations and powers of this passThrough.

transparentWindow

artificiel:
I wonder if `.transparentInput` is the correct base name for the click through feature? adopting GLFW naming would make things more auto-documented, so `.mousePassthrough` to shadow `GLFW_MOUSE_PASSTHROUGH`?

likewise, `ofSetWindowMousePassthrough(bool)` instead of `ofSetWindowTransparencyInput(bool)`?

unless "transparent input" is a term commonly used?

danoli3 (commented):
Updated with better variables / function names

danoli3 (commented):
Awesome this is looking ready, now!

danoli3:
Okay rechecked this bad boi going to merge

PR #8097, opened 2024-08-29
[idea / wip] - ofShader::setUniformBufferObject to easily pass a struct to a shader

dimitre:
relevant discussion: https://forum.openframeworks.cc/t/multiple-lights-limit-with-opengl-3-2-how-to-increase/43777/26

What do you think @NickHardeman  ?


NickHardeman:
Should probably be wrapped in the defines next to `getUniformBlockIndex, getUniformBlockBinding`, etc. 
I don't have much experience with Uniform Blocks, so would be good to test with openGL ES. Testing with Emscripten would be helpful. Could be wrapped in 
```
#if !defined(TARGET_OPENGLES) || defined(TARGET_EMSCRIPTEN)
#ifdef GLEW_ARB_uniform_buffer_object
```

PR #8118, opened 2024-09-10
OF nightly and manual without wget2

dimitre:
downloader is interrupting with wget2, 
this is a temporary workaround to *hopefully* allow nightly to complete


danoli3:
Probably better to just use cURL until wget2 is fixed properly

PR #8119, opened 2024-09-10
ofFileDialogResult using fs::path

dimitre:
This is an API breaking change, but I think it is OK


dimitre:
another one OK to go. opinons on this? @ofTheo @danoli3 ?

ofTheo:
Also looks great! Thanks @dimitre !!

PR #8123, opened 2024-09-15
XCode project changes - script to compile OF now in Project.xcconfig

dimitre:
It is easier to write and read outside of that small text area in XCode
And parameters for project compilation can be defined now per project using App.xcconfig.

Useful for me, I'll always compile just for arm, Intel can be compiled just if needed

Other small change, xcodebuild is less verbose because it doesn't show env vars repeatedly


dimitre:
@danoli3

ofTheo:
fine with me! :)

PR #8129, opened 2024-09-21
XCode 16 error fix : no type named 'lock_guard' in namespace 'std'

dimitre:


roymacdonald:
This shouldnt hurt if not needed but it does if it is so I would simply merge this. Unless you think otherwise

PR #8142, opened 2024-10-11
[wip] ofVideoPlayer string and path test

dimitre:


libs/openFrameworks/video/ofVideoBaseTypes.h
```diff
@@ -209,6 +209,10 @@ class ofBaseVideoPlayer: virtual public ofBaseVideo{
 	/// \sa loadAsync()
 
 	virtual bool load(const of::filesystem::path & fileName) = 0;
```
ofTheo, line 211: I think this might need to be not pure virtual as inherited classes from older addons won't have this implemented

PR #8149, opened 2024-10-16
No fs::path return

roymacdonald:
@ofTheo

### DIFF
--- a/libs/openFrameworks/utils/ofSystemUtils.h
+++ b/libs/openFrameworks/utils/ofSystemUtils.h
@@ -11,13 +11,13 @@ public:
 
 	/// \return the name of the selected file or directory, if set
 	/// currently returns only 1 file, this may change in the future
-	of::filesystem::path getName();
+	std::string getName();
 
 	/// \return the full path of the selected file or directory, if set
-	of::filesystem::path getPath();
+	std::string getPath();
 
-	of::filesystem::path filePath { "" }; ///< full path to selected file or directory
-	of::filesystem::path fileName { "" }; ///< selected file or directory name
+	std::string filePath = ""; ///< full path to selected file or directory
+	std::string fileName = ""; ///< selected file or directory name
 	bool bSuccess = false; ///< true if the dialog action was successful, aka file select not cancel
 };
 
@@ -29,7 +29,7 @@ void ofSystemAlertDialog(std::string errorMessage);
 /// \param bFolderSelection set to true to allow folder selection
 /// \param defaultPath optional default directory path to start the dialog in, ie. ofFilePath::getUserHomeDir()
 /// \return dialog result with selection (if any)
-ofFileDialogResult ofSystemLoadDialog(std::string windowTitle="", bool bFolderSelection = false, of::filesystem::path defaultPath="");
+ofFileDialogResult ofSystemLoadDialog(std::string windowTitle="", bool bFolderSelection = false, std::string defaultPath="");
 
 /// \brief show a file save dialog box
 /// \param defaultName suggested filename to start dialog, ie "screenshot.png"
--- a/libs/openFrameworks/graphics/ofImage.cpp
+++ b/libs/openFrameworks/graphics/ofImage.cpp
@@ -251,7 +251,7 @@ static bool loadImage(ofPixels_<PixelType> & pix, const of::filesystem::path & _
 		if(fif == FIF_JPEG) {
 			option = getJpegOptionFromImageLoadSetting(settings);
 		}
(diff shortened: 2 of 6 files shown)


libs/openFrameworks/utils/ofSystemUtils.h
```diff
@@ -11,13 +11,13 @@ class ofFileDialogResult{
...
 
 	/// \return the full path of the selected file or directory, if set
-	of::filesystem::path getPath();
+	std::string getPath();
 
-	of::filesystem::path filePath { "" }; ///< full path to selected file or directory
-	of::filesystem::path fileName { "" }; ///< selected file or directory name
+	std::string filePath = ""; ///< full path to selected file or directory
+	std::string fileName = ""; ///< selected file or directory name
 	bool bSuccess = false; ///< true if the dialog action was successful, aka file select not cancel
```
ofTheo, line 21: I think the above could all be of::filesystem::path internally and then `std::string getName(); / std::string getPath();` just uses the `filePath.string()` in its return.

libs/openFrameworks/utils/ofSystemUtils.h
```diff
@@ -29,7 +29,7 @@ void ofSystemAlertDialog(std::string errorMessage);
 /// \param bFolderSelection set to true to allow folder selection
 /// \param defaultPath optional default directory path to start the dialog in, ie. ofFilePath::getUserHomeDir()
 /// \return dialog result with selection (if any)
-ofFileDialogResult ofSystemLoadDialog(std::string windowTitle="", bool bFolderSelection = false, of::filesystem::path defaultPath="");
+ofFileDialogResult ofSystemLoadDialog(std::string windowTitle="", bool bFolderSelection = false, std::string defaultPath="");
```
ofTheo, line 33: As with the previous comment I think this argument could be left as is.

ofTheo:
Thanks @roymacdonald - I added a couple of comments but apart from that looks good to go. 
I am going to take a stab at the ofBaseVideoPlayer stuff now.

roymacdonald:
@ofTheo should we merge?

libs/openFrameworks/utils/ofSystemUtils.h
dimitre: Yes or ofPathToString(filePath)

ofTheo:
@roymacdonald - If you can address the two comments I left then I think its good to merge! 
Thanks!

roymacdonald:
@ofTheo as for the public members, I think it is better to leave as strings, since changing to ::path will break stuff (actually, the ofDropInfo class whose public members were changed to path break backwards compatibility, such as the example in ofxHapPlayer). So, I think we should leave those as strings. 

As of the function parameter, changing string to ::path requires some implementation reworking  so it is an useful change and to make sure it works properly. 
In the implementation, `defaultPath` could be renamed to `_defaultPath` and then internally declare `std::string defaultPath` so there is no need to rework any of the code. 
```
ofFileDialogResult ofSystemLoadDialog(std::string windowTitle, bool bFolderSelection, of::filesystem::path _defaultPath){
    std::string defaultPath = ofPathToString(_defaultPath);
...
}
```

But that would give the false impresion that the use of std::filesystem::path is done properly, and because of such I think it is better to keep the function parameter as std::string, and only change it when it is properly handled.

ofTheo:
Thanks @roymacdonald ! 

@dimitre - I am tempted to just merge this due to the public member issue Roy mentioned.  

Would that be fine with you and then we can PR in some improvements for `ofFileDialogResult` where we can make the `std::string filePath` a copy of the real internal path ( which can be protected ) and then getter functions for Path and String?

dimitre:
Hey @ofTheo great. if this makes a smoother transition
I still consider your comment in ```ofSystemUtils.h``` valid, preserving internally as fs::path. @roymacdonald can comment if he has any objection to this

libs/openFrameworks/utils/ofSystemUtils.h
dimitre: hey @ofTheo just for your information, this was reverted back to string . just unsure if it was the idea

ofTheo: Yes - I think the issue was that because the members were public they had to be string for usage not to break on Windows. 

My thought was we could do a new PR and maybe have these std::string variables as legacy support but have a `protected` `of::filesystem::path` internally be used by the class with the string values copies of the path data. 

And then a method for accessing as a file path. 

But it felt like it needed a bit of thought and we needing to revert to string because of the direct variable usage anyway.

Issue #8143, opened 2024-10-11
ofToDataPath returning std::filesystem::path


roymacdonald:
So, 
on windows I get an error when trying to automatically convert from std::filesystem::path to string
![Screenshot 2024-10-11 173742](https://github.com/user-attachments/assets/c771496c-9f4a-495b-9bd3-1f6ab2a9775e)

and I have found it being an issue with ofxCv

@artificiel @ofTheo @dimitre

artificiel:
this happens because of an seemingly complicated (according to stack overflow etc) to navigate number of parameters relating to the setup of the IDE project options, encoding, manifest, BOM, filesystem, etc, the implicit conversion done by ofToDataPath's argument (which takes a path) will upgrade to wstring on Windows, even if it's explicitly passed a narrow std::string. (this is yet another side quest (testing) but it seems the OF tests in GitHub are setup in a way that does not trigger this problem, and/or the above test should be added).

so it's trying to convert wstring to a string in order to assign and not finding that converter.

don't have windows here; can you splice this in ofFileUtils.h at 1229 (just after ofToDataPath):
```C++
inline of::filesystem::path ofToDataPath2(std::string str, bool absolute = false) {
	return ofToDataPath(of::filesystem::path(std::string(str)), absolute);
}
```
and try your failure with `ofToDataPath2`? if that works we can work around by explicitly pushing std::string into filesystem::path. (and a potential "future" OF app that wants to be unicode-windows-friendly will anyway have to carefully use std::wstring or ::path and fall into the non-converted signature).

[NB the last phrase is another argument to "raise awareness of path vs string": if we expect some mac-developed app to compile in windows and support wide unicode, it means std::string is never used as a path container, even though on Mac it does not make an actual difference]

[so to be clear, std::string should never be used on windows for a path containers, but obviously legacy-compatibility requires a solution]

roymacdonald:
@artificiel 
I tested the snippet of code you posted and it does not work. same error.
the problem is implicitly converting from of::filesystem::path to std::string, not the other way around. Actually the problem seems to be that there is no implicit conversion from std::wstring to std::string.

I think that at least in the case of ofToDataPath and any class where the return type of a function which used to be std::string and got changed to of::filesystem::path we should roll back to std::string until we find a better solution as this will break on windows a lot of things. As I said elsewhere, we rollback the master branch and we make a dev branch where we have this changes and we test those.

artificiel:
[...]

PR #8150, opened 2024-10-17
Fix of video player path

ofTheo:
As discussed in #8143 this adds the load( std::string ) function to ofBaseVideoPlayer class so that older video players will still work when used with 0.13 and on. 

Tested both with using via ofVideoPlayer and directly. 
Seems to work great and should allow for both newer path implementation and older legacy approaches. 

cc @dimitre @roymacdonald


libs/openFrameworks/types/ofBaseTypes.cpp
```diff
@@ -47,7 +47,24 @@ ofBaseVideoPlayer::~ofBaseVideoPlayer(){
...
+
+//---------------------------------------------------------------------------
+bool ofBaseVideoPlayer::load(const std::string & fileName){
+	ofLogError("ofBaseVideoPlayer") << " One of the two load functions need to be implemented ";
+	return false;
+}
+
+//---------------------------------------------------------------------------
 void ofBaseVideoPlayer::loadAsync(const of::filesystem::path & fileName){
+	loadAsync(fileName.string());
```
dimitre, line 63: we can use ```ofPathToString(fileName)```  here

roymacdonald: Indeed. the string() function can throw exceptions and because of that the ofPathToString handles it. In the switch to use fs::path we should enforce the use of ofPathToString rather than calling string() directly.

ofTheo: ahh good catch! thanks guys!

PR #8152, opened 2024-10-19
Removal of FMOD linking and dylib copying on osx, since it is not being used

roymacdonald:
FMOD is NOT being used on osx and still the dylib is being linked and copied into the app's bundle.
this PR removes such.

It is not only disabled by default, it is not possible to enable simply by modifiing the Project.xconfig file as  OF_NO_FMOD is being defined in ofConstants.h . Even if I wanted to use FMOD, it will not be that straight forwards.

OF_NO_FMOD defined in ofConstants.h is not the same as the one that is read by [xcode_project.sh](https://github.com/openframeworks/openFrameworks/blob/master/scripts/osx/xcode_project.sh#L26). The one that the latter reads is the one defined in Project.xcconfig, that is why it needs to be set in it.

This will not break anything as not using FMOD is the default behaviour and not really possible to change by just passing a flag. 

Although, I think it is still reasonable to keep the dylib in the core for legacy purposes. (Eventually we should move it into an addon with all fmod related classes)

I guess that it was supposed to be the same variable but it is not.


NickHardeman:
+1 For moving FMOD out of the core either to an add on or removing it all together.

ofTheo:
Let's do it! 
Hallelujah!!! There has got to be an issue we can close with this.  

Oh yes - and it's an oldey in the 3 digits!! #167

ofTheo:
only took 14 years ;-)

PR #8154, opened 2024-10-20
iso646.h in windows — and / or templates fix

dimitre:
Subsequent changes for this PR https://github.com/openframeworks/openFrameworks/pull/8153


artificiel:
this is now closed/merged, but in addition to the local spots it is needed, I suggest keeping it into ofMain.h also as using `and` `or` etc is ISO standard C++, but end-user code working in clang will suddenly break on MSVC with no good reason. having the include in ofMain makes things "more compatible".

PR #8161, opened 2024-10-27
Added support for ADDON_DEFINES from addon_config.mk for make files

roymacdonald:
This a adds support for ADDON_DEFINES when using make. It follows the pattern already set for the other addon_config.mk files and it was also partly implemented, so it should not break nor bother when not present.


roymacdonald:
I will merge this since it is really useful to have the new PG working nicely, but also it does not interfere with what was already there

Issue #5640, opened 2017-05-25
Replace ofVec*, ofMatrix*, ofQuaternion with glm in examples.


bakercp:
@arturoc Has anyone done this yet?

toliaferrode:
Why don't you wrap GLM with oF? Excuse me if this has already been thought of.

hamoid:
To get an idea of how many files there are:

```bash
$ cd examples

$ ag ofQuat -l --ignore obj
3d/cameraParentingExample/src/ofApp.cpp
3d/quaternionLatLongExample/src/ofApp.cpp
3d/quaternionArcballExample/src/ofApp.cpp
3d/quaternionArcballExample/src/ofApp.h
3d/orientationExample/src/ofApp.cpp
gl/pixelBufferExample/src/ofApp.cpp
gl/threadedPixelBufferExample/src/ofApp.cpp
ios/iosCoreMotionCameraExample/src/ofApp.mm
ios/iosCoreMotionExample/src/ofApp.mm

$ ag ofMatr -l --ignore obj
3d/advanced3dExample/src/ofApp.cpp
3d/advanced3dExample/src/OrthoCamera.cpp
gl/textureBufferInstancedExample/src/ofApp.h
android/androidAdvanced3DExample/src/OrthoCamera.cpp
android/androidAdvanced3DExample/src/ofApp.cpp
ios/iosCoreMotionExample/src/ofApp.mm

$ ag ofVec -l --ignore obj
windowing/multiWindowExample/src/main.cpp
windowing/multiWindowOneAppExample/src/main.cpp
graphics/floatingPointImageExample/README.md
graphics/floatingPointImageExample/src/ofApp.cpp
graphics/floatingPointImageExample/src/ofApp.h
graphics/lutFilterExample/README.md
graphics/lutFilterExample/src/ofApp.cpp
graphics/lutFilterExample/src/ofApp.h
3d/cameraParentingExample/src/ofApp.cpp
3d/normalsExample/src/ofApp.cpp
3d/pointCloudExample/src/ofApp.cpp
3d/quaternionLatLongExample/README.md
3d/quaternionLatLongExample/src/ofApp.cpp
3d/3DPrimitivesExample/src/ofApp.cpp
3d/meshFromCameraExample/src/ofApp.cpp
3d/cameraLensOffsetExample/src/ofApp.cpp
3d/cameraLensOffsetExample/src/ofApp.h
3d/modelNoiseExample/src/ofApp.cpp
[...]

PR #8178, opened 2024-11-03
[Idea] - Test removing legacy math from core

dimitre:
This is a proposition for discussion. 
I think we can do well removing ofPoint ofVec{2, 3, 4}f, ofMatrix{3x3,4x4} usage from core and core addons and substitute for glm equivalent ones. removing toOf / toGlm conversions too.

We can keep all functions related to this functions inside an ifdef, so we can test compile OF without legacy math.

### DIFF
--- a/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
+++ b/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp
@@ -5,7 +5,7 @@
 #include "ofPixels.h"
 #include "ofGraphics.h"
 #include "ofConstants.h"
-#include "ofMatrix4x4.h"
+//#include "ofMatrix4x4.h"
 #include "ofUtils.h" // ofGetElapsedTimef
 
 #include <assimp/cimport.h>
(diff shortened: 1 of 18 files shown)


dimitre:
related: https://github.com/openframeworks/openFrameworks/issues/5640

roymacdonald:
As far as I can recall, there was a lot of work and though put into how to fully go into using glm only but not breaking projects using OF vec classes. The solution at which we got is quite nice and it works (thus, I dont see the need to remove at all if it is not bothering). Having `#ifdef OF_USE_LEGACY_MATH` is nice, but I would leave it enabled by default and only if someone really knew what they are doing to disable. I tend to work with old projects and it really becomes painful and annoying when things dont work as they used to. 

As for the examples, all those should have been converted to use glm long ago and I dont see any issue on doing so. (I suggest making a different PR just with the updated examples). But you need to test these and compare, visually that the behavior is the same. I remember seeing odd stuff,  like with matrix decomposition.

I would keep using the toOf and toGlm functions within the core in order to enforce their usage and still keep compatibility, which essentially implies dont change much. 

> I think we should not use implicit conversion in OF Core,
Implicit conversion is actually nice and useful, what problems do you see with it? 

> I think today some interoperability between glm::vec2 -> glm::vec3 is achieved through some indirect mean like glm::vec2 -> ofVec2f -> ofVec3f -> glm::vec3 which is not ideal.
if it is possible to remove this, and make it direct it would be great.

As of the addons. I think it is ok to update their internal use but I would not change their signatures (public and protected members and methods) as that can potentially break projects. I would change those only using OF_USE_LEGACY_MATH is undefined.


This is a BIG change which I really think it needs to be taken with calm. there is no need to rush. there is no need to break what is already working.  


Last but not least:
Please, make a new branch for this, it makes it a lot easier for other to check. And  dont merge to master until agreed to do so.

dimitre:
@roymacdonald This PR should not change any functionality. (it needs testing)
what I did was finish porting all Core functions and examples to glm and having the option of not using "legacy math"

roymacdonald:
I know what you did, I checked your commits, and while seems like a tempting idea, it really needs to be checked and tested.

dimitre:
There are some nice things about removing from core interoperability between glm and "classic math" 
so we discover sometimes includes, interoperability can be avoided only by taking better care of types.
I'm working on a fork that has some of  this things addressed here:
https://github.com/dimitre/openFrameworks/commit/2b533769a877fe53d3dfd795345f1c8bd9ff0c5a

dimitre:
@ofTheo what do you think of this one? 
it updates everything ofMath to glm in core.

dimitre:
Implemented in ofWorks fork. all legacy OF math was moved to ofxMath, and will be included in "compatibility mode"
Closing the PR in OF main as too old

PR #8187, opened 2024-11-12
more .h includes simplification

dimitre:


dimitre:
good to go. should we re-add includes in ofMain.h to keep all of them available ?

dimitre:
I've brought this PR to be compatible with master
if it is a good idea lets merge or discuss it 
if not merged I'll be closing by November 24 (PR birthday)

cc @ofTheo @NickHardeman @artificiel

ofTheo:
Thanks @dimitre - have you benchmarked the compile time difference with this change and without? 
I'm all for speeding up compile times and  it would be good to know if this PR has a significant impact.  

Generally ofConstants.h is where these things are meant to live, so moving to ofMain.h would only make sense if we can see a big speed up. 

Thanks!
Theo

dimitre:
@ofTheo I didnt think it would make much differente but in fact if we compare the runners in this one against another normal PR like this we see improvement in compile times in all runners. sometimes less than 50% original time.
take a look:

https://github.com/openframeworks/openFrameworks/pull/8187/checks
https://github.com/openframeworks/openFrameworks/actions/runs/19430409178/job/55587700525

This PR removes lots of unneeded system includes (sometimes they are huge) so compiler has much less lines to process

ofTheo:
@dimitre wow, yes, I see the difference too! 
Weird that VS seems unaffected but msys2 has the same 50% speed up. 

Seems worth it for the speed up alone and people typically include ofMain.h so seems fairly safe

Issue #4538, opened 2015-11-08
ofAVFoundationPlayer slow setPosition


i-n-g-o:
calling setPosition() on ofAVFoundationPlayer is slow.
this is due to synchron disposal of resources (AVAssetReader).

disposing AVAssetReader on a backgroundThread (dispatch_async) would solve that issue, but has problems when called too quickly after each other. (in my case updating the position every 100ms works, but leaks memory (??!!) when calling every 50ms)

this needs further investigation.

fast-seek (async disposal) can be found in this branch:
https://github.com/i-n-g-o/openFrameworks/tree/ofAVFoundationPlayer_fast_seek

i-n-g-o:
i see the memoryleaks with a 1080i50 file encoded with ProRes422

```
if ((ofGetElapsedTimeMillis() - timeLastLoaded > 50)) {
    fingerMovie.setPosition(ofRandomuf());
    timeLastLoaded = ofGetElapsedTimeMillis();
}
```

dimitre:
Potentially fixed by
https://github.com/openframeworks/openFrameworks/pull/6797
https://github.com/openframeworks/openFrameworks/pull/6791
one addresses the "infinite threads" issue and the other on some seek position issues.

PR #8189, opened 2024-11-13
ofVideoPlayer AVFoundation, setPosition can work with arbitrary frames.

dimitre:
by lowering seekToTime tolerance to zero it will force the video frame to be displayed correctly.

Today if I load a video recorded with a camera, lets say Sony alpha 6500 in h264, it makes one keyframe each second. 
if I'm trying to setPosition in an arbitrary position it will only display one frame per second (the keyframe)

with this change we assure it will be positioned in the correct frame even if it takes more cpu cycles to calculate (in the case of random access or backwards access). 
it doesn't change anything for normal play because frame difference will be accumulating.

quick way of testing
```c++
	float pos = ofMap(mouseX, 0, ofGetWindowWidth(), 0, 1);
	video.setPosition(pos);
	video.update();
```

### DIFF
--- a/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
+++ b/libs/openFrameworks/video/ofAVFoundationVideoPlayer.m
@@ -1180,7 +1180,7 @@ static const void *PlayerRateContext = &ItemStatusContext;
 }
 
 - (void)seekToTime:(CMTime)time {
-	[self seekToTime:time withTolerance:kCMTimePositiveInfinity];
+	[self seekToTime:time withTolerance:kCMTimeZero];
 }
 
 - (void)seekToTime:(CMTime)time


artificiel:
it's great to be able to do that!

however it might also be good to have the option to maintain the current (default) to retrieve only keyframes as inter-frame decode presumably takes more CPU it will change the performance of existing code. ex:

```c++
video.setTolerance(ofVideoPlayer::TOLERANCE_ZERO);
video.setTolerance(ofVideoPlayer::TOLERANCE_INFINITY); // historical default
```

dimitre:
We can optionally expose something like that if it is useful
I can only see a setPosition that just works. 
the historical implementation is not good, I know people gave up using videoPlayer on macOS (which performs great btw)

dimitre:
@ofTheo greattt news about the release! 
take a look at this one when you have time
I know a project was moved from OF to another platform because of this one

dimitre:
closes https://github.com/openframeworks/openFrameworks/issues/4538

dimitre:
Closing this one as stalled. implemented in ofWorks

Issue #8193, opened 2024-11-15
Nightly scripts aren't running. 


ofTheo:
https://github.com/openframeworks/openFrameworks/actions/runs/11761061978 

```
[Invalid workflow file: .github/workflows/nightly-linux64.yml#L48](https://github.com/openframeworks/openFrameworks/actions/runs/11761061978/workflow)
You have an error in your yaml syntax on line 48
```

<img width="972" alt="image" src="https://github.com/user-attachments/assets/da3a6e6c-4892-4c42-9a00-715f37bb50cb">

dimitre:
https://github.com/openframeworks/openFrameworks/issues/8194

dimitre:
I can fix that yml syntax

ofTheo:
closed by #8195

PR #8195, opened 2024-11-15
[actions] fix yml format for nightly

dimitre:


ofTheo:
closes #8193  - thanks @dimitre !!

PR #8204, opened 2024-11-20
[actions] test fix libraries setting env var RELEASE=latest

dimitre:


dimitre:
@ofTheo I'll be merging this one so we can keep making things work.  
@danoli3  I've moved nightly and manual-nightly actions to .github/disable so you can keep working later

PR #8212, opened 2024-11-25
[test] ofTexture more straightforward way to see if it support GL_RGB16F

dimitre:


dimitre:
@ofTheo I think this one is good to go. it fixes things for RPI and completes compilation on aarch64.
it reveals apothecary or system libraries issues (GLEW) on previous RPI versions

libs/openFrameworks/gl/ofTexture.h
```diff
@@ -177,13 +177,12 @@ class ofTextureData {
...
-			textureTarget = GL_TEXTURE_2D;
-		} else {
-			glInternalFormat = GL_RGB;
-			textureTarget = GL_TEXTURE_2D;
-		}
+	#if defined(GL_RGB16F)
+		glInternalFormat = GL_RGB16F;
+	#else
+		glInternalFormat = GL_RGB;
+	#endif
```
ofTheo, line 184: Feels like this would maybe be a safer change as it keeps the original logic, but also checks for the define

```
		if(ofGLESVersionFromGL() >= 300) {
				#if defined(GL_RGB16F)
						glInternalFormat = GL_RGB16F;
				#else
						glInternalFormat = GL_RGB;
				#endif
			textureTarget = GL_TEXTURE_2D;
		} else {
			glInternalFormat = GL_RGB;
			textureTarget = GL_TEXTURE_2D;
		}

**Comment by ofTheo** [comment]
Thanks @dimitre - unrelated curious why we still have the armv7l CI checks - thought we dropped armv7 builds?

**Comment by dimitre** [comment]
@ofTheo I've put back working runners for armv6l and armv7l like 7 months ago. we talked about dropping but there were other ideas, like 32 bits and rpi zero (or something like this)

File: libs/openFrameworks/gl/ofTexture.h (hunk 1 above)
Comment (ofTheo) [reply libs/openFrameworks/gl/ofTexture.h:184]: yeah - I think I was just imagining possible regressions where it might now default to GL_RGB16F instead of GL_RGB - one is a floating point texture and the other is unsigned char. 

So that could mean some shaders might behave a little differently - curious if that would be very noticeable or not really an issue. 

I think in general bug fixes should correct breaking bugs or existing issues without changing behavior, hence my caution.

**Comment by ofTheo** [comment]
@dimitre can you double check / fix the apothecary changes that are pulled into this PR and if it all looks good feel free to merge?

File: libs/openFrameworks/app/ofMainLoop.cpp [hunk 2]
```diff
@@ -23,8 +23,6 @@
 	#include "ofAppGLFWWindow.h"
 #elif defined(TARGET_RASPBERRY_PI)
 	#include "ofAppEGLWindow.h"
-	#include "ofAppGLFWWindow.h"
```
ofTheo: @dimitre - is this safe to remove?
I know a lot of people use `ofAppGLFWWindow.h` on Pi - or does it get included elsewhere ?

dimitre: Yes safe. this commit in fact reverts ofMainLoop.cpp to original state. I added this one in previous commits and removed now

dimitre:
@ofTheo I've cleaned up this PR to reflect the needed changes, so it is good to go in my opinion.
it can potentially help with emscripten issues

ofTheo:
Amazing - thanks @dimitre !

Issue #8220, opened 2024-11-28
Cannot compile OF libs with msys2 on nightly build


Scylardor:
Hello,

I'm failing at building openFrameworks for mingw64 using msys2.

I found this issue: #8034 
and realized there seems to be an issue with 0.12.0, so I tried installing the nightly build instead like what this issue suggested.

But at the step:

```
cd your_oF_directory/libs/openFrameworksCompiled/project
make
```

it starts compiling, but compilation fails quickly with this message:

```
Compiling ../../../libs/openFrameworks/gl/ofFbo.cpp
In file included from C:/msys64/mingw64/include/stdio.h:1302,
                 from C:/msys64/mingw64/include/c++/14.2.0/cstdio:42,
                 from ../../../libs/openFrameworks/utils/ofConstants.h:11,
                 from ../../../libs/openFrameworks/communication/ofSerial.h:4,
                 from ../../../libs/openFrameworks/communication/ofSerial.cpp:1:
C:/msys64/mingw64/include/swprintf.inl: In function 'int swprintf(wchar_t*, const wchar_t*, ...)':
C:/msys64/mingw64/include/swprintf.inl:37:19: error: invalid use of '__builtin_va_arg_pack ()'
   37 |   return _swprintf( __stream, __format, __builtin_va_arg_pack() );
      |          ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

There seems to be a wrong compiler flag used here, or a missing one, as suggested by this SO post:
https://stackoverflow.com/questions/78503173/how-to-build-when-builtin-va-arg-pack-is-used

Is it a known issue? How could I try to fix this?

PS: my "real" issue is that I'm trying to use Visual Studio Code with openFrameworks on Windows, but it's unusable because the project generator doesn't generate the .vscode folder, so there are no DEBUG/RELEASE build tasks available. I don't know what to do. Hence I'm trying to install openFrameworks using msys2 instead of msbuild.

Scylardor:
FYI: It seems a bug related to a similar issue is on the GCC tracker, but I cannot be sure it's the exact same one: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110834
It looks to me like a compiler bug.
[...]

PR #8222, opened 2024-12-03
[msys2] - ofxGui avoid fs::path / string implicit conversion. fix for __builtin_va_arg_pack issue in swprintf

dimitre:
it can close 
- https://github.com/openframeworks/openFrameworks/issues/8220

### DIFF
--- a/libs/openFrameworks/communication/ofSerial.cpp
+++ b/libs/openFrameworks/communication/ofSerial.cpp
@@ -446,7 +446,7 @@ bool ofSerial::setup(string portName, int baud){
 		cfgSize = sizeof(cfg);
 		GetCommConfig(hComm, &cfg, &cfgSize);
 		int bps = baud;
-		swprintf(buf, L"baud=%d parity=N data=8 stop=1", bps);
+		swprintf(buf, 80, L"baud=%d parity=N data=8 stop=1", bps);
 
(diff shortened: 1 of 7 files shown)


libs/openFrameworks/communication/ofSerial.cpp
```diff
@@ -446,7 +446,7 @@ bool ofSerial::setup(string portName, int baud){
 		cfgSize = sizeof(cfg);
 		GetCommConfig(hComm, &cfg, &cfgSize);
 		int bps = baud;
-		swprintf(buf, L"baud=%d parity=N data=8 stop=1", bps);
+		swprintf(buf, 80, L"baud=%d parity=N data=8 stop=1", bps);
```
ofTheo, line 450: this is the only thing needed for the fix right?

ofTheo:
fine to merge with me - as long as the failing checks were failing before :)

libs/openFrameworks/communication/ofSerial.cpp
dimitre: For the specific fix yes, but the system complained about string - fs::path conversion here, so I had to change other things to be able to find this error. if you prefer I can split this PR in two and merge this specific one

ofTheo: ah cool! - no that's all good - thanks for the fix @dimitre !

Issue #8225, opened 2024-12-07
ofPixels copy broken


prisonerjohn:
I posted a comment in the problematic commit but I'll add it here too as it's a 7 month old change...

Here's the commit: https://github.com/openframeworks/openFrameworks/commit/f7522b9cc0e304d03735ca4e0ad80d7aa28d2360#diff-7d727932e9b85218273507defcbf4b9e9c9a300c10d62867e1459b2f167b45f8L399

And the issue:
`ofPixels.size()` was changed to return the number of bytes instead of the number of pixels, and that breaks copying with different data formats.

If you try something like this:
```
ofFloatPixels fPix;
fPix.allocate(512, 512, 1);
ofPixels cPix;
cPix = fPix;
```
then you get a buffer overrun in `copyFrom()` because the loop runs on `ofPixels.size()`.
```
for(size_t i = 0; i < mom.size(); i++){
	pixels[i] = ofClamp(mom[i], 0, 1) * factor;
}
```
There's already an `ofPixels.getTotalBytes()` that returns the total bytes, why make `size()` return the same thing?

dimitre:
Good catch. I've made a PR to potentially address this one
- https://github.com/openframeworks/openFrameworks/pull/8226

but it should be tested to see if it works well on this situation also
- https://github.com/openframeworks/openFrameworks/issues/6226

dimitre:
Good catch. 
thoughts: 
There are inconsistencies here as pixels is made of PixelType, and we have some formats that doesn't have an uniform pixeltype, like RGB565, which the total number of bits would be a more correct manner of doing the correct allocation. OR we can mark as unimplemented if pixel format is RGB565.

https://github.com/openframeworks/openFrameworks/pull/8226/files
in this PR size() is again reflecting only the pixel count, ignoring channel bit depth or channel count.
if we decide this is the correct usage of size we should update doxygen to reflect. 
today it hints returning total bytes, but only considering one byte per channel.
yes there is historical confusion needing clarification and cleanup.

```
	/// \brief Get the number of values that the ofPixels object
	/// contains, so an RGB data 400x400 would be 480,000, whereas RGBA data
	/// of the same dimensions would be 640,000.
```

artificiel:
not that the doxygen is necessarily "truth" but in this case `the number of values ... contained` is pretty clearly independent from the resolution of the said values (for example it makes a difference between RGB and RGBA but it makes no difference between 8 or 10 or 16 bits RGB)

[...]

PR #8226, opened 2024-12-08
ofPixels memory allocation size fix

dimitre:
This one will need the tests updated to reflect what pixelSize represents now.
closes https://github.com/openframeworks/openFrameworks/issues/8225


dimitre:
This one will need tests updated because pixelSize represented a different thing in the past.
I think we are going to a more complete and correct way of treating pixels
ideas? suggestions?

ofTheo:
Don't we want a simple revert of:
https://github.com/openframeworks/openFrameworks/commit/f7522b9cc0e304d03735ca4e0ad80d7aa28d2360#diff-7d727932e9b85218273507defcbf4b9e9c9a300c10d62867e1459b2f167b45f8L399 

Looking at this PR I am not sure if it's actually reverting and might be changing behavior even more. 

I'll note in the lines what I mean. 

cc @prisonerjohn

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -393,10 +393,10 @@ template<typename PixelType>
 void ofPixels_<PixelType>::setFromExternalPixels(PixelType * newPixels, size_t w, size_t h, ofPixelFormat _pixelFormat){
 	clear();
 	pixelFormat = _pixelFormat;
-	width= w;
+	width = w;
 	height = h;
 
-	pixelsSize = bytesFromPixelFormat(w,h,_pixelFormat);
+	pixelsSize = w * h;
```
ofTheo, line 399: pixels size used to be: 
`	pixelsSize = bytesFromPixelFormat(w,h,_pixelFormat) / sizeof(PixelType);`

bytesFromPixelFormat is effectively: w * h * 24 / 8 ( for RGB ) 
so: newSize = w * h * 3;  for RGB

```
size_t ofPixels_<PixelType>::bytesFromPixelFormat(size_t w, size_t h, ofPixelFormat format){
	return w * h * pixelBitsFromPixelFormat(format) / 8;  <-- pixelBitsFromPixelFormat for rgb is 24 
}
```
a simpler fix would be: 

`pixelsSize = w*h * getNumChannels()`  

But maybe safer to first revert the commit where the regression was introduced and then do a PR with the bit depth support after?

libs/openFrameworks/graphics/ofPixels.cpp
```diff
@@ -530,9 +530,12 @@ void ofPixels_<PixelType>::allocate(size_t w, size_t h, ofPixelFormat format){
 	width 		= w;
 	height 		= h;
 
-	pixelsSize = newSize;
+	pixelsSize = w * h;
```
ofTheo, line 533: again good to note that pixelsSize is not `w*h` but `w*h * getNumChannels()`; 
pixelsSize is what ofPixels::size() returns and if you look here, it needs to represent the total number of bytes not just w*h. 

```
void ofPixels_<PixelType>::copyFrom(const ofPixels_<SrcType> & mom){
	if(mom.isAllocated()){
		allocate(mom.getWidth(),mom.getHeight(),mom.getNumChannels());

		const float srcMax = ( (sizeof(SrcType) == sizeof(float) ) ? 1.f : std::numeric_limits<SrcType>::max() );
		const float dstMax = ( (sizeof(PixelType) == sizeof(float) ) ? 1.f : std::numeric_limits<PixelType>::max() );
		const float factor = dstMax / srcMax;

		if(sizeof(SrcType) == sizeof(float)) {
			// coming from float we need a special case to clamp the values
			for(size_t i = 0; i < mom.size(); i++){
				pixels[i] = ofClamp(mom[i], 0, 1) * factor;
			}
		} else{
			// everything else is a straight scaling
			for(size_t i = 0; i < mom.size(); i++){
				pixels[i] = mom[i] * factor;
			}
		}
	```

dimitre: I propose we gave more thought and try to solve incongruences that always existed in ofPixels instead of reverting last PR. When you say the total number of bytes it is not true, it is the total number of channels, each of them can be one byte (as it always supported)  or more.

I'll take a look soon at this PR. so size() is back to original behavior

libs/openFrameworks/graphics/ofPixels.cpp
dimitre: Part of the problem (historical ofPixels problem) is we have two different units
one is PixelType (it can be char, float, etc)

and the other are formats that doesn't have an uniform number of bytes per channel like RGB565 (5 + 6 + 5 total 16 bits) or NV12 (12 bits per channel)

but the array that stores pixel information will be always a PixelType (so we consider an "uncompressed" pixel type, to be worked on shaders).

we can go one way or another. 

if we support all the images format I suggest changing the Pixels array to uint8_t instead of PixelType and allocate everything following "bytesPerPixel"

Or keep everything as it is, and comment out all image types that doesn't follow PixelType bit depth.
makes sense?

dimitre:
I reverted size to the behavior it always was, and changed the PR title.
this now handles well new Pixels allocation that is based on "total number of channels" instead of total number of bytes, which is only true if we have a 8 bit PixelType (ofPixels)

libs/openFrameworks/graphics/ofPixels.cpp
dimitre: Please take a look again and see the difference. it is now not based in bytes for pixels allocation.

ofTheo:
Awesome thanks @dimitre !

Issue #8238, opened 2024-12-26
[bug] ofGetLastFrameTime() always returns 0


artificiel:
on macOS14 with current OF-git, doing this always prints 0:
```c++
void ofApp::update(){
	ofLogNotice("last") << ofGetLastFrameTime();
}
```
```c++
[notice ] last: 0.0
[notice ] last: 0.0
[notice ] last: 0.0
[notice ] last: 0.0
```

perhaps it's local to macOS — could other platforms confirm?
if it's really a regression it's traceable to #7966 and reverting it fixes the problem:
```c++
[notice ] last: 0.00400008
[notice ] last: 0.0123949
[notice ] last: 0.00437546
[notice ] last: 0.0117662
```

danoli3:
Can confirm Visual Studio 2022 / Windows so all platforms affected.

I noticed this ages ago although was convinced it was a local project issue and forgot as I fixed with (in setup):
```
ofSetFrameRate(60);
ofSetTimeModeFixedRate(ofGetFixedStepForFps(60));
```

danoli3:
Okay to assist debugging:

Problem only occurs on System.

Example failing: code (update)
```
float timeFromLast = ofGetLastFrameTime();
currentTime += timeFromLast;

ofLog(OF_LOG_NOTICE) << "ofGetLastFrameTime:[" << timeFromLast << "] currentFadeIn:" << currentFadeIn << " < fadeInTotal:[" << fadeInTotal << "]";
if(isFadedIn == false) {
 if(currentFadeIn >= fadeInTotal) {
     isFadedIn = true;
		ofLog(OF_LOG_NOTICE) << "FadedIn timeFromLast:" << timeFromLast;
 }
 }

```
Logs:
```
[notice ] ofGetLastFrameTime:[0] currentFadeIn:0 < fadeInTotal:[0.417]
[notice ] ofGetLastFrameTime:[0] currentFadeIn:0 < fadeInTotal:[0.417]
[notice ] ofGetLastFrameTime:[0] currentFadeIn:0 < fadeInTotal:[0.417]
[notice ] ofGetLastFrameTime:[0] currentFadeIn:0 < fadeInTotal:[0.417]
[notice ] ofGetLastFrameTime:[0] currentFadeIn:0 < fadeInTotal:[0.417]
[notice ] ofGetLastFrameTime:[0] currentFadeIn:0 < fadeInTotal:[0.417]
[...]

PR #8240, opened 2024-12-31
Fix: ofGetLastFrameTime() always returns 0 (Filtered/System timeMode) Bug

danoli3:
Fixes: https://github.com/openframeworks/openFrameworks/issues/8238

### ofFPSCounter 
- fixed default values
- fixed ofGetLastFrameTimeSecs() to return double duration as to spec 
- fixed ofGetLastFrameFilteredTimeSecs()  to return double duration as to spec 
- fixed timestamp calculation for < 2
- fixed filteredTime calculation with clamps for anomalies 
- fixed filteredTime calculation to only occur with timeMode FilteredTime
- passing timeMode as int in ctor / setter added
- queue to deque (for size optimisation)
- allowed for FPS change with functions added

### ofEvents
- fixed setting of ofFPSCounter and timeMode 
- fixed changing of FPS not changing fps counter

### DIFF
--- a/.github/workflows/of.yml
+++ b/.github/workflows/of.yml
@@ -32,7 +32,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Docker Step
-        run: "docker run -di --name emscripten -v $PWD:/src emscripten/emsdk:3.1.73 bash"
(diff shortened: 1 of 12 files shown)


danoli3:
Okay this is tested working for System, Filtered and normal working. 

- Also fixed emscripten build ci for split for addons
- fixed linux rtAudio redefine issues 
- Linux - GDB enabled for tests

PR #8262, opened 2025-02-05
Enable shared libs for Linux

Jonathhhan:
Not sure if this is the right position of code and if there are any drawbacks. But when I want to compile OF as a shared lib with Linux I need to compile with -fPIC. There are also some apothecary libs that need to be rebuild with -fPIC for that (kiss and pugixml for example). My use case is, that I want to compile https://github.com/Jonathhhan/ofxOfeliaExtended as an external for Pure Data, which works if I compile OF core with -fPIC and leave the mentioned libs out.


danoli3:
New toolchains for Linux use -fPIC

Jonathhhan:
@danoli3 thats nice. I deleted my PR, because in ofxOfelia is a script, which adds that flag...

PR #8265, opened 2025-02-14
ofRectangle changes to avoid including ofVectorMath.h

dimitre:
with some small changes here we can avoid including ofVectorMath.h with all templates and other includes

### DIFF
--- a/libs/openFrameworks/3d/of3dPrimitives.cpp
+++ b/libs/openFrameworks/3d/of3dPrimitives.cpp
@@ -12,6 +12,7 @@
 #include "ofVboMesh.h"
 #include "ofTexture.h"
 #include "of3dUtils.h"
+#include "ofVectorMath.h"
 
 using std::vector;
 using std::shared_ptr;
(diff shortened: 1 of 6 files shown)


libs/openFrameworks/3d/of3dPrimitives.cpp
```diff
@@ -242,11 +243,11 @@ void of3dPrimitive::drawNormals(float length, bool bFaceNormals) const{
...
 			for(size_t i = 0; i < normals.size(); i++ ) {
                 if(i % 3 == 0) {
-                    vert = (vertices[i]+vertices[i+1]+vertices[i+2]) / 3;
+                    vert = (vertices[i]+vertices[i+1]+vertices[i+2]) / 3.0f;
                 } else if(i % 3 == 1) {
-                    vert = (vertices[i-1]+vertices[i]+vertices[i+1]) / 3;
+                    vert = (vertices[i-1]+vertices[i]+vertices[i+1]) / 3.0f;
                 } else if ( i % 3 == 2) {
-                    vert = (vertices[i-2]+vertices[i-1]+vertices[i]) / 3;
+                    vert = (vertices[i-2]+vertices[i-1]+vertices[i]) / 3.0f;
```
danoli3, line 250: yeah good idea to force the float!

danoli3 (approved):

PR #8277, opened 2025-02-23
osx / iOS Configuration build dir fix 

danoli3:


dimitre:
hello @danoli3 I think this line has an absolute path of your machine
https://github.com/openframeworks/openFrameworks/blob/aa344f6feebee1a8eb4e35b9c8f19f17fb6ca03f/libs/openFrameworksCompiled/project/macos/openFrameworksLib.xcodeproj/project.pbxproj#L709

/Users/one/SOURCE/openFrameworksiOSTIPS/libs/brotli/lib/macos/brotli.xcframework

PR #8278, opened 2025-02-23
tvOS Config fix. OSX Clean fix

danoli3:


danoli3:
@dimitre had to change those commands to sh as they blocked on newer osx !

danoli3:
image

image

libs/openFrameworksCompiled/project/tvOS/tvOS+OFLib.xcodeproj/project.pbxproj
```diff
@@ -1245,16 +1220,20 @@
 		844639661BC343E000F24926 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CONFIGURATION_TEMP_DIR = "$(SRCROOT)/../../lib/tvos/build/debug/";
+				BUILD_DIR = "~/Library/Developer/Xcode/DerivedData";
```
ofTheo, line 1223: Should we be hard coding the build dir? 
I think before it respected the user setting right?   

image

Issue #8267, opened 2025-02-16
ofxAssimp vs. ofxAssimpModelLoader - Differences and performance


dimitre:
I'm using back and forth both addons so I have the best of both worlds.
But I've noticed ofxAssimpModelLoader is much faster at loading complex models, like 11.36 seconds vs 32.35 seconds.

I'm sharing here the model if it helps debugging.
[cr1.zip](https://github.com/user-attachments/files/18815661/cr1.zip)

I'm not sure if ofxAssimpModelLoader will be mantained, 
but other differences I've noticed:
Y axis is inverted from one to another, 
differences in scale
ofxAssimpModelLoader has some problems loading complex animations ( I can share the specic model if useful )

this tests are just the model inside a default easycam
![Image](https://github.com/user-attachments/assets/570b7c65-cad3-4f7a-97fb-f3215b78cf06)
![Image](https://github.com/user-attachments/assets/e854d290-9ce2-4bbe-8273-333a88706dad)

ofTheo:
@dimitre 
Are you noticing those differences in Debug and Release or just Debug? 

I thought it might be due to the differences between the two optimization flags, but they look the same: 

ofxAssimp: 
https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxAssimp/src/Source/ofxAssimpSrcScene.cpp#L115-L128 

And ofxAssimpModelLoader:
https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxAssimpModelLoader/src/ofxAssimpModelLoader.cpp#L137-L150

dimitre:
Everything tested on Release

ofTheo:
@dimitre is it possible to share the textures that are missing?

I see with both version of ofxAssimp addons that it is looking for texture files. 
The newer addon seems to be getting really caught up with the missing textures which might be the real cause of the issue with the slower loading times. 

For example if I try two complex GLTF models,  the newer version loads the FlightHelmet in 2.988 seconds. 
The older ofxAssimpModelLoader addon loads the FlightHelmet in 4.03441 seconds, so 25% slower. 

Here's the missing texture messages from the CR1 model: 
```
[verbose] ofxAssimpModelLoader: loadGLResources(): loading mesh 2
[verbose] ofxAssimpModelLoader::loadGLResources: : mesh is one sided
[verbose] ofxAssimpModelLoader: loadGLResource(): loading Diffuse image from "Cr1/Concreto___01.png"
[...]

PR #8281, opened 2025-02-25
create a map of the bones instead of searching for all the meshes 

NickHardeman:
Fixes load times mentioned in #8267 by caching the bone list instead of searching through all the meshes and all the bones for every isBone() function call.


dimitre:
Lets merge this one

NickHardeman:
This fix greatly reduces scene parsing by limiting recursion, though it's not really a bug, it's a vast improvement for loading large, complex scenes. Would be nice to get in for v0.12.1 :) What do you think @ofTheo

ofTheo:
Was the issue in #8267 also related to textures or was this the main issue?

NickHardeman:
There may still be a texture issue, but this was the main slow down ( 30 seconds ), now resulting in load times comparable to the older ofxAssimpModelLoader ~13 seconds.

Issue #7903, opened 2024-02-23
Coredump on wayland


kflak:
Hi,

Stumbled on a new wayland-related issue: I'm suddenly no longer able to run any oF-apps on wayland (tried both sway and Hyprland). xorg works fine. The error I'm getting from a clean, empty default project created with the projectGenerator is:
```
 [ error ] ofAppGLFWWindow: 65550: X11: Platform not initialized
make: *** [../../../libs/openFrameworksCompiled/project/makefileCommon/compile.project.mk:184: run] Segmentation fault (core dumped)
```
Running `of_v0.12.0_linux64gcc6_release` on Arch Linux, and I'm using the `glfw-git` package as mentioned in [this issue](https://github.com/openframeworks/openFrameworks/issues/6306). This used to work just fine, but now no more...

cgxeiji:
I am having the same issue with 0.12.0

kflak:
Any news on this? It's getting a bit tedious having to switch back and forth between xorg and wayland...

tobiashuttinger:
As a temporary fix, this can be circumvented by building glfw with the cmake flag `DGLFW_BUILD_WAYLAND=OFF`.

To do this on Arch Linux when installing the glfw-git version from AUR, edit PKGBUILD and set `DGLFW_BUILD_WAYLAND` from `ON` to `OFF` in Line 44 (currently).

kflak:
> As a temporary fix, this can be circumvented by building glfw with the cmake flag `DGLFW_BUILD_WAYLAND=OFF`.
> 
> To do this on Arch Linux when installing the glfw-git version from AUR, edit PKGBUILD and set `DGLFW_BUILD_WAYLAND` from `ON` to `OFF` in Line 44 (currently).

Just getting back into oF now after a bit of a pause, and your solution works!!!! Thank you so much :-) Let's hope it doesn't break something else on my system....

ofTheo:
Thanks for sharing this temporary workaround! 
Added this to our release checklist. #7588 

I am guessing the changes needed for ofAppGLFWWindow won't be huge, as most of our calls are GLFW calls, so I would assume there might be just a few assumptions we might need to address to make it work for both.

danoli3:
Fixed with Linux merge @ofTheo ! I'm just cleaning up cross compilers today

kflak:
> Fixed with Linux merge [@ofTheo](https://github.com/ofTheo) ! I'm just cleaning up cross compilers today

[...]

PR #8283, opened 2025-02-26
Linux GLFW Wayland support

danoli3:
Fixes https://github.com/openframeworks/openFrameworks/issues/7903#issuecomment-2681991557

Adds support for Wayland


danoli3:
added to linux_2025

PR #8288, opened 2025-02-26
Bugfix assimp addons

### LINKED ISSUE #8273: Emscripten addons not linking
Building with emscripten latest on macOS with current nightly. 
Running emmake make Release I see a bunch of errors relating to metalangle. 

- 1 Metal Angle Errors

```
In file included from In file included from ../../../libs/metalangle/include/third_party/jsoncpp/source/include/json/features.h../../../libs/metalangle/include/third_party/jsoncpp/source/include/json/forwards.h::10:
In file included from 10../../../libs/metalangle/include/third_party/jsoncpp/source/include/json/forwards.h:
:In file included from 10../../../libs/metalangle/include/third_party/jsoncpp/source/include/json/config.h:
:In file included from 10../../../libs/metalangle/include/third_party/jsoncpp/source/include/json/config.h:
:In file included from 10/Users/theo/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/istream:
:In file included from 163/Users/theo/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/istream:
:In file included from 163/Users/theo/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__fwd/istream.h:
:In file included from 13/Users/theo/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__fwd/istream.h:
:13:
/Users/theo/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__fwd/string.h:/Users/theo/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__fwd/string.h22:22::11::  error: error: templates must have C++ linkagetemplates must have C++ linkage
```

Removing the metalangle folder from libs/ fixed it. 

- 2 ASSERTIONS warning - not critical 
`-s ASSERTIONS=2`  
[These flags ](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/emscripten/config.emscripten.default.mk#L78-L79) need to be commented out.

Once the metalangle issue is addressed Emscripten builds and runs again!! 
Yay. Huge thanks to @danoli3 @Jonathhhan and anyone else who contributed to getting it working. Huge to have it back!

Image

ofTheo:
closes #8273 

this PR allows for % wildcard patterns for libs which weren't supported before in makefiles. 
so this now works: 
`	ADDON_LIBS=../ofxAssimpModelLoader/libs/assimp/lib/emscripten/%/libassimp.a`


dimitre:
with this change a similar change in PG is needed.
if we just used usual folders it won't be needed even declare ADDON_LIBS because PG and Make will look there automatically

ofTheo:
@dimitre yeah - I think its an edge case just of ofxAssimp as its relying on libs from ofxAssimpModelLoader. 

PG solves this with `ADDON_ADDITIONAL_LIBS = ../ofxAssimpModelLoader/libs` which finds the libs as needed, but ADDON_ADDITIONAL_LIBS doesn't work with makefiles. 

So it needs this set: 
```
	ADDON_LIBS=../ofxAssimpModelLoader/libs/assimp/lib/emscripten/%/libassimp.a
```

Anyway yeah, probably best not to rely on the wild card. 

The main issue is there is so much code in apothecary now doing the sub folder stuff it would be a lot of work to revert, so I just make the smallest possible fix now. 

probably better to just have:
```
	ADDON_LIBS=../ofxAssimpModelLoader/libs/assimp/lib/emscripten/WASM/libassimp.a
```

ofTheo:
@dimitre the wildcard pattern also fixes OpenCV 
https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxOpenCv/addon_config.mk#L154-L186

But again, this could be switched to WASM or this section could be removed and picked up by the folder parser.

danoli3:
need the extra folder to support WASM64 as well, there is two targets
WASM
WASM64 

Nice

danoli3:
again not using subfolders was the root cause of the issue

Issue #7588, opened 2023-08-07
OF 0.12.1 Release Checklist


dimitre:
Complete List - https://github.com/openframeworks/openFrameworks/milestone/27
Just starting a sketch of things that would be nice to work after 0.12 release, feel free to edit / suggest items

### general / all ###
- [x] https://github.com/openframeworks/openFrameworks/issues/8271
- [x] https://github.com/openframeworks/openFrameworks/issues/8282
- [x] remove Boost ( remove from ofConstants.h )  #8297 #8303 #8304 #8306 #8307
- [x] remove Poco ( push to 0.13.0 ) (keep poco 12.1.0) Win/OSX
- [x] rethink math constants
- [x] merge @danoli3 's apothecary PR https://github.com/openframeworks/apothecary/pull/231
- [x] warning for unused function natural - see: https://github.com/openframeworks/openFrameworks/issues/6692#issuecomment-1698778254
- [x] remove `using namespace std;` as the default fixing #7628 
- [x] Add GLM define to project files - #6530 #8352 
- [x] #5820
- [ ] consider moving FS functions to internal only #8410 
- [x] #6246 
 
### arm ### 
- [ ] #7641
- [ ] Fix this example #7653
- [x] #8355 
- [x] #8356 
- [x] #8359 

### android ### 
- [x] get android working again 
- [ ] fix all examples (some deprecated code)

### iOS ### 
- [ ] new examples #7678
- [x] get simulator builds working with xcframework solution 
- [ ]  #8305
- [x] #8308 
- [x] #8310
- [x] #8312
- [x] #8314 
- [x] #8315 
- [x] #8316

### macOS ###
- [x] curl regression - #7656
- [x] macOS Clean deletes bin folder. #8276 
- [x] #8321 
- [x] #8377 

### msys2 ### 
- [x] #7668 

### vs ###  
- [x] Projects outside of OF folder broken when using addons - #7560
- [x] Make PG allow for retarget project on update ( as well as the currently working generate ) see: https://github.com/openframeworks/openFrameworks/issues/7625#issuecomment-1714407390 
- [x] Dll issue with ofxLaser - #7652
- [x] #7665
- [ ] #8302 
- [x] https://github.com/openframeworks/apothecary/issues/506 
- [x]  #8343  https://github.com/openframeworks/apothecary/pull/525

### linux ###  
- [x] #7633 
- [x] #7677
- [x] #7794 
- [x] #8290
- [ ] #8425

[...]

PR #8297, opened 2025-02-27
Remove boost as alternative in ofConstants.h

NickHardeman:
boost is also referenced in a few make files; like openFrameworksCompiled / project / osx / config.osx.default.mk; since it's an exclusion, it's been left there.
Referenced for v0.12.1 here: #7588

### DIFF
--- a/libs/openFrameworks/utils/ofConstants.h
+++ b/libs/openFrameworks/utils/ofConstants.h
@@ -436,37 +436,23 @@ typedef TESSindex ofIndexType;
 #endif
 
 
-#if defined(OF_USING_STD_FS)
-    #if defined(OF_USE_EXPERIMENTAL_FS)
-        // C++17 experimental fs support
-        #include <experimental/filesystem>
-		namespace std {
-			namespace experimental{
-				namespace filesystem {
-					using path = v1::path;
-				}
+#if defined(OF_USE_EXPERIMENTAL_FS)
+	// C++17 experimental fs support
+	#include <experimental/filesystem>
+	namespace std {
+		namespace experimental{
+			namespace filesystem {
+				using path = v1::path;
 			}
 		}
-
-		namespace of {
-			namespace filesystem = std::experimental::filesystem;
-		}
-    #else
-		#include <filesystem>
-		namespace of {
-			namespace filesystem = std::filesystem;
-		}
-    #endif
-#else //not OF_USING_STD_FS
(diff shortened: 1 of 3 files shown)


dimitre:
Nice, maybe we can remove experimental also
I've been using like this:
```c++
#include <filesystem>
namespace of {
	namespace filesystem = std::filesystem;
}
#define OF_USING_STD_FS
```

NickHardeman:
Since experimental is already in there, maybe we can leave it for now to support older compilers?

ofTheo:
@NickHardeman might be good to add to the PR removing the VS boost references too?

NickHardeman:
@ofTheo ah yes, just added that, then did a test building a vs code project on macOS and it compiled successfully. There are some lingering references to boost via make files, ie.
https://github.com/openframeworks/openFrameworks/blob/a94e7e9573e1f2bd14cd860b16e07c65c4138749/libs/openFrameworksCompiled/project/osx/config.osx.default.mk#L270

But seems a little safer to leave the exclusions in those types of files.

ofTheo:
Sorry meant these ones: 

https://github.com/openframeworks/openFrameworks/blob/a94e7e9573e1f2bd14cd860b16e07c65c4138749/libs/openFrameworksCompiled/project/vs/openFrameworksDebug.props#L13 

https://github.com/openframeworks/openFrameworks/blob/a94e7e9573e1f2bd14cd860b16e07c65c4138749/libs/openFrameworksCompiled/project/vs/openframeworksLib.vcxproj#L78

https://github.com/openframeworks/openFrameworks/blob/a94e7e9573e1f2bd14cd860b16e07c65c4138749/libs/openFrameworksCompiled/project/vs/openFrameworksRelease.props#L13

https://github.com/openframeworks/openFrameworks/blob/a94e7e9573e1f2bd14cd860b16e07c65c4138749/libs/openFrameworksCompiled/project/vs/openFrameworksRelease.props#L13

https://github.com/openframeworks/openFrameworks/blob/a94e7e9573e1f2bd14cd860b16e07c65c4138749/libs/openFrameworksCompiled/project/vs2019/openframeworksLib.vcxproj#L78

Def worth doing a search: 
https://github.com/search?q=repo%3Aopenframeworks%2FopenFrameworks+boost&type=code

I'd do these as a separate PR but it would be nice to have it all in one squashed  commit.

PR #8299, opened 2025-02-28
Ios template fix closes #8298

### LINKED ISSUE #8298: [iOS] all xcodeproj are broken in of_v20250226_ios_release
also, using the PG within of_v20250226_ios_release (0.86.0) generates broken xcodeproj with same msg:

Image

ofTheo:
Ios template fix closes #8298


ofTheo:
will merge once iOS test passed.

PR #8301, opened 2025-02-28
Fix to json scripts

ofTheo:
does the pretty formatting with the plutil command avoiding the need for brew


dimitre:
Much better now! thanks
I just prefer the .command extension so we can invoke directly from finder (double click)

danoli3:
> Much better now! thanks I just prefer the .command extension so we can invoke directly from finder (double click)

same I used to as well but unless the files of that type are signed, Apple legit will flag them randomly and u need to go to Settings / allow execution if its a command, many more clicks overall, I just drag the .sh into the Terminal icon in dock instead of click now

Issue #8316, opened 2025-03-01
[iOS] assimp has wrong lib architecture


artificiel:
<img width="750" alt="Image" src="https://github.com/user-attachments/assets/5813b83b-8e81-4636-8cda-1cc90c3d0bd9" />

danoli3:
https://github.com/openframeworks/projectGenerator/issues/603

danoli3:
also a oF iOS Core clean issue.

<img width="754" alt="Image" src="https://github.com/user-attachments/assets/d6ac6cae-2443-4f2f-8ef4-738084efe4a3" />

```

Showing All Messages
Building for 'iOS', but linking in object file (/Users/one/SOURCE/openFrameworksiOSTIPS/libs/openFrameworksCompiled/lib/ios/openFrameworksiOS.a[arm64][2](of3dPrimitives.o)) built for 'iOS-simulator'

**Comment by danoli3** [comment]
yeah this one it seems to be adding the library directly in projectGenerator .a for simulator and not relying on xcframework system

<img width="872" alt="Image" src="https://github.com/user-attachments/assets/c935b7a9-fa77-48ca-ad4e-44059893caff" />

**Comment by danoli3** [comment]
looks like removing the addon_config iOS stuff fixes it 

<img width="801" alt="Image" src="https://github.com/user-attachments/assets/43d26cf3-330d-43da-9e98-ce4baec00528" />

```
[notice ] { "openFrameworks projectGenerator": "0.86.0" }
[notice ] ofPath set and valid using ["/Users/one/SOURCE/openFrameworksiOSTIPS"]

[verbose]  projectPath path: ["/Users/one/SOURCE/openFrameworksiOSTIPS/apps/myApps/mySketchAss"] root_path: ["/"]
[verbose]  ofPath path: ["/Users/one/SOURCE/openFrameworksiOSTIPS"]
[verbose]  ofRoot path: ["/Users/one/SOURCE/openFrameworksiOSTIPS"]
[verbose]  projectPath path: ["/Users/one/SOURCE/openFrameworksiOSTIPS/apps/myApps/mySketchAss"]
[verbose]  The project path exists.
[verbose]   and it is a directory.
[notice ] setting OF path to: ["/Users/one/SOURCE/openFrameworksiOSTIPS"]
[notice ] from -o option

[notice ] -----------------------------------------------
[notice ] target platform is: [ios]
[...]

PR #8319, opened 2025-03-01
assimp ios addon_config

danoli3:
Fixes for iOS assimp addon device target:
https://github.com/openframeworks/openFrameworks/issues/8316

### DIFF
--- a/addons/ofxAssimp/addon_config.mk
+++ b/addons/ofxAssimp/addon_config.mk
@@ -102,11 +102,6 @@ android/x86:
 	ADDON_LIBS+=../ofxAssimpModelLoader/libs/assimp/lib/android/x86/libassimp.a
 	ADDON_LIBS+=../ofxAssimpModelLoader/libs/assimp/lib/android/x86/libIrrXML.a
 
-ios:
-	ADDON_LIBS=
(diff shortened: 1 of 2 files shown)


addons/ofxAssimp/addon_config.mk
```diff
@@ -102,11 +102,6 @@ android/x86:
...
 	ADDON_LIBS+=../ofxAssimpModelLoader/libs/assimp/lib/android/x86/libIrrXML.a
 
-ios:
-	ADDON_LIBS=
-	ADDON_LIBS=../ofxAssimpModelLoader/libs/assimp/lib/macos/assimp.xcframework/ios-arm64/assimp.a
-	ADDON_LIBS=../ofxAssimpModelLoader/libs/assimp/lib/macos/assimp.xcframework/ios-arm64_x86_64-simulator/assimp.a
-	
 osx:
 	ADDON_LIBS=
 	ADDON_LIBS=../ofxAssimpModelLoader/libs/assimp/lib/macos/assimp.xcframework/macos-arm64_x86_64/assimp.a
```
dimitre, line 107: is hardcoded ADDON_LIBS needed for osx? maybe something to help building with makefile?

ofTheo:
Yes! 
Needed for makefile as the depends code that PG uses doesn’t work with makefile.

danoli3:
can confirm a test when osx removing 
image

in theory though the makefile should be able to find it automatically via PC files if we package them

Issue #8329, opened 2025-03-03
[idea for macOS] using retina capable window with regular framebuffer


dimitre:
I'm using like this:
HIGH_RESOLUTION_CAPABLE = YES
and
glfwWindowHint(GLFW_COCOA_RETINA_FRAMEBUFFER, false);
which results in the same resolution everything the same

But the window decoration and title aren't low res anymore

danoli3:
If can make a variable to toggle this on an off (pass to the false there), so its as easy to toggle full retina on or off, could this be then changed dynamically... if so.. yes !

dimitre:
Working great for the projects I'm working on.
If you want this for 0.12.1 I can submit the PR

danoli3:
Yes please ! macOS non-retina apps look so broke without header at retina

PR #8338, opened 2025-03-05
Retina window please {do not merge yet}

dimitre:
closes https://github.com/openframeworks/openFrameworks/issues/8329

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -146,7 +146,7 @@ void ofAppGLFWWindow::setup(const ofGLESWindowSettings & settings) {
             setup(ofGLFWWindowSettings(settings));
         }
     }
-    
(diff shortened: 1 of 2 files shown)


artificiel:
how does that interact with applications that rely on high-resolution rendering? (i.e. what is currently happening with the `HIGH_RESOLUTION_CAPABLE = YES` flag).

dimitre:
I think there should be a boolean in ofWindowSettings. 
like bool highResolutionCapable = false;
that can be changed in ofGLWindowSettings.

This PR isnt' working now anyway, there is something off on window size on master.
I've been using this other code, that is simplified and doesn't rely on pixelScreenCoordScale
https://github.com/dimitre/openFrameworks/blob/ofworks/libs/openFrameworks/app/ofAppGLFWWindow.cpp

dimitre:
I'll close this one as OF needs some deeper change in GLFW window and other places

PR #8340, opened 2025-03-05
Restore target build dir behavior for Xcode. Closes #8321

### LINKED ISSUE #8321: [bug :: macOS] bin/data in DerivedData breaks user interaction with ofToDataPath()
so the actual `ofToDataPath()` is buried in DerivedData (ex: `/Users/brtn/Library/Developer/Xcode/DerivedData/myDeliberateSketch-aqwtovqchnmcfrazakooakhzrkco/Build/Products/Debug/data`) so the actual data dir is not bin/data...  might no be apparent for static ressources, but things like ofxWatchFile, ofxAutoReloadedShaders do not work as expected because watched ressources are those within DerivedData. and of course anything saved to disk will be thrown in DerivedData.

ofTheo:
Narrowly scoped just for osx/ template. 
Might be needed for macOS/ template too, but that is not currently being used by PG. 
I believe iOS and tvOS is not needed as they deploy to the device. 

Move the setting to Project.xcconfig with a comment, to give people the option to easily override it if they want. 

Closes #8321

### DIFF
--- a/scripts/osx/xcode_project.sh
+++ b/scripts/osx/xcode_project.sh
@@ -46,13 +46,13 @@ bundle_data_folder() {
 
     if [ -z "$OF_BUNDLE_DATA_FOLDER" ]; then
         msg 'Bundle data folder disabled \ncan be enabled with OF_BUNDLE_DATA_FOLDER=1 in Project.xcconfig ';
-        if [ "$ARCHIVING" = 0 ]; then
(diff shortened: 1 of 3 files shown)


danoli3:
Awesome I actually was thinking of how to Clean if we can't clean and can just add it to of script for the locations! 

I'm a clean fiend since cache creates gremlins

Issue #8271, opened 2025-02-19
ofGetFrameRate is always off by +0.5 fps


dimitre:
if you set framerate to 30 and read framerate it shows something like
30.5003 in macOS

I know it is about the reading because my alternative FPS display function is showing correct fps

![Image](https://github.com/user-attachments/assets/dbcd2fa8-e3b6-4ac5-b995-49bf5332a030)

Maybe it is related to this commit
https://github.com/openframeworks/openFrameworks/commit/a34ab6017a48b2d439698c0b9d7fb30114e1f057
cc: @danoli3

Jonathhhan:
I can confirm this with Windows. Setting to 60 (or any other rate) is the same, always + 0.5.

danoli3:
What time mode are you using?

Those changes were only for one specific mode that was non functional

dimitre:
I have no idea of time mode. 
steps to reproduce are just set framerate, and write results to the screen

dimitre:
I think this one is really important to be addressed before 0.12.1 release @ofTheo @danoli3

danoli3:
Can you give me your sample FPS code as above, im not seeing this with System I shall do analysis might be the change around update storing, meaning it might be superficial only

danoli3:
send on slack if private just snipped so I test exactly ur scenario

dimitre:
just use;
ofSetFrameRate(30)
and print the result

ofTheo:
@dimitre it does seem related to that [commit](https://github.com/openframeworks/openFrameworks/commit/a34ab6017a48b2d439698c0b9d7fb30114e1f057). 
I reverted to the previous commit and the fps went from 60.5 to 60.00024

master:

![Image](https://github.com/user-attachments/assets/47ecb117-ca6e-4765-a531-0a266b2027aa)


previous ofFpsCounter version:
![Image](https://github.com/user-attachments/assets/bba35b36-ebdd-401f-8237-4e361022c1dd)

dimitre:
yes I suspect the 0.5s removed in this commit was important
```
if(!timestamps.empty() && timestamps.front() + 0.5s < now){
```

ofTheo:
I tried reverting just that function and no change. 
Struggling to track it down. 

Only thing that works so far is revert both files and then make changes to ofCoreEvents to match the older api.

danoli3:
just a reminder that the FPS counter completely broke the most important ofGetLastFrameTime() prior to these changes. 

[...]

PR #8347, opened 2025-03-06
Fix ofFpsCounter value offset update(now)

danoli3:
Fixes: https://github.com/openframeworks/openFrameworks/issues/8271


danoli3:
image

Issue #6530, opened 2020-01-28
GLM_FORCE_CTOR_INIT needed for preprocessor macros 


neilmendoza:
I just updated to the latest version of OF and commit a9e962c is giving me some strange issues. In the example for [ofxTinynurbs](https://github.com/neilmendoza/ofxTinynurbs), if I add...

```
glm::vec3 a, b;
glm::vec3 c = 0.5f * (a + b);
```    

...at the start of ofApp::setup(), the data in ofxTinynurbs gets corrupted with OF commit a9e962c or later (where GLM_FORCE_CTOR_INIT is added).

ofTheo:
hmm. 
`#define GLM_FORCE_CTOR_INIT`

Is meant to set a and b to 0,0,0. 

can you cout a and b and share? 
both with GLM_FORCE_CTOR_INIT defined and commented out? 

I am assuming you have the glm libs updated from apothecary or from the latest release?

bakercp:
So, I ran into this problem and it took me several days to figure it out.  

I found that it happened in my addons where I was including glm headers _without_ calling including  `ofConstants.h` (where `GLM_FORCE_CTOR_INIT` is defined)  immediately before those includes. Further, I found that depending on the compile state, that wasn't reliable and I had to include `#include "ofVectorMath.h"` in all addons where glm was used directly.

Anyway, it was a huge headache and I'd strongly advocate that we not define `GLM_FORCE_CTOR_INIT` in OF, but rather make sure we manually initialize glm elements within the core and document that addons need to do the same as modern glm requires.

ofTheo:
Initially I thought we shouldn't include GLM_FORCE_CTOR_INIT. But in the OF core it took forever to track down and fix bugs related to quats and vecs not being set.  I think it could bring a bunch of misery to people's projects and break addons ( which are 0.11.0 compatible but aren't being updated regularly ) to not have that include.  

At the same time I understand that you'll end up with two types of glm code - some which assumes auto initialization and some which doesn't. I am not sure if there is a graceful way to transition from it ( like deprecation )?  Maybe there is? 

The code @neilmendoza posted above wouldn't be valid without it: 
```
glm::vec3 a, b;
glm::vec3 c = 0.5f * (a + b);
```

neilmendoza:
[...]

PR #8352, opened 2025-03-06
GLM Preprocessor Definitions to Project

NickHardeman:
Wrapping glm defines in if statements to avoid re-definition using project preprocessor definitions.
Will be editing projects to add `GLM_FORCE_CTOR_INIT` and `GLM_ENABLE_EXPERIMENTAL` preprocessor defines to projects.
As discussed in this thread #6530


NickHardeman:
Defines were not added for tvOS since the project/tvOS/CoreOF.xcconfig did not have GCC_PREPROCESSOR_DEFINITIONS and I do not have a proper way to test. 
And did not update the qtCreator project since I am not familiar with qtCreator.

ofTheo:
It would be good to test that the defines come through at the project level. 
I think the Xcode ones will as the Project.xcconfig includes the OF config, but not sure about VS and makefile common.

NickHardeman:
I tested Xcode and Emscripten on macOS using the following class included in ofApp.h that does not include ofMain.h. 
Emsripten worked after I edited the makefile common. 
VS and iOS still need to be tested properly.

[TestGLMSrc.zip](https://github.com/user-attachments/files/19130136/TestGLMSrc.zip)


```
#pragma once
#include <string>
#include <iostream>
#include <glm/vec3.hpp>
#include <sstream>

class TestGLM {
public:
	
	TestGLM() {
		std::stringstream ss;
#if defined(GLM_FORCE_CTOR_INIT)
		ss << "GLM_FORCE_CTOR_INIT defined.";
#else
		ss << "xxxx GLM_FORCE_CTOR_INIT xxxx NOT defined.";
#endif
		std::cout << ss.str() << std::endl;
		
		add("constructor");
	}
	
	void add(std::string aname) {
		glm::vec3 a, b;
		glm::vec3 c = 0.5f * (a + b);
		std::cout << aname << " c: " << c.x << " a: " << a.x << " b: " << b.x;
	}
	
	glm::vec3 ma, mb;
	glm::vec3 mc = 0.5f * (ma + mb);
    
};

```

NickHardeman:
Checking on VS 2022, the preprocessor macros are coming in and it prints as defined in the sample TestGLMSrc code above.
image

ofTheo:
Thanks!!!

PR #8373, opened 2025-03-13
Nightly - Android - Linux

danoli3:
Changes to nightly create package 
- Add Android Release - Windows / macOS 
- Linux create package to use built project generator rather than build 
- non verbose on extraction libs from download_libs
- disabled build android examples PG until fixed
- Allowed for nightly failure on one target to not break others.
- export build summary now found on action (click it) with table of success

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,14 +1,26 @@
 name: nightly-linux64
 
 on:
- schedule:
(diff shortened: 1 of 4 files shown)


danoli3:
build summary now found in action and will continue on fail 
image

danoli3:
Looks good shall figure out whats going on after live test, will put up another PR shortly

PR #8381, opened 2025-03-19
Android Updates

danoli3:
## Android 
- Template rework - fixed the source depth issue for template 
- CMakeList fixes for NEON / static/shared output for oF core 
- Core Gradle fixes for ^


danoli3:
- Fixed Project ofApp.h position to src 
in the past this was an issue with default Android source folders in Android Studio following src/main/cc/openframeworks/src/cpp/ or Java. Now just default openFrameworks just fixed up the CMake paths

danoli3 (commented):
Looks good

PR #8385, opened 2025-03-20
ofxAssimp: Fix for incorrectly over writing textures.

NickHardeman:
Material set metalness, roughness and clear coat values. Somewhat related: #8267
Clean up of commented code.
Recalculate scene bounds per request, instead of cache and requiring calling recalculate bounds per mesh.

### DIFF
--- a/addons/ofxAssimp/src/Source/ofxAssimpSrcBone.cpp
+++ b/addons/ofxAssimp/src/Source/ofxAssimpSrcBone.cpp
@@ -24,99 +24,6 @@ void SrcBone::setAiBone(aiBone* aAiBone, aiNode* aAiNode) {
 	mAiMatrix = mAiNode->mTransformation;
 }
 
-////--------------------------------------------------------------
-//void SrcBone::update() {
-//	if( mAiNode == nullptr ) {
-//		ofLogWarning("SrcBone::update mAiNode nullptr!!!");
-//		return;
-//	}
-//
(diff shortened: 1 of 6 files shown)


addons/ofxAssimp/src/Source/ofxAssimpSrcBone.cpp
```diff
@@ -24,99 +24,6 @@ void SrcBone::setAiBone(aiBone* aAiBone, aiNode* aAiNode) {
...
-////	for(auto it = childBones.begin(); it != childBones.end(); ++it ) {
-////		it->second->update();
-////	}
-////	for( auto& kid : childBones ) {
-////		kid->update();
-////	}
-//
-////	getGlobalTransformMatrix();
-//}
-
```
ofTheo, line 119: should all this be commented out?

NickHardeman: I think so, it was already commented out, I just removed it. I tested with the examples and they work as expected.
https://github.com/openframeworks/openFrameworks/blob/b4118a0c86a1bf5989f24216e4766efcdd8f3c4f/addons/ofxAssimp/src/Source/ofxAssimpSrcBone.cpp#L27-L118

ofTheo: oops - sorry I misread the diff - haha!

Issue #8396, opened 2025-03-26
ofCubeMapShaders.h not included by default, is this on purpose?


danomatika:
From checking ofxLua with the of_v20250319_osx_release nightly:

A question, should gl/ofCubeMapShaders.h be included anywhere? Currently, it is not included through ofMain and only included by its implementation file. Is this an oversight?

I know this as I told SWIG to read the header to generate wrappers however the class is then undefined upon building since the header is not included by default.

danomatika:
Also, there appears to be a typo: static class function `irriadianceCubeMap()` should probably be `irradianceCubeMap()`.

NickHardeman:
This PR addresses the typo #8399, so maybe this can be closed?

danomatika:
Yes thanks. Sorry for not making the PR myself. I found a bunch of things and realized my fork is a bit out of sync ATM.

Issue #8395, opened 2025-03-26
ofCubeMap::regenerateAllTextures() needs dummy implementation or Android-only declaration


danomatika:
From checking ofxLua with the of_v20250319_osx_release nightly:

The ofCubeMap regenerateAllTextures static class function is declared in the header, however the implementation is only available for Android. There should be dummy implementation for other platforms or the declaration should be wrapped by `#ifdef TARGET_ANDROID`.

PR #8399, opened 2025-03-26
ofCubeMap: Android define around regenerateAllTextures()

NickHardeman:
spelling correction for irradianceCubeMap()

#8395 and #8396


danoli3 (approved):
looks good - bytesize fix

PR #8407, opened 2025-03-30
ofxAssimp small suggestions

dimitre:
The important change here is getAsString
the base class has a different parameter type of the derived class (int / unsigned int), so it complains about shadowing.
I've changed the parameter types and added override
removed virtual from derived classes

and some small changes to avoid int64 -> int32 implicit conversion warnings


NickHardeman (approved):
Looks good to me!

Issue #8408, opened 2025-03-31
OF_SERIAL_ERROR and OF_SERIAL_NO_DATA no longer public


danomatika:
With `of_v20250330_osx_release`:

I notice that the `OF_SERIAL_ERROR` and `OF_SERIAL_NO_DATA` defines have been moved into ofSerial.cpp however the documentation and example code in ofSerial.h still refers to them. Both reading and writing as well as `available()` will return these values, however now the user cannot use the previous define. Is this on purpose?

PR #8413, opened 2025-03-31
serial error codes fix - back to public

dimitre:
closes https://github.com/openframeworks/openFrameworks/issues/8408


danomatika:
Yup, thanks

PR #8417, opened 2025-04-02
Small log and docs cleanup

danomatika:
Gathering some commits I made a few years ago:
* consistent whitespace within log messages, mainly no leading whitespace as one is added after the log module
* consistent log module naming, at least matching ones that are already there
* one small oxygen inline comment style set

This is not complete, just some small things I noticed and fixed on my fork, but forgot to make a PR before.


danomatika:
@ofTheo As requested for 0.13 ;)

dimitre:
it seems great for merging. any reason for waiting 0.13 ?

PR #8432, opened 2025-04-19
ofTrueTypeFont mesh insert at once

dimitre:


dimitre:
I'm loving to look at the errors here:
```
../../../libs/openFrameworks/graphics/ofTrueTypeFont.cpp:1021:3: error: non-constant-expression cannot be narrowed from type 'unsigned int' to 'unsigned short' in initializer list [-Wc++11-narrowing]
 1021 |                 firstIndex + 1U,
```
ofIndexType is an alias to TESSIndex. does it need to be? 
TESSIndex vary from platform to platform? Ins't unsigned short is too short to represent indices? 
What about this was fixed std::size_t for all platforms?

ofTheo:
@dimitre there is definitely a historical reason for ofIndexType being TESSIndex - something with max buffer sizes on memory constrained devices? 
I can't remember 100%, but  I wouldn't change it unless there is bug that it solves.

ofTheo:
@dimitre def feel free to merge. 
Curious what the reason behind this was, did you notice a big speed up for font loading?

dimitre:
the idea at the time was to optimize tiny things on the core, like addVertices to resize memory allocation once. 
Performance wise never tested, I suppose it is equal or better, not worse.
Now I don't think OF even should deal directly with typography, there are much better tools out there to do complicated things like glyph positioning.

PR #8450, opened 2025-06-01
Don't load non-existent JSON settings file when saving a new file

smeyfroi:
Prevent this error when a new JSON settings is saved:
```
[ error ] ofLoadJson: Error loading json from "/Users/steve/Documents/MarkSynth/settings/synth1/settings-4.json": file doesn't exist
```

I'm also suspicious of L235-237 for the same reason, but there's no error message when loading a non-existent XML so not sure.


ofTheo:
Thanks!!

Issue #8452, opened 2025-06-10
Fix offset on ofDrawArrow


elliotwoods:
fix code so that the point of the arrow head has its tip on the position that is input (previously the middle of the arrow head not the tip would be at the position given as the input argument)

proposed code:

```
//--------------------------------------------------------------
void of3dGraphics::drawArrow(const glm::vec3& start, const glm::vec3& end, float headSize) const{

	//draw line
	renderer->drawLine(start.x,start.y,start.z, end.x,end.y,end.z);

	// Note that `glm::rotation` requires its parameters to be normalized `glm::vec3`s
	const glm::mat4 cone_height_offset = glm::translate(glm::vec3(0, headSize, 0));
	const glm::mat4 cone_rotation    = glm::mat4_cast(glm::rotation(glm::vec3(0,1,0), glm::normalize(start-end)));
	const glm::mat4 cone_translation = glm::translate(end);
	const glm::mat4 cone_transform   = cone_translation * cone_rotation * cone_height_offset;

	renderer->pushMatrix();
	renderer->multMatrix(cone_transform);
    drawCone(headSize, headSize*2.);
    renderer->popMatrix();
}
```

Corrected behaviour:
![Image](https://github.com/user-attachments/assets/7fe434c8-8798-4722-83ef-6660488aef59)

Previous behaviour:
![Image](https://github.com/user-attachments/assets/0bc6744a-8135-49c2-bf8c-1eb79a68442e)

I spent a few minutes trying to make this into a pull request but a few extra edits showed up in the PR so I've spent my time limit there and am doing it as an issue instead. It's tested here so I'd be happy to just edit it in directly on the oF repo on github, but I know that isn't favoured.

Elliot

ofTheo:
Awesome - @elliotwoods 
Sounds like a good plan - want to send a PR?

PR #8453, opened 2025-06-11
Fix the cone head on an arrow as per Fix offset on ofDrawArrow #8452

elliotwoods:


elliotwoods:
I tried a couple of times to do this the right way (branch in my own repo and PR from there), but this is the closest I could get to work (branch in oF main repo). I'll delete the branch after,
Sorry about this. Feel free to kick me off from admin privileges from oF repo

ofTheo:
Haha!! 
All good @elliotwoods - I often struggle to get a clean PR. 

My current approach to avoid all the merge commits is simply clone a fresh repo and do the changes there. 
Then delete. :) 
Not efficient but probably saves me time in the end.

PR #8468, opened 2025-09-12
[bugfix] linux64 linker

themancalledjakob:
The gold linker is slowly fading out of existence on some Linux distros (e.g. Void linux).
Compilation therefore fails (unnecessarily).

[This is very likely gonna happen for any modern Linux distro.](https://www.phoronix.com/news/GNU-Gold-Linker-Deprecated)

But even apart from this, I think a better approach would be to only add `-fuse-ld=gold` if `gold` exists.

And while we're at it.. use `mold`, if `mold` exists.

### DIFF
--- a/libs/openFrameworksCompiled/project/linux64/config.linux64.default.mk
+++ b/libs/openFrameworksCompiled/project/linux64/config.linux64.default.mk
@@ -27,4 +27,8 @@
 
 include $(OF_SHARED_MAKEFILES_PATH)/config.linux.common.mk
 
-PLATFORM_LDFLAGS += -fuse-ld=gold
+ifneq (, $(shell command -v mold))
+ PLATFORM_LDFLAGS += -fuse-ld=mold
+else ifneq (, $(shell command -v gold))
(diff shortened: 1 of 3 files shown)


ofTheo:
Thanks!! @themancalledjakob

dimitre:
Thank you @themancalledjakob 
I've followed your logic and added other linkers in a different order, in a fork I'm working.
mold, lld, bfd (we can check in shell by invoking ld, but -fuse-ld parameter is bfd) and last of all, gold.

```
include $(OF_SHARED_MAKEFILES_PATH)/config.linux.common.mk

ifneq (, $(shell command -v mold))
	PLATFORM_LDFLAGS += -fuse-ld=mold
else ifneq (, $(shell command -v lld))
	PLATFORM_LDFLAGS += -fuse-ld=lld
else ifneq (, $(shell command -v ld))
	PLATFORM_LDFLAGS += -fuse-ld=bfd
else ifneq (, $(shell command -v gold))
	PLATFORM_LDFLAGS += -fuse-ld=gold
endif

```

PR #8483, opened 2025-11-05
Add filesystem support to Emscripten config

ddgg-el:
Enable filesystem support in Emscripten build configuration.


dimitre:
Thank you @ddgg-el

PR #8484, opened 2025-11-12
Added Yunet face detection example

brunovianna:
This is a contributed example of a face detection application using neural networks in opencv.

It was discussed in this forum thread: https://forum.openframeworks.cc/t/face-detection-with-neural-networks/45209


ofTheo:
Thanks! I imagine the models are too big to bundle with the example?

PR #8490, opened 2026-01-09
MIN -> std::min in examples

dimitre:
Updating examples replacing MIN macro to more up to date std::min.
in my opinion all math macros should be removed in favor to std:: ones


ofTheo:
looks good to me!!

Issue #8472, opened 2025-09-13
libImath  cannot open shared object [ linux ]


CarloCattano:
running ProjectGenerator or attempting to run any compiled _oF_ project, I get the same message:
```bash
projectGenerator: error while loading shared libraries: libImath-3_1.so.29: cannot open shared object file: No such file or directory
```
it kind of makes sense, since I only seem to have 
```
❯ /usr/bin/ls libImath*
libImath-3_2.so  libImath-3_2.so.30  libImath-3_2.so.30.3.2.1  libImath.so
```

This is with latest git version

```
 pacman -Qs imath     
local/imath 3.2.1-1
    A C++ and python library of 2D and 3D vector, matrix, and math operations for computer graphics
                
```

any clues  ?

# TLDR fix 
the arch script uses old freeimg package which was dropped.
remove freeimg package if you have it installed
then reinstall with paru, yay or makepkg -si
`aur/freeimage 3.18.0-23 [+13 ~1.11] `

artificiel:
this is the classic dylib problem, where the system upgrades while the app refers to an older lib that's not there anymore. the proper binary distribution is to bundle everything into a flatpack/appImage, so the app carries it's libs.

as a hack you can try `sudo ln -s /usr/lib/libImath-3_2.so.30 /usr/lib/libImath-3_1.so.29` and depending on the actual interaction with the lib, it may not segfault (the number coming after `.so` is related to the lib's ABI, different numbers indicates incompatibilities, but I doubt Project Generator makes any calls into that lib). you can also procure a copy of the older version of the lib (I'm attaching one, just in case it works). if you don't want to mess with `/usr/lib` you can use `LD_LIBRARY_PATH` to find libs elsewhere at runtime. I think putting the lib besides the compiled app or in it's `data/` may also work.

contrarily to other distros that provide ways to install older versions of libs, Arch's rolling release approach is to clean slate everything, and there is no way to strictly tie a build to a "distribution version". the "Arch way" would be to recompile the packages that rely on outdated libs, but getting Project Generator to compile these days appears to be tricky...

[libImath-3_1.so.29.zip](https://github.com/user-attachments/files/22434341/libImath-3_1.so.29.zip)

CarloCattano:
Thanks , that was my first attempt to manually sym link it, but unfortunately it throws
```bash
[...]

PR #8497, opened 2026-01-25
Fix Arch install scripts

CarloCattano:
After the [recent drop of freeimage](https://archlinux.org/todo/drop-freeimage/) on arch repos, we need an alternative to install it on arch. 

This PR hopes to keep the `/scripts/linux/arch` in a working state.

Uses yay, or paru if available. If not the user should know how to do it with makepkg on its _own_

Added few missing dependencies in the install from main arch repositories .

> Hope to keep it up to date to keep enjoying oF on Arch `btw` !!

Related issues have raised in the past and will still do but we can keep an eye.

https://github.com/openframeworks/openFrameworks/issues/8472

### DIFF
--- a/scripts/linux/archlinux/install_dependencies.sh
+++ b/scripts/linux/archlinux/install_dependencies.sh
@@ -11,7 +11,7 @@ fi
 
 ROOT=$(cd $(dirname $0); pwd -P)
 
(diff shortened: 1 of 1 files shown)


ofTheo:
Looks great @CarloCattano ! 
Thanks so much!!!

CarloCattano:
Thanks to you for all the work :-)  Hope someone can verify soon on a fresh install. There has to be said that if one misses the script `download_libs.sh` , can run into problems too. Will think about it @ofTheo

PR #8502, opened 2026-02-03
update alpha blend, smoothing, antialiasing and point sprites to be set with bool

surrealism21:
Previously these were set with `ofEnable[property]` or `ofDisable[property]`. I have changed that so they're now set with a bool, which I feel most users will prefer. All the examples have also been updated, and this code has been tested.
Failed to mention in the commit message but point sprites have also been changed to a bool set function


ofTheo:
Hi there! 

I am sorry @surrealism21 but we don't accept huge Pull Request changes like this without discussion and approval from the @openframeworks/core  team. 

If you are interested in contributing, PRs that close issues with clear fixes are most welcome!

PR #8515, opened 2026-05-21
Forward addon -std=* from ADDON_CFLAGS into CXXFLAGS

GitBruno:
Forward addon -std=* from ADDON_CFLAGS into CXXFLAGS
ADDON_CFLAGS are merged into OF_PROJECT_CFLAGS only, so addon
-std=gnu++20 (etc.) never reached C++ builds. Extract -std flags
from PROJECT_ADDONS_CFLAGS, dedupe against core flags, and apply
the last addon value to both CFLAGS and CXXFLAGS.


danoli3:
Thank you tested working

PR #8518, opened 2026-05-30
Improve ofxSvg

GitBruno:
Extract attribute conversion with proper arc and Catmull-Rom curve handling; add Inkscape layer and fill-rule support; fix use href and x/y placement. And preserve Inkscape layers.

There is more work to do but this is what I needed for my ofxDXF addon.

For full SVG spec compliance we would still need gradients, filters, clipping, dashed strokes, skew, animations, etc :)

I would like support of dashed strokes but not quite sure how to go about. Any comment from you guys would be helpful regarding this.

### DIFF
--- a/addons/ofxSvg/src/ofxSvg.cpp
+++ b/addons/ofxSvg/src/ofxSvg.cpp
@@ -1,5 +1,6 @@
 #include "ofxSvg.h"
 #include "ofUtils.h"
+#include <deque>
 #include <regex>
 #include "ofGraphics.h"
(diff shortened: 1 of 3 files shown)


danoli3:
Nice can you confirm ofxSVG example is working with these changes

danoli3:
@GitBruno https://github.com/ofKitty closed this by deleting the head repository [1 hour ago](https://github.com/openframeworks/openFrameworks/pull/8518#event-29153659154)

PR #8530, opened 2026-08-09
GLFW Window Focus Fix

danoli3:
## ofxGLFWWindow Focus
- Add `bWindowFocused` member to `ofAppGLFWWindow`
- Add `focus_cb` static callback (matches `GLFWfocusfun`), registered with
  `glfwSetWindowFocusCallback` alongside the other window callbacks in `setup()`
- Seed initial state from `glfwGetWindowAttrib(windowP, GLFW_FOCUSED)` right
  after registration, since a window isn't guaranteed to start focused
- Clear the callback in `close()` for symmetry with the other teardown calls
- `isWindowActive()` now returns `bWindowFocused` instead of the stub

ofAppGLFWWindow::isWindowActive()` was hardcoded to `return true`, so window now can be detected if active or not active... useful for backgrounding apps running machines, so you can now actively pause or sleep draw or update

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -94,6 +94,7 @@ void ofAppGLFWWindow::close() {
 		glfwSetScrollCallback( windowP, nullptr );
 		glfwSetDropCallback( windowP, nullptr );
   	glfwSetWindowRefreshCallback(windowP, nullptr);
+		glfwSetWindowFocusCallback(windowP, nullptr);
 
 		//hide the window before we destroy it stops a flicker on OS X on exit.
 		glfwHideWindow(windowP);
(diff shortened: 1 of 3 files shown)


danoli3:
Ideally this should be a windowResized event in coreEvents, &ofBaseApp::windowFocusChanged... however I think for complexity it's easier for the moment to just add it as a ofEvent<bool> windowFocusChanged where opt in is:

```
ofAddListener(ofEvents().windowFocusChanged, this, &MyApp::onFocusChanged)
```

easy

danoli3:
Added ofEvent<bool> windowFocusChanged; to ofEvents

danoli3:
@ofTheo is ofEvent good idea for now vs change core events virtual function etc... I think this is good to merge

danoli3 (commented):
good changes

dimitre:
Thanks @danoli3 I'm not using this anymore in my day to day use.
but I've been using a simplified version of this for the last two years
if you want to take a look it is this code:
https://github.com/dimitre/openFrameworks/blob/ofworks/libs/openFrameworks/app/ofAppGLFWWindow.cpp

I removed some code that counted frames since state change and others simplifications. but there is api change there so near zero change of being pulled to OF main :/

Other two niceties there is ofWindowSettings is only one object for all kinds of windows (I think it is best) and also we can choose to span one same window to multiple monitors (poor's man Nvidia Mosaic) but I think it is very useful to span 4 outputs in one single window, as I do often.

danoli3:
I'm going to put this in for now as Focus / pause event quite useful in general for apps for power usage

PR #8539, opened 2026-08-14
ofxSvg updates

NickHardeman:


addons/ofxSvg/src/ofxSvg.cpp
```diff
@@ -185,7 +185,7 @@ ofxSvg::ofxSvg(const of::filesystem::path & fileName) {
 
 //--------------------------------------------------------------
 bool ofxSvg::load( const of::filesystem::path& fileName ) {
-	ofFile mainXmlFile( fileName, ofFile::ReadOnly );
+	ofFile mainXmlFile( fileName, ofFile::ReadWrite );
```
danoli3: Does it need write?

danoli3 (changes requested):

addons/ofxSvg/src/ofxSvg.cpp
NickHardeman: Hmm, doesn't seem like it should, especially since it's in load and is a temporary variable. I can switch it back to ReadOnly.

NickHardeman: Reverted via https://github.com/openframeworks/openFrameworks/pull/8539/commits/48e0cc3ab1adbf864f157432aae32b5c3d3bb71e

danoli3 (approved):
Looks good

PR #8541, opened 2026-08-15
CWE-22 - Fix Android ZIP path traversal

### LINKED ISSUE #8509: [BUG] Zip Path Traversal in openFrameworks Android Extraction Helper
### SUMMARY
Latest-source `openframeworks/openFrameworks` still exposes ZIP path traversal in the Android extraction helper. I cloned the latest upstream tree, compiled the in-tree `OFZipUtil.java` inside Docker together with minimal Android stub classes, and drove the real `OFZipUtil.extract(InputStream, File)` API with a crafted ZIP entry named `../escaped.txt`. It seems like this issue has a similar root cause to CVE-2021-23520.

### DETAILS
Reachable target source file:
- `openFrameworks/addons/ofxAndroid/Java/cc/openframeworks/OFZipUtil.java`

Relevant vulnerable fragment from the latest target source:

```java
public static void extractFile(ZipInputStream in, File outdir, String name) throws IOException
{
  byte[] buffer = new byte[BUFFER_SIZE];
  BufferedOutputStream out = new BufferedOutputStream(new FileOutputStream(new File(outdir,name)));
  int count = -1;
  while ((count = in.read(buffer)) != -1)
    out.write(buffer, 0, count);
  out.close();
}

public static void extract(InputStream zipfile, File outdir)
{
  try
  {
    ZipInputStream zin = new ZipInputStream(zipfile);
    ZipEntry entry;
    String name, dir;
    Log.i("OF", "uncompressinggggg ");
    while ((entry = zin.getNextEntry()) != null)
    {
      name = entry.getName();
      if( entry.isDirectory() )
      {
        mkdirs(outdir,name);
        continue;
      }
      dir = dirpart(name);
      if( dir != null )
        mkdirs(outdir,dir);

      extractFile(zin, outdir, name);
    }
    zin.close();
  }
  catch (IOException e)
  {
    e.printStackTrace();
  }
}
```

This is the core issue. The latest source still passes the ZIP entry name directly into new File(outdir, name) without removing .., canonicalizing the path, or verifying that the final destination remains under the intended extraction root. As a result, an entry such as ../escaped.txt is resolved outside the extraction directory.

### Comparison with the original vulnerability:

The upstream JUCE fix addressed the same archive-entry trust boundary by rejecting extraction targets that escape the intended root directory.

```cpp
auto targetFile = targetDirectory.getChildFile (entryPath);

if (! targetFile.isAChildOf (targetDirectory))
    return Result::fail ("Entry " + entryPath + " is outside the target directory");

if (followSymlinks == FollowSymlinks::no && hasSymbolicPart (targetDirectory, targetFile.getParentDirectory()))
    return Result::fail ("Parent directory leads through symlink for target file: " + targetFile.getFullPathName());
```

The difference is clear:

JUCE constructs the target path and then verifies root containment.
openFrameworks constructs the target path and immediately opens the file.
Therefore, the original archive entry path -> filesystem write outside root primitive remains present in the current target.

### Full-path execution result:

```text
LOG[OF] uncompressinggggg
ZIP_PATH /tmp/openframeworks_fullpath_case/traversal.zip
OUTDIR /tmp/openframeworks_fullpath_case/extract-root
ESCAPED_PATH /tmp/openframeworks_fullpath_case/escaped.txt
ESCAPED_EXISTS true
ESCAPED_CONTENT OPENFRAMEWORKS_ZIP_TRAVERSAL
```
This result shows that the real helper API path in the latest source wrote /tmp/openframeworks_fullpath_case/escaped.txt outside the intended extraction root /tmp/openframeworks_fullpath_case/extract-root. In other words, the traversal primitive is confirmed as a real write outside the destination directory.

### Why this is vulnerable:
- ZIP entry names are attacker-controlled input.
- The current helper treats the entry path as if it were a trusted child path.
- As a result, it can write files outside the extraction root.

Impact:
- If an attacker-controlled ZIP is passed to the Android-side extraction helper, files can be written outside the intended extraction directory.

Please check the information and fix the code.

Also, let me know if there is anything you need to know when you proceed with the patch.

danoli3:
Fixes #8509

Prevents arbitrary file writes outside the extraction directory (CWE-22 / Zip Slip).

Test:
`tests/android/OFZipUtil/run.sh`

Output:
`OFZipUtil security regression tests passed`

### DIFF
--- a/addons/ofxAndroid/Java/cc/openframeworks/OFZipUtil.java
+++ b/addons/ofxAndroid/Java/cc/openframeworks/OFZipUtil.java
@@ -16,21 +16,49 @@ public class OFZipUtil
 {
   private static final int  BUFFER_SIZE = 4096;
 
+  private static File resolveEntry(File outdir, String name) throws IOException
+  {
+    File canonicalOutdir = outdir.getCanonicalFile();
+    File destination = new File(canonicalOutdir, name).getCanonicalFile();
+    String outdirPath = canonicalOutdir.getPath();
+    String destinationPath = destination.getPath();
+
+    if (!destinationPath.equals(outdirPath)
+        && !destinationPath.startsWith(outdirPath + File.separator))
+      throw new IOException("Zip entry is outside the target directory: " + name);
+
+    return destination;
+  }
+
   public static void extractFile(ZipInputStream in, File outdir, String name) throws IOException
   {
     byte[] buffer = new byte[BUFFER_SIZE];
-    BufferedOutputStream out = new BufferedOutputStream(new FileOutputStream(new File(outdir,name)));
-    int count = -1;
-    while ((count = in.read(buffer)) != -1)
-      out.write(buffer, 0, count);
-    out.close();
+    File destination = resolveEntry(outdir, name);
+    try (BufferedOutputStream out = new BufferedOutputStream(new FileOutputStream(destination)))
+    {
+      int count;
+      while ((count = in.read(buffer)) != -1)
+        out.write(buffer, 0, count);
+    }
+  }
+
+  private static void createDirectories(File outdir, String path) throws IOException
+  {
+    File directory = resolveEntry(outdir, path);
+    if (!directory.exists() && !directory.mkdirs() && !directory.isDirectory())
+      throw new IOException("Could not create directory: " + directory);
   }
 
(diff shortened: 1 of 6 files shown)


danoli3 (commented):
Looks good

danoli3:
Push for v12.2

PR #8550, opened 2026-08-31
Use X11/Wayland GLFW on linux aarch64, not EGL/GLES

danoli3:
## Why

Native ARM64 Linux builds GLFW with **both** X11 and Wayland backends (the ARM64 setup already installs Wayland, xkbcommon, and X11 headers). `armv6l` / `armv7l` still disable those and use EGL/GLES.

`linuxaarch64` was still pulling `glesv1_cm` / `glesv2` / `egl`, and `InitGLFW()` always forced `GLFW_PLATFORM_X11`. That meant a packaged `libglfw3.a` with both backends could not actually run Wayland, and aarch64 looked like the Pi GLES target.

openFrameworks only needs to **link** the right windowing libs and **select** the backend at runtime.

## Changes

- `ofConstants.h`: `__aarch64__` / `__ARM64__` → desktop `TARGET_LINUX` (GLFW), not `TARGET_LINUX_ARM` + `TARGET_OPENGLES`
- `config.linuxaarch64.default.mk`: drop EGL/GLES pkg-config (desktop GL comes from `config.linux.common.mk` because `LINUX_ARM` is not set for aarch64)
- `config.linux.common.mk`: on desktop linux (x86_64 and aarch64), pkg-config-link `wayland-client`, `wayland-cursor`, `wayland-egl`, `xkbcommon`, `libdecor-0` when present
- `ofAppGLFWWindow`: if `WAYLAND_DISPLAY` is set, try Wayland, then fall back to X11
- Arch deps: `wayland`, `libxkbcommon`, `libdecor` (Debian/Ubuntu/Fedora already had them)

armv6l / armv7l are unchanged (EGL/GLES).

### DIFF
--- a/libs/openFrameworks/app/ofAppGLFWWindow.cpp
+++ b/libs/openFrameworks/app/ofAppGLFWWindow.cpp
@@ -33,6 +33,8 @@
 	#include <GLFW/glfw3native.h>
 #endif
 
+#include <cstdlib>
+
 using std::numeric_limits;
 using std::shared_ptr;
 using std::vector;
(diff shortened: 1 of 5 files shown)


libs/openFrameworks/app/ofAppGLFWWindow.cpp
```diff
@@ -33,17 +33,27 @@
...
+			const char *waylandDisplay = getenv("WAYLAND_DISPLAY");
+			if (waylandDisplay && waylandDisplay[0] != '\0') {
+				glfwInitHint(GLFW_PLATFORM, GLFW_PLATFORM_WAYLAND);
+				if (glfwInit()) {
+					return true;
+				}
+				glfwTerminate();
+				ofLogNotice("ofAppGLFWWindow") << "Wayland init failed, trying X11";
+			}
+			glfwInitHint(GLFW_PLATFORM, GLFW_PLATFORM_X11);
```
ofTheo, line 55: the GLFW_PLATFORM_WAYLAND lines above can't be approved as Native Wayland breaks core ofSetWindowPos ofSetWindowShape functionality.

ofTheo:
Just a heads up @danoli3 - totally fine for ARM64 Linux builds to be treated like regular TARGET_LINUX, but we should always be doing GLFW_PLATFORM_X11 and never GLFW_PLATFORM_WAYLAND even for systems running Wayland, as Wayland has no way to programmatically move or resize windows. 

So as is, please don't merge this.

libs/openFrameworks/app/ofAppGLFWWindow.cpp
danoli3: I'll test it out. Yeah moved to draft

PR #8554, opened 2026-09-04
Decouple GL from ofGraphics

NickHardeman:
Removed GL specific code in ofGraphics and required the following definitions in renderers. 
```
void drawBackgroundGradient(const ofFloatColor& start, const ofFloatColor& end, ofGradientMode mode) const;
void enablePointSprites();
void disablePointSprites();
```

Removed `ofVboMesh` from `ofPath` and replaced with `ofMesh`. 
This makes it a bit easier to create new renderers without having certain functions and variables tied to GL specific code.

### DIFF
--- a/addons/ofxGui/src/ofxToggle.h
+++ b/addons/ofxGui/src/ofxToggle.h
@@ -1,8 +1,8 @@
 #pragma once
-
 #include "ofParameter.h"
 #include "ofxBaseGui.h"
 #include "ofPath.h"
+#include "ofVboMesh.h"
 
 class ofxToggle : public ofxBaseGui{
 public:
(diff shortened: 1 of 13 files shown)


ofTheo:
Awesome! 
And I think we could do some more to consolidate, where the GL functions get abstracted out of drawBackgroundGradient so the mesh generation etc for the backgroundGradient could be universal and in one place.  
But this is great for the timebeing!!  

Nice to see some steps towards real swappable renderers

NickHardeman:
@ofTheo agree, it's not great that the code is duplicated between the GLRenderer and the programmable renderer. 
Renderers like ofCairoRenderer don't support mesh colors
[ofLogWarning("ofCairoRenderer") << "draw(): cairo mesh rendering doesn't support colors, textures, or normals. drawing wireframe ...";](https://github.com/openframeworks/openFrameworks/blob/18565f66997d40c704e10c8400fcbba4c80e480c/libs/openFrameworks/graphics/ofCairoRenderer.cpp#L334)

And ofNoopRenderer doesn't draw anything :) 
https://github.com/openframeworks/openFrameworks/blob/18565f66997d40c704e10c8400fcbba4c80e480c/libs/openFrameworks/app/ofAppNoWindow.cpp#L69

danoli3:
Very nice!!!

PR #8533, opened 2026-08-11
OpenGLES 3.1

danoli3:
### OpenGLES 3.1
Updated PR to https://github.com/openframeworks/openFrameworks/pull/8512
Merged in @mruegenberg changes and patched them into the core.
Support for openGLES 3.0
Support for openGLES 3.1
Fixed up Emscripten defines to allow for ES3
This feature allows for emulation more easily in Dawn / Emscripten / Metal / Vulkan for Mobile platforms for ES
Merged with Tess2 fixes. Auto Short / Int determination (less defines)

## TODO: 
- [x] Review Code changes
- [x] Smoke Test on iOS / Android / Emscripten 
- [x] OpenGL ES 1 compatibility 
- [x] OpenGL ES 2 compatibility 
- [x] OpenGL compatibility
- [ ] Simulator Test on iOS / Android 
- [ ] Device Test on iOS / Android

PR #8557, opened 2026-09-11
Add GL smoke tests with CI headless fixes

danoli3:
Add tests/gl/glRenderSmokeTest (desktop GLFW) with shared GLSmokeTestCore.h exercising indexed VBO, instanced draw, tessellated ofPath, FBO readback.

- Add tests/emscripten/glRenderSmokeTest and tests/ios/glRenderSmokeTest (symlinked core)
- Fix GLES CI: keep CC as C under ccache (was CXX), skip foreign tests (android/ios/emscripten/tvOS) on desktop runners, run linux tests under xvfb when headless, skip GLFW smoke test if window never came up

Split from #8533 OpenGLES 3.1 to land tests on master before main GLES merge, so CI can validate them independently.


scripts/ci/linux64/install.sh
danoli3, line 1: Virtual Frame buffer - allows for headless linux to run GPU tests

danoli3 (commented):
Isolated to test cases. fixed some mk file issues

PR #8562, opened 2026-09-14
Link split OpenSSL XCFrameworks (libssl / libcrypto)

danoli3:
Follows openframeworks/apothecary#596.

Apple packages now include `libssl.xcframework` and `libcrypto.xcframework` alongside the legacy combined `openssl.xcframework`. OF was still linking only the combined product, which is the wrong layout for POCO and CMake FindOpenSSL.

## Changes

- macOS, osx, and iOS OF lib Xcode projects link `libssl.xcframework` and `libcrypto.xcframework`
- makefile discovery skips `openssl.xcframework` so the combined archive is not also pulled in (duplicate symbols)
- leftover `openssl.xcframework` file refs stay in the project groups for the transition package
- ofxPoco extra config points at the split products instead of `ssl.a` / `crypto.a`

Headers are unchanged (`libs/openssl/include`).


danoli3 (commented):
Looks good (review)

danoli3:
Fixed headers in lib crypto only. Should fix the last issue there

danoli3:
Will merge in a few hours after running some tests

